PR #17, opened 2022-01-14
CI: Export platform specific env variables

certik:


certik:
Linux and macOS now work, Windows is pretty far, but not quite yet.

PR #23, opened 2022-01-15
Remove all the Fortran tests and reference

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
This PR removes all the test files in both `tests`(including error, warning, reference and interop) and `integration_test`(*.f90) folders.

PR #28, opened 2022-01-15
Make CI to run only on main `lpython` repo

Smit-create:
Currently, CI runs on every commit on `lcompilers/lpython` as well as forked repo(`user/lpython`) which might be a waste of resources. So we could just run these tests when opened a PR in the `main` branch or pushed to the `main` branch.

cc @certik

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,10 +1,17 @@
-name: CI
+name: LPython CI
 
-on: [push, pull_request]
+on:
+  push:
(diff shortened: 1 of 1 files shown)


Smit-create:
For reference CI also ran on my fork for the PRs I opened previously: https://github.com/Smit-create/lpython/actions

certik:
I think that's fine. If somebody wants to run a CI, they have to open a PR, or enable it in their fork.

PR #29, opened 2022-01-15
Implement string repetition

### LINKED ISSUE #13: Implement repeat operator (strop)
This should work:
```python
def main():
    s: str
    s = 'a' * 2
    s = 4 * 'bb'
    s = 3 * 'a' * 3
```

namannimmo10:
Fixes #13

### DIFF
--- a/grammar/ASR.asdl
+++ b/grammar/ASR.asdl
@@ -238,7 +238,7 @@ binop = Add | Sub | Mul | Div | Pow
 
 unaryop = Invert | Not | UAdd | USub
 
-strop = Concat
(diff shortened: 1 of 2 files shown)


Thirumalai-Shaktivel:
Can you please add a test in the `tests` folder??

namannimmo10:
Since we discussed in our last group meeting, that this is one obvious solution to string repetition, I think this can go in.

certik:
I think this is fine for now. We need to setup tests in the tests folder and then test all these things. For now I think we can merge it.

PR #31, opened 2022-01-15
CI: test the new Python frontend

certik:


certik:
@Thirumalai-Shaktivel, @namannimmo10, @Smit-create: you can now use this file: https://github.com/lcompilers/lpython/blob/270ec24518506de0cc169404bb56e494154f8f52/doconcurrentloop_01.py as a test, as it is executed by the CI on all 3 platforms. Simply add new functions to it and in the functions test whatever statements or expressions you implement. Not ideal but at least we know that it will compile to ASR (even if the ASR might not be correct). We'll setup regular tests soon.

PR #32, opened 2022-01-15
Implement `complex` intrinsic function

namannimmo10:
expr2.py contains:
```python
def main():
    c: c128
    c = complex()
    c = complex(3.4)
    c = complex(5., -4.3)
```
ASR:
```bash
(TranslationUnit
   (SymbolTable
      1
      {
         main:
            (Subroutine
               (SymbolTable
                  2
                  {
                     c:
                        (Variable
                           2
                           c
                           Local () ()
                           Default
                           (Complex 8 [])
                           Source
                           Public
                           Required .false.)
                  })
               main [] [
               (=
                  (Var 2 c)
                  (ConstantComplex 0.000000 0.000000
                     (Complex 8 [])) ())
               (=
                  (Var 2 c)
                  (ConstantComplex 3.400000 0.000000
                     (Complex 8 [])) ())
               (=
                  (Var 2 c)
                  (ConstantComplex 5.000000 -4.300000
                     (Complex 8 [])) ())]
               Source
               Public
      })
   [])
```

### DIFF
--- a/doconcurrentloop_01.py
+++ b/doconcurrentloop_01.py
@@ -2,7 +2,10 @@ def test_complex():
     c: c128
     c = complex()
     c = complex(3.4)
-    c = complex(5., -4.3)
+    c = complex(5., 4.3)
+    c = complex(1)
+    c = complex(3, 4)
+    c = complex(2, 4.5)
 
 def triad(a: f32[:], b: f32[:], scalar: f32, c: f32[:]):
     N: i32
--- a/doconcurrentloop_01.py
+++ b/doconcurrentloop_01.py
@@ -1,11 +1,20 @@
 def test_complex():
     c: c128
+    c1: c128
+    c2: c128
+    c3: c128
+    b: bool
+
+    # constant real or int as args
     c = complex()
     c = complex(3.4)
     c = complex(5., 4.3)
     c = complex(1)
-    c = complex(3, 4)
-    c = complex(2, 4.5)
+    c1 = complex(3, 4)
+    c2 = complex(2, 4.5)
+    c3 = complex(3., 4.)
+    b = c1 != c2
+    b = c1 == c3
 
 def triad(a: f32[:], b: f32[:], scalar: f32, c: f32[:]):
     N: i32
--- a/doconcurrentloop_01.py
+++ b/doconcurrentloop_01.py
@@ -1,3 +1,9 @@
+def test_complex():
+    c: c128
+    c = complex()
+    c = complex(3.4)
+    c = complex(5., -4.3)
+
 def triad(a: f32[:], b: f32[:], scalar: f32, c: f32[:]):
     N: i32
     i: i32
--- a/doconcurrentloop_01.py
+++ b/doconcurrentloop_01.py
@@ -1,21 +1,3 @@
-def test_complex():
-    c: c128
-    c1: c128
-    c2: c128
-    c3: c128
-    b: bool
-
-    # constant real or int as args
-    c = complex()
-    c = complex(3.4)
-    c = complex(5., 4.3)
-    c = complex(1)
-    c1 = complex(3, 4)
-    c2 = complex(2, 4.5)
-    c3 = complex(3., 4.)
-    b = c1 != c2
-    b = c1 == c3
-
 def triad(a: f32[:], b: f32[:], scalar: f32, c: f32[:]):
     N: i32
     i: i32
--- a/src/lfortran/semantics/python_ast_to_asr.cpp
+++ b/src/lfortran/semantics/python_ast_to_asr.cpp
@@ -1212,6 +1212,29 @@ public:
             } else {
                 throw SemanticError("chr() must have one integer argument", x.base.base.loc);
             }
+        } else if (call_name == "complex") {
+            LFORTRAN_ASSERT(ASRUtils::all_args_evaluated(args));
+            int16_t n_args = args.size();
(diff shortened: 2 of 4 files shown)


certik:
Do you mind adding the test case into `doconcurrentloop_01.py`? That way it will get tested by the CI.

src/lfortran/semantics/python_ast_to_asr.cpp
```diff
@@ -1212,6 +1212,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            } else if (n_args == 2) {
+                ASR::expr_t* expr1 = args[0];
+                ASR::expr_t* expr2 = args[1];
+                double c1 = ASR::down_cast<ASR::ConstantReal_t>(ASRUtils::expr_value(expr1))->m_r;
+                double c2 = ASR::down_cast<ASR::ConstantReal_t>(ASRUtils::expr_value(expr2))->m_r;
+                tmp = ASR::make_ConstantComplex_t(al, x.base.base.loc, c1, c2, type);
+                return;
+            } else {
+                throw SemanticError("complex() must have at most two real arguments", x.base.base.loc);
+            }
```
czgdp1807, line 1237:
```suggestion
            if( n_args > 2 || n_args < 0 ) { // n_args shouldn't be less than 0 but added this check for safety
                throw SemanticError("complex() must have at most two real arguments", x.base.base.loc);
            }

            double c1 = 0.0, c2 = 0.0; // Default if n_args = 0
            if (n_args >= 1) { // Handles both n_args = 1 and n_args = 2
                c1 = ASR::down_cast<ASR::ConstantReal_t>(ASRUtils::expr_value(args[0]))->m_r;
            }

            if (n_args == 2) { // Extracts imaginary component if n_args = 2
                double c2 = ASR::down_cast<ASR::ConstantReal_t>(ASRUtils::expr_value(args[1]))->m_r;
            }

            tmp = ASR::make_ConstantComplex_t(al, x.base.base.loc, c1, c2, type);
```

czgdp1807:
Was going through the PRs and left the above suggestion.

czgdp1807:
Just a minor question, Is it legal to do something like the following,
```
r = 1
i = 2
c = complex(r, i)
```
In words, I mean passing variables as arguments to `complex`?

Thirumalai-Shaktivel:
This is not mentioned in the [Docs](https://docs.python.org/3/library/functions.html#complex), But the interactive python shell accepts it!
```
Python 3.8.5 (default, Sep  4 2020, 07:30:14) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
>>> r = 1
>>> i = 2
>>> c = complex(r, i)
>>> c
(1+2j)
>>> type(c)
<class 'complex'>
```
which means it is legal to pass the variables as parameters.

src/lfortran/semantics/python_ast_to_asr.cpp
```diff
@@ -1212,6 +1212,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             } else {
                 throw SemanticError("chr() must have one integer argument", x.base.base.loc);
             }
+        } else if (call_name == "complex") {
+            LFORTRAN_ASSERT(ASRUtils::all_args_evaluated(args));
```
czgdp1807, line 1216: I think that if variables are allowed then this check should be removed as this accepts only those arguments which are constants or can be evaluated to a constant at compile time.

Thirumalai-Shaktivel: Yup, I agree with you! 
But, I think naman is planning to support only constant values for now. (I don't know)

Thirumalai-Shaktivel: In that case, we can add a comment saying "Only constant numeric values as parameter are supported for now". Iteratively we can support for both strings(`complex('1+2j')` or `complex('1')`) and variables (`complex(x)`)

namannimmo10: Thanks.  

> I think that if variables are allowed then this check should be removed as this accepts only those arguments which are constants or can be evaluated to a constant at compile time.

Yes, this should be done. I guess we also need to modify other intrinsic functions in this file (Python AST->ASR), like `chr` and `ord`, in order to make support for variables.

certik (approved):
I think this is fine. This is a compile time function, which can be used to construct complex numbers. Once we have a runtime library working, we can add a runtime implementation.

Smit-create:
The CI passes for this PR. I'm just re-running the jobs after we have a few updates in CI to verify if that fails here.

Thirumalai-Shaktivel:
This PR is not updated (I mean, not pushed) since PR #33  is merged, that's why it uses the old CI...

certik:
What is the status of this PR --- can it go in?

namannimmo10:
I think this is ready once the build is a success. We can extend this later.

doconcurrentloop_01.py
```diff
@@ -1,3 +1,21 @@
...
+    # constant real or int as args
+    c = complex()
+    c = complex(3.4)
+    c = complex(5., 4.3)
+    c = complex(1)
+    c1 = complex(3, 4)
+    c2 = complex(2, 4.5)
+    c3 = complex(3., 4.)
+    b = c1 != c2
+    b = c1 == c3
```
certik, line 17: Add these changes to `lpython_tests.py`, after #37 is merged. Then it should work.

Issue #21, opened 2022-01-14
CI: setup caching


certik:
Here is how to do it: https://github.com/epassaro/cache-conda-envs

Smit-create:
Working on it!

PR #33, opened 2022-01-16
CI: Enable caching

Smit-create:
Related issue: #21

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,4 +1,4 @@
-name: LPython CI
+name: CI
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -9,27 +9,59 @@ on:
...
           auto-activate-base: false
 
+      - name: Set cache date
+        run: echo "DATE=$(date +'%Y%m%d')" >> $GITHUB_ENV
+
+      - uses: actions/cache@v2
+        with:
+          path: ${{ matrix.prefix }}
+          key: ${{ matrix.label }}-conda-${{ hashFiles('ci/environment.yml') }}-${{ env.DATE }}-${{ env.CACHE_NUMBER }}
+        id: cache
```
Smit-create, line 59: @certik  Here, the cache will be used for 24 hours. More clearly, the cache of each day will be updated on the first run and can be used for further runs on the same day.

Smit-create: or may be we can change to all-time cache with no date:
```
key:
          ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{ hashFiles('ci/environment.yml') }}
```

certik:
I think that works, thanks! Do you know what happens when we change the Conda dependencies, say in the `environment.yml` file? Will it automatically rebuild it, or do we have to trigger it manually somehow to update the cache?

Smit-create:
> Will it automatically rebuild it,

I will automatically update accordingly.

certik:
I essentially reverted this in #39. We'll have to try again on this.

PR #34, opened 2022-01-17
Debugging CI

Smit-create:


Smit-create:
ubuntu and mac are working fine.

.github/workflows/CI.yml
```diff
@@ -30,10 +30,10 @@ jobs:
...
             python-version: "3.10"
 
-          - os: windows-latest
-            label: win-64
-            prefix: C:\Miniconda3\envs\test
-            python-version: "3.10"
+          # - os: windows-latest
+          #   label: win-64
+          #   prefix: C:\Miniconda3\envs\test
+          #   python-version: "3.10"
```
Thirumalai-Shaktivel, line 37: But How?? These changes are unrelated, right!!

Smit-create: Yeah, just debugging which platforms fails.

ci/environment.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
   - toml
   - pytest
   - jupyter
-  - xeus=1.0.1
+  - xeus
```
Smit-create, line 10: Is it necessary to pin?

Thirumalai-Shaktivel: Yes, I believe... The modules or functions used by LFortran might be taken from the xeus 1.0.1 version.

Smit-create:
Now, Linux was working fine but some issue with mac failed the whole job. Using 3 scripts for 3 platforms should be good going as they won't affect each other's failure and might help to fix platform-specific issues. @certik  What do you think? Caching was working fine too, I checked it.

Thirumalai-Shaktivel:
Hi @Smit-create. If you don't mind, Can I work on this PR??

Smit-create:
Sure, you can!

Smit-create:
Till you debug here, I'm working parallelly on separating 3 ci files.

.github/workflows/CI.yml
```diff
@@ -42,25 +36,27 @@ jobs:
...
+        env:
+          CACHE_NUMBER: 0
         with:
           path: ${{ matrix.prefix }}
           key: ${{ matrix.label }}-conda-${{ hashFiles('ci/environment.yml') }}-${{ env.DATE }}-${{ env.CACHE_NUMBER }}
         id: cache
 
       - name: Update environment
-        run: conda env update -n test -f ci/environment.yml
-        if: steps.cache.outputs.cache-hit != 'true'
```
Thirumalai-Shaktivel, line 63: I believe the issue was with this `if` statement

Smit-create:
Thanks @Thirumalai-Shaktivel. I think we can merge this. Separating each OS in the different files can be discussed with @certik

.github/workflows/CI.yml
Smit-create: Yeah, I think that too. I removed it in the other PR

certik:
I essentially reverted this in #39. We'll have to try again on this.

PR #35, opened 2022-01-17
Separating CI for each OS

Smit-create:


.github/workflows/linux.yml
```diff
@@ -0,0 +1,71 @@
...
+env:
+  # Increase this value to reset cache if ci/environment.yml has not changed
+  CACHE_NUMBER: 1
+
+
+jobs:
+  Build:
+    strategy:
+      matrix:
+        include:
```
Thirumalai-Shaktivel, line 21: If we test only one OS, then there is no need for a matrix. Instead, we can use like this: https://github.com/marketplace/actions/setup-miniconda#caching-packages

Smit-create: I'll do those clean-ups later on. Trying to get it run once.

Thirumalai-Shaktivel:
> Linux CI works fine with caching, notice the time difference: https://github.com/lcompilers/lpython/runs/4841084418?check_suite_focus=true#logs and https://github.com/lcompilers/lpython/runs/4840593367?check_suite_focus=true

I think time can still be reduced by using `mamba` instead of `conda`...

Thirumalai-Shaktivel:
Hey @Smit-create, Debugging work has been completed in that PR. If you need any help with this PR, please let me know...

Smit-create:
All CIs go green!!!

Smit-create:
> It seems separating the CI makes things more complicated. I would prefer to consolidate things in such a way that we can use exactly (or almost exactly) the same setup on all platforms.

Yeah, I agree with this. The only problem with this is that if there is some issue with one OS, it will cancel the others too.

certik:
I am going to close this for now, but we can always resurrect it. I would suggest to keep the CI in one file and as simple as possible. Yes, we should switch to mamba, that should be safe. We should experiment with caching, but we have to be very careful to ensure things continue working, it's actually no easy to do that: one must test that things properly update when we change the configuration.

PR #39, opened 2022-01-17
CI: Revert to a version that works

certik:


certik:
I don't know what the issue is, but this PR seems to work. One possibility is that the caching cached the old "good" version, and so we did not notice right away when we changed the CI scripts (to a broken state), until some time after when the cache tried to rebuild and failed.

I would suggest to first convert to mamba, but do not use any cache, as it seems to be fragile.

certik:
Ok, so this seems to work, so I'll merge it so that we have a CI that works and we can keep working on other PRs. And then in the meantime we need to re-enable the improvements, but in some way that will be more robust.

PR #47, opened 2022-01-19
Rename lfortran folder & cpp file to lpython

Thirumalai-Shaktivel:


certik (approved):
I think that looks good, thanks for this!

PR #51, opened 2022-01-20
Move the Python parser into the runtime library

certik:
lpython now works from any directory.


Smit-create:
Hi,
I built on the current master and I get the following error:
```
$ lpython --show-asr expr3.py 
python: can't open file '/home/admin-pc/Smitlunagariya/lpython/inst/bin/../share/lpython/lib/lpython_parser.py': [Errno 2] No such file or directory
The command 'python /home/admin-pc/Smitlunagariya/lpython/inst/bin/../share/lpython/lib/lpython_parser.py expr3.py' failed.
```
Is this related?

certik:
It might be related. Try to rebuild from scratch, if it is not fixed, please open up a new issue.

PR #53, opened 2022-01-21
Compute the result of the binary operations

namannimmo10:


tests/reference/llvm-expr_01-54467c1.json
```diff
@@ -2,12 +2,12 @@
...
     "outfile_hash": null,
     "stdout": "llvm-expr_01-54467c1.stdout",
-    "stdout_hash": "cedd6bf3b0932220945df1b9405f9f7263ed61f134621c73bc08c124",
-    "stderr": null,
-    "stderr_hash": null,
-    "returncode": 0
+    "stdout_hash": "b8199ba2afc838ae6571fb8d46fe9ebbec96c8e58e3f057b67364ca7",
+    "stderr": "llvm-expr_01-54467c1.stderr",
+    "stderr_hash": "347f1bde6e1f1bc4308747ce8f6d07d8c69e9ee0e23f4127d82d7fc4",
+    "returncode": 3
```
certik, line 12: There is an error message that is being printed for `--show-llvm` possibly this is not implemented yet in ASR to LLVM. You can simply create a new test for these complex numbers and simply only test ASR, but not LLVM in this new test. We'll fix LLVM soon, but for now that's how you can get things going.

certik: However, it's very concerning that tests pass in this PR at the CI. They should fail, because the file `llvm-expr_01-54467c1.stderr` is not committed...

certik: I tested your PR locally, indeed tests pass even without the `.stderr` file. I think that's a bug in our tester, although maybe I did it so that you don't actually have to have the file committed, as long as the hash matches with the `.stderr` file that is generated.

certik: I see, even the `stdout` file doesn't have to be there. It seems the only thing that matters is the json file. The stdout and stderr files are for us, to know what files are corresponding to the hashes, so that we can see what has changed. Ok, so no bug.

certik: I did this so that we can also check binary files, without committing the binary files into the git.

namannimmo10: > Ok, so no bug. It seems the only thing that matters is the json file

Yes, I agree. Let's add this test to a new file and test only the AST and ASR for now.

namannimmo10:
Will raise another PR soon to include the implementation for string repetition. Thanks for the review.

certik:
+1

PR #57, opened 2022-01-22
Fixes `range` not to include upper limit

### LINKED ISSUE #54: Bug in `range`
Reference Code:
```py
def test():
    i: i32
    a : i32
    a = 0
    for i in range(1, 10):
        a += i
```

<details>
<summary> ASR </summary>
<p>

```bash
(TranslationUnit 
   (SymbolTable 
      1 
      {
         test: 
            (Subroutine 
               (SymbolTable 
                  2 
                  {
                     a: 
                        (Variable 
                           2 
                           a 
                           Local () () 
                           Default 
                           (Integer 4 []) 
                           Source 
                           Public 
                           Required .false.), 
                     i: 
                        (Variable 
                           2 
                           i 
                           Local () () 
                           Default 
                           (Integer 4 []) 
                           Source 
                           Public 
                           Required .false.)
                  }) 
               test [] [
               (= 
                  (Var 2 a) (ConstantInteger 0 
                  (Integer 4 [])) ()) 
               (DoLoop 
                  (
                     (Var 2 i) (ConstantInteger 1 
                     (Integer 4 [])) (ConstantInteger 10 
                     (Integer 4 [])) (ConstantInteger 1 
                     (Integer 4 []))) [
                  (= 
                     (Var 2 a) 
                     (BinOp 
                        (Var 2 a) 
                        Add 
                        (Var 2 i) 
                        (Integer 4 []) () ()) ())])] 
               Source 
               Public 
      }) 
   [])

```
</p>
</details>

<details>
<summary> CPP </summary>
<p>

```cpp
#include <iostream>
#include <string>
#include <vector>
#include <Kokkos_Core.hpp>

template <typename T>
{
        r(i) = v[i];
    }
    return r;
}

void test()
{
    int a;
    int i;
    a = 0;
    for (i=1; i<=10; i++) {
        a = a + i;
    };
}
```

</p>
</details>

So the issue is `range(1, 10)` considers it as `[1, 10]` instead of `[1, 10)`

Smit-create:
Fixes #54


certik (approved):
I think that looks great! Thanks.

certik:
The failure seems like the tests were not updated? Not sure right now.

Smit-create:
Thanks for the review. Tests now pass after rebasing

PR #58, opened 2022-01-22
Implement Function in ASR

Smit-create:
Discussions in: #9


Smit-create:
Presently, function do work at ASR level but with a few bugs:
On this reference code:
```py
def test(x: i32) -> i32:
    a: i32
    a = 1
    return a
```
We have ASR:
```
(TranslationUnit 
   (SymbolTable 
      1 
      {
         test: 
            (Function 
               (SymbolTable 
                  2 
                  {
                     a: 
                        (Variable 
                           2 
                           a 
                           Local () () 
                           Default 
                           (Integer 4 []) 
                           Source 
                           Public 
                           Required .false.), 
                     i32: 
                        (Variable 
                           2 
                           i32 
                           ReturnVar () () 
                           Default 
                           (Integer 4 []) 
                           Source 
                           Public 
                           Required .false.), 
                     x: 
                        (Variable 
                           2 
                           x 
                           In () () 
                           Default 
                           (Integer 4 []) 
                           Source 
                           Public 
                           Required .false.)
                  }) 
               test [
               (Var 2 x)] [
               (= 
                  (Var 2 a) (ConstantInteger 1 
                  (Integer 4 [])) ()) 
               (Return)] 
               (Var 2 i32) 
               Source 
               Public 
               Implementation ())
      }) 
   [])
```
The bug will be more clear in CPP code below:
```cpp
#include <iostream>
#include <string>
#include <vector>
#include <Kokkos_Core.hpp>

template <typename T>
{
        r(i) = v[i];
    }
    return r;
}

int test(int x)
{
    int a;
    int i32;
    a = 1;
    return;
    return i32;
}

````

**Comment by Smit-create** [comment]
I think the double return appears because of function_t and return_t both are visited

File: src/lpython/semantics/python_ast_to_asr.cpp [hunk 1]
```diff
@@ -255,33 +268,45 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                if(current_scope->scope.find(return_var_name) == current_scope->scope.end()) {
+                    ASR::ttype_t *type = ast_expr_to_asr_type(x.m_returns->base.loc, *x.m_returns);
+                    ASR::asr_t *return_var = ASR::make_Variable_t(al, x.m_returns->base.loc,
+                        current_scope, s2c(al, return_var_name), ASRUtils::intent_return_var, nullptr, nullptr,
+                        ASR::storage_typeType::Default, type,
+                        current_procedure_abi_type, ASR::Public, ASR::presenceType::Required,
+                        false);
+                    current_scope->scope[return_var_name]
+                             = ASR::down_cast<ASR::symbol_t>(return_var);
+                }
```
Smit-create, line 290: I fixed the double `return` issue. This code also correctly finds the return variable name. Now the issue is if we create the return variable while visiting the `FunctionDef`, there will be a re-declaration error in `AnnAssign`. Any suggestions to tackle this?

certik:
I need to review this carefully. Why do tests fail?

Smit-create:
> Why do tests fail?

I have canceled them because the tests will fail since the code isn't perfect yet and fails on example in OP

certik:
Cool. I marked it as Draft. Once this is ready for review, just remove the `Draft: ` prefix from the PR name.

Smit-create:
> Once this is ready for review, just remove the `Draft: ` prefix from the PR name.

Okay, thanks!
Any alternatives/suggestions on [this](https://github.com/lcompilers/lpython/pull/58#discussion_r790682580)?

certik:
So I think one clean way to do it is as follows:

* In the symbol table visitor, see the function signature and create a return variable and just name it with some unique name and the type from the Python annotation
* In the body visitor, when you visit the return statement:
  * Check that it returns an expression and the expression has the same type as the return variable, otherwise give an error message
  * Assign the expression (which can be an actual expression like `1+2` or another variable like `a`) into the return variable
  * Return

Smit-create:
It's working fine now.

Python code:
```py
def test(x: i32) -> i32:
    a: i32
    a = 1
    return a

```
CPP code:

```cpp
#include <iostream>
#include <string>
#include <vector>
#include <Kokkos_Core.hpp>

template <typename T>
{
        r(i) = v[i];
    }
    return r;
}

int test(int x)
{
    int _lpython_return_variable;
    int a;
    a = 1;
    _lpython_return_variable = a;
}

```

tests/expr_01.py
```diff
@@ -1,7 +1,8 @@
-def main():
+def main() -> i32:
```
certik, line 1: Should main return i32 or i64? I actually don't know.

tests/reference/cpp-loop1-0a8cf3b.stdout
```diff
@@ -14,28 +14,32 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
...
 }
 
-void test_factorial_1(int x)
+int test_factorial_1(int x)
 {
+    int _lpython_return_variable;
     int result;
     if (x < 0) {
-        return;
+        _lpython_return_variable = 0;
```
certik, line 22: This should first assign to `_lpython_return_variable`, which it does. But then it needs to actually return it using `return _lpython_return_variable`.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1282,7 +1265,25 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+        if (!ASRUtils::check_equal_type(target_type, value_type)) {
+            std::string ltype = ASRUtils::type_to_str(target_type);
+            std::string rtype = ASRUtils::type_to_str(value_type);
+            throw SemanticError("Type Mismatch in return, found (" +
+                    ltype + " and " + rtype + ")", x.base.base.loc);
+        }
+
+        ASR::stmt_t *overloaded=nullptr;
+        tmp = ASR::make_Assignment_t(al, x.base.base.loc, target, value,
+                                overloaded);
```
certik, line 1286: Everything looks correct to me in this PR, except here after the assignment we also have to emit Return in ASR, see my other comment in the C++ generated code.

The question becomes how to do that, since our visitor is assuming that every AST node transforms to 0 or 1 ASR nodes, but here we need to emit 2 ASR nodes. I think that the cleanest way is to save the `body` Vector as an internal variable, and then insert the instruction from here.

certik (changes requested):
Looks very good and clean, just one last thing to fix.

certik:
I pushed in the main fix, but I noticed some tests now `return` twice, so I am looking into it. The main bug should be fixed now though.

certik:
I think it's fixed now.

tests/reference/cpp-expr9-48868e9.stdout
```diff
@@ -0,0 +1,45 @@
...
+    return _lpython_return_variable;
+}
+
+int test_return_3(int a)
+{
+    int _lpython_return_variable;
+    a = 3;
+    _lpython_return_variable = a;
+    return _lpython_return_variable;
+    return _lpython_return_variable;
```
namannimmo10, line 43: Why does it return twice?

tests/expr_01.py
certik: @Thirumalai-Shaktivel said that it should return i32, so this should be correct.

tests/reference/cpp-loop1-0a8cf3b.stdout
certik: That's now fixed.

src/lpython/semantics/python_ast_to_asr.cpp
certik: Fixed now.

certik (approved):
I think this is good to merge if tests pass.

There is still one issue left: I think just a `return` from a function that returns `None` does not work. It should translate to just `Return` ASR node, with no assignment, inside a subroutine. If you want, you can implement that in a subsequent PR.

certik:
All tests pass. Thanks!

Smit-create:
Thanks for updating and merging the PR.

PR #59, opened 2022-01-22
Compute the result of string repetition

namannimmo10:


namannimmo10:
So far so good... Let's try to refactor it.

Smit-create (approved):

namannimmo10:
Thanks, @Smit-create for having a look.  @certik, can you also take a quick look so that this can be merged?

certik (approved):
I think that looks good. The same comment as in the previous PR applies: we will have to improve it in the future to represent the operation exactly as written, but providing compile time value if available as an option.

PR #61, opened 2022-01-22
Implement `abs` intrinsic function

namannimmo10:


certik (approved):
This is fine for now. Just remember this is only the compile time optimization. We will keep this even after we implement an actual `abs` function in the runtime library, and the compile time value will be provided in the `value` member.

Also, the current implementation is replacing `abs` with the compile time value. Ok for now, but the correct solution is to represent it using the `abs` runtime function, and then provide compile time evaluation (if available) in the `value` member of the `FunctionCall` node.

certik:
Also, would you please mind sending it from a branch in your fork, not in the `lcompilers/lpython` repository? That way we only have the `main` branch in the main repository.

namannimmo10:
Sure, I will make sure that will not happen again.

PR #63, opened 2022-01-23
Implement `visit_Dict`

namannimmo10:


grammar/ASR.asdl
```diff
@@ -200,6 +200,7 @@ expr
     | NamedExpr(expr target, expr value, ttype type)
     | Compare(expr left, cmpop op, expr right, ttype type, expr? value, expr? overloaded)
     | IfExp(expr test, expr body, expr orelse, ttype type)
+    | Dict(expr* keys, expr* values, ttype type)
```
certik, line 203: I think this has to have a type for "keys" and a separate type for "values". Also this has to be a type, not an expression.

Summary: add a new type "Dictionary" that has a type for keys and another type for values.

Then add a new `expr` node called `ConstantDictionary` which represents a constant (compile time) dictionary (and of type `Dictionary`), just like `ConstantArray` or `ConstantInteger`.

PR #64, opened 2022-01-24
Make `len` work for strings

namannimmo10:


certik (approved):
I think this looks good. The usual caveat that we actually have to represent this as a function call and only provide a compile time evaluation if available in the `value` member. But for now it's good.

PR #66, opened 2022-01-24
Allow negative integers in `bin()`, `hex()`, and `oct()`

namannimmo10:


certik:
Thanks for the PR. I think that looks good. Does it give the same answers as Python?

Next time, if you wouldn't mind, let's try to review each PR before merging. I'll try to review quickly.

namannimmo10:
Yes, it gives the same output as Python.

> Next time, if you wouldn't mind, let's try to review each PR before merging. I'll try to review quickly.

Sure. This patch was trivial, so I pushed it.  Thanks for letting me know!

PR #70, opened 2022-01-25
Implement `not` unary operator for boolean types

namannimmo10:
Reference code:
```python
def test_UnaryOp():
    a: i32
    a = +4
    a = -500

    f: f32
    f = +1.0
    f = -183745.534

    b1: bool
    b2: bool
    b1 = True
    b2 = not False
```

ASR:
```bash
(TranslationUnit
   (SymbolTable
      1
      {
         test_UnaryOp:
            (Subroutine
               (SymbolTable
                  2
                  {
                     a:
                        (Variable
                           2
                           a
                           Local () ()
                           Default
                           (Integer 4 [])
                           Source
                           Public
                           Required .false.),
                     b1:
                        (Variable
                           2
                           b1
                           Local () ()
                           Default
                           (Logical 1 [])
                           Source
                           Public
                           Required .false.),
                     b2:
                        (Variable
                           2
                           b2
                           Local () ()
                           Default
                           (Logical 1 [])
                           Source
                           Public
                           Required .false.),
                     f:
                        (Variable
                           2
                           f
                           Local () ()
                           Default
                           (Real 4 [])
                           Source
                           Public
                           Required .false.)
                  })
               test_UnaryOp [] [
               (=
                  (Var 2 a) (ConstantInteger 4
                  (Integer 4 [])) ())
               (=
                  (Var 2 a) (ConstantInteger -500
                  (Integer 4 [])) ())
               (=
                  (Var 2 f)
                  (ConstantReal 1.000000
                     (Real 8 [])) ())
               (=
                  (Var 2 f)
                  (ConstantReal -183745.534000
                     (Real 8 [])) ())
               (=
                  (Var 2 b1)
                  (ConstantLogical .true.
                     (Logical 1 [])) ())
               (=
                  (Var 2 b2)
                  (UnaryOp
                     Not
                     (ConstantLogical .false.
                        (Logical 1 []))
                     (Logical 1 [])
                     (ConstantLogical .true.
                        (Logical 1 []))) ())]
               Source
               Public
      })
   [])
```

CPP codegen:
```c++
#include <iostream>
#include <string>
#include <vector>
#include <Kokkos_Core.hpp>

template <typename T>
{
        r(i) = v[i];
    }
    return r;
}

void test_UnaryOp()
{
    int a;
    bool b1;
    bool b2;
    float f;
    a = 4;
    a = -500;
    f = 1.000000;
    f = -183745.534000;
    b1 = true;
    b2 = !false;
}

```


tests/expr10.py
```diff
@@ -0,0 +1,13 @@
...
+    a = -500
+
+    f: f32
+    f = +1.0
+    f = -183745.534
+
+    b1: bool
+    b2: bool
+    b1 = True
+    b2 = not False
```
certik, line 13: Can you add a test here:
```
b3: bool
b3 = not b2
```
I think you'll get an assert error, because you assume that the operand is a `ConstantLogical`.

Instead, `not` must produce the `UnaryOp` ASR operation. In addition, if there is a compile time value, it should be assign to the `value` member. If there is no compile time value, the `value` should be `nullptr`.

tests/reference/cpp-expr10-76fb526.stdout
```diff
@@ -0,0 +1,30 @@
...
+    int a;
+    bool b1;
+    bool b2;
+    float f;
+    a = 4;
+    a = -500;
+    f = 1.000000;
+    f = -183745.534000;
+    b1 = true;
+    b2 = !false;
```
namannimmo10, line 28: I'm gonna explore more of the c++ codegen part for some time, from now on. This can help me give a break from Python AST->ASR. :)
Let me know if this looks good.

certik: I am unsure, as it seems the line `result = !op_value;` is not evaluated in this case. It seems it is actually represented as UnaryOp, as it should be. But your code in this PR seems to instead return the actual answer so not `!false`, but `true`.

namannimmo10: > I am unsure, as it seems the line `result = !op_value;` is not evaluated in this case. It seems it is actually represented as UnaryOp, as it should be. But your code in this PR seems to instead return the actual answer so not `!false`, but `true`.

That line, `result = !op_value;`, is being evaluated since here is the node that was created:
```
               (=
                  (Var 2 b2)
                  (UnaryOp
                     Not
                     (ConstantLogical .false.
                        (Logical 1 []))
                     (Logical 1 [])
                     (ConstantLogical .true.
                        (Logical 1 []))) ())]
```
With `.true.` passed in as the value member. But I'm not sure what we want the cpp code to produce: `true` or `!false`.

tests/expr10.py
namannimmo10: > I think you'll get an assert error because you assume that the operand is a `ConstantLogical`

It does produce an ASR node:
```bash
               (=
                  (Var 2 b3)
                  (UnaryOp
                     Not
                     (Var 2 b2)
                     (Logical 1 []) ()) ())]
```
But no value member added in `tmp` since we assume operand is `ConstantLogical`. So, no assert error.

namannimmo10: > Can you add a test here

Done.

tests/reference/cpp-expr10-76fb526.stdout
namannimmo10: Am I missing something?

certik: Ah, so it is already doing what it should! Ok, that's fine. The C++ code should return `!false`. The backend returns the full expression by default.

certik (approved):
Ok, I think this works well. Sorry about my comment, I think it already does what I was suggesting.

namannimmo10:
No problem, thanks for the review.

PR #73, opened 2022-01-26
Fix double return in cpp

Smit-create:
Python code:
```py
def test_factorial_1(x: i32) -> i32:
    if x < 0:
        return 0
    result: i32
    result = 1
    while x > 0:
        result = result * x
        x -= 1
    return result
```

CPP code:

```cpp
#include <iostream>
#include <string>
#include <vector>
#include <Kokkos_Core.hpp>

template <typename T>
{
        r(i) = v[i];
    }
    return r;
}

int test_factorial_1(int x)
{
    int _lpython_return_variable;
    int result;
    if (x < 0) {
        _lpython_return_variable = 0;
    };
    result = 1;
    while (x > 0) {
        result = result*x;
        x = x - 1;
    };
}

```

### DIFF
--- a/src/libasr/codegen/asr_to_cpp.cpp
+++ b/src/libasr/codegen/asr_to_cpp.cpp
@@ -447,10 +447,6 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
         }
         current_function = nullptr;
 
-        body += indent + "return "
-            + LFortran::ASRUtils::EXPR2VAR(x.m_return_var)->m_name
-            + ";\n";
-
         if (decl.size() > 0 || body.size() > 0) {
             sub += "{\n" + decl + body + "}\n";
         } else {
--- a/src/libasr/codegen/asr_to_cpp.cpp
+++ b/src/libasr/codegen/asr_to_cpp.cpp
@@ -441,11 +441,20 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
 
         current_function = &x;
         std::string body;
+        bool visited_return = false;
+
         for (size_t i=0; i<x.n_body; i++) {
             this->visit_stmt(*x.m_body[i]);
+            visited_return = visited_return || (is_a<ASR::Return_t>(*x.m_body[i]));
             body += src;
         }
         current_function = nullptr;
+
+        if(!visited_return) {
+            body += indent + "return "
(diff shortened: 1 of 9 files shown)


Smit-create:
Trying to address: https://github.com/lcompilers/lpython/pull/58#discussion_r791938280

certik:
This works in LPython, but will not work in LFortran; in ASR currently the way functions are designed are that you save the result into a variable, and a function get finish without ever calling "return". So this PR will stop working for such functions.

I thought about this design today, if we really need it, but I think we do: it allows to return an array, that you can set using a for loop, and you don't have to copy it on the "return" line; this in return makes it easy to have an ASR pass that transforms functions returning an array into a subroutine (with `intent(inout)` argument), which is very efficient: no costly array copy being done. If we follow the usual "return A" approach (with `A` being an array), it seems it would be much harder for functions with multiple such `return` statements to convert them to a subroutine. So I think the current design is actually not bad.

But how to fix this issue? Well, I can see two approaches forward:

* Always require for a function to end with a "return", and check it in the ASR verify pass. Then your PR would work.
* In the C++ backend check if the function exited with a "return", if so, do nothing (as in this PR), and if not, then add the return statement.

It almost seems the first approach is easier on the backends. Let me think about this.

src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -441,15 +441,20 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
...
         current_function = &x;
         std::string body;
+        bool visited_return = false;
+
         for (size_t i=0; i<x.n_body; i++) {
             this->visit_stmt(*x.m_body[i]);
+            visited_return = visited_return || (is_a<ASR::Return_t>(*x.m_body[i]));
             body += src;
         }
         current_function = nullptr;
```
certik, line 451: What if there are some statements after the return, and reachable via goto? Not recommended, but possible.

How about this:
```
bool visited_return = false;
if (x.n_body > 0 && is_a<ASR::Return_t>(*x.m_body[x.n_body-1])) visited_return = true;
```
That way if the last statement was a return, then we don't repeat it, otherwise we add a return.

Smit-create: That sounds good.

certik (approved):
I think that works!

PR #74, opened 2022-01-26
Implement `visit_Assert` codegen for cpp

Smit-create:


namannimmo10 (approved):
Looks great, thanks!

certik:
I think you have to also include `cassert`, according to: https://en.cppreference.com/w/cpp/error/assert

certik (approved):
I think that will work. Did you test if it compiles? If it does, then you can merge it after tests pass.

certik:
I guess you should merge the latest main and regenerate tests.

Smit-create:
Thanks for the reviews. Merging!!

PR #75, opened 2022-01-26
Handle casting in cpp

Smit-create:
We need Integer to real cast in cpp because of this failing example:
On python code:
```py
def test_cast():
    b: f32
    b = 3/4
    print(b) # b= 0.75
```
Presently, we get the following cpp(without handling cast in cpp) which is not correct:
```cpp
void test_cast()
{
    float b;
    b = 3/4;
    std::cout << b << std::endl; // this will print 0
}
```


certik (approved):
Excellent. Later, we can implement more complicated solution that checks what the operations are around it, to see if things get casted automatically, so for example Python's `3/4` would get translated to `(float)(3) / 4` instead of the current `(float)(3) / (float)(4)`. We can even check the integer, and if it is a constant, then we can translate to `3. / 4`.

But for now the current solution is the simplest to implement, and it works.

PR #76, opened 2022-01-26
Generate c++ code for complex numbers

namannimmo10:


certik:
Does the generated C++ code compile?

tests/reference/cpp-complex2-cc31ff8.stdout
```diff
@@ -0,0 +1,23 @@
...
+        r(i) = v[i];
+    }
+    return r;
+}
+
+void test_complex_cpp()
+{
+    std::complex<float> c;
+    c = {423.000000, -23423.540000};
+    c = {1.000000, 2.000000};
```
namannimmo10, line 21: This generated portion does compile, and I check that separately.

certik (approved):

tests/reference/cpp-complex2-cc31ff8.stdout
certik: Perfect!

certik:
See also #77, which should make it much easier to test (we should allow generating code that only requires standard C++ compiler no 3rd party dependencies).

PR #79, opened 2022-01-27
Improve c++ codegen

namannimmo10:


src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -759,15 +760,15 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
...
         last_unary_plus = false;
     }
 
     void visit_BoolOp(const ASR::BoolOp_t &x) {
         this->visit_expr(*x.m_left);
-        std::string left_val = "(" + src + ")";
+        std::string left_val = src;
         this->visit_expr(*x.m_right);
-        std::string right_val = "(" + src + ")";
+        std::string right_val = src;
```
certik, line 771: I think you have to keep the parentheses there, in case the `src` is a more complicated expression with a lower precedence than the current operation.

We have to implement proper optional parentheses like in LFortran's AST to SRC printer, using precedence. Until then I think the only correct (robust) solution is to put parentheses everywhere, otherwise there is always some corner case that is incorrect.

namannimmo10: > We have to implement proper optional parentheses like in LFortran's AST to SRC printer, using precedence.

I see. I will do that soon.

certik: You can do exactly what we did in LFortran. It seems it's a very simple and yet correct way to do it.

src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -759,7 +760,15 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
...
         this->visit_expr(*x.m_right);
         std::string right_val = src;
-        src = "(" + left_val + ") + (" + right_val + ")";
+        switch (x.m_op) {
+            case (ASR::stropType::Concat): {
+                src = "std::string(" + left_val + ") + std::string(" + right_val + ")";
+                break;
+            }
+            case (ASR::stropType::Repeat): {
+                // TODO: implement
```
namannimmo10, line 769: There's no such direct idiomatic way to repeat a string in C++ equivalent to the `*` operator we have in Python.
What I did for the implementation of string repetition (Python frontend):
```c++
                std::ostringstream os;
                std::fill_n(std::ostream_iterator<std::string>(os), repeat, std::string(str));
                std::string result_s = os.str();
```
Should we generate something like this now? Any other ideas are appreciated! @certik @Smit-create

certik: I think we need a "runtime" library in C++ that this backend then uses. It would be a header file and a cpp file. Some of the Kokkos conversion functions would go there too, that we currently generate at the top of the file.

namannimmo10: Good idea. Let's do that once we get the basic functionality up and running.

namannimmo10:
This is ready for now. Will try to address https://github.com/lcompilers/lpython/pull/79#discussion_r793827132 in a separate PR.

certik (approved):

PR #80, opened 2022-01-27
Improve `UnaryOp` visitor to support operation on complex

namannimmo10:
Reference code:
```python
def test_UnaryOp():
    c: c128
    c = +complex(1, 2)
    c = -complex(3, -4.0)

```

ASR:
```stdout
(TranslationUnit
   (SymbolTable
      1
      {
         test_UnaryOp:
            (Subroutine
               (SymbolTable
                  2
                  {
                     c:
                        (Variable
                           2
                           c
                           Local () ()
                           Default
                           (Complex 8 [])
                           Source
                           Public
                           Required .false.)
                  })
               test_UnaryOp [] [
               (=
                  (Var 2 c)
                  (UnaryOp
                     UAdd
                     (ConstantComplex 1.000000 2.000000
                        (Complex 8 []))
                     (Complex 8 [])
                     (ConstantComplex 1.000000 2.000000
                        (Complex 8 []))) ())
               (=
                  (Var 2 c)
                  (UnaryOp
                     USub
                     (ConstantComplex 3.000000 -4.000000
                        (Complex 8 []))
                     (Complex 8 [])
                     (ConstantComplex -3.000000 4.000000
                        (Complex 8 []))) ())]
               Source
               Public
      })
   [])
```


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1030,7 +1031,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     case (ASR::unaryopType::UAdd): { result = op_value; break; }
                     case (ASR::unaryopType::USub): { result = -op_value; break; }
                     default: {
-                        throw SemanticError("Unary operator not implemented yet for compile time evaluation",
+                        throw SemanticError("Bad operand type for unary operation",
```
namannimmo10, line 1034: Also: I'm trying to improve the error messages. I just don't see a way to convert `ttypeType` to string, otherwise, we would have nice and self-explaining messages, something like this:
> "Bad operand type for unary operation ~: real"

namannimmo10: Okay, I found `type_to_str` inside ASR utils. That will work!

certik: Yes. And if you can't find what you need, just write it yourself and put it somewhere in `libasr`.

src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -668,23 +668,34 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
...
-                return;
             } else if (x.m_op == ASR::unaryopType::USub) {
                 src = "-" + src;
                 last_unary_plus = true;
                 last_binary_plus = false;
-                return;
+            } else if (x.m_op == ASR::unaryopType::Not) {
+                src = "!" + src;
+                last_unary_plus = false;
+                last_binary_plus = false;
```
certik, line 694: Btw, these `last_unary_plus` do not work well. I did it like this because I thought it would work, but it doesn't work. The way to implement this is using this approach: 

* https://gitlab.com/lfortran/lfortran/-/blob/127b40f5c6c074a763b104e86fe599c3e77ef798/src/lfortran/ast_to_src.cpp#L114

* https://gitlab.com/lfortran/lfortran/-/blob/127b40f5c6c074a763b104e86fe599c3e77ef798/src/lfortran/ast_to_src.cpp#L3367

* https://gitlab.com/lfortran/lfortran/-/blob/127b40f5c6c074a763b104e86fe599c3e77ef798/src/lfortran/ast_to_src.cpp#L3371

certik (approved):

PR #82, opened 2022-01-28
Integration tests

Smit-create:


integration_tests/run_tests.py
```diff
@@ -3,16 +3,22 @@
...
+    "run_tests.py"
 ]
 
 def main():
+    tests = []
+    for file in os.listdir(CUR_DIR):
+        if file in skip_list:
+            continue
+        if file.endswith(".py") and not file.startswith("py_"):
+            tests.append(file)
```
certik, line 17: This works now, but it will not work once we add modules --- as we only want to call LPython on the "main" program, not the modules it imports. So we should explicitly list all the tests that we want to execute by hand, and that way we can have other Python files that can be imported, but they will not be directly tested. 

If you agree, I would leave the original explicit list of tests in there.

certik:
The tests look good. 

In general, let's separate changes to the test runner and adding tests. If you can separate the two into two PRs, then we can merge the PR that adds tests, as it looks good.

The changes to the test runner require more work, and also I always have to carefully review any such changes, as in my experience it's easy to make a mistake and then the runner does not work anymore.

integration_tests/run_tests.py
Smit-create: Okay, agreed!!

certik (approved):

Issue #78, opened 2022-01-27
Enable function to be called from another function


Smit-create:
Reference code:
```py
def test(a:i32, b:i32):
    return a**b

def check():
    a: i32
    a = test(2, 2)
```

<details>
<summary> AST </summary>
<p>

```bash
(Module [
   (FunctionDef 
      test 
      ([] [
         (
            a 
            (Name 
               i32 
               Load) ()) 
         (
            b 
            (Name 
               i32 
               Load) ())] [] [] [] [] []) [
      (Return 
         (BinOp 
            (Name 
               a 
               Load) 
            Pow 
            (Name 
               b 
               Load)))] [] () ()) 
   (FunctionDef 
      check 
      ([] [] [] [] [] [] []) [
      (AnnAssign 
         (Name 
            a 
            Store) 
         (Name 
            i32 
            Load) () 1) 
      (Assign [
         (Name 
            a 
            Store)] 
         (Call 
            (Name 
               test 
               Load) [
            (ConstantInt 2 ()) 
            (ConstantInt 2 ())] []) ())] [] () ())] [])
```

</p>
</details>

<details>
<summary> ASR </summary>
<p>

```bash
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/admin-pc/Smitlunagariya/lpython/src/bin/lpython", in _start()
  File "/build/glibc-S9d2JN/glibc-2.27/csu/../csu/libc-start.c", line 310, in __libc_start_main()
  File "/home/admin-pc/Smitlunagariya/lpython/src/bin/lpython.cpp", line 1485, in main()
    with_intrinsic_modules, compiler_options);
  File "/home/admin-pc/Smitlunagariya/lpython/src/bin/lpython.cpp", line 561, in emit_asr()
    r = LFortran::Python::python_ast_to_asr(al, *ast, diagnostics);
  File "/home/admin-pc/Smitlunagariya/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 1664, in LFortran::Python::python_ast_to_asr(Allocator&, LFortran::Python::AST::ast_t&, LFortran::diag::Diagnostics&)
    auto res2 = body_visitor(al, *ast_m, diagnostics, unit);
[...]

PR #83, opened 2022-01-28
Raise error if return type not defined

Smit-create:
Helpful in cases such as #78


certik (approved):
This is an improvement, so +1 to merge.

However, we still need to test just "return" from a "subroutine", that is, a Python function that returns `None`.

PR #84, opened 2022-01-29
Add support for `callable` inbuilt function

Smit-create:


tests/constants1.py
```diff
@@ -39,3 +39,15 @@ def test_bool():
     a = bool('t')
     a = bool(2.3)
     assert a == True
+
+
+def test_callable():
+    a: bool
+    b: i32
+    b = 2
+    a = callable(test_len)
```
namannimmo10, line 48: We could also test `callable` on other built-in functions:
```python
callable(print) == True
callable(len) == True
```

Smit-create: Ohh, yeah thanks for pointing. One solution to that can be to manually store built-in functions' names in a vector and then check for them.

certik (approved):
I think this is fine.

tests/constants1.py
certik: The current check is fine.

certik: We have to move all these built-in functions into `FunctionCall` with a compile time value. Then this check would work for most. Finally, there are some like "print", which would need to be handled in a different way --- probably the LPython frontend would know. There might be some corner cases that would need some design work, such as:
```
a = print
callable(a)
```
And passing `a` at runtime...

PR #86, opened 2022-01-29
Implement `quit()` and `exit()` built-in functions to terminate a script early

namannimmo10:


certik (approved):
I think this is fine for now, although I assume this will break if a user defines an `exit` function. For all these intrinsic functions we have to actually look them up, and only when we fail, then we should determine based on the name what to do.

PR #87, opened 2022-01-30
Add integration test for function call

Smit-create:


Smit-create:
Thanks for the review

PR #90, opened 2022-01-31
CPP backend clean ups

Smit-create:


namannimmo10 (approved):

certik:
Isn't it more readable to just write `is_a` instead of `ASR::is_a`? In AST to ASR we have to use the prefix, as we have both AST and ASR, but in libasr, the only `is_a` is in ASR, there is no other representation. You don't want to do this in header files, but in cpp files I think it's fine to shorten it. What do you think?

certik:
I think `is_something` is good if the check is not trivial. If it is as trivial as `is_a<Subroutine_t>` then I think it's not worth it, is it?

Smit-create:
> Isn't it more readable to just write `is_a` instead of `ASR::is_a`?

IMHO, `ASR::is_a` seems to be more apt as I think using namespaces makes it clearer and easier while debugging.

certik (approved):
I think it's fine to merge.

PR #91, opened 2022-01-31
Implement Set

namannimmo10:


grammar/ASR.asdl
```diff
@@ -209,6 +209,7 @@ expr
     | ConstantInteger(int n, ttype type)
     | ConstantReal(float r, ttype type)
     | ConstantComplex(float re, float im, ttype type)
+    | ConstantSet(expr* elts, ttype type)
```
certik, line 212: I would call it:
```suggestion
    | ConstantSet(expr* elements, ttype type)
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1365,6 +1365,28 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            } else {
+                if (!ASRUtils::check_equal_type(ASRUtils::expr_type(value), type)) {
+                    throw SemanticError("All Set values must be of the same type for now",
+                                        x.base.base.loc);
+                }
+            }
+            elts.push_back(al, value);
+        }
+
+        tmp = ASR::make_ConstantSet_t(al, x.base.base.loc, elts.p, elts.size(), type);
```
certik, line 1387: The `type` is the type of the elements, and it is correct that you check that they are all of the same type.

However, the type of `ConstantSet` is not the type of the element (say `i32`), but rather a new type `Set<i32>`. So we have to extend our types with a new type `Set`, and that type has to have a member of a type of the element. Then we need to make sure all backends and tests work, so we really need to do this in LFortran to ensure everything works with the new type(s). While at it, we should also add a type Dict.

certik (changes requested):
We should start with extending our type system with Set and Dict, and then go from there. See above.

src/lpython/semantics/python_ast_to_asr.cpp
namannimmo10: > The `type` is the type of the elements, and it is correct that you check that they are all of the same types.

Down the road, we ought to make support for different types of elements inside a set, right?
```python
>>> {1, -2, -3.4, "a"}
{1, 'a', -3.4, -2}
```

certik: It's the same issue as with a list of different elements. I wouldn't worry about it for now, for performance reasons I would focus on the same element type.

namannimmo10:
ASR:
```stdout
               test_Set [] [
               (=
                  (Var 2 a)
                  (ConstantSet [(ConstantInteger 1
                     (Integer 4 [])) (ConstantInteger 2
                     (Integer 4 [])) (ConstantInteger 3
                     (Integer 4 []))]
                     (Set
                        (Integer 4 []))) ())
               (=
                  (Var 2 a)
                  (ConstantSet [(ConstantInteger 2
                     (Integer 4 [])) (ConstantInteger 3
                     (Integer 4 [])) (ConstantInteger 4
                     (Integer 4 [])) (ConstantInteger 5
                     (Integer 4 [])) (ConstantInteger 5
                     (Integer 4 []))]
                     (Set
                        (Integer 4 []))) ())
               (=
                  (Var 2 a)
                  (ConstantSet [
                     (ConstantString "a"
                        (Character 1 1 () []))
                     (ConstantString "b"
                        (Character 1 1 () []))
                     (ConstantString "c"
                        (Character 1 1 () []))]
                     (Set
                        (Character 1 1 () []))) ())
               (=
                  (Var 2 b)
                  (ArrayRef 2 a [
                     (() (ConstantInteger 0
                        (Integer 4 [])) ())]
                     (Integer 4 []) ()) ())]
```

namannimmo10:
> However, the type of `ConstantSet` is not the type of the element (say `i32`), but rather a new type `Set<i32>`

Seems like this is true in the above-produced ASR.

tests/set1.py
```diff
@@ -0,0 +1,9 @@
+def test_Set():
+    a: i32
+    b: i32
+
+    a = {1, 2, 3}
+    a = {2, 3, 4, 5, 5}
+    a = {"a", "b", "c"}
```
certik, line 7: All these should give an error message that "Set[i32] (or character) cannot be assigned to i32". 

Then let's add tests like:
```
a: Set[i32]
a = {1, 2, 3}
```
And we need to improve the assignment to do this check.

certik:
Yes, this looks very good. The RHS seems correct in ASR. The Assignment is incorrect, as you cannot assign a set to an integer, see my comment above. You can only assign a set to a set.

namannimmo10:
> The Assignment is incorrect, as you cannot assign a set to an integer, see my comment above. You can only assign a set to a set.

I know, just I was planning to add the `Set` annotation.

certik:
Here are examples of typing: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#built-in-types, let's use the same syntax

namannimmo10:
Let's move the improvement of the assignment to a separate PR?

I know many tests will fail once I apply this diff:
```diff
diff --git a/src/lpython/semantics/python_ast_to_asr.cpp b/src/lpython/semantics/python_ast_to_asr.cpp
index bc7bb209..72fe972d 100644
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -515,7 +515,11 @@ public:
         }
         this->visit_expr(*x.m_value);
         ASR::expr_t *value = ASRUtils::EXPR(tmp);
-        ASR::stmt_t *overloaded=nullptr;
+        ASR::stmt_t *overloaded = nullptr;
+        if (!ASRUtils::check_equal_type(ASRUtils::expr_type(target), ASRUtils::expr_type(value))) {
+            throw SemanticError("Assignment types do not match",
+                x.base.base.loc);
+        }
         tmp = ASR::make_Assignment_t(al, x.base.base.loc, target, value,
                                 overloaded);
     }
```

certik:
We can't merge this PR anyway until we run all tests in LFortran, so we have to do that.

namannimmo10:
> until we run all tests in LFortran

I did so -- no test failures.

certik:
Awesome. Would you mind please sending a Merge Request against LFortran with this change in ASR? Let's get it merged there first. Then we'll come back and finish this PR.

namannimmo10:
MR opened: [!1650](https://gitlab.com/lfortran/lfortran/-/merge_requests/1650).

certik (approved):
The same comment as for the dictionary. I think this is ok to merge.

The ConstantSet creation is correct. But declaring a set and assignment to it does not generate a correct ASR yet, but we can fix that later.

tests/set1.py
```diff
@@ -0,0 +1,9 @@
...
+    # a = {-1, -2, "c"} -> semantic error for now
+    b = a[0]
```
Smit-create, line 9: Should this test be valid?
In python:
```py
>>> s={1,2,5,3,4}
>>> s[2]
TypeError: 'set' object is not subscriptable
```

namannimmo10: Thanks for pointing. We should remove this.

Smit-create: Removed in https://github.com/lcompilers/lpython/pull/98

PR #93, opened 2022-02-01
Return none inside subroutine now works

Smit-create:
Addressed: https://github.com/lcompilers/lpython/pull/58#pullrequestreview-862625867


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1317,8 +1317,13 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
     void visit_Return(const AST::Return_t &x) {
         std::string return_var_name = "_lpython_return_variable";
         if(current_scope->scope.find(return_var_name) == current_scope->scope.end()) {
-            throw SemanticError("Return type of function is not defined",
+            if (x.m_value) {
+                throw SemanticError("Return type of function is not defined",
                                 x.base.base.loc);
+            }
+            // this may be a case with void return type (like subroutine)
```
certik, line 1324: We should add a check here (we can do that in a subsequent PR) that we are actually in a subroutine, otherwise give an error message.

certik (approved):
I think that works, thanks!

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: Yes!

Smit-create:
Thanks for reviewing!

PR #94, opened 2022-02-01
Remove jupyter nb lfortran tests

Smit-create:


certik (approved):
That's fine. We should get interactive LPython working and add tests for LPython based Jupyter.

PR #95, opened 2022-02-01
Get "import" working

certik:


certik:
It now correctly puts every python file into a module. The module that `lpython` was called on is called `__main__`, the other modules are named by their names. 

What remains now is to:

* [ ] Get the "main" function working again
* [ ] Import the symbols from a module (`from b import f`) as `ExternalSymbol`s
* [ ] Test recursive module imports, and fix bugs if any

Then everything should start working.

certik:
The recursive import does not work yet, but a simple import now works.

This should be enough to get the runtime library working.

integration_tests/run_tests.py
```diff
@@ -9,15 +9,19 @@
...
 def main():
     print("Compiling...")
     for pyfile in tests:
         basename = os.path.splitext(pyfile)[0]
-        cmd = os.path.join("src", "bin", "lpython") + " integration_tests/%s -o integration_tests/%s" % (pyfile, basename)
+        cmd = os.path.join("..", "src", "bin", "lpython") + " %s -o %s" % (pyfile, basename)
         print("+ " + cmd)
+        os.chdir("integration_tests")
         r = os.system(cmd)
+        os.chdir("..")
```
certik, line 24: This part is needed because the module import is looking for the module in the current directory instead of the directory relative to the file that contains the "import ..." statement. We have to fix that too in subsequent PRs.

PR #96, opened 2022-02-01
Get Python lists working

namannimmo10:


certik:
I think that looks good.

namannimmo10:
I will also do tuples. Let's introduce `ConstantTuple` in the ASR.

certik:
You also have to do a type Tuple

PR #97, opened 2022-02-01
Get tuples working

namannimmo10:
```python
def test_Tuple():
    a: i32
    b: i32
    b1: i32

    a = (1, 2, 3)
    a = (-3, -2, -1)
    a = ("a", "b", "c")

    b = a[0]
    b, b1 = a[2], a[1]
```
```bash
(TranslationUnit
   (SymbolTable
      1
      {
         test_Tuple:
            (Subroutine
               (SymbolTable
                  2
                  {
                     a:
                        (Variable
                           2
                           a
                           Local () ()
                           Default
                           (Integer 4 [])
                           Source
                           Public
                           Required .false.),
                     b:
                        (Variable
                           2
                           b
                           Local () ()
                           Default
                           (Integer 4 [])
                           Source
                           Public
                           Required .false.),
                     b1:
                        (Variable
                           2
                           b1
                           Local () ()
                           Default
                           (Integer 4 [])
                           Source
                           Public
                           Required .false.)
                  })
               test_Tuple [] [
               (=
                  (Var 2 a)
                  (ConstantTuple [(ConstantInteger 1
                     (Integer 4 [])) (ConstantInteger 2
                     (Integer 4 [])) (ConstantInteger 3
                     (Integer 4 []))]
                     (Tuple
                        (Integer 4 []))) ())
               (=
                  (Var 2 a)
                  (ConstantTuple [
                     (UnaryOp
                        USub (ConstantInteger 3
                        (Integer 4 []))
                        (Integer 4 []) (ConstantInteger -3
                        (Integer 4 [])))
                     (UnaryOp
                        USub (ConstantInteger 2
                        (Integer 4 []))
                        (Integer 4 []) (ConstantInteger -2
                        (Integer 4 [])))
                     (UnaryOp
                        USub (ConstantInteger 1
                        (Integer 4 []))
                        (Integer 4 []) (ConstantInteger -1
                        (Integer 4 [])))]
                     (Tuple
                        (Integer 4 []))) ())
               (=
                  (Var 2 a)
                  (ConstantTuple [
                     (ConstantString "a"
                        (Character 1 1 () []))
                     (ConstantString "b"
                        (Character 1 1 () []))
                     (ConstantString "c"
                        (Character 1 1 () []))]
                     (Tuple
                        (Character 1 1 () []))) ())
               (=
                  (Var 2 b)
                  (ArrayRef 2 a [
                     (() (ConstantInteger 0
                        (Integer 4 [])) ())]
                     (Integer 4 []) ()) ())
               (=
                  (ConstantTuple [
                     (Var 2 b)
                     (Var 2 b1)]
                     (Tuple
                        (Integer 4 [])))
                  (ConstantTuple [
                     (ArrayRef 2 a [
                        (() (ConstantInteger 2
                           (Integer 4 [])) ())]
                        (Integer 4 []) ())
                     (ArrayRef 2 a [
                        (() (ConstantInteger 1
                           (Integer 4 [])) ())]
                        (Integer 4 []) ())]
                     (Tuple
                        (Integer 4 []))) ())]
               Source
               Public
      })
   [])
```


namannimmo10:
Once again I see no test failures in LFortran.

grammar/ASR.asdl
```diff
@@ -233,6 +234,7 @@ ttype
     | Character(int kind, int len, expr? len_expr, dimension* dims)
     | Logical(int kind, dimension* dims)
     | Set(ttype type)
+    | Tuple(ttype type)
```
certik, line 237: Let's do it like this:
```suggestion
    | Tuple(ttype* type)
```
This specifies a type for each element of the tuple. A tuple in Python is always a constant (although it can be created at runtime). We'll impose a limitation that the type has to be known at compile time. The type consists of a list of types for each element. So you know both the length and the types at compile time.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -573,6 +573,27 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             list.size(), type);
     }
 
+    void visit_Tuple(const AST::Tuple_t &x) {
+        Vec<ASR::expr_t*> tuple;
+        tuple.reserve(al, x.n_elts);
+        ASR::ttype_t *type = nullptr;
```
certik, line 579: Change this to:
```
Vec<ASR::ttype_t> tuple_type;
tuple_type.reserve(al, x.n_elts);
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -573,6 +573,27 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            this->visit_expr(*x.m_elts[i]);
+            ASR::expr_t *expr = ASRUtils::EXPR(tmp);
+            if (type == nullptr) {
+                type = ASRUtils::expr_type(expr);
+            } else {
+                if (!ASRUtils::check_equal_type(ASRUtils::expr_type(expr), type)) {
+                    throw SemanticError("All Tuple elements must be of the same type for now",
+                        x.base.base.loc);
+                }
+            }
```
certik, line 590: And then change this to:
```suggestion
            tuple_type.push_back(al, ASRUtils::expr_type(expr));
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -573,6 +573,27 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            tuple.push_back(al, expr);
+        }
+        ASR::ttype_t *tuple_type = ASRUtils::TYPE(ASR::make_Tuple_t(al, x.base.base.loc, type));
```
certik, line 593:
```suggestion
        ASR::ttype_t *tuple_type = ASRUtils::TYPE(ASR::make_Tuple_t(al, x.base.base.loc, tuple_type_vec.n, tuple_type_vec.p));
```

src/runtime/lpython_parser.py
```diff
@@ -35,6 +35,8 @@ def visit_Constant(self, node):
                         node.value.imag, node.kind)
         elif isinstance(node.value, dict):
             return python_ast.Dict(node.keys, node.values)
+        elif isinstance(node.value, tuple):
+            return python_ast.Tuple(node.elts)
```
certik, line 39: The `Tuple` node is defined as:
```
         | Tuple(expr* elts, expr_context ctx)
```
Isn't it missing an argument?

namannimmo10: Sorry, I missed it.  Thanks for pointing!

grammar/ASR.asdl
namannimmo10: Good idea!

src/runtime/lpython_parser.py
```diff
@@ -35,6 +35,8 @@ def visit_Constant(self, node):
                         node.value.imag, node.kind)
         elif isinstance(node.value, dict):
             return python_ast.Dict(node.keys, node.values)
+        elif isinstance(node.value, tuple):
+            return python_ast.Tuple(node.elts, node.ctx)
```
certik, line 39: The other thing I noticed is that I think you have to convert "node.elts" from Python's AST to `python_ast` (our module). Otherwise they still refer to Python's AST, which does not know how to serialize, so it should fail... I don't understand why it does not fail. Perhaps because the types are simple. 

Try constructing a tuple of tuple.

certik (approved):
I think this looks good now and perhaps we can merge it now.

I still think there are some bugs, see my comment above.

src/runtime/lpython_parser.py
namannimmo10:
```python
a = ((1, 2), (3, 4))
```
The above code leaves us with:
```bash
LFortranException: Not implemented
``` 
because it fails to get the type when it hits this line:
```c++
tuple_type_vec.push_back(al, ASRUtils::expr_type(expr))
```
So it works because the types are simple.

certik: The code was not being executed, so we removed it.

certik (approved):
I think this is ready to go in once tests pass.

PR #98, opened 2022-02-02
CPP backend for constantset

Smit-create:


Smit-create:
Currently, doesn't generate a correct cpp code because we still need a notation for set in `lpython` but `ConstantSet` seems to be working for now:
```cpp
void test_Set()
{
    int a;
    int b;
    a = {1, 2, 3};
    a = {2, 3, 4, 5, 5};
    a = {"a", "b", "c"};
    b = a[0 + 1 - 1];
}
```

tests/reference/cpp-doconcurrentloop_01-4e9f274.stdout
```diff
@@ -28,8 +28,8 @@ int main()
     scalar = 10;
     nsize = 1234;
     Kokkos::parallel_for(Kokkos::RangePolicy<Kokkos::DefaultExecutionSpace>(1, nsize - 1+1), KOKKOS_LAMBDA(const long i) {
-        a[i-1] = 5;
-        b[i-1] = 5;
+        a[i + 1 - 1] = 5;
+        b[i + 1 - 1] = 5;
```
certik, line 32: I think this change is not correct.

certik: The `i` runs from 1 to `nsize` (I think) and the array `a` is indexed from 0 to 10000-1.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -541,7 +541,14 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         ASR::array_index_t ai;
         ai.loc = x.base.base.loc;
         ai.m_left = nullptr;
-        ai.m_right = index;
+        // ASR uses 1-based indexing
```
certik, line 544: Actually, ASR does not use 1 based indexing. Rather, you have to explicitly specify lower and upper bound.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -541,7 +541,14 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+        ASR::ttype_t *a_type = ASRUtils::TYPE(ASR::make_Integer_t(al, x.base.base.loc,
+            4, nullptr, 0));
+        ASR::expr_t *constant_one = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantInteger_t(
+                                            al, x.base.base.loc, 1, a_type));
+        make_BinOp_helper(index, constant_one, ASR::binopType::Add,
+                            x.base.base.loc, false);
+        ai.m_right = ASRUtils::EXPR(tmp);
```
certik, line 551: I think this change is not correct --- when we index, we represent it exactly as it is. 

Then for each array we store the lower and upper bound.

tests/reference/cpp-doconcurrentloop_01-4e9f274.stdout
Smit-create: Yeah, that's true for this case.
But this fails for the following case:
Py code:
```
def test_Set():
    a: i32
    b: i32

    a = {1, 2, 3}
    a = {2, 3, 4, 5, 5}
    a = {"a", "b", "c"}
    b = a[0]
```

CPP:
```cpp
void test_Set()
{
    int a;
    int b;
    a = {1, 2, 3};
    a = {2, 3, 4, 5, 5};
    a = {"a", "b", "c"};
    b = a[0-1];
}
```

certik: That's because the "a" should not be an array ref, it's not an array.

Smit-create: Yep, make sense!

certik:
I don't see any test of C++ output for the new function. Did I miss it?

Smit-create:
> I don't see any test of C++ output for the new function. Did I miss it?

I haven't added since the generation is not accurate until we support type annotation for the set. But that's okay, we can improve that part as a follow up

certik (approved):

PR #99, opened 2022-02-03
cpp backend for constant dictionary

Smit-create:


src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -593,6 +593,21 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
         last_binary_plus = false;
     }
 
+    void visit_ConstantDictionary(const ASR::ConstantDictionary_t &x) {
+        LFORTRAN_ASSERT(x.n_keys == x.n_values);
```
namannimmo10, line 597: I don't think we need the check here since we already do this in AST->ASR.

certik: Yes, we don't need to check, but it doesn't hurt either. The assert statements have no overhead in Release mode.

certik:
And add tests that show the newly implemented C++ output.

certik (approved):

Smit-create:
Thanks for the review!

PR #100, opened 2022-02-03
Improve C++ codegen by using last precedence

namannimmo10:


src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -98,6 +98,9 @@ class ASRToCPPVisitor : public ASR::BaseVisitor<ASRToCPPVisitor>
     std::string src;
     int indentation_level;
     int indentation_spaces;
+    // The precedence of the last expression, using the table:
+    // https://en.cppreference.com/w/cpp/language/operator_precedence
+    int last_expr_precedence;
     bool last_unary_plus;
     bool last_binary_plus;
```
certik, line 105: You can remove these now.

src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -781,30 +784,42 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
...
+            case (ASR::boolopType::Eqv): {
+                last_expr_precedence = 10;
                 break;
             }
             default : throw CodeGenError("Unhandled switch case");
         }
-        last_binary_plus = false;
-        last_unary_plus = false;
+
+        if (left_precedence >= last_expr_precedence) {
```
certik, line 812: The lower the precedence number, the higher the precedence. So I think it has to be like this:
```suggestion
        if (left_precedence <= last_expr_precedence) {
```

tests/reference/cpp-loop2-0686fc4.stdout
```diff
@@ -20,10 +20,10 @@ void test_for()
 {
     int i;
     for (i=0; i<=10 - 1; i++) {
-        if (i == 0) {
+        if ((i) == (0)) {
             continue;
         }
-        if (i > 5) {
+        if ((i) > (5)) {
```
namannimmo10, line 26: This looks weird.

certik:
You have to update test results.

tests/reference/cpp-doconcurrentloop_01-4e9f274.stdout
```diff
@@ -41,8 +41,8 @@ void triad(const Kokkos::View<float*> &a, const Kokkos::View<float*> &b, float s
 {
     int N;
     N = 1234;
-    Kokkos::parallel_for(Kokkos::RangePolicy<Kokkos::DefaultExecutionSpace>(1, N - 1+1), KOKKOS_LAMBDA(const long i) {
-        c[i-1] = a[i-1] + scalar*b[i-1];
+    Kokkos::parallel_for(Kokkos::RangePolicy<Kokkos::DefaultExecutionSpace>(1, N - (1)+1), KOKKOS_LAMBDA(const long i) {
```
certik, line 44: This change seems wrong.

tests/reference/cpp-doconcurrentloop_01-4e9f274.stdout
```diff
@@ -41,8 +41,8 @@ void triad(const Kokkos::View<float*> &a, const Kokkos::View<float*> &b, float s
...
+        c[i-1] = a[i-1] + (scalar) * (b[i-1]);
```
certik, line 45: This change seems wrong for two reasons:
* there should be no parentheses
* There shouldn't be space around `*`. This is to follow SymPy's approach, of putting space around + and -, but not * and /. It seems the expression is generally more readable that way.

src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -535,26 +534,18 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
 
     void visit_ConstantInteger(const ASR::ConstantInteger_t &x) {
         src = std::to_string(x.m_n);
-        last_unary_plus = false;
-        last_binary_plus = false;
```
certik, line 539: I think you have to set the last precedence for each such place that you removed `last_unary_plus`. Essentially you have to go into each expression node and ensure the last precedence is set. That should fix the issue with extra (inconsistent) parentheses.

src/libasr/asr_utils.h
```diff
@@ -149,14 +149,34 @@ static inline std::string unop_to_str(const ASR::unaryopType t) {
     }
 }
 
+static inline std::string binop_to_str(const ASR::binopType t) {
+    switch (t) {
+        case (ASR::binopType::Add): { return " + "; }
+        case (ASR::binopType::Sub): { return " - "; }
+        case (ASR::binopType::Mul): { return " * "; }
+        case (ASR::binopType::Div): { return " / "; }
+        default : throw LFortranException("Not implemented");
```
czgdp1807, line 158: A better error message?

certik:
@namannimmo10, if you want, you can try to finish this one. It shouldn't be difficult.

namannimmo10:
I'll try to complete this one soon.

namannimmo10:
Ready?

src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -518,8 +519,6 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
             }
             src = fn_name + "(" + args + ")";
         }
-        last_unary_plus = false;
-        last_binary_plus = false;
```
certik, line 522: `FunctionCall` must set precedence. I don't see it set in the visitor itself. Is it somehow implicitly set elsewhere? I don't know of a good way to ensure that it is set in every `expr` visitor, but for now a good way is to literally explicitly assign to `last_expr_precedence` in every single visitor, or add a comment why that is not needed in the source code. So since I don't see it being assigned, I think it might be a bug? Let's either assign or add a comment here why not.

src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -695,45 +680,30 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
...
-                last_binary_plus = false;
             } else {
                 throw CodeGenError("Unary type not implemented yet for Real");
             }
             return;
         } else if (x.m_type->type == ASR::ttypeType::Logical) {
             if (x.m_op == ASR::unaryopType::Not) {
                 src = "!" + src;
-                last_unary_plus = false;
-                last_binary_plus = false;
```
certik, line 736: Let's assign to `last_expr_precedence` in this function or add a comment why it is not needed.

certik:
I haven't gone over the other expr visitors, but we need to carefully go over every expr visitor and ensure we assign to `last_expr_precedence`, or add a comment.

src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -813,35 +811,46 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
                 // TODO: implement
             }
         }
-        last_unary_plus = false;
```
certik, line 816: This has to assign `last_expr_precedence`.

certik:
Last minor update I found. After that, do you mind rebasing this PR to a more logical set of commits? It looks like we don't have any comments, we always assign the last precedence, so you can probably squash the last few commits together and rewrite the commit log.

certik (approved):
I think this looks great, thanks! After this is merged, I'll port it to LFortran where we have a lot more C++ tests and if there are any bugs, I'll report back.

PR #102, opened 2022-02-04
Create main program properly

certik:
This creates a main program for the "main Python script" and includes any global commands in it (if any).


certik:
Things to fix before merging:
* [ ] The C++ output seems to be skipping some stuff, investigate / fix
* [ ] Make all Python script call the "main" function, as you would in Python

certik:
I noticed tests in `tests/expr12.py` are not present in the C++ output, not sure if this is caused by the "unused function optimization".

certik:
The C++ tests didn't have a main program, so the functions were not called, thus the "unused" optimization optimized them out. This is now fixed and all the C++ tests now look good.

certik:
Ok, this now allows us to write proper programs in LPython.

namannimmo10:
Awesome, thanks for getting it done!!

PR #104, opened 2022-02-05
Make `len` work for all supported containers

namannimmo10:


tests/constants1.py
```diff
@@ -28,6 +28,12 @@ def test_len():
     a: i32
     a = len('test')
     a = len("this is a test")
+    a = len((1, 2, 3))
+    a = len((("a", "b", 3.4), ("c", 3, 5.6)))
+    a = len([1, 2, 3])
```
namannimmo10, line 33: This is a compile-time function for now.

certik (approved):
I think that's fine. It's only compile time, but it's consistent with what we have. We'll make it all work at runtime soon.

PR #112, opened 2022-02-08
Initial implementation of the math.py runtime library

certik:


namannimmo10:
Looks great :)

Smit-create (approved):
This looks great!! Thanks

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -46,6 +46,23 @@ LFortran::Result<LFortran::Python::AST::ast_t*> parse_python_file(Allocator &al,
...
+    } else {
+        std::string filename_intrinsic = runtime_library_dir + "/" + filename;
+        bool status = read_file(filename_intrinsic, input);
+        if (status) {
+            return filename_intrinsic;
+        } else {
+            return LFortran::Error();
+        }
+    }
+}
```
certik, line 64: This is obviously inefficient, we should simply read it but not throw the source out, but rather give it as a string to `parse_python_file` (which should be renamed to `parse_python`). This function for now would still save it to a disk and run Python on it, but later we will rewrite it with our own parser that would do everything in memory.

But we can do this later.

PR #118, opened 2022-02-08
Add an initial sys module

certik:


src/runtime/sys.py
```diff
@@ -0,0 +1,6 @@
+def exit(error_code: i32):
+    """
+    Exits the program with an error code `error_code`.
+    """
+
+    print("Not implemented yet.")
```
certik, line 6: To implement this, we need to be able to exit with a nonzero exit code. I think we can piggy back on the "intrinsic" exit function, and translate it correctly to "stop" with a proper error code, or to "error stop" in ASR. Then we might be able to implement this in pure Python such that it works both in CPython and LPython. If it cannot be done, then we need to implement it by calling into C (#3).

PR #121, opened 2022-02-09
Implement initial ltypes.py

### LINKED ISSUE #115: Implement `ltypes.py`
With `i32`, `i64`, etc. types. This file should probably be in `src/runtime` also. We should be importing it, and ensure that things just work with CPython with the same source code.

certik:
Fixes #115.


src/runtime/ltypes/ltypes.py
```diff
@@ -0,0 +1,4 @@
+i32 = []
+i64 = []
+f32 = []
+f64 = []
```
namannimmo10, line 4:
```suggestion
f64 = []
c128 = []
```

certik: I merged it, as tests pass and this was an important PR. Can you submit this as a new PR?

PR #131, opened 2022-02-11
Completes implicit cast int->complex and int->logical

namannimmo10:


certik:
Is there some way to write a test for this?

namannimmo10:
Now that I look at it, seems like `ImplicitCastRules::set_converted_value` is not used much:
```bash
$ git grep set_converted_value
src/lpython/semantics/asr_implicit_cast_rules.h:  static void set_converted_value(Allocator &al, const Location &a_loc,
src/lpython/semantics/ast_body_visitor.cpp:            ImplicitCastRules::set_converted_value(al, x.base.base.loc, &value,
src/lpython/semantics/ast_common_visitor.h:      ImplicitCastRules::set_converted_value(
src/lpython/semantics/ast_common_visitor.h:    ImplicitCastRules::set_converted_value(al, x.base.base.loc, conversion_cand,
src/lpython/semantics/ast_common_visitor.h:    ImplicitCastRules::set_converted_value(al, loc, &result,
src/lpython/semantics/ast_common_visitor.h:    ImplicitCastRules::set_converted_value(al, loc, &result,
src/lpython/semantics/ast_common_visitor.h:                ImplicitCastRules::set_converted_value(
src/lpython/semantics/ast_common_visitor.h:        ImplicitCastRules::set_converted_value(al, x.base.base.loc, conversion_cand,
src/lpython/semantics/ast_symboltable_visitor.cpp:                    ImplicitCastRules::set_converted_value(al, x.base.base.loc, &init_expr, init_type, type);
```
We didn't use it for Python AST->ASR.  Thought these two guys were missing, so I completed them. Not sure how to test this separately.

certik:
I wouldn't merge it unless it influences some tests. If there is no way to test this, then I would say it is not needed.

PR #123, opened 2022-02-09
Add type checking to assignment

certik:

PR #133, opened 2022-02-11
Refactor implicit cast and use it in compare 

Smit-create:


namannimmo10 (approved):
LGTM

tests/reference/cpp-doconcurrentloop_01-4e9f274.stdout
```diff
@@ -28,11 +28,11 @@ void main0()
     Kokkos::View<float[10000]> c("c");
     int nsize;
     float scalar;
-    scalar = 10;
+    scalar = (float)(10);
```
certik, line 31: This is actually incorrect.  This is a bug in a test, it should give a compiler error message saying that you cannot assign integer to a variable that is declared as a float:
```python
    scalar: f32
    scalar = 10
```
And the fix is to use `scalar = 10.0`.

tests/reference/cpp-doconcurrentloop_01-4e9f274.stdout
```diff
@@ -28,11 +28,11 @@ void main0()
...
     nsize = 1234;
     Kokkos::parallel_for(Kokkos::RangePolicy<Kokkos::DefaultExecutionSpace>(1, nsize - 1+1), KOKKOS_LAMBDA(const long i) {
-        a[i-1] = 5;
-        b[i-1] = 5;
+        a[i-1] = (float)(5);
+        b[i-1] = (float)(5);
```
certik, line 35: This however is correct, as the integer gets upgraded to a float, because the LHS is an array, so things get implicitly cast.

tests/reference/cpp-expr3-9c516d4.stdout
```diff
@@ -28,6 +28,9 @@ void test_cast()
     a = (float)(a) - 3.900000;
     a = (float)(a)/b;
     b = (float)(3)/(float)(4);
+    if ((float)(a) < b) {
+        std::cout << "a < b" << std::endl;
+    }
```
certik, line 33: This is correct.

tests/reference/cpp-doconcurrentloop_01-4e9f274.stdout
Smit-create: > This is actually incorrect.

I think this should be allowed to assign an integer to float. Python also allows to do that:
```py
>>> d: float
>>> d = 2
>>> print(d)
2

```

certik: Fortran and C allows to do that. But not Python:
```
>>> d: float
>>> d = 2
>>> print(d)
2
>>> type(d)
<class 'int'>
```
And we should emulate what Python does. In python, `d` will be an integer. So that's why if we declare `d: float`, we need to give an exception in LPython.

Smit-create: Oh, okay, that makes sense!

Smit-create:
Will update the PR once #136  is merged

tests/doconcurrentloop_01.py
```diff
@@ -12,11 +12,11 @@ def main0():
...
     i: i32
     nsize: i32
-    scalar = 10
+    scalar = 10.0
     nsize = size(a)
     for i in range(nsize): # type: parallel
-        a[i] = 5
-        b[i] = 5
+        a[i] = 5.0
+        b[i] = 5.0
```
certik, line 19: This part doesn't need to be changed, as even in Python `a[i] = 5` will be properly cast to floating point as long as `a[i]` is a NumPy array.

Smit-create: Oh, okay! We can fix that as a follow up(as a special case of assign)

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2234,10 +2229,16 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+
+        if(ASR::is_a<ASR::Function_t>(*s)) {
+            ASR::Function_t *func = ASR::down_cast<ASR::Function_t>(s);
+            ASR::ttype_t *a_type = ASRUtils::expr_type(func->m_return_var);
+            tmp = ASR::make_FunctionCall_t(al, x.base.base.loc, s,
+                nullptr, args.p, args.size(), nullptr, 0, a_type, nullptr, nullptr);
+        } else {
+            tmp = ASR::make_SubroutineCall_t(al, x.base.base.loc, s,
+                nullptr, args.p, args.size(), nullptr);
+        }
```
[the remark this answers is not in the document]
Smit-create: This is fixed now.

Smit-create:
Tests are passing, we can work on a new PR for fixes related to `implicitcast` and patch #123

certik (approved):
This looks great. After rebasing this can be merged.

Smit-create:
Windows CI fails due to some reason that does not seem to be related to this PR. @certik Have a look if this can be merged.

certik:
The Windows error is:
```
C:\ProgramData\chocolatey\bin\c++.exe -DHAVE_TARGET_X86=1 -ID:/a/lpython/lpython/lpython-0.1.0-470-g951cb6d27/test-bld/src/libasr/.. -ID:/a/lpython/lpython/lpython-0.1.0-470-g951cb6d27/src/libasr/.. -isystem C:/Miniconda/envs/test/Library/include -Wall -Wextra -O3 -march=native -funroll-loops -DNDEBUG -fno-rtti -std=gnu++17 -Wno-deprecated-declarations -MD -MT src/libasr/CMakeFiles/asr.dir/codegen/evaluator.cpp.obj -MF src\libasr\CMakeFiles\asr.dir\codegen\evaluator.cpp.obj.d -o src/libasr/CMakeFiles/asr.dir/codegen/evaluator.cpp.obj -c D:/a/lpython/lpython/lpython-0.1.0-470-g951cb6d27/src/libasr/codegen/evaluator.cpp
In file included from C:/Miniconda/envs/test/Library/include/llvm/Object/ObjectFile.h:23,
                 from C:/Miniconda/envs/test/Library/include/llvm/DebugInfo/DIContext.h:18,
                 from C:/Miniconda/envs/test/Library/include/llvm/ExecutionEngine/RuntimeDyld.h:19,
                 from C:/Miniconda/envs/test/Library/include/llvm/ExecutionEngine/JITEventListener.h:19,
                 from C:/Miniconda/envs/test/Library/include/llvm/ExecutionEngine/Orc/RTDyldObjectLinkingLayer.h:19,
                 from D:/a/lpython/lpython/lpython-0.1.0-470-g951cb6d27/src/libasr/codegen/KaleidoscopeJIT.h:23,
                 from D:/a/lpython/lpython/lpython-0.1.0-470-g951cb6d27/src/libasr/codegen/evaluator.cpp:53:
C:/Miniconda/envs/test/Library/include/llvm/Object/SymbolicFile.h: In function 'OStream& llvm::object::operator<<(OStream&, const llvm::object::DataRefImpl&)':
C:/Miniconda/envs/test/Library/include/llvm/Object/SymbolicFile.h:46:30: error: expected ')' before 'PRIxPTR'
   OS << "(" << format("0x%08" PRIxPTR, D.p) << " (" << format("0x%08x", D.d.a)
                      ~       ^~~~~~~~
                              )
```
I might have seen this before. It might be caused by some missing or extra include that confuses things.

Smit-create:
Fails on #144

czgdp1807:
Is there any specific for avoiding https://gitlab.com/lfortran/lfortran/-/blob/master/src/lfortran/semantics/asr_implicit_cast_rules.h and re-writing the whole type handling logic again in `python_ast_to_asr.cpp`?

certik:
The semantic rules might be slightly different. I do feel there might be a way to somehow reuse parts of this in LFortran and LPython.

czgdp1807:
The APIs in the above linked file accept only ASR nodes, so it can be integrated in lpython with almost no or very few changes AFAICT. ASR in lpython and lfortran is same, right?

certik:
The ASR is the same, but the semantics how the ASR is obtained from the surface language is different. I don't know if the common code can be refactored into common parts, but we should attempt it.

czgdp1807:
Well, at least for implicit cast purposes one only needs to update, `ImplicitCastRules::rule_map` and `ImplicitCastRules::type_priority` in the [above linked file](https://gitlab.com/lfortran/lfortran/-/blob/master/src/lfortran/semantics/asr_implicit_cast_rules.h). The functions, `ImplicitCastRules::set_converted_value` and `ImplicitCastRules::find_conversion_candidate` shouldn't need any change.

Re-using code will help in reducing the code surface and hence the bugs eventually. Otherwise it will double the amount of efforts to first figure out the bugs in two separate codes and then eventually fixing them separately.

certik:
I agree, no doubt. Another place for reuse is the runtime library handling and module loading. It's slightly different, but I still feel we might be able to refactor common parts.

PR #135, opened 2022-02-11
Change types of complex numbers to c32 and c64

### LINKED ISSUE #134: Change types of complex numbers to c32 and c64
It seems much more natural to me, and consistent with quite a few codes and libraries:

* https://arrayfire.org/docs/gettingstarted.htm
* https://michaelhatherly.github.io/NumericSuffixes.jl/stable/
* https://github.com/JuliaLang/julia/pull/24647
* https://docs.rs/cauchy/latest/cauchy/type.c64.html

Essentially, it means the complex number is composed of the 32 or 64 bit float. One says double precision complex number, and double precision is 64 bit, so it's 64 bit complex number, `c64` (yes, it is represented by 128bits).

namannimmo10:
Fixes #134


namannimmo10:
Down the road, we can also add c16 if needed.

certik (approved):
Thanks for implementing this!

PR #137, opened 2022-02-12
Rename `parse_python_file` to `parse_python`

namannimmo10:


certik:
I think that this is fine. But isn't it better to call it `parse_python_file` and then later once we can parse in memory, we implement this function using `parse_python` that parses a string?

namannimmo10:
Yes, that's even better.

PR #139, opened 2022-02-13
Parse `list` type annotation

namannimmo10:


certik (approved):
That looks good to me.

PR #144, opened 2022-02-17
Update README.md

Smit-create:
testing for windows failure


certik:
Ok. Since the master fails too, let's fix the master here first. Then we can merge your PR.

PR #145, opened 2022-02-17
Fix windows build failure

Smit-create:
See https://github.com/lcompilers/lpython/runs/5227998621?check_suite_focus=true

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,6 +187,9 @@ if (WITH_FMT)
     find_package(fmt)
 endif()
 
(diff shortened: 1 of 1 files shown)


Smit-create:
Tried 2-3 alternative solutions, but still fails!

certik:
Thanks for trying! We'll have to figure out what is going on.

certik:
I think you fixed it, now there is another issue.

Smit-create:
Closing this in favour of #151

PR #147, opened 2022-02-17
Computes the result of string comparison

namannimmo10:


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1703,7 +1703,45 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     }
                 }
                 value = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantLogical_t(
-                    al, x.base.base.loc, result, source_type));
+                    al, x.base.base.loc, result, type));
+            } else if (ASRUtils::is_character(*source_type)) {
+                char* left_value = ASR::down_cast<ASR::ConstantString_t>(
+                                        ASRUtils::expr_value(left))->m_s;
```
czgdp1807, line 1709: What is `ASRUtils::expr_value(left)` is nullptr? Is that case possible?

namannimmo10: Yes!

tests/expr13.py
```diff
@@ -7,3 +7,12 @@ def test_Compare():
     a = 3.3 == 3.3
     a = 3.3 != 3.4
     a = complex(3, 4) == complex(3., 4.)
+
+    # string comparison
+    a = "abc" > "abd"
```
certik, line 12:
```suggestion
    a = "abc" > "abd"
    assert a
```

certik: And the same for all the other new tests.

certik:
Otherwise it looks good.

tests/expr13.py
namannimmo10: What's the use of an `assert` statement here right now? These are not integration tests. It will create an `Assert` node in ASR but, it doesn't actually fail if `a` is False with the above-suggested change.

certik: Good point, I forgot this is not actually tested in integration tests.

certik:
This looks good. Couple points:

* Try to rebase instead of merging master, since now there are two merges with master in this PR, unnecessarily complicating the history.
* You can start adding tests into integration_tests and actually test them with assert. If you want to test ASR, you can still do that, simply add the test in `tests/tests.toml`.

namannimmo10:
Sure!  BTW, we already use `assert` in `integration_tests`.

PR #148, opened 2022-02-17
Add gcd and lcm math functions

Smit-create:


src/runtime/math.py
```diff
@@ -109,3 +109,30 @@ def pow(x: f64, y: f64) -> f64:
...
+    Returns a%b
+    """
+    return a - (a//b)*b
+
+
+def gcd(a: i32, b: i32) -> i32:
+    """
+    Returns greatest common divisor of `a` and `b`
+    """
+    temp: i32
```
namannimmo10, line 125:
```suggestion
    temp: i32
    if a == 0: return b if b >= 0 else -b
    if b == 0: return a if a >= 0 else -a
```
And we can use `abs` builtin once we have that.

certik (approved):
Thanks!

Issue #1, opened 2021-12-29
Add str() builtin function


certik:
Once LPython can use ASR's generic function facilities, it should just be a generic function. In the meantime, let's implement:

* `str_int(x)` ... converts an integer to a string
* `str_float(x)` ... converts a floating point to a string

These will be implemented in pure Python, such as:
```python
def str_int(x):
    ....
    # s = ...
    return s
```
For now, let's put this into one file, and print the result at the end. Some things will not be implemented in LPython yet, so let's open up issues for things that are not implemented yet, and we'll fix them.

Smit-create:
I tried implementing `str_int(x)` using different ways:
1. Using `x: i32` 
   
<details>
    
<summary> Reference code </summary>
    
<p>

   ```py
 def str_int(x: i32) -> str:
        if x == 0:
            return '0'
        result: str
        result = ''
        if x < 0:
            result += '-'
            x = -x
        rev_result: str
        rev_result = ''
        pos_to_str: list
        pos_to_str = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']
        while x > 0:
            rev_result += pos_to_str[x%10]
            x = x//10
        for pos in range(len(rev_result) - 1, -1, -1):
            result += rev_result[pos]
        return result
```
</p>
</details>

<details>

<summary> AST </summary>

<p>

```py
$ lfortran --show-python-ast --indent ser.txt
(Module [
   (FunctionDef 
      str_int 
      ([] [
         (
            x 
            (Name 
               i32 
               Load) ())] [] [] [] [] []) [
      (If 
         (Compare 
            (Name 
               x 
               Load) 
            Eq [
            (ConstantInt 0 ())]) [
         (Return 
            (ConstantStr "0" ()))] []) 
      (AnnAssign 
         (Name 
            result 
            Store) 
         (Name 
            str 
            Load) () 1) 
      (Assign [
         (Name 
            result 
            Store)] 
         (ConstantStr "" ()) ()) 
      (If 
         (Compare 
            (Name 
               x 
               Load) 
            Lt [
            (ConstantInt 0 ())]) [
         (AugAssign 
            (Name 
               result 
               Store) 
            Add 
            (ConstantStr "-" ())) 
         (Assign [
            (Name 
               x 
               Store)] 
            (UnaryOp 
               USub 
               (Name 
                  x 
                  Load)) ())] []) 
      (AnnAssign 
         (Name 
            rev_result 
            Store) 
         (Name 
            str 
[...]

PR #152, opened 2022-02-20
Implement `str` and `bool` runtime functions

namannimmo10:
Inspiration: https://github.com/lcompilers/lpython/issues/1#issuecomment-1002880770


namannimmo10:
We should do something about this line: https://github.com/lcompilers/lpython/blob/e1b2778dc30d413a6ce3abb698ff52c1598240f9/src/libasr/codegen/asr_to_llvm.cpp#L1518

Nonetheless, the ASR looks good to me.

certik:
Is this an intrinsic function? I don't see it here: https://docs.python.org/3/library/functions.html

Also it's time to refactor this whole function into LFortran's style. Let's not put anymore functions here until we refactor, as it is unmaintainable longterm. Let me know if you know how to do the refactor.

namannimmo10:
> Is this an intrinsic function?

No. We will add a generic function, `str()` down the road, but right now I'm trying to make `str_int` and `str_float` work.

I have two ideas for doing that:
1. Using decorators... 
2. Using type guard:
```python
>>> def str(val: i32 | f64):
...   # "isinstance" type guard
...   if isinstance(val, i32):
...     # type of `val` is narrowed to `i32`
...     return str_int(val)
...   elif isinstance(val, f64):
...     # else, type of `val` is narrowed to `f64`
...     return str_float(val)
...
>>>
```
You are talking about [this file](https://github.com/lcompilers/lpython/blob/main/src/lpython/semantics/comptime_eval.h), right?

namannimmo10:
Once we merge this, I will add the rest.

certik (approved):
I think that this looks great!

PR #153, opened 2022-02-21
Implement `visit_import`

Smit-create:


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -595,6 +595,37 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+        }
+        if (!main_module) {
+            st = st->parent;
+        }
+        for (auto &mod_sym : mods) {
+            bool ltypes;
+            t = (ASR::symbol_t*)(load_module(al, st,
+                mod_sym, x.base.base.loc, false, rl_path, ltypes,
+                [&](const std::string &msg, const Location &loc) { throw SemanticError(msg, loc); }
+                ));
```
Smit-create, line 614: We visit import in symbol table visitor and store the module name in `current_scope`. But still the following snippet gives the error:
```py
import math

def f():
    i: i32
    i = math.factorial(3)
    print(i)
```
ASR:
```
semantic error: module ''math' is not imported
 --> expr3.py:1:1
  |
1 | import math
  | ^ 
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -595,6 +595,37 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                // importing it.
+                tmp = nullptr;
+                continue;
+            }
+            if (!t) {
+                throw SemanticError("The module '" + mod_sym + "' cannot be loaded",
+                        x.base.base.loc);
+            }
+        }
+    }
```
certik, line 627: This code I think gets executed on `import math`. I think it adds the module into the symbol table. Unless there are bugs, this looks ok.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1845,20 +1880,84 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                }
+
+            } else {
+                throw SemanticError("Only Name type supported in Call",
+                    x.base.base.loc);
+            }
+        } else {
+            throw SemanticError("Only Name or Attribute type supported in Call",
+                x.base.base.loc);
+        }
```
certik, line 1960: This also looks great, that's how I would do it. I think what remains are bugs to identify and fix.

Smit-create: Yep. The main bug is now why `current_scope` is unable to find the module name in `visit_Call`, when we have stored it during `visit_Import`

certik: I think that's because `current_scope` is the local scope in the subroutine, and that scope does not have the module. The module lives at the top level.

certik: We should figure out how to handle nested modules, so that you can actually import a math module into a subroutine's scope (if you do `import math` inside a subroutine), in Fortran modules are global, and in Python I am currently not sure, they might also be global in some sense, but they can definitely be nested inside other modules and subroutines. But for now, let's implement the global approach. For that, just do something like:
```
while (symtab->parent != nullptr) symtab = symtab->parent;
```
That gives you the global scope symbol table. And then from there, just find the `math` module, it should be in this table.

Smit-create: Okay, trying that!

Smit-create:
yep, fixed the scope issue.

Smit-create:
Now, we can refactor the creation of `ExternalSymbol_t` so that we can use it in `Import` and `ImportFrom`. Will do that once this PR look good.

certik:
> Now, we can refactor the creation of `ExternalSymbol_t` so that we can use it in `Import` and `ImportFrom`. Will do that once this PR looks good.

Yes, we can do that in subsequent PRs.

Smit-create:
Good to go in?

certik (approved):
I think that's good to be merged. We can improve upon this in subsequent PRs.

Smit-create:
I'll refactor the code tomorrow to make it look cleaner.

PR #158, opened 2022-02-22
Refactor import

Smit-create:
See https://github.com/lcompilers/lpython/pull/153#issuecomment-1047005825


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -310,6 +310,75 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+            (ASR::symbol_t*)mv,
+            m->m_name, nullptr, 0, mv->m_name,
+            ASR::accessType::Public
+            );
+        return ASR::down_cast<ASR::symbol_t>(v);
+    } else {
+        throw SemanticError("Only Subroutines, Functions and Variables are currently supported in 'import'",
+            loc);
+    }
+    // should not reach here
```
certik, line 378:
```suggestion
    // should not reach here
    LFORTRAN_ASSERT(false)
```

certik (approved):
That looks good.

PR #165, opened 2022-02-23
Implement proper intrinsic functions handling

certik:
This uses the same design as in LFortran. In this PR we implemented an `abs` function, with proper compile time evaluation. Now we just need to port all the other functions to use this design, see the comments in the code.


namannimmo10 (approved):
Looks great! Thanks.

Issue #141, opened 2022-02-14
Figure out how to handle generic functions


certik:
So that you can implement a function `sin(x)` in LPython, that would work both for `f32` and `f64`.

certik:
So one good design might be:
```python
from ltypes import generic_function

@generic_function("sin")
def sin_f32(x: f32) -> f32:
    ...

@generic_function("sin")
def sin_f64(x: f64) -> f64:
    ...
```

And we implement the `generic_function` decorator in `ltypes.py` to be used by CPython. Then LPython has understanding of this decorator and it creates `GenericProcedure` ASR type for `sin`.

certik:
Python has `typing.overload` decorator: https://docs.python.org/3/library/typing.html#typing.overload, which is almost what we want. We could do `ltypes.overload`:
```python
from ltypes import overload

@overload
def sin(x: f32) -> f32:
    ...

@overload
def sin(x: f64) -> f64:
    ...
```

Smit-create:
Here is the example of python's overload:
```py
from typing import overload

@overload
def foo(x: int) -> int: ...

@overload
def foo(x: str) -> str: ...

def foo(x: str | int) -> str | int:
    return x

print(type(foo(2)), type(foo('2')))
```
Output: 
`<class 'int'> <class 'str'>`

For this we need to add the ellipsis (#146) both in AST and ASR.

Smit-create:
I tried implementing the `overload` operator in pure python which works in much similar fashion using this following script:
<details>
<summary> Python Script </summary>
<p>

```py
from inspect import getfullargspec, getcallargs

global_map = {}

class Dummy:
    def __init__(self, name):
        self.func_name = name

    def __call__(self, *args, **kwargs):
        func_map_list = global_map.get(self.func_name, False)
        if not func_map_list:
            raise Exception("Function not defined")
        for item in func_map_list:
            func, key = item
            try:
                ann_dict = getcallargs(func, *args, **kwargs)
            except:
                continue
            f = True
            for k, v in ann_dict.items():
                if not key.annotations.get(k, False):
                    f = False
                    break
                else:
                    if type(v) != key.annotations.get(k):
                        f = False
                        break
            if f:
                return func(*args, **kwargs)
        raise Exception("Function not found with matching signature")


def overload(f):
    dummy = Dummy(f.__name__)
    f_list = global_map.get(f.__name__, [])
    f_list.append((f, getfullargspec(f)))
    global_map[f.__name__] = f_list
[...]

PR #166, opened 2022-02-24
Implement Overload decorator in pure python

Smit-create:
* References: #141, https://github.com/lcompilers/lpython/issues/141#issuecomment-1049164124

### DIFF
--- a/src/runtime/ltypes/ltypes.py
+++ b/src/runtime/ltypes/ltypes.py
@@ -1,6 +1,53 @@
+from inspect import getfullargspec, getcallargs
+
+# data-types
+
 i32 = []
 i64 = []
 f32 = []
 f64 = []
 c32 = []
 c64 = []
+
+# overloading support
+
+class OverloadedFunction:
+    """
+    A wrapper class for allowing overloading.
+    """
+    global_map = {}
+
+    def __init__(self, func):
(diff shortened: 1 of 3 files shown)


src/runtime/ltypes/ltypes.py
```diff
@@ -1,6 +1,53 @@
...
+        func_map_list = self.global_map.get(self.func_name, False)
+        if not func_map_list:
+            raise Exception("Function not defined")
+        for item in func_map_list:
+            func, key = item
+            try:
+                # This might fail for the cases when arguments don't match
+                ann_dict = getcallargs(func, *args, **kwargs)
+            except TypeError:
+                continue
```
certik, line 36: If the arguments don't match, shouldn't we rather raise an exception that the arguments don't match?

Smit-create: Suppose, the list contains two functions `foo(a)` and `foo(a, b)`, then, `foo(1, 2)` will raise an error if we use `getcallargs` on the first function.

certik: Right. We should determine which of the overloads contains two integer arguments in this case. And just call that.

Smit-create: Yes, that's right!

PR #167, opened 2022-02-24
Implement C interoperation

certik:
For now in CPython only.

### DIFF
--- a/src/runtime/ltypes/ltypes.py
+++ b/src/runtime/ltypes/ltypes.py
@@ -10,7 +10,7 @@ f64 = types.new_class("f64")
 c32 = types.new_class("c32")
 c64 = types.new_class("c64")
 
-# overloading support
+# Overloading support
 
 def ltype(x):
     """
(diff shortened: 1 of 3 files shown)


certik:
What remains is to extract the argument types from annotations and pass them to ctypes. Then it should work.

src/runtime/ltypes/ltypes.py
```diff
@@ -51,3 +51,31 @@ def __call__(self, *args, **kwargs):
...
+    """
+    A wrapper class for interfacing C via ctypes.
+    """
+
+    def __init__(self, f):
+        self.name = f.__name__
+        self.args = f.__code__.co_varnames
+        self.annotations = f.__annotations__
+        lib = "/Users/ondrej/repos/lpython/src/runtime/liblfortran_runtime.dylib"
+        import ctypes
```
Smit-create, line 69: Can we move this `ctypes` as the class variable so it isn't imported at every object creation?

src/runtime/ltypes/ltypes.py
```diff
@@ -51,3 +51,31 @@ def __call__(self, *args, **kwargs):
...
+
+class CTypes:
+    """
+    A wrapper class for interfacing C via ctypes.
+    """
+
+    def __init__(self, f):
+        self.name = f.__name__
+        self.args = f.__code__.co_varnames
+        self.annotations = f.__annotations__
```
Smit-create, line 67: We can skip them if they aren't used anywhere in the class?

src/runtime/ltypes/ltypes.py
```diff
@@ -70,3 +71,45 @@ def __call__(self, *args, **kwargs):
...
+            if arg == f64:
+                return ctypes.c_double
+            elif arg == f32:
+                return ctypes.c_float
+            else:
+                raise NotImplementedError("Type not implemented")
+        self.name = f.__name__
+        self.args = f.__code__.co_varnames
+        self.annotations = f.__annotations__
+        lib = "/Users/ondrej/repos/lpython/src/runtime/liblfortran_runtime.dylib"
```
certik, line 95: Just the library lookup needs to be fixed, otherwise it works (on my computer).

src/runtime/ltypes/ltypes.py
certik: They are used now.

src/runtime/ltypes/ltypes.py
certik: I moved it to the top of the file.

certik:
This seems to work. We'll build upon it with subsequent PRs.

PR #168, opened 2022-02-24
Implement `ord()` and `chr()` functions

namannimmo10:
Included these in a separate PR because we have to fix the string length issue in LLVM. 👇 
 https://github.com/lcompilers/lpython/blob/e1b2778dc30d413a6ce3abb698ff52c1598240f9/src/libasr/codegen/asr_to_llvm.cpp#L1518


src/lpython/semantics/python_comptime_eval.h
```diff
@@ -168,6 +170,43 @@ struct PythonIntrinsicProcedures {
...
     }
 
+    static ASR::expr_t *eval_chr(Allocator &al, const Location &loc, Vec<ASR::expr_t*> &args) {
+        LFORTRAN_ASSERT(ASRUtils::all_args_evaluated(args));
+        ASR::expr_t* arg = args[0];
+        ASR::ttype_t* type = ASRUtils::expr_type(arg);
+        if (ASR::is_a<ASR::Integer_t>(*type)) {
+            int64_t c = ASR::down_cast<ASR::ConstantInteger_t>(arg)->m_n;
+            ASR::ttype_t* str_type =
+                LFortran::ASRUtils::TYPE(ASR::make_Character_t(al,
```
namannimmo10, line 180:
```suggestion
                ASRUtils::TYPE(ASR::make_Character_t(al,
```

integration_tests/test_builtin.py
```diff
@@ -1,20 +1,20 @@
-from ltypes import i32
-
```
certik, line 2: I think you need to keep these here, because both LPython and CPython requires it below for `i: i32`. CPython might ignore it by accident (I think it ignores local variables, but requires it for function signatures) and LPython might currently ignore it due to us not being strict about reporting errors, but the right solution is for LPython to always give an error message when `i32` is not imported.

certik (approved):
Otherwise this looks great. Much cleaner now!

PR #169, opened 2022-02-25
Add more runtime implementations of builtins

namannimmo10:


src/runtime/lpython_builtin.py
```diff
@@ -75,3 +75,18 @@ def bool(x: i32) -> bool:
...
+    """
+    Return the length of the string `s`.
+    """
+    if s == '':
+        return 0
+    count: i32
+    count = 0
+    end: i32
+    end = -1
+    while s[count:end] != '':
```
namannimmo10, line 90: This raises a codegen error when it hits this line 👇 

https://github.com/lcompilers/lpython/blob/19dd8a71ad4262587ff6e3c80482fdefdde86344/src/libasr/codegen/asr_to_llvm.cpp#L1022

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -207,6 +210,64 @@ struct PythonIntrinsicProcedures {
...
+        ASR::expr_t* arg1 = ASRUtils::expr_value(args[0]);
+        ASR::expr_t* arg2 = ASRUtils::expr_value(args[1]);
+        ASR::ttype_t* arg1_type = ASRUtils::expr_type(arg1);
+        ASR::ttype_t* arg2_type = ASRUtils::expr_type(arg2);
+        ASR::ttype_t *int_type = ASRUtils::TYPE(ASR::make_Integer_t(al, loc, 4, nullptr, 0));
+        ASR::ttype_t *real_type = ASRUtils::TYPE(ASR::make_Real_t(al, loc, 8, nullptr, 0));
+        if (!ASRUtils::check_equal_type(arg1_type, arg2_type)) {
+            throw SemanticError("The arguments to pow() must have the same type.", loc);
+        }
+        if (ASRUtils::is_integer(*arg1_type) || ASRUtils::is_integer(*arg1_type)) {
```
namannimmo10, line 253:
```suggestion
        if (ASRUtils::is_integer(*arg1_type) && ASRUtils::is_integer(*arg2_type)) {
```

src/runtime/lpython_builtin.py
certik: Right. `len` is a truly "builtin" function. It does not have an implementation using LPython. Implement it like this:
```python
def len(s: str) -> i32:
    """
    Return the length of the string `s`.
    """
    pass
```
The LLVM backend then has to provide its own implementation of `len` (which it already does).

certik: Other functions in this category are numpy's `size`, `shape`, etc. They all have to be implemented by the backend directly. Other functions like `chr` can (and should) be implemented in LPython.

src/runtime/lpython_builtin.py
```diff
@@ -75,3 +75,25 @@ def bool(x: i32) -> bool:
...
+        count += 1
+    return count + 1
+
+
+def pow(x: i32, y: i32) -> i32:
+    """
+    Returns x**y.
+    """
+    return (x)**(y)
```
certik, line 99: I would just write it like this:
```suggestion
    return x**y
```
but it's a matter of style.

certik:
This looks great. Do the change for `len` as I suggested above, and see if this works. If not, I can help fix it.

namannimmo10:
> This looks great. Do the change for `len` as I suggested above, and see if this works. If not, I can help fix it.

I tried using `pass`, but now it gives:
`code generation error: Unsupported len value in ASR`

certik:
Just do the change anyway, that's what has to happen. We might need to modify our AST to ASR to generate an interface for "len", not an implementation. Python doesn't have syntax for declaring an interface, we might perhaps later do it with a decorator.

certik:
Looks good. If there is a bug with `len`, please open up an issue.

PR #170, opened 2022-02-25
Support lpython types in overloading 

Smit-create:


certik (approved):
I think this is fine. The tests seem to pass.

I will note that we need to support some kind of an array notation such as `i32[:,:]` for a 2D array. Do you know how to do it with the `types.new_class()` approach?

Smit-create:
> Do you know how to do it with the `types.new_class()` approach?

I don't know exactly but will figure that out.

PR #174, opened 2022-02-26
Make pi, e, tau variables

certik:


certik:
Depends on #178 to be merged first.

PR #176, opened 2022-02-26
Implement `int` intrinsic function

namannimmo10:


certik (approved):
I think this is fine for now.

This will have to be later improved, and if the input is a variable of type float (or int), it should generate a cast at compile time, not a function call. This will require unevaluated argument, so we might need to add the true/false approach in the intrinsic table as in LFortran.

PR #177, opened 2022-02-26
Implement `float` intrinsic function

namannimmo10:


certik (approved):
I think this is fine. The same caveat as for `int()` to convert it to a cast, but we'll fix that later.

PR #179, opened 2022-02-27
Work on recognizing the @ccall decorator

certik:


certik:
@Smit-create here is how to recognize the `ccall` decorator by LPython and do something.

Smit-create:
Okay! Will have a look at it. Thanks!

Issue #175, opened 2022-02-26
String length issue in LLVM


namannimmo10:
```python
def test_len():
    s: str
    s = "abcd"
    assert len(s) == 4
    s = ''
    assert len(s) == 0
    assert len("abcd") == 4
    assert len("") == 0
```
⬆️  When we run this under integration tests, it returns a codegen error, as it hits this line 👇 
 https://github.com/lcompilers/lpython/blob/e1b2778dc30d413a6ce3abb698ff52c1598240f9/src/libasr/codegen/asr_to_llvm.cpp#L1518

certik:
I am reopening this, as it doesn't work yet in main, but I have a fix for this in my local branch.

PR #180, opened 2022-02-27
Implement compile time and runtime `bin`

namannimmo10:


integration_tests/run_tests.py
```diff
@@ -26,7 +26,8 @@
 # CPython tests only
 test_cpython = [
     "test_generics_01.py",
-    "test_c_interop_01.py"
+    "test_c_interop_01.py",
+    "test_builtin_bin.py"
```
namannimmo10, line 30: I will move this to "LPython + CPython tests" once we fix #175

certik (approved):
I think this looks good, it adds a compile time as well as a runtime implementation.

Would you please mind waiting until #181 is merged, and then resolving conflicts? I need that one to be merged, as I made a mistake and forgot to test it.

PR #184, opened 2022-02-28
Implement sin, cos, tan, sqrt

certik:
This interfaces C implementation from `src/runtime/impure/lfortran_intrinsics.c`.

### DIFF
--- a/integration_tests/test_math_02.py
+++ b/integration_tests/test_math_02.py
@@ -1,0 +1,10 @@
+from math import sin, pi
+
+def test_sin():
+    # TODO: importing from `math` doesn't work here yet:
+    pi: f64 = 3.141592653589793238462643383279502884197
+    eps: f64 = 1e-12
+    assert abs(sin(0.0)-0) < eps
+    assert abs(sin(pi/2)-1) < eps
+
+test_sin()
--- a/integration_tests/test_math_02.py
+++ b/integration_tests/test_math_02.py
@@ -1,10 +1,19 @@
-from math import sin, pi
+from math import sin, cos, tan, pi, sqrt
 
-def test_sin():
+def test_trig():
     # TODO: importing from `math` doesn't work here yet:
     pi: f64 = 3.141592653589793238462643383279502884197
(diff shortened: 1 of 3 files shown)


certik:
@Smit-create, se can use this approach to implement all other runtime functions. If you want, go ahead and implement the rest of the trig functions.

Smit-create:
This looks super good! Thanks.

Smit-create:
yeah, I'll add the remaining ones soon.

Smit-create (approved):

src/runtime/math.py
```diff
@@ -139,3 +139,27 @@ def lcm(a: i32, b: i32) -> i32:
...
+
+def sin(x: f64) -> f64:
+    return _lfortran_dsin(x)
+
+@ccall
+def _lfortran_dcos(x: f64) -> f64:
+    pass
+
+def cos(x: f64) -> f64:
+    return _lfortran_dcos(x)
```
namannimmo10, line 158: Why are we mixing `pure` and `impure` implementations?

certik: Right now everything is in one directory. This would be impure, because it depends on C.

integration_tests/test_math_02.py
```diff
@@ -0,0 +1,19 @@
...
+    assert abs(cos(pi/2)-0) < eps
+    assert abs(tan(0.0)-0) < eps
+    assert abs(tan(pi/4)-1) < eps
+
+def test_sqrt():
+    eps: f64 = 1e-12
+    assert abs(sqrt(2.0) - 1.4142135623730951) < eps
+    assert abs(sqrt(9.0) - 3) < eps
+
+test_trig()
```
Smit-create, line 19:
```suggestion
test_trig()
test_sqrt()
```

certik: Good catch!

src/runtime/math.py
namannimmo10: OK, we can separate them down the road.

PR #187, opened 2022-02-28
Add overload support in lpython

Smit-create:


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -809,9 +832,33 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         v.n_body = body.size();
     }
 
+    ASR::symbol_t* overloaddef_find_helper(std::string func_name, Vec<ASR::ttype_t*> args,
+                const Location &loc) {
+        for(auto &t: overload_defs[func_name]) {
+            bool ok = ASRUtils::select_func_subrout(t, args, loc,
+            [&](const std::string &msg, const Location &loc) { throw SemanticError(msg, loc); });
```
Smit-create, line 839: I'm trying to figure how to find the `overload_defs` in `BodyVisitor`, since it was defined in `SymbolTable` visitor.

certik:
I think the symbol table visitor can do all of this, correct? Or do we need to handle this in the body visitor also?

If just symbol table visitor, then I think what you can do is to start the ASR GenericProcedure node, and just keep adding to it. I think. You can see how this is done in LFortran, I can't remember right now. Otherwise using a temporary structure like you have done in this PR also works.

Smit-create:
> ASR GenericProcedure node,

Will have a look how that is done.

Smit-create:
On the current diff, I get:
```py
from ltypes import overload, i32

@overload
def foo(a: i32, b: i32) -> i32:
    return a*b

@overload
def foo(a: i32) -> i32:
    return a**2
```

ASR:
```bash
(TranslationUnit 
   (SymbolTable 
      1 
      {
         __lpython_overloaded_0__foo: 
            (Function 
               (SymbolTable 
                  2 
                  {
                     _lpython_return_variable: 
                        (Variable 
                           2 
                           _lpython_return_variable 
                           ReturnVar () () 
                           Default 
                           (Integer 4 []) 
                           Source 
                           Public 
                           Required .false.), 
                     a: 
                        (Variable 
                           2 
                           a 
                           In () () 
                           Default 
                           (Integer 4 []) 
                           Source 
                           Public 
                           Required .false.), 
                     b: 
                        (Variable 
                           2 
                           b 
                           In () () 
                           Default 
                           (Integer 4 []) 
                           Source 
                           Public 
                           Required .false.)
                  }) 
               __lpython_overloaded_0__foo [
               (Var 2 a) 
               (Var 2 b)] [
               (= 
                  (Var 2 _lpython_return_variable) 
                  (BinOp 
                     (Var 2 a) 
                     Mul 
                     (Var 2 b) 
                     (Integer 4 []) () ()) ()) 
               (Return)] 
               (Var 2 _lpython_return_variable) 
               Source 
               Public 
               Implementation ()), 
         __lpython_overloaded_1__foo: 
            (Function 
               (SymbolTable 
                  3 
                  {
                     _lpython_return_variable: 
                        (Variable 
                           3 
                           _lpython_return_variable 
                           ReturnVar () () 
                           Default 
                           (Integer 4 []) 
                           Source 
                           Public 
                           Required .false.), 
                     a: 
                        (Variable 
                           3 
                           a 
                           In () () 
                           Default 
                           (Integer 4 []) 
                           Source 
                           Public 
                           Required .false.)
                  }) 
               __lpython_overloaded_1__foo [
               (Var 3 a)] [
               (= 
                  (BinOp 
                     (Var 3 a) 
                     Pow (ConstantInteger 2 
                     (Integer 4 [])) 
                     (Integer 4 []) () ()) ()) 
               (Return)] 
               Source 
               Public 
               Implementation ()), 
         main_program: 
            (Program 
               (SymbolTable 
                  4 
                  {
                  }) 
               main_program [] [])
      }) 
   [])
```

Smit-create:
We will now need to add support for overload in `Call`

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -139,6 +139,121 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
     return mod2;
 }
 
+bool types_equal(const ASR::ttype_t &a, const ASR::ttype_t &b) {
```
certik, line 142: Use `ASRUtils::check_equal_type()`, as in:
```c++
        if (ASRUtils::check_equal_type(ASRUtils::expr_type(target),
                                    ASRUtils::expr_type(value)))`
```
If this `types_equal` does anything else, then let's improve the `ASRUtils::check_equal_type` function.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -139,6 +139,121 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
...
+        ASR::Function_t *fn
+            = ASR::down_cast<ASR::Function_t>(proc);
+        if (argument_types_match(args, *fn)) {
+            result = true;
+        }
+    } else {
+        err("Only Subroutine and Function supported in generic procedure", loc);
+    }
+    return result;
+}
```
certik, line 253: We should abstract this out from LFortran and LPython and put into `libasr` to be used by both.

certik:
The ASR example you posted is almost ok, but it needs to also add `foo: GenericProcedure(__lpython_overloaded_0__foo, __lpython_overloaded_1__foo)` into the symbol table.

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: Yeah, it's just the modified version of https://github.com/lcompilers/libasr/blob/82d115cf2e0886e794c8ebe22ff04c094370acc0/src/libasr/asr_utils.cpp#L630. It is present in `libasr`, but isn't included in any namespace.

Smit-create:
> The ASR example you posted is almost ok, but it needs to also add `foo: GenericProcedure(__lpython_overloaded_0__foo, __lpython_overloaded_1__foo)` into the symbol table.

I thought since we are changing the name of the function/subroutine, it might not be that important to use `GenericProcedure`.

src/lpython/semantics/python_ast_to_asr.cpp
certik: What modifications were needed? Let's improve the function from `libasr` and use it instead.

In order to reduce complexity of the frontend, we need to try to figure out how to "offload" as much as functionality into libasr, where it is shared across multiple frontends (LPython and LFortran for now, more in the future).

Effectively, these are all auxiliary routines to make building ASR easier. In this case, building GenericProcedures and using them easier.

certik:
What is your idea/design of making this work without a GenericProcedure in ASR? By storing the list of overloaded procedures in `overload_defs` inside AST to ASR, but not in ASR itself?

How would that work with independent compilation of modules to "mod" files, and then loading them in a new `lpython` invocation? 

The idea that we have followed so far is to keep ASR self-contained, so that it contains all the information that you might need later. It seems in this PR we are not storing all the information for generic subroutines in ASR, but keeping it outside of it.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -139,6 +139,53 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
     return mod2;
 }
 
+template <typename T>
+bool argument_types_match(const Vec<ASR::expr_t*> &args,
+        const T &sub) {
+    if (args.size() <= sub.n_args) {
+        size_t i;
+        for (i = 0; i < args.size(); i++) {
+            ASR::Variable_t *v = LFortran::ASRUtils::EXPR2VAR(sub.m_args[i]);
```
namannimmo10, line 148:
```suggestion
            ASR::Variable_t *v = ASRUtils::EXPR2VAR(sub.m_args[i]);
```

src/lpython/semantics/python_ast_to_asr.cpp
namannimmo10: I think they are almost identical, so I would use this from the `libasr` directly.

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: Code refactoring is still in progress, thanks!

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2227,15 +2331,36 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
             throw SemanticError("Unsupported call type for " + call_name,
                 x.base.base.loc);
         }
     }
+    int select_generic_procedure(const Vec<ASR::expr_t*> &args,
+            const ASR::GenericProcedure_t &p, Location loc) {
+        for (size_t i=0; i < p.n_procs; i++) {
+
+            if( ASR::is_a<ASR::ClassProcedure_t>(*p.m_procs[i]) ) {
+                ASR::ClassProcedure_t *clss_fn
```
Smit-create, line 2348: @certik  As discussed, we should add these helper functions in `asr_utils` right?

certik: Yes.

certik: No, simply modify files in the `libasr` subdirectory as part of this PR.

Smit-create:
The PR looks all good to go once the tests passes. Just need to figure out how to get rid of the global variable `ast_overload`.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -139,6 +139,8 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
     return mod2;
 }
 
+std::map<int, ASR::symbol_t*> ast_overload;
```
certik, line 142: The last part to do is to pass this via arguments to the constructor of the visitor classes and have it as a local variable in `python_ast_to_asr`. Then it will not clash when modules are imported recursively.

certik:

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: > Then it will not clash when modules are imported recursively.

That's fine, but I guess it won't be a big issue since we are storing the memory reference of each function which will indeed be unique.

certik: Only if you get lucky. When you load multiple modules, the visitor gets created, then destroyed. I think the nodes are allocated using our linear allocator, so I guess they would still be unique. But then when you compile multiple stuff from the main driver and happen to create a new allocator, then it will possibly collide. So we should not rely on that and have no global state.

Smit-create: How about using it as a class variable and clearing it completely when a symbol table visitor is instantiated?

certik: That will erase it when you import other modules recursively, wouldn't it?

Smit-create: Ahh, yes! One probable approach can be to add an optional arguement in `FunctionDef` to store the index where it will be present in `GenericNode`.

certik:
Let me know if you need any help with implementing the passing of the `ast_overload` into the visitor patterns via their constructors.

certik (approved):
I moved the `ast_overload` into a local variable, so it should be good to go in now. Thanks @Smit-create !

Smit-create:
Thanks for updating the PR and merging!!

PR #189, opened 2022-02-28
Improve compile time `abs` implementation

namannimmo10:


src/lpython/semantics/python_comptime_eval.h
```diff
@@ -96,24 +96,30 @@ struct PythonIntrinsicProcedures {
         if (args.size() != 1) {
             throw SemanticError("Intrinsic abs function accepts exactly 1 argument", loc);
         }
-        ASR::expr_t* trig_arg = args[0];
+        ASR::expr_t* arg = ASRUtils::expr_value(args[0]);
```
certik, line 99: Currently we evaluate all arguments and only if `value` is present, we call `eval_abs` on this value. Thus `ASRUtils::expr_value(args[0])` is equivalent to just `args[0]`.
```suggestion
        ASR::expr_t* arg = args[0];
```

tests/constants1.py
```diff
@@ -19,15 +19,15 @@ def test_ord_chr():
 
 
 def test_abs():
-    # a: i32
-    # a = abs(5)
-    # a = abs(-500)
-    # a = abs(False)
-    # a = abs(True)
+    a: i32
+    a = abs(5)
```
certik, line 23: I think this should create the correct "value" (at compile time). However, I think it also creates a `FunctionCall` that calls `abs(float)` with `abs(5)` and that will fail verify (once we update verify to be more strict).

What we need is to first fix #141, and then implement a runtime version of `abs` for other types than float.

certik: I think this is still producing an incorrect ASR, see the reference results. So it should not be committed in my opinion.

certik: The new reference ASR in this PR contains:
```
(FunctionCall 4 abs () [(ConstantInteger 5 (Integer 4 []))] [] (Real 8 []) (ConstantInteger 5 (Integer 4 [])) ()) ()) 
```
Which is incorrect, because it is calling an abs function that expects a float, but passing in integer as an argument. So we should not commit this test, as this is not correct.

I'll try to robustify our verify very soon, so that it triggers an error, then it will be obvious.

namannimmo10:
I think you're right, we definitely need to implement a generic `abs` function, so as to generate the correct ASR.

namannimmo10:
I will decorate the `abs` with overload (as added in https://github.com/lcompilers/lpython/pull/166) and get this PR merged soon.

certik:
Overloads don't work in LPython yet, see #187.

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -96,24 +96,30 @@ struct PythonIntrinsicProcedures {
         if (args.size() != 1) {
             throw SemanticError("Intrinsic abs function accepts exactly 1 argument", loc);
         }
-        ASR::expr_t* trig_arg = args[0];
+        ASR::expr_t* arg = args[0];
         ASR::ttype_t* t = ASRUtils::expr_type(args[0]);
-        if (ASR::is_a<ASR::Real_t>(*t)) {
-            double rv = ASR::down_cast<ASR::ConstantReal_t>(trig_arg)->m_r;
+        ASR::ttype_t *int_type = ASRUtils::TYPE(ASR::make_Integer_t(al, loc, 4, nullptr, 0));
+        ASR::ttype_t *real_type = ASRUtils::TYPE(ASR::make_Real_t(al, loc, 8, nullptr, 0));
```
czgdp1807, line 102: Can we take the kind from `args[0]`'s type instead of defaulting it to `8` or `4`?

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -96,24 +96,30 @@ struct PythonIntrinsicProcedures {
...
+            int64_t rv = ASR::down_cast<ASR::ConstantInteger_t>(arg)->m_n;
             int64_t val = std::abs(rv);
-            return ASR::down_cast<ASR::expr_t>(ASR::make_ConstantInteger_t(al, loc, val, t));
-        } else if (ASR::is_a<ASR::Complex_t>(*t)) {
-            double re = ASR::down_cast<ASR::ConstantComplex_t>(trig_arg)->m_re;
-            double im = ASR::down_cast<ASR::ConstantComplex_t>(trig_arg)->m_im;
+            return ASR::down_cast<ASR::expr_t>(ASR::make_ConstantInteger_t(al, loc, val, int_type));
+        } else if (ASRUtils::is_logical(*t)) {
+            bool rv = ASR::down_cast<ASR::ConstantLogical_t>(arg)->m_value;
+            int8_t val = rv ? 1 : 0;
```
czgdp1807, line 113:
```suggestion
            int8_t val = rv;
```

I think auto cast happens in C++?

namannimmo10:
This was included in https://github.com/lcompilers/lpython/pull/250. Thanks!

src/lpython/semantics/python_comptime_eval.h
namannimmo10: Yes, we should, perhaps everywhere in the codebase.

PR #192, opened 2022-03-02
Reallocate LHS for allocatable strings

certik:


certik:
The basic use case was fixed in #194, here we need to implement proper reallocation to actually free the memory correctly to fix all leaks.

certik:
I think the way to do this initially is to have the `tmp` llvm value to "own" the data and if it is allocatable, it must release it. Constant strings must always be copied, so that they can be destroyed.

Later, we can add optimizations to this scheme.

czgdp1807:
I think we need to work on this. @certik

PR #194, opened 2022-03-03
LLVM: get allocatable string declaration working

certik:
Basic assignment also works, tests added.

This implementation leaks memory, the strings are never deallocated at runtime. This has to be fixed properly by #192.


namannimmo10:
Thanks, this looks good!

PR #199, opened 2022-03-04
Implement `round` built-in function

namannimmo10:
We can implement a generic `round` once `overload` is supported in LPython (#187), but for now this works for float arguments.

TODO:

- [x]  fix the case when the compile-time value is 0.5 (Outputs with Python and C++ are different: 0 and 1 respectively)


src/runtime/lpython_builtin.py
```diff
@@ -173,3 +173,13 @@ def oct(n: i32) -> str:
...
+
+
+def round(value: f64) -> i32:
+    """
+    Rounds a floating point number to the nearest integer.
+    """
+    if abs(value - int(value)) <= 0.5:
+        return int(value)
+    else:
+        return int(value) + 1
```
certik, line 185: Isn't round the same thing as `int(value+0.5)`? Possibly ±1 for negative values?

certik:
Looks good so far.

src/runtime/lpython_builtin.py
namannimmo10: > Isn't `round` the same thing as `int(value+0.5)`? 

Actually, this gives 1 when the `value` is 0.5, but Python returns 0 for `round(0.5)`

namannimmo10:
Gives garbage values for the 6 `print` statements:
```
1075230278
1090940434
1078307718
1071644672
1078575105
1071644672
```
I think that's because we are calling other built-in functions in our current runtime implementation.

certik:
I usually get the function (`round` in this case) working first in a simple script. Then put it into the runtime library.

src/runtime/lpython_builtin.py
certik: Interesting, you are right:
```
>>> round(0.5)
0
```

certik: Here is the CPython's implementation: https://github.com/python/cpython/blob/5ab745fc51e159ead28b523414e52f0bcc1ef353/Objects/floatobject.c#L1058, it looks like they do something like this:
```c
        rounded = round(x); // the same as int(x+0.5) ?
        if (fabs(x-rounded) == 0.5)
            rounded = 2.0*round(x/2.0);
```

namannimmo10:
OK, this is ready for now; let me know if you have any suggestions or I'll merge this. Generic versions of these built-in functions will be added soon.

certik (approved):
I think this looks great!

PR #201, opened 2022-03-05
Implement `math.hypot` and `math.copysign`

namannimmo10:


integration_tests/test_math.py
```diff
@@ -84,6 +84,24 @@ def test_lcm():
...
+
+
+def test_hypot():
+    f: f64
+    f = hypot(3, 4)
+    assert f == 5.0
+    f = hypot(-3, 4)
+    assert f == 5.0
+    f = hypot(6, 6)
+    assert f == 8.48528137423857
```
certik, line 102: Test these floating point numbers using the `abs(x-y) < eps` idiom. Otherwise it is very fragile.

certik:
Otherwise this looks good.

integration_tests/test_math.py
```diff
@@ -1,5 +1,5 @@
 from math import (factorial, isqrt, perm, comb, degrees, radians, exp, pow,
-                  ldexp, fabs, gcd, lcm)
+                  ldexp, fabs, gcd, lcm, copysign, hypot)
```
certik, line 2: You can remove this change now.

certik:
👍

PR #202, opened 2022-03-06
Initial implementation of the `random` module

namannimmo10:


certik (approved):
I think that looks good. The `rand()` function we can probably implement in LPython itself later on, but calling it from C is fine for now.

PR #205, opened 2022-03-08
Make runtime string comparison work 

namannimmo10:
TODO:
- [ ] Insert these functions by the frontend


certik:
Yes, the tricky part is to now insert these functions for string comparisons in ast to asr, but not for character comparison, as those have to be still handled by the backend, because the implementation of these functions uses it.

namannimmo10:
I will have to see how to do that -- this can go in, though.

certik (approved):
Yes, I think that this is good.

PR #209, opened 2022-03-10
Add initial statistics support

Smit-create:


certik:
It's probably the lists that don't work yet. I would not worry about it until we implement lists.

You can try to implement the numpy versions of these:

https://numpy.org/doc/stable/reference/routines.statistics.html

At least the 1D f64 versions might work.

namannimmo10:
@Smit-create -- can we close this PR? #901 is taking over.

PR #210, opened 2022-03-10
implement builtin sorted function

Smit-create:


src/runtime/lpython_builtin.py
```diff
@@ -183,3 +183,43 @@ def round(value: f64) -> i32:
...
+        else:
+            break
+    tmp = array[start]
+    array[start] = array[high]
+    array[high] = tmp
+
+    return high
+
+
+def quick_sort(array: list[f64], start: i32, end: i32):
```
namannimmo10, line 215:
```suggestion
def _quick_sort(array: list[f64], start: i32, end: i32):
```

src/runtime/lpython_builtin.py
```diff
@@ -183,3 +183,43 @@ def round(value: f64) -> i32:
...
+    if start >= end:
+        return
+    p: i32
+    p = _partition(array, start, end)
+    quick_sort(array, start, p-1)
+    quick_sort(array, p+1, end)
+
+
+def sorted(array: list[f64]):
+    quick_sort(array, 0, len(array)-1)
```
czgdp1807, line 225: Well Python uses [Tim sort](https://en.wikipedia.org/wiki/Timsort), a combination of merge and insertion sort. Quick sort's time complexity depends on pivot selection and the ordering of input elements. Merge sort's time complexity is stable in that sense. So, I believe there must be good reasons to use quick sort and not Tim sort here.

czgdp1807: In addition I think this recursive implementation is limiting (due to the maximum number of recursions one can do on a system). For large arrays this will fail. For small arrays, insertion sort should be faster anyways. So my suggestion - Quick sort (or a better choice Tim sort) can be implemented recursively. See, https://github.com/codezonediitj/pydatastructs/blob/3abc0264cb97fcb90b82e95eed4b1bc6b1148049/pydatastructs/linear_data_structures/algorithms.py#L751-L790 for example.

certik: I think `quick_sort` is easy to implement, so we can start with that. I think Timsort is better, but much harder to implement. Once we have it, we should switch it.

czgdp1807: Sure. @Smit-create Is there any blocker which is preventing from implementing this feature? Or is it low on the priority list?

integration_tests/sorted1.py
```diff
@@ -0,0 +1,13 @@
...
+    x: list[f64] = [35.9, 24.9, 23.1, 4223.5, 64.34, 3095.34]
+    x = sorted(x)
+    ans: list[f64] = [23.1, 24.9, 35.9, 64.34, 3095.34, 4223.5]
+    eps: f64 = 1e-12
+    e: f64
+    i: i32
+    for i in range(len(x)):
+        assert abs(x[i]-ans[i]) < eps
+
+f()
```
czgdp1807, line 13: Take a larger list of integers say 100-200, fill the list as `[n, n - 1, n - 2, ...., 0]` and then sort it. Then check for correctness. May be benchmark your algorithm separately for comparisons with C++ sorting algorithms. Let's see where we stand.

czgdp1807 (changes requested):

Smit-create:
This PR will require CPP backend support for lists to make the tests pass.

Smit-create:
I can try adding Lists in the CPP backend once we finalize the design for the same. I will open a separate issue for this.

integration_tests/sorted1.py
certik: I think that would be fine, as long as it runs quickly (=immediately).

certik:
I think tests have to be updated, so that they pass.

Smit-create:
> I think tests have to be updated, so that they pass.

I'll update that once we implement the `List` in CPP.

czgdp1807:
> I'll update that once we implement the List in CPP.

We have the lists in CPP. What else is needed to complete this PR?

Smit-create:
I see some strange behaviour with this PR:
Doing this works fine:
```
% lpython integration_tests/test_builtin_sorted.py
ok
```
But while running the integration tests, I get an error for the same file:
```
  store %complex_8 %4, double* %pivot, align 8
 doubleBoth operands to FCmp instruction are not of the same type!
  %21 = fcmp ule %complex_8 %19, double %20
  store %complex_8 %26, double* %tmp, align 8
 doubleStored value type does not match pointer operand type!
  store double %50, %complex_8* %49, align 8
 %complex_8 = type { double, double }Stored value type does not match pointer operand type!
  store %complex_8 %57, double* %tmp, align 8
 doubleStored value type does not match pointer operand type!
  store double %81, %complex_8* %80, align 8
 %complex_8 = type { double, double }
```

PR #212, opened 2022-03-10
Implement numpy.empty()

certik:
Test added.


certik:
The numpy library has to be added to the CI I am guessing, tests will probably fail.

certik:
Now it should work. The next step is to implement `zeros`, `ones` as regular LPython functions.

PR #216, opened 2022-03-11
Rename `_lfortran_main_program` to `_lpython_main_program` in the ASR

namannimmo10:


src/libasr/pass/global_stmts_program.cpp
```diff
@@ -19,7 +19,7 @@ using LFortran::ASRUtils::EXPR;
  */
 void pass_wrap_global_stmts_into_program(Allocator &al,
             ASR::TranslationUnit_t &unit) {
-    std::string fn_name_s = "_lfortran_main_program";
+    std::string fn_name_s = "_lpython_main_program";
```
certik, line 22: I see. That's not going to work, as it will break LFortran. What we need is to pass the name as an argument to this function, then LPython can set it to `_lpython_main_program`, and LFortran can set it to `_lfortran_main_program`.

certik: Everything in `libasr` is shared with LFortran, so it needs to be language neutral.

namannimmo10:
Let me know if this is fine -- and I will make the changes in LFortran.

src/libasr/pass/global_stmts_program.cpp
```diff
@@ -18,15 +18,14 @@ using LFortran::ASRUtils::EXPR;
  *
  */
 void pass_wrap_global_stmts_into_program(Allocator &al,
-            ASR::TranslationUnit_t &unit) {
-    std::string fn_name_s = "_lfortran_main_program";
+            ASR::TranslationUnit_t &unit, std::string name) {
```
certik, line 21: Let's make the argument more descriptive, such as:
```suggestion
            ASR::TranslationUnit_t &unit, std::string program_fn_name) {
```

certik:
Otherwise this looks great!

src/libasr/pass/global_stmts_program.cpp
certik: Also, pass it with `&`.

certik: And `const`.

certik (approved):
I think this looks good to be merged now.

namannimmo10:
Thanks!

PR #219, opened 2022-03-12
Fix implicit cast issue

### LINKED ISSUE #217: Needs Implementation: Increment in Python
Sample:
```
def main0():
    i:i64
    i = 1
    i+=1
    print(i)

main0()
```

Exception:
```
; ModuleID = 'LFortran'
source_filename = "LFortran"

@0 = private unnamed_addr constant [6 x i8] c"%lld\0A\00", align 1

define void @_lfortran_main_program() {
.entry:
  call void @main0()
  br label %return

  ret void
}

define void @main0() {
.entry:
  %i = alloca i64, align 8
  store i64 1, i64* %i, align 4
  %0 = load i64, i64* %i, align 4
  %1 = trunc i64 %0 to i32
  %2 = add i32 %1, 1
  store i32 %2, i64* %i, align 4
  %3 = load i64, i64* %i, align 4
  call void (i8*, ...) @_lfortran_printf(i8* getelementptr inbounds ([6 x i8], [6 x i8]* @0, i32 0, i32 0), i64 %3)
  br label %return

  ret void
}

define i32 @main() {
.entry:
  call void @_lfortran_main_program()
  ret i32 0
}
  File "/Users/anesa00/owork/lpython/src/bin/lpython.cpp", line 1540
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir, compiler_options);
  File "/Users/anesa00/owork/lpython/src/bin/lpython.cpp", line 742
    res = fe.get_llvm3(*asr, diagnostics);
  File "/Users/anesa00/owork/lpython/src/lpython/fortran_evaluator.cpp", line 309
    if (res.ok) {
  File "/Users/anesa00/owork/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 4073
    diagnostics.diagnostics.push_back(diag::Diagnostic(msg,
  store i32 %2, i64* %i, align 4
 i64


```

Smit-create:
Fixes #217


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -987,15 +987,15 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         if (ASRUtils::is_integer(*left_type) && ASRUtils::is_integer(*right_type)) {
             bool is_l64 = ASR::down_cast<ASR::Integer_t>(left_type)->m_kind == 8;
             bool is_r64 = ASR::down_cast<ASR::Integer_t>(right_type)->m_kind == 8;
-            if (is_l64 != is_r64) {
+            if (is_l64 && !is_r64) {
```
certik, line 990: I think this breaks binary operations like `a + b` if right is 64 bit and left is 32 bit.

In binary operations, the cast has to happen from 32 bit to 64 bit no matter if it is LHS or RHS that needs casting.

Smit-create: I think it won't be an issue as it will be handled by:
https://github.com/lcompilers/lpython/blob/b3154087db644b951b6fef745408c3a875a9a207/src/lpython/semantics/python_ast_to_asr.cpp#L1379-L1383

certik: ok. Let me carefully review this to make sure we don't introduce an error here.

Smit-create: It can be verified from the following example:
```py
def main0():
    i: i64
    j: i32
    i = 0
    j = 1
    print(i + j)
    print(j + i)

main0()
```

<details>
<summary> ASR </summary>
<p>

```bash
(TranslationUnit 
   (SymbolTable 
      1 
      {
         _lpython_main_program: 
            (Subroutine 
               (SymbolTable 
                  4 
                  {
                  }) 
               _lpython_main_program [] [
               (SubroutineCall 1 main0 () [] ())] 
               Source 
               Public 
         main0: 
            (Subroutine 
               (SymbolTable 
                  2 
                  {
                     i: 
                        (Variable 
                           2 
                           i 
                           Local () () 
                           Default 
                           (Integer 8 []) 
                           Source 
                           Public 
                           Required .false.), 
                     j: 
                        (Variable 
                           2 
                           j 
                           Local () () 
                           Default 
                           (Integer 4 []) 
                           Source 
                           Public 
                           Required .false.)
                  }) 
               main0 [] [
               (= 
                  (Var 2 i) 
                  (ImplicitCast (ConstantInteger 0 
                     (Integer 4 [])) 
                     IntegerToInteger 
                     (Integer 8 []) ()) ()) 
               (= 
                  (Var 2 j) (ConstantInteger 1 
                  (Integer 4 [])) ()) 
               (Print () [
                  (BinOp 
                     (Var 2 i) 
                     Add 
                     (ImplicitCast 
                        (Var 2 j) 
                        IntegerToInteger 
                        (Integer 8 []) ()) 
                     (Integer 8 []) () ())]) 
               (Print () [
                  (BinOp 
                     (ImplicitCast 
                        (Var 2 j) 
                        IntegerToInteger 
                        (Integer 8 []) ()) 
                     Add 
                     (Var 2 i) 
                     (Integer 8 []) () ())])] 
               Source 
               Public 
         main_program: 
            (Program 
               (SymbolTable 
                  3 
                  {
                  }) 
               main_program [] [
      }) 
   [])
```

</p>
</details>

certik:
P.S. Every time reference tests are modified, one has to very carefully check that each change is correct and what we want. I wrote up in the previous comment in detail how I do it. Often times, the changes in reference tests are relatively trivial (such as adding some new intrinsic function, or symbol table ID changes, etc.), so it's fast to review. But changes in implicit casts are usually slow to review, one must ensure that every change is correct.

We should actually improve our verify to check this for us, and it would fail above. Right now verify is not strong enough. We should do that soon, it's not difficult.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1001,15 +1001,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
             bool is_r64 = ASR::down_cast<ASR::Integer_t>(right_type)->m_kind == 8;
-            if (is_l64 != is_r64) {
+            if (is_assign) {
+                if (is_l64 != is_r64) {
+                    // here we need to cast to left type strictly
+                    return ASR::down_cast<ASR::expr_t>(ASR::make_ImplicitCast_t(
+                        al, right->base.loc, right, ASR::cast_kindType::IntegerToInteger,
+                        left_type, nullptr));
+                }
+            } else if (is_l64 && !is_r64) {
```
certik, line 1011: Would this work?
```suggestion
            if ((is_assign && is_l64 != is_r64) || (is_l64 && !is_r64)) {
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1001,15 +1001,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
             bool is_r64 = ASR::down_cast<ASR::Real_t>(right_type)->m_kind == 8;
-            if (is_l64 != is_r64) {
+            if (is_assign) {
+                if (is_l64 != is_r64) {
+                    // here we need to cast to left type strictly
+                    return ASR::down_cast<ASR::expr_t>(ASR::make_ImplicitCast_t(
+                        al, right->base.loc, right, ASR::cast_kindType::RealToReal,
+                        left_type, nullptr));
+                }
+            } else if (is_l64 && !is_r64) {
```
certik, line 1026: And the same here.

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: Done thanks!

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1001,15 +1001,15 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         if (ASRUtils::is_integer(*left_type) && ASRUtils::is_integer(*right_type)) {
             bool is_l64 = ASR::down_cast<ASR::Integer_t>(left_type)->m_kind == 8;
             bool is_r64 = ASR::down_cast<ASR::Integer_t>(right_type)->m_kind == 8;
-            if (is_l64 != is_r64) {
+            if ((is_assign && is_l64 != is_r64) || (is_l64 && !is_r64)) {
```
certik, line 1004: I think I wrote it incorrectly, sorry about that. This should be correct:
```suggestion
            if ((is_assign && (is_l64 != is_r64)) || (is_l64 && !is_r64)) {
```
I think the `&&` has precedence over `!=` otherwise, which we do not want.

Smit-create: Okay, I'll fix that.

Smit-create:
Looks good to go in?

certik:
Ok, so the only change in tests is in `loop1`, which is the correct fix, instead of first casting 64 bit to 32 bit, and then back to 64 bit, it casts 32 bit right away to 64 bit and stays 64 bit. No other change, so this looks good.

Before you merge, can you please rework the git history to a few polished commits, such as adding a new test (1), fixing the bug (2), and updating reference tests (3)?

certik:
Awesome, thanks!

PR #220, opened 2022-03-12
Make runtime string repetition work

namannimmo10:


integration_tests/test_builtin_str_02.py
```diff
@@ -38,6 +40,17 @@ def _lpython_strcmp_gt(a: str, b: str) -> bool:
 def _lpython_strcmp_gteq(a: str, b: str) -> bool:
     return _lpython_strcmp_eq(a, b) or _lpython_strcmp_gt(a, b)
 
+def _lpython_str_concat(a: str, b: str) -> str:
+    return a + b
```
certik, line 44: This function must be implemented by copying the strings. Doing "+" won't work, since "+" will be implemented by calling this function.

Thirumalai-Shaktivel: https://github.com/lcompilers/lpython/blob/main/src/runtime/impure/lfortran_intrinsics.c#L532-L550
I guess we can use this or implement the same as this.

certik (approved):

PR #221, opened 2022-03-13
Allow binop on logical types

namannimmo10:


src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -764,7 +764,7 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
             case (ASR::binopType::Mul) : { last_expr_precedence = 5; break; }
             case (ASR::binopType::Div) : { last_expr_precedence = 5; break; }
             case (ASR::binopType::Pow) : {
-                src = "std::pow(" + left + ", " + right + ")";
+                src = "pow(" + left + ", " + right + ")";
```
certik, line 767: I think you need `std::pow`, don't you? Just like `std::abs`.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1536,6 +1539,24 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                                                    ASRUtils::expr_value(left))->m_value;
+                int8_t right_value = ASR::down_cast<ASR::ConstantLogical_t>(
+                                                    ASRUtils::expr_value(right))->m_value;
+                int8_t result;
+                switch (op) {
+                    case (ASR::binopType::Add): { result = left_value + right_value; break; }
+                    case (ASR::binopType::Sub): { result = left_value - right_value; break; }
+                    case (ASR::binopType::Mul): { result = left_value * right_value; break; }
+                    case (ASR::binopType::Pow): { result = std::pow(left_value, right_value); break; }
+                    case (ASR::binopType::Div): { } // TODO: Handle division of logicals
```
certik, line 1553: I learned something new, I didn't know you can do this in Python, but I guess you can:
```
In [3]: True + True
Out[3]: 2

In [4]: True * True
Out[4]: 1

In [5]: True / True
Out[5]: 1.0
```

certik:
Otherwise looks good.

src/lpython/semantics/python_ast_to_asr.cpp
namannimmo10: Yes.
```python
>>> isinstance(True, int)
True
```

src/libasr/codegen/asr_to_cpp.cpp
namannimmo10: Well, actually both will compile, but I can make the changes.

certik: The `<cmath>` has `std::pow`. It seems it compiles without `std::` in front, but then it can also be the `pow` from `<math.h>` which we do not want, so it's safer to always use the `std` namespace.

namannimmo10: I see, thanks!

PR #222, opened 2022-03-13
Extending more Math functions

Abdullahjavednesar:
Adding few more math functions.


integration_tests/test_math.py
```diff
@@ -83,6 +83,34 @@ def test_lcm():
...
 
+def test_log1p():
+    i: f64
+    i = 10.0
+    assert log1p(i) == log(i+1)
+
+def test_expm1():
+    i: f64
+    i = 2.0
+    assert expm1(i) == exp(i) - 1
```
Smit-create, line 94: Hi, thanks for the PR. Just some suggestion here:
```suggestion
def test_expm1():
    i: f64
    eps: f64
    eps = 1e-12
    i = 2.0
    assert abs(expm1(i) - exp(i) + 1) < eps
```

Abdullahjavednesar: Hi @Smit-create, Thanks for the review!

certik:
I think that this is good, but now we need to improve LPython to be able to compile it.

certik:
Regarding the last commit, would you please mind splitting formatting changes into a separate commit? It makes it hard to review exactly what the change is, if it is mixed with formatting.

I think `ceiling` and `floor` can both be implemented in Python itself.  I think we just need casting from a floating point to integer, which I think we do, or if not, should implement. That way, the compiler then then inline it and optimize it out if requested, while if it calls into C, it can't optimize the call out.

Abdullahjavednesar:
>Regarding the last commit, would you please mind splitting formatting changes into a separate commit? It makes it hard to review exactly what the change is, if it is mixed with formatting.

Sure, that makes sense. I'll split the above commit accordingly.

>I think `ceiling` and `floor` can both be implemented in Python itself. I think we just need casting from a floating point to integer, which I think we do, or if not, should implement. That way, the compiler then then inline it and optimize it out if requested, while if it calls into C, it can't optimize the call out.

For both `ceil` and `floor` I initially was trying to return `f64` type itself, but I see python `ceil`, `floor` functions returns integer values. 
```
>>> math.ceil(2.4)
3
>>> math.floor(2.4)
2
```
What was failing in the previous commit was the floor division operator `//` for `f64` types. Anyways I'll try to figure out how to implement casting of floating point to integer,  any leads on the same would be appreciated.

certik:
I think the `floor` function is effectively just `int()`:
```
>>> int(3.5)
3
>>> int(3.3)
3
>>> int(3.)
3
>>> int(3.9)
3
>>> int(4.1)
4
```
So you can try that. Perhaps for negative numbers you have to subtract or add 1.

The ceil is like floor but you might need to add one or something like that.

certik:
Just implement it in CPython, and then if it doesn't work in LPython, report a bug and we'll fix it.

Abdullahjavednesar:
>I think the floor function is effectively just int():

Yes, I'll implement that way and keep it straight. But before that we need to work on #254.

Abdullahjavednesar:
@certik can you please review this.

src/runtime/math.py
```diff
@@ -296,3 +296,42 @@ def _lfortran_datanh(x: f64) -> f64:
...
+def ceil(x: f64) -> i32:
+    if x > 0 and fmod(x, 1.0) != 0:
+        return int(x)+1
+    return int(x)
+
+
+def floor(x: f64) -> i32:
+    if x < 0 and fmod(x, 1.0) != 0:
+        return int(x)-1
+    return int(x)
```
certik, line 310: Why is the `fmod` part needed? LFortran has this implementation of `floor`: https://gitlab.com/lfortran/lfortran/-/blob/28fa2448c6a1c950d2eb7ca1ecc8f0c9b84293b1/src/runtime/pure/lfortran_intrinsic_math2.f90#L135

Abdullahjavednesar: In condition ` if x < 0 and fmod(x, 1.0) != 0:`,  for floor calculation if the number, say x is negative and has non zero fractional part it should return int(x)-1. `fmod(x, 1.0)` is required to compute the fractional part.

Abdullahjavednesar:
Removed `floor`, `ceil` implementation as already merged.

integration_tests/test_math.py
```diff
@@ -106,6 +106,12 @@ def test_ceil():
     assert i == -13
 
 
+def test_remainder():
```
namannimmo10, line 109: We should run this test inside `check()`.

Abdullahjavednesar: Good catch, added it to check.

certik (approved):

certik:
Can you please add tests for `fmod`, `expm1` and `logp1`?

Abdullahjavednesar:
@certik, I have added test cases for all the function, can you please review.

Smit-create (approved):

certik (approved):

Issue #200, opened 2022-03-04
Which modules to implement from the Python standard library


certik:
The Python standard library has a lot of modules: https://docs.python.org/3/library/index.html, but many of them are rarely used. First we should concentrate on the most commonly used modules, probably:

* `sys`
* `os`
* `string`
* `datetime`
* `cmath`
* `decimal`
* `numbers`
* `fractions`
* `statistics`

Smit-create:
I can work on implementing `statistics`.

namannimmo10:
I can do the `random` module

Oshanath:
Should these modules be implemented in C++ or in LPython itself?

Oshanath:
Basically if I were to implement something from this list, what I should do is refer to the Python documentation of that module and implement those same functions the same way using LPython. Am I correct?

certik:
Yes, and add them to `src/runtime` and add tests to `integration_tests` and register them in `integration_tests/run_tests.py`, they will be run via CPython (to ensure the test script is written correctly) as well as LPython (to ensure everything actually compiles).

rajpratyush:
I think itertools, datetime, sys and os should also be added to the lpython since they are used extensively in scientific computing. i can start the work for sys and os modules.

namannimmo10:
> I think itertools, datetime, sys and os should also be added to the lpython since they are used extensively in scientific computing. i can start the work for sys and os modules.

Great. Let us know if you need any help!

rajpratyush:
sure could you provide me up links to the development guide so that I can set up the project locally. or any kind of documentation website

namannimmo10:
You can find the installation guide in the README.md file.  For now, see https://docs.lfortran.org/developer_tutorial/ for more information -- LPython follows the same design as in LFortran.

rajpratyush:
so it doesn't have a documentation website ?

certik:
We currently don't have dedicated LPython documentation. We will eventually add it.

rajpratyush:
how about building one?

certik:
Yes, we should.

Umesh-JNU:
I would like to work for argparse. Can you please help me how to start?

kmr-srbh:
Hello! I got to know about LPython through the GSoC page of Python. I am interested in helping LPython develop these modules for science. I have a strong command over Python and a keen interest in programming languages.

[...]

PR #228, opened 2022-03-14
Implement cmath python module

Smit-create:
#200

### DIFF
--- a/integration_tests/test_cmath.py
+++ b/integration_tests/test_cmath.py
@@ -1,4 +1,4 @@
-from cmath import exp, log, sqrt
+from cmath import exp, log, sqrt, acos, asin, atan, cos, sin, tan
 from ltypes import c64
 
 def test_power_logarithmic():
@@ -9,4 +9,18 @@ def test_power_logarithmic():
     y = log(x)
     y = sqrt(x)
 
+
+def test_trigonometric():
+    x: c64
+    y: c64
+    x = complex(3, 3)
+    y = acos(x)
+    y = asin(x)
+    y = atan(x)
+    y = cos(x)
+    y = sin(x)
+    y = tan(x)
+
+
 test_power_logarithmic()
+test_trigonometric()
--- a/integration_tests/test_cmath.py
+++ b/integration_tests/test_cmath.py
@@ -1,4 +1,5 @@
-from cmath import exp, log, sqrt, acos, asin, atan, cos, sin, tan
+from cmath import (exp, log, sqrt, acos, asin, atan, cos, sin, tan,
+                   acosh, asinh, atanh, cosh, sinh, tanh)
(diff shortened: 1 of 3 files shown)


src/runtime/impure/lfortran_intrinsics.c
```diff
@@ -225,6 +225,16 @@ LFORTRAN_API double _lfortran_dlog10(double x)
     return log10(x);
 }
 
+LFORTRAN_API float_complex_t _lfortran_clog10(float_complex_t x)
+{
+    return log10(x);
```
certik, line 230: This line fails to compile on Windows with the error:
```
D:\a\lpython\lpython\lpython-0.1.0-788-gcb5b5f606\src\runtime\impure\lfortran_intrinsics.c(230): error C2440: 'function': cannot convert from 'float_complex_t' to 'double'
```

certik: Typically those complex functions are prefixed with `c`.

certik: Also for single precision, they use the `f` suffix.

However, it seems that `log10` does not have a complex version? The `clog10` and `clog10f` don't seem to compile for me.

integration_tests/test_cmath.py
```diff
@@ -0,0 +1,41 @@
...
+from ltypes import c64
+
+def test_power_logarithmic():
+    x: c64
+    y: c64
+    x = complex(3, 3)
+    y = exp(x)
+    y = log(x)
+    y = sqrt(x)
+    y = log10(x)
```
certik, line 12: Let's also test the results in `y` to make sure we are getting the right answer and not 0. I tested locally and it seems to work. I think we need to implement this in LPython first:
```
>>> y = 3+3j
>>> y
(3+3j)
>>> y.real
3.0
>>> y.imag
3.0
```

And absolute value for complex numbers:
```
>>> abs(y - (3+3j))
0.0
>>> abs(y - (3+3.001j))
0.0009999999999998899
```

So let's do it in subsequent PRs.

src/runtime/impure/lfortran_intrinsics.c
certik: So if this is needed in Python, you can compute it from Python using the complex log of base `e`.

Smit-create: > you can compute it from Python using the complex log of base `e`.

Yeah, that should work too. Thanks!

certik (approved):

certik:
This looks good, thanks! Let's get the real/imag parts of complex numbers working, so that we can add tests here that test the results.

Smit-create:
Yep, sounds good!

PR #234, opened 2022-03-14
typo corrected

Oshanath:
Corrected a typo in a comment in the first line of grammar/AST.asdl file.


czgdp1807:
IMO, we can also an `environment.yml` file separately under the project root. Reason being, it helps new devs to quickly setup by just executing, `conda create env -f environment.yml`, instead of going through a bunch links on the web to figure how to do it all. In any case having such a file would be really helpful.

certik:
Sure, created a dedicated issue #350 for that.

PR #235, opened 2022-03-15
Further extend the random module

namannimmo10:


src/runtime/random.py
```diff
@@ -31,10 +46,30 @@ def randint(lower: i32, upper: i32) -> i32:
...
     u: f64
     u = 1.0 - random()
     return u ** (-1.0 / alpha)
+
+def expovariate(l: f64) -> f64:
+    """
+    Return a random number from an exponential distribution with parameter
+    `l` (lambda).
+    """
+    return -_log(1.0 - random()) / l
```
Smit-create, line 68: We should check if `l!=0`.

src/runtime/random.py
```diff
@@ -31,10 +46,30 @@ def randint(lower: i32, upper: i32) -> i32:
...
+
+def weibullvariate(alpha: f64, beta: f64) -> f64:
+    """
+    Return a random number from a Weibull distribution with parameters `alpha`
+    and `beta`.
+    """
+    return alpha * (-_log(1.0 - random())) ** (1.0 / beta)
```
Smit-create, line 75: Same here that `b!=0`

src/runtime/random.py
```diff
@@ -31,10 +46,30 @@ def randint(lower: i32, upper: i32) -> i32:
 def _lfortran_random_int(lower: i32, upper: i32) -> i32:
     pass
 
+def uniform(a: f64, b: f64) -> f64:
+    """
+    Get a random number in the range [a, b) or [a, b] depending on rounding.
+    """
+    return a + (b - a) * random()
```
Smit-create, line 53: Check that `b>=a`

namannimmo10: Hmm... It looks like they don't check this.
```python
>>> import random
>>> random.uniform(100., 50.)
94.69713154566529
```

src/runtime/random.py
```diff
@@ -65,11 +66,13 @@ def expovariate(l: f64) -> f64:
     Return a random number from an exponential distribution with parameter
     `l` (lambda).
     """
+    assert l != 0.0
     return -_log(1.0 - random()) / l
```
Smit-create, line 70: Instead of `assert` we can raise `ValueError` with some error message so that it is more clear.

src/runtime/random.py
Smit-create: Ahh, then can leave it as it is.

certik:
I noticed the `random_float` must return `f32`, not `f64`:
```c
LFORTRAN_API float _lfortran_random_float()
{                                             
    return ((float) rand() / (float) RAND_MAX);
}                                        
```

src/runtime/random.py
```diff
@@ -31,10 +46,32 @@ def randint(lower: i32, upper: i32) -> i32:
...
     u: f64
     u = 1.0 - random()
     return u ** (-1.0 / alpha)
+
+def expovariate(l: f64) -> f64:
+    """
+    Return a random number from an exponential distribution with parameter
+    `l` (lambda).
+    """
+    assert l != 0.0
```
certik, line 68: Do `abs(l) > eps` and choose `eps = 1e-16` or something like that.

src/runtime/random.py
```diff
@@ -31,10 +46,32 @@ def randint(lower: i32, upper: i32) -> i32:
...
+    return -_log(1.0 - random()) / l
+
+def weibullvariate(alpha: f64, beta: f64) -> f64:
+    """
+    Return a random number from a Weibull distribution with parameters `alpha`
+    and `beta`.
+    """
+    assert beta != 0.0
```
certik, line 76: Same here.

certik (approved):
This looks good otherwise.

PR #236, opened 2022-03-15
Use colourful logs in integration tests

Smit-create:


namannimmo10 (approved):
looks nice!

certik (approved):
Looks good. Might not work well on Windows, but we can fix that later.

PR #237, opened 2022-03-15
CI: Build and Test separately

Smit-create:


certik:
What is the motivation for this change?

Smit-create:
So that we can easily find if it is a test failure or a build failure. Presently to find the test failure we see many unnecessary build logs.

certik:
I see. Good point!

certik (approved):
I think this looks good.

PR #241, opened 2022-03-16
Add comp-time `divmod` in python_comptime_eval.h

namannimmo10:


certik:
Would you please mind separating such formatting changes into a separate commit? It's really hard to review what was actually changed.

Also, I suggest we do not indent inside namespaces. That's the style we follow and that's quite common in C++. One loses valuable space on the left, and the namespace starts at the top and ends at the bottom, so it doesn't need to be indented.

namannimmo10:
This is ready to go in.

certik (approved):
This looks great, thank you!

PR #247, opened 2022-03-17
Overload `bool` runtime function

namannimmo10:


namannimmo10:
Let me know if this looks fine, and I will make all other builtin functions generic.

certik (approved):
Looks good!

certik:
The indentation error is caused by a bug in `--indent`, so some time ago at least I added an assert statement there, to prevent a segfault. We should fix it.

PR #248, opened 2022-03-18
Remove LFortran frontend

certik:
TODO:

* [x] We just need to load lbound/ubound correctly, it almost works, but the backend does not recognize them correctly


certik:
The backend now seems to generate valid LLVM code, but the `test_numpy_02` fails assert:
```
Running: integration_tests/tmp/test_numpy_02
Assertion failed
```
So there is still some bug to fix. Otherwise it seems to work.

PR #250, opened 2022-03-18
Overload all existing built-in functions

namannimmo10:
- [x]  overload `abs`
- [x] overload `round` 
- [x] overload `pow`


certik (approved):
I think this looks good! Very helpful, this makes the intrinsic functions pretty much usable now.

certik (approved):
I think that looks good.

I wonder what the purpose of the `pow` function is, when we have `**` in Python.

namannimmo10:
Perhaps once we add support for the third (optional) argument in `pow`, it can more efficiently compute `a` to the power `b` modulo `c`, than `pow(a, b)%c`

Have you heard about aliasing and chunking? This Pytalk by Raymond is really good: https://www.youtube.com/watch?v=UANN2Eu6ZnM (start from 8:50)

certik:
> Have you heard about aliasing and chunking?

Not using this name. How does it apply here in your opinion?

namannimmo10:
> Chunking refers to the process of taking individual pieces of information and grouping them into larger units

I think the fact what Raymond wants to convey is that instead of doing computationally-expensive tasks repeatedly, we can have functions (that would do it efficiently), thereby reducing the "registers" and maintaining our cognitive capacity. I am sure he can explain it much better!

certik:
Yes, you don't want to have too large loop body, otherwise you run out of registers. The compiler unrolling facility must unroll in order to vectorize well, but not too much to run our of registers. Regarding using functions to structure your code, that seems obvious. I don't now how it applies to this PR though. :)

PR #251, opened 2022-03-18
Add overload support with attribute call

Smit-create:


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2311,22 +2297,26 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             } // end of "comment"
         }
 
+        make_call_helper(al, s, current_scope, args, call_name, x.base.base.loc);
+    }
+
+    void make_call_helper(Allocator &al, ASR::symbol_t* s, SymbolTable *current_scope,
```
certik, line 2303: Can you please add a comment what the function does: I assume it takes a remote symbol `s` and arguments `args` and creates a proper call, adding ExternalSymbol as needed?

What exactly is `call_name` -- is it the same as the name of the symbol `s`, or something else?

certik:
Otherwise it looks good.

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: > I assume it takes a remote symbol s and arguments args and creates a proper call, adding ExternalSymbol as needed?

Yes, exactly.

> What exactly is call_name -- is it the same as the name of the symbol s, or something else?

It is the original name of the function/subroutine call. It will be only used during intrinsic case.

certik: If you can document that, that would be awesome. There might be a way to simplify further, but that can be done later. In this PR, just document the current approach, and we can merge it.

Smit-create: Okay!

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2311,22 +2297,28 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             } // end of "comment"
         }
 
+        make_call_helper(al, s, current_scope, args, call_name, x.base.base.loc);
+    }
+
+    // Function to create appropriate call based on symbol type. If it external
```
certik, line 2303:
```suggestion
    // Function to create appropriate call based on symbol type. If it is external
```

certik (approved):
Just left a minor typo fix.

PR #257, opened 2022-03-19
fix for float floor division operator

### LINKED ISSUE #255: Needs Implementation: floor division operator for float types
Example:
```
def f():
    i: f64
    i = 300.27
    print(i//1.0)

f()
```
Error Stacktrace:
```
>>./src/bin/lpython integration_tests/test_math1.py --show-stacktrace
; ModuleID = 'LFortran'
source_filename = "LFortran"

@0 = private unnamed_addr constant [9 x i8] c"%23.17f\0A\00", align 1

define void @_lfortran_main_program() {
.entry:
  call void @f()
  br label %return

  ret void
}

define void @f() {
.entry:
  %i = alloca double, align 8
  store double 3.002700e+02, double* %i, align 8
  %0 = load double, double* %i, align 8
  %1 = fptosi double %0 to i64
  %2 = fdiv i64 %1, 1
  %3 = fpext i64 %2 to double
  call void (i8*, ...) @_lfortran_printf(i8* getelementptr inbounds ([9 x i8], [9 x i8]* @0, i32 0, i32 0), double %3)
  br label %return

  ret void
}

define i32 @main() {
.entry:
  call void @_lfortran_main_program()
  ret i32 0
}
  File "/Users/anesa00/owork/lpython/src/bin/lpython.cpp", line 1540
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir, compiler_options);
  File "/Users/anesa00/owork/lpython/src/bin/lpython.cpp", line 742
    res = fe.get_llvm3(*asr, diagnostics);
  File "/Users/anesa00/owork/lpython/src/lpython/fortran_evaluator.cpp", line 309
    if (res.ok) {
  File "/Users/anesa00/owork/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 4073
    diagnostics.diagnostics.push_back(diag::Diagnostic(msg,
Floating-point arithmetic operators only work with floating-point types!
  %2 = fdiv i64 %1, 1
FPExt only operates on FP
  %3 = fpext i64 %2 to double



```

Abdullahjavednesar:
closes #255 

Fix suggested by @Smit-create, see #255


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1410,11 +1410,17 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     left = ASR::down_cast<ASR::expr_t>(ASR::make_ImplicitCast_t(
                         al, left->base.loc, left, ASR::cast_kindType::RealToInteger, dest_type,
                         value));
+                    left = ASR::down_cast<ASR::expr_t>(ASR::make_ImplicitCast_t(
+                        al, left->base.loc, left, ASR::cast_kindType::IntegerToReal, dest_type,
+                        value));
```
certik, line 1415: I'll try to test this soon. It seems it is casting real2int and then int2real. Wouldn't it be better to do something like `float(int(a) // int(b))`?

Smit-create: I was trying to do the same initially(in the original code), but since `a` and `b` are float, llvm will use `fdiv` and raise verification error.

certik:
So the way forward seems to be to treat a // b for floating point operands as `int(floor(a / b))`, where `a/b` is floating point division. If one operand is an integer, it will get cast to real. Otherwise if bot are integer, then the result is integer division in ASR.

Abdullahjavednesar:
Can you also add some tests for floor division.

src/runtime/lpython_builtin.py
```diff
@@ -314,3 +314,26 @@ def _lpython_imag(x: c64) -> f64:
...
+def _lpython_floordiv(a: i64, b: i32) -> i64:
+    return int(a/b)
+
+@overload
+def _lpython_floordiv(a: i64, b: i64) -> i64:
+    return int(a/b)
+
+@overload
+def _lpython_floordiv(a: f64, b: f64) -> f64:
+    return floor(int(a/b))
```
certik, line 339: I think it has to be just:
```suggestion
    return floor(a/b)
```
No cast to `int` first.

certik: I would just keep the following 4 functions:
* `def _lpython_floordiv(a: i32, b: i32) -> i32`
* `def _lpython_floordiv(a: i64, b: i64) -> i64`
* `def _lpython_floordiv(a: f32, b: f32) -> f32`
* `def _lpython_floordiv(a: f64, b: f64) -> f64`

And the ast2asr frotnend would do the correct casting first of int to float, or of 32 to 64 bit.

certik (commented):
I think this is a good approach, I left some comments. Also tests have to be added.

src/runtime/lpython_builtin.py
Smit-create: > I think it has to be just:

oh, it's a typo, I meant `float(int(a/b))`. Thanks. using `floor` hangs in ASR (maybe multiple module issue)

src/runtime/lpython_builtin.py
```diff
@@ -314,3 +314,26 @@ def _lpython_imag(x: c64) -> f64:
...
+def _lpython_floordiv(a: i32, b: i32) -> i32:
+    return int(a/b)
+
+@overload
+def _lpython_floordiv(a: i32, b: i64) -> i64:
+    return int(a/b)
+
+@overload
+def _lpython_floordiv(a: i64, b: i32) -> i64:
+    return int(a/b)
```
certik, line 331: I would remove all these. These should be represented in ASR as a division (BinOp) of two integers. The LLVM backend then implements this as integer division and it does what we want. Also the frontend takes care of casting 32 to 64, etc. if needed.

src/runtime/lpython_builtin.py
```diff
@@ -314,3 +314,26 @@ def _lpython_imag(x: c64) -> f64:
...
+
+@overload
+def _lpython_floordiv(a: i64, b: i64) -> i64:
+    return int(a/b)
```
certik, line 335: This should also be removed.

src/runtime/lpython_builtin.py
```diff
@@ -314,3 +314,26 @@ def _lpython_imag(x: c64) -> f64:
...
+
+@overload
+def _lpython_floordiv(a: f64, b: f64) -> f64:
+    return float(int(a/b))
```
certik, line 339: I think this actually has to be `floor(a/b)`.

certik:
There are couple issues. Currently there is an infinite recursion that gets "fixed" by:
```
--- a/src/runtime/lpython_builtin.py
+++ b/src/runtime/lpython_builtin.py
@@ -97,9 +97,9 @@ def str(x: i32) -> str:
     pos_to_str: list[str]
     pos_to_str = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']
     while x > 0:
-        rev_result += pos_to_str[x - (x//10)*10]
+        rev_result += pos_to_str[x - (x)*10]
         rev_result_len += 1
-        x = x//10
+#        x = x//10
     pos: i32
     for pos in range(rev_result_len - 1, -1, -1):
         result += rev_result[pos]
@@ -193,10 +193,10 @@ def bin(n: i32) -> str:
         prep = '-0b'
     res: str
     res = ''
-    res += '0' if (n - (n//2)*2) == 0 else '1'
+    res += '0' if (n - (n)*2) == 0 else '1'
     while n > 1:
-        n = n//2
-        res += '0' if (n - (n//2)*2) == 0 else '1'
+        n = n
+        res += '0' if (n - (n)*2) == 0 else '1'
     return prep + res[::-1]
 
 
@@ -218,9 +218,9 @@ def hex(n: i32) -> str:
     res = ""
     remainder: i32
     while n > 0:
-        remainder = n - (n//16)*16
+        remainder = n - (n)*16
         n -= remainder
-        n = n//16
+        n = n
         res += hex_values[remainder]
     return prep + res[::-1]
 
@@ -243,9 +243,9 @@ def oct(n: i32) -> str:
     res = ""
     remainder: i32
     while n > 0:
-        remainder = n - (n//8)*8
+        remainder = n - (n)*8
         n -= remainder
-        n = n//8
+        n = n
         res += _values[remainder]
     return prep + res[::-1]
 
```
Essentially the builtin module uses integer `//` operation which then recursively calls itself. The builtin module has to be constructed in a way that it does not need itself. There are two ways:

* Either all `//` operations must be converted to an explicit function call inside the builtin module
* Or (better) the integer `//` operation should just be represented as regular integer division in ASR, as I commented above. That also takes care of this recursion issue because we only use integer `//` here. The specialized floating point `//` operation can still be dispatched to a function call, that is fine. It is rarely used anyway, and as long as we don't use it in the builtin module, we are fine.

Abdullahjavednesar:
I think it doesn't work correctly for negative numbers
Example:
```
def f():
    print(5//2)
    print(-5//2)

f()
```
Result:
```
2
-2
```
Correct result should be:
```
>>> -5//2
-3
>>> 5//2
2
```

certik:
@Abdullahjavednesar good point, I created #318 for this.

PR #271, opened 2022-03-22
Remove unnecessary calls to ASR utils

namannimmo10:


tests/constants1.py
```diff
@@ -35,12 +35,13 @@ def test_len():
...
-    a = len([[-4, -5, -6], [-1, -2, -3]])
-    a = len({1, 2, 3})
-    a = len({1: "a", 2: "b", 3: "c"})
+    # TODO(namannimmo10): These commented out tests should work
+    # a = len((1, 2, 3))
+    # a = len((("c", "b", 3.4), ("c", 3, 5.6)))
+    # a = len([1, 2, 3])
+    # a = len([[-4, -5, -6], [-1, -2, -3]])
+    # a = len({1, 2, 3})
+    # a = len({1: "c", 2: "b", 3: "c"})
```
certik, line 44: Why don't they work anymore?

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -131,12 +131,13 @@ struct PythonIntrinsicProcedures {
     }
 
     static ASR::expr_t *eval_str(Allocator &al, const Location &loc, Vec<ASR::expr_t*> &args) {
+        LFORTRAN_ASSERT(ASRUtils::all_args_evaluated(args));
```
certik, line 134: This assert might stay --- once we allow unevaluated arguments that we seem to need for some intrinsic functions, then the callback can sometimes have arguments which are not evaluated. So having this asssert statement there is a good idea, to catch bugs early.

tests/constants1.py
namannimmo10: Looks like the argument is not a constant, and returns false when it runs across...

https://github.com/lcompilers/lpython/blob/2dd1e4ddd3e40ed65f2db9db3fe30c1247ef41dc/src/libasr/asr_utils.h#L417-L466

src/lpython/semantics/python_comptime_eval.h
certik: I made a mistake, I thought this was removing the assert. Yes, adding the asserts is a good idea.

certik (approved):
This looks good to merge.

PR #274, opened 2022-03-22
Adds a List to ASR as a dedicated container type 

### LINKED ISSUE #272: Adding a List to ASR as a dedicated container type
Per my comment here:

https://github.com/lcompilers/lpython/issues/270#issuecomment-1075213149

The fundamental semantic difference between an array and a list:

Array: It seems that semantically a multidimensional array is not a container in a sense that it only can contain elements of basic type like integer or real, but not "allocatable type", so you cannot have an array of allocatable character (array of strings), or allocatable arrays (array of different length arrays, so called "ragged array". The main purpose of a multidimensional array is a very efficient representation of uniform types and the fastest operations, but it is not meant as a container of "diverse types".

List: List can contain allocatable types, so it can represent a list of strings as a list of of allocatable characters, or ragged arrays (a list of lists or a list of allocatable arrays). You pay for this extra flexibility with a lowered performance. Also a list is fundamentally 1D.

It seem that whenever you can use an array, you should use it because it is faster, but it you need flexibility with each element having a different size, then you can use a list. Since the type of the list must be known at compile time (for efficiency and type checking), it seems the elements have to be of the same type, but they can be allocatable, thus having different size. That might cover most use cases. If different types are needed, one would have to use a list of structs.

In Python, it seems the most common use cases for lists that cannot be represented as arrays, are:

* List of strings: any string manipulation needs it, such as split(), str.join(), readlines(), etc.
* Ragged arrays: e.g. reading rows of numbers (or strings) from a file, where each row can have a different length

As a container (just like a struct), when the list goes out of scope, it deallocates all its elements (if they are allocatable).

namannimmo10:
Fixes #272


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -342,13 +342,33 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                            if (AST::is_a<AST::Name_t>(*ss->m_slice)) { // list[list[i32]]
+                                type = ast_expr_to_asr_type(loc, *ss->m_slice);
+                                return ASRUtils::TYPE(ASR::make_List_t(al, loc, type));
+                            } else {
+                                throw SemanticError("Only Name in Subscript supported for now in `list`"
+                                    " annotation", loc);
+                            }
+                        } else {
+                            throw SemanticError("Only list supported inside a list for now.", loc);
+                        }
```
namannimmo10, line 365: This is just a draft. We might not need this code block -- recursion could help us do the job. I'll see.

namannimmo10:
I wonder why the tests pass.

certik:
Hm, we have some issues in our CI test suite, that even when LPython fails, the test suite still passes. We have to fix it.

namannimmo10:
OK, so finally the build was not a success, and that's good. All we need to do is to fix the error and get this merged.

Smit-create:
> OK, so finally the build was not a success, and that's good. All we need to do is: fix the error and get this merged.

Rebasing will also fix that.

src/runtime/lpython_builtin.py
```diff
@@ -104,27 +104,28 @@ def str(x: i32) -> str:
...
-        result += '-'
-        x = -x
-    rev_result: str
-    rev_result = ''
-    rev_result_len: i32
-    rev_result_len = 0
-    pos_to_str: list[str]
-    pos_to_str = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']
-    while x > 0:
-        rev_result += pos_to_str[x - _lpython_floordiv(x, 10)*10]
```
namannimmo10, line 121: Had to comment out this implementation, otherwise, it gives a type mismatch error (character and list).  We should fix this soon.

PR #276, opened 2022-03-24
Implement abs for complex

### LINKED ISSUE #230: Implement abs() for complex numbers
And absolute value for complex numbers:
```
>>> abs(y - (3+3j))
0.0
>>> abs(y - (3+3.001j))
0.0009999999999998899
```
We probably need #229 first in order to implement it.

Smit-create:
Fixes #230

### DIFF
--- a/integration_tests/test_complex.py
+++ b/integration_tests/test_complex.py
@@ -40,8 +40,21 @@ def test_complex():
     print(x.real)
     print(x.imag)
 
+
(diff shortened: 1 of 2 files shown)


certik (approved):
I think this works.

Once we get the multiple modules working, we can use `c.imag`.

namannimmo10 (approved):

PR #278, opened 2022-03-24
XX Example how to improve print for arrays

certik:
Just to show how to do it.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -63,6 +63,8 @@
 #include <libasr/codegen/llvm_utils.h>
 #include <libasr/codegen/llvm_array_utils.h>
 
+#include <lpython/pickle.h>
+
 
 namespace LFortran {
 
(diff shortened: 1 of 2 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3464,6 +3466,8 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
             int a_kind = ASRUtils::extract_kind_from_ttype_t(t);
             if (ASRUtils::is_integer(*t) ||
                 ASR::is_a<ASR::Logical_t>(*ASRUtils::type_get_past_pointer(t))) {
+                ASR::Integer_t *t2 = ASR::down_cast<ASR::Integer_t>(t);
+                LFORTRAN_ASSERT(t2->n_dims == 0);
```
certik, line 3470: Here, extract the dimensions from `t2->m_dims[0].m_start`, you can do `ASR::down_cast<ASR::ConstantInteger_t>(t2->m_dims[0].m_start)`.

czgdp1807: The start of an array is always 0 in Python so this is correct. However, this might not work for end of an array. We might need to load the ending point of an array from the array descriptor. Or arrays are always constant sized in Python (i.e., we have start and end always available at compile time)?

PR #283, opened 2022-03-25
Fix lpython builtin `float` and `int`

### LINKED ISSUE #254: float to int conversion doesn't work
Example:
```
def f():
    i: f64
    j: i32
    i = 300.27
    j = int(i)
    print(i, j)

f()
```

output:

```
~/owork/lpython> ./src/bin/lpython integration_tests/test_math1.py --show-stacktrace  
ld: warning: dylib (/Users/anesa00/owork/lpython/src/runtime/liblfortran_runtime.dylib) was built for newer macOS version (11.5) than being linked (11.0)
~/owork/lpython> ./a.out                                                            
  300.26999999999998181 1
```

Smit-create:
Fixes #254


certik:
I am not convinced this is the way to do it. Rather, definitely for int/float/complex/logical conversions, we should simply insert the "ImplicitCast" node, which works for both compile time and runtime. We should not have any `int` or `float` functions.

Smit-create:
> we should simply insert the "ImplicitCast" node, which works for both compile time and runtime

Oh okay, you mean to handle them directly in `visit_call` rather than putting them in `lpython_builtin`?

namannimmo10:
What is the problem with the CI?

Smit-create:
> What is the problem with the CI?

GitHub is facing issues with GA.

Smit-create:
> we should simply insert the "ImplicitCast" node,

One problem with this approach I see is that we can only convert `RealToInteger` but the conversions like `string` to `int`, `bool` to `int` may not be possible only with `ImplicitCast`.

Smit-create:
Closing this in favor of #287

PR #285, opened 2022-03-25
initialization fix

### LINKED ISSUE #279: BUG: Binop with complex
On the following python code:
```py
def test():
    y: c64
    y = 6 + 8j
    print(y)

test()
```
The result is:
```
% ./a.out 
(6.000000,0.000000)
```

Does [this](https://github.com/lcompilers/lpython/blob/808ff0247e4f510c438ee602d27527a815444c59/src/lpython/semantics/python_ast_to_asr.cpp#L675) have anything to do with this bug because since `left` will be an implicitcast, `expr_value(left)` is `nullptr`?

Oshanath:
I implemented python-like initialization for lpython. 
` 
def test():

    y: c64
    y = 3j + 4

    print(y)

test() 
`

The above code for example will work now. Fixes #279 

But the real component has to be either integer or real for now. I will look at how to achieve this kind of initialization with variable real parts next.

Please note that I skipped the casting of the real part altogether and just created a complex asr node and used the value of the real part directly inside. We might have to tweak this if we are to enable this behavior for variable real parts.

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -1395,7 +1395,7 @@ public:
             } else { // real divison in python using (`/`)
                 dest_type = ASRUtils::TYPE(ASR::make_Real_t(al, loc,
                     8, nullptr, 0));
-                if (ASRUtils::is_integer(*left_type)) {
+                if (ASRUtils::is_integer(*left_type)) {              
                     left = ASR::down_cast<ASR::expr_t>(ASR::make_ImplicitCast_t(
                         al, left->base.loc, left, ASR::cast_kindType::IntegerToReal, dest_type,
                         value));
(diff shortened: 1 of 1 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1395,7 +1395,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             } else { // real divison in python using (`/`)
                 dest_type = ASRUtils::TYPE(ASR::make_Real_t(al, loc,
                     8, nullptr, 0));
-                if (ASRUtils::is_integer(*left_type)) {
+                if (ASRUtils::is_integer(*left_type)) {
```
namannimmo10, line 1398:
```suggestion
                if (ASRUtils::is_integer(*left_type)) {
```

namannimmo10:
Would you mind adding a test?

Smit-create:
I think you haven't rebased your branch with the main one. Some of the fixes were handled already.

Oshanath:
> I think you haven't rebased your branch with the main one. Some of the fixes were handled already.

I see that's why there are conflicts. I will close this pull request and try again. Thank you.

Oshanath:
@Smit-create is the method I used to solve this problem inadvisable? If not shall I rebase with the main branch and create another pull request?

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1481,8 +1481,54 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+
+            // case: (complex j + real or integer)
+            if(ASRUtils::is_complex(*left_type) && (ASRUtils::is_integer(*right_type) || ASRUtils::is_real(*right_type))){
+                dest_type = left_type;
+                double right_value;
+
+                // right operand is real
+                if(ASRUtils::is_real(*right_type)){
+                    right_value = ASR::down_cast<ASR::ConstantReal_t>(
+                                                        ASRUtils::expr_value(right))->m_r;
```
certik, line 1496: This would fail if the right hand side is a variable, not a number, correct?

certik:
Thanks @Oshanath for the PR. It's great that you were able to do this. The general approach you took is all there is to it to get LPython working for this case. The particular approach I think might be possible to simplify. Why don't you send a new PR based on the latest master, and add a test that currently fails in master but your PR fixes. I'll then think about how to simplify your code. It's an iterative process, so just submit what you can, and we'll  iterate on it until it's ready to merge.

src/lpython/semantics/python_ast_to_asr.cpp
Oshanath: Yes you are correct. For it to work with variables, I think I need to insert a check to avoid throwing an exception when it is a variable. Then it will work at runtime for now. Later when we implement compile time value deduction for implicit casts, it will work in compile time.

PR #287, opened 2022-03-26
Alternative fix for `int` and `float`

### LINKED ISSUE #254: float to int conversion doesn't work

Smit-create:
https://github.com/lcompilers/lpython/pull/283#issuecomment-1079212318

Fixes #254


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2303,6 +2303,53 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         tmp = nullptr;
     }
 
+    void make_type_conversions(Allocator &al, std::string call_name, Vec<ASR::call_arg_t> args,
+                                const Location &loc) {
```
certik, line 2307: Make this function return an `expr_t*`, and just return the value instead of assigning it to `tmp`. Then assign it to `tmp` in the caller.

certik, line 2307: Given that there is no common code for int/float, let's create two functions:

* `handle_intrinsic_int`
* `handle_intrinsic_float`

And in the caller call the appropriate one.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2303,6 +2303,53 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+        if (args.size() > 1) {
+            throw SemanticError("Either 0 or 1 argument is expected in '" + call_name + "'",
+                loc);
+        }
+
+        ASR::expr_t *arg = (args.size() > 0) ? args[0].m_value : nullptr;
+        ASR::ttype_t *type = arg ? ASRUtils::expr_type(arg) : nullptr;
```
certik, line 2314: Let's use if statements, don't use the `?` operator which is less readable.

tests/constants1.py
```diff
@@ -85,10 +85,10 @@ def test_int():
     a = int(4.56)
     a = int(5)
     a = int(-5.00001)
-    a = int(True)
-    a = int(False)
-    a = int("5346")
-    # a = int(complex(3.45, 5.6)) -> semantic error
+    # commented tests should work
+    # a = int(True)
+    # a = int(False)
```
certik, line 90: Not in this PR, but later we should add casting from Logical to Integer. We already have integer to logical. And implement it everywhere.

tests/constants1.py
```diff
@@ -85,10 +85,10 @@ def test_int():
...
+    # a = int("5346")
```
certik, line 91: For a string I think we need to call a runtime function to do the string to int conversion.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2303,6 +2303,53 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            ASR::ttype_t *to_type = ASRUtils::TYPE(ASR::make_Real_t(al, loc,
+                                        8, nullptr, 0));
+            if (!arg) {
+                tmp = ASR::make_ConstantReal_t(al, loc, 0, to_type);
+                return;
+            }
+            if (ASRUtils::is_integer(*type)) {
+                tmp = (ASR::asr_t *)ASR::down_cast<ASR::expr_t>(ASR::make_ImplicitCast_t(
+                al, loc, arg, ASR::cast_kindType::IntegerToReal,
+                to_type, nullptr));
```
certik, line 2343: Currently you assign `nullptr` as the compiler time value. However, in many cases (including many of our tests) the value is known at compile time, so we have to compute it here. I think we have to logic in `comptime_eval`, so you can possibly even reuse it for the compile time evaluation.

certik:
I think this looks excellent, this is exactly how to do it. I left some comments to fix before we can merge. Just ping me once it is updated and I'll do another review. Thanks for this work.

namannimmo10 (approved):
Looks great. Thanks, @Smit-create !

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2303,6 +2303,70 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                                    4, nullptr, 0));
+        if (!arg) {
+            return ASR::make_ConstantInteger_t(al, loc, 0, to_type);
+        }
+        if (ASRUtils::is_real(*type)) {
+            if (ASRUtils::expr_value(arg) != nullptr) {
+                int32_t ival = ASR::down_cast<ASR::ConstantReal_t>(
+                                        ASRUtils::expr_value(arg))->m_r;
+                value =  ASR::down_cast<ASR::expr_t>(make_ConstantInteger_t(al,
+                                loc, ival, to_type));
```
certik, line 2324: So right now `int` always returns 32bit integer. So if you do:
```
i: i64
i = int(3.0)
```
It will fail, because `int` returns a 32 bit integer, but `i` is a 64bit.

The way we do this in LPython is that on assignment `a = b`, we require both `a` and `b` to be exactly the same type (int/float/complex/...), but they don't have to be the same kind. Rather, LPython will insert implicit cast to cast from 64bit to 32bit and vice versa. 

For now I think int() should return a 64 bit integer, and then LPython implicitly casts it to 32bit if needed. The same with float(), which I think should return a 64bit float, and LPython casts it to 32bit float if needd.

This can be implemented in a future PR.

certik (approved):
So I think this is fine to merge as is. I left a very important caveat that we will have to fix (soon). But I think this PR is an improvement over master, so it can be merged.

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: Sure, will do that.

Smit-create:
Thanks for the review @namannimmo10 @certik

PR #290, opened 2022-03-27
Add ImplicitCast LogicalToInteger

Smit-create:
https://github.com/lcompilers/lpython/pull/287#discussion_r835771562


certik:
I think that this is fine. It might need some support in the LLVM backend also.

Smit-create:
On the current diff, the following gives:
```py
def f():
    a: i32
    b: i32
    a = True + True
    b = False - True
    print(a, b)

f()
```
Output as:
```
-2 1
```

I think `llvm` considers `True` as `-1`? whereas Python considers `True` as `1`.

Thirumalai-Shaktivel:
Hi @Smit-create, while testing your implementation: I came across this error!
```console
✗ lpython examples/expr2.py; ./a.out
semantic error: Not Implemented: type mismatch in binary operator; only Integer, Real, Complex, Logical combinations and string concatenation/repetition are implemented for now.
 --> examples/expr2.py:4:11
  |
4 |     a = ((True + True) - True)
  |           ^^^^^^^^^^^    ^^^^ type mismatch (integer and logical)
```
which might be a bug

Thirumalai-Shaktivel:
> On the current diff, the following gives:
> 
> ```python
> def f():
>     a: i32
>     b: i32
>     a = True + True
>     b = False - True
>     print(a, b)
> 
> f()
> ```
> 
> Output as:
> 
> ```
> -2 1
> ```
> 
> I think `llvm` considers `True` as `-1`? whereas Python considers `True` as `1`.

I don't think so, if it does then True + True ie., -1 + -1 will become 0, right?

Smit-create:
> , -1 + -1

That is equal to -2.

Thirumalai-Shaktivel:
Sorry my bad

Thirumalai-Shaktivel:
is this a valid syntax??
```console
$ lpython examples/expr2.py; ./a.out; rm ./a.out; echo "\nPython 3"; python examples/expr2.py
semantic error: Type mismatch in assignment, the types must be compatible
 --> examples/expr2.py:4:5
  |
4 |     a = True
  |     ^   ^^^^ type mismatch (integer and logical)
```
Currently, I'm testing things and also will be reporting issues here. After finding all of them I will create a new issue.

Thirumalai-Shaktivel:
> I think `llvm` considers `True` as -1?

Yes, I think the same
```python
def f():
    b: i32
    b = True - True
    print(b)
    b = False - False
    print(b)
    b = False - True
    print(b)
    b = True - False
    print(b)
    b = True + True
    print(b)
    b = False + False
    print(b)
    b = False + True
    print(b)
    b = True + False
    print(b)

f()
```

```console
$ lpython examples/expr2.py; a.out; rm ./a.out; echo "\nPython 3"; python examples/expr2.py
0
0
1
-1
-2
0
-1
-1

Python 3
0
0
-1
1
2
0
1
1
```

Smit-create:
> is this a valid syntax??

Nope, assigning `bool` to `int` is not supported.
```py
>>> x: int
>>> x = True
>>> x
True
>>> type(x)
<class 'bool'>
```

tests/reference/cpp-expr2-09c05ad.stdout
```diff
@@ -25,8 +25,8 @@ void test_boolOp()
     a = a && b;
     b = a || true;
     a = a || b;
-    a = a && b == b;
-    a = a && b != b;
+    a = a && (int)(b) == (int)(b);
+    a = a && (int)(b) != (int)(b);
```
Thirumalai-Shaktivel, line 29: `b` should not be explicitly converted to `int`, right??

Smit-create: yeah, I see, there are many such similar issues with `cpp` `codegen`. We can try to address all of them together.

Smit-create:
@certik  Does this look good now?

tests/reference/cpp-expr2-09c05ad.stdout
certik: I think in this particular place it's a bug in ASR, there should not be a cast to int.

certik:
This in general looks ok, but there seem to be some bugs:
```console
(lp) repos/lpython(pr-origin-290) $ cat tests/expr2.py 
def test_boolOp():
    a: bool
    b: bool
    a = False
    b = True
    a = a and b
    b = a or True
    a = a or b
    a = a and b == b
    a = a and b != b
    a = b or b
(lp) repos/lpython(pr-origin-290) $ lpython --show-asr tests/expr2.py
(TranslationUnit (SymbolTable 1 {main_program: (Program (SymbolTable 3 {}) main_program [] []), test_boolOp: (Subroutine (SymbolTable 2 {a: (Variable 2 a Local () () Default (Logical 1 []) Source Public Required .false.), b: (Variable 2 b Local () () Default (Logical 1 []) Source Public Required .false.)}) test_boolOp [] [(= (Var 2 a) (ConstantLogical .false. (Logical 1 [])) ()) (= (Var 2 b) (ConstantLogical .true. (Logical 1 [])) ()) (= (Var 2 a) (BoolOp (Var 2 a) And (Var 2 b) (Logical 1 []) ()) ()) (= (Var 2 b) (BoolOp (Var 2 a) Or (ConstantLogical .true. (Logical 1 [])) (Logical 1 []) ()) ()) (= (Var 2 a) (BoolOp (Var 2 a) Or (Var 2 b) (Logical 1 []) ()) ()) (= (Var 2 a) (BoolOp (Var 2 a) And (Compare (ImplicitCast (Var 2 b) LogicalToInteger (Integer 4 []) ()) Eq (ImplicitCast (Var 2 b) LogicalToInteger (Integer 4 []) ()) (Logical 4 []) () ()) (Logical 1 []) ()) ()) (= (Var 2 a) (BoolOp (Var 2 a) And (Compare (ImplicitCast (Var 2 b) LogicalToInteger (Integer 4 []) ()) NotEq (ImplicitCast (Var 2 b) LogicalToInteger (Integer 4 []) ()) (Logical 4 []) () ()) (Logical 1 []) ()) ()) (= (Var 2 a) (BoolOp (Var 2 b) Or (Var 2 b) (Logical 1 []) ()) ())] Source Public Implementation () .false. .false.)}) [])
```
This casts the bool to int here: `(ImplicitCast (Var 2 b) LogicalToInteger (Integer 4 []) ())`, but it should not do that.

tests/reference/cpp-expr2-09c05ad.stdout
Smit-create: Fixed that.

Smit-create:
Okay, thanks for reporting that. I fixed that.
Python code:
```py
def test_boolOp():
    a: bool
    b: bool
    a = False
    b = True
    a = a or b == b
```

<details>
<summary> ASR </summary>
<p>

```
(TranslationUnit 
   (SymbolTable 
      1 
      {
         main_program: 
            (Program 
               (SymbolTable 
                  3 
                  {
                  }) 
               main_program [] []), 
         test_boolOp: 
            (Subroutine 
               (SymbolTable 
                  2 
                  {
                     a: 
                        (Variable 
                           2 
                           a 
                           Local () () 
                           Default 
                           (Logical 1 []) 
                           Source 
                           Public 
                           Required .false.), 
                     b: 
                        (Variable 
                           2 
                           b 
                           Local () () 
                           Default 
                           (Logical 1 []) 
                           Source 
                           Public 
                           Required .false.)
                  }) 
               test_boolOp [] [
               (= 
                  (Var 2 a) 
                  (ConstantLogical .false. 
                     (Logical 1 [])) ()) 
               (= 
                  (Var 2 b) 
                  (ConstantLogical .true. 
                     (Logical 1 [])) ()) 
               (= 
                  (Var 2 a) 
                  (BoolOp 
                     (Var 2 a) 
                     Or 
                     (Compare 
                        (Var 2 b) 
                        Eq 
                        (Var 2 b) 
                        (Logical 4 []) () ()) 
                     (Logical 1 []) ()) ())] 
               Source 
               Public 
      }) 
   [])
```

</p>
</details>

certik (approved):

PR #291, opened 2022-03-27
Rename LFortran to LPython in CMakeList, NameSpace

Thirumalai-Shaktivel:


CMakeLists.txt
```diff
@@ -70,8 +70,8 @@ endif()
 set(WITH_LFORTRAN_ASSERT ${WITH_LFORTRAN_ASSERT_DEFAULT}
```
Thirumalai-Shaktivel, line 70: Some of the variables like `WITH_LFORTRAN_ASSERT` are connected to libasr, so I decided not to modify it (as it also affects LFortran)

Thirumalai-Shaktivel:
This PR is ready to go in.

grammar/asdl_cpp.py
```diff
@@ -1675,7 +1675,7 @@ def add_masks(fields, node):
 
 FOOT = r"""} // namespace LFortran::%(MOD)s
```
Thirumalai-Shaktivel, line 1676: Removing  `LFortran` from the `namespace` is messing with the things in `libasr`.

certik: With the current PR, do you see any failures in libasr?

certik:
I think these changes generally look fine, but I need to test this manually to ensure everything still works.

grammar/asdl_cpp.py
Thirumalai-Shaktivel: Nah, it doesn't fail. 
I said **if** we remove `LFortran` from `namespace` then it fails: in turn, we have to alter the content in the `Libasr`

PR #295, opened 2022-03-28
Support LLVM 11, 12, 13, 14; Add a Spack environment

certik:
Activate it with:

    spack env activate .

The first time you need to build all the packages in it:

    spack install

### DIFF
--- a/spack.yaml
+++ b/spack.yaml
@@ -13,5 +13,5 @@
 
 spack:
   # add package specs to the `specs` list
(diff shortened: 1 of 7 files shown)


certik:
This PR skips the JIT for LLVM >= 12, so `ctest` fails for LLVM >= 12 too. For LLVM 11 everything works as before.

What remains to be done is to port the JIT part to LLVM >= 12, but we can do it later.

PR #297, opened 2022-03-29
Automatically calculated value member of expr nodes in asr

Oshanath:
- Created the function `make_ImplicitCast_t_value()` in `ASRUtils` that will create an implicit cast node and calculate the `value` attribute if possible. If not possible, function is identical to `make_ImplicitCast_t()`.
- Replaced the `make_ImplicitCast_t()` calls in `implicitcast_helper()` with `make_ImplicitCast_t()`.
- Commented test case in [integration_tests/test_complex.py](integration_tests/test_complex.py) now passes. (Does not pass in the main branch)
- Made the function `make_ImplicitCast_t_value()` non-inline since it will be large.

- `IntegerToReal` cast fails test cases in [integration_tests/test_builtin_pow.py](integration_tests/test_builtin_pow.py). When both arguments of the pow() are integer and right arg is negative, IntegerToReal cast node gets created with a constantReal_t as the argument.
- `IntegerToInteger` cast fails test cases in [integration_tests/test_types_01.py](integration_tests/test_types_01.py). 
- So both of the above casts along with `IntegerToLogical` remain unimplemented for now.


certik:
Yes, this is exactly how to do it. I have to review the details of the new function carefully, but overall this is what I had in mind. I'll review it carefully hopefully tomorrow.

integration_tests/test_complex.py
```diff
@@ -11,7 +11,7 @@ def test_real_imag():
     b = x.imag
     assert abs(a - 2) < eps
     # TODO: below test should work
-    # assert abs(b - 3) < eps
+    assert abs(b - 3) < eps
```
czgdp1807, line 14: Some more tests may be? For better coverage.

Oshanath: Do you have any tips as to how I can write some tests to check whether an implicit cast node has a compile time value?

czgdp1807:
Why is this PR closed?

integration_tests/test_complex.py
czgdp1807: I think value calculation will happen while visiting `ImplicitCast` node. So, adding tests which will require implicit cast should do the job. Like, `c = a + b` where `a`, `b` and `c` all are of different types.

certik:
I don't know. I just reviewed PR #319 which seems to be a succession of this one.

PR #299, opened 2022-03-30
CI: Enable failing on error

certik:
This was introduced when we split the CI script into two, the second one
does not fail on error.


certik:
@Thirumalai-Shaktivel I had to implement your original solution, as I forgot that we have to have exactly the same API as CPython. Our tests caught it, but the CI was broken, so we didn't notice. This PR fixes both bugs.

Thirumalai-Shaktivel:
Thanks for fixing this!!

PR #303, opened 2022-03-30
Add test_lfortran to gitignore

Smit-create:


Thirumalai-Shaktivel:
Pull the main from lpython,
Do `git clean -dfx` and then build (`./build0.sh; ./build1.sh`), I think `test_lfortran` won't be generated. Actually I renamed test_lfortran to test_lpython in #291

Smit-create:
That works, thanks!!

PR #309, opened 2022-03-31
Change `doctest.h` to newer version

Abdelrahman-Kh-Fouad:
I changed doctest.h file to [last version](https://github.com/doctest/doctest/blob/master/doctest/doctest.h) ,because older one have problems with building (i don't know why )


certik:
Thanks!

Issue #298, opened 2022-03-29
Write a re2c+Bison parser for Python


certik:
The re2c should track indentation level and insert INDENT / DEDENT tokens. Each INDENT will have a matching DEDENT.

Links:

* https://docs.python.org/3/reference/compound_stmts.html#grammar-token-suite
* https://cppsecrets.com/users/136429711412197110461031111219710852575364103109971051084699111109/Python-token-INDENT-and-DEDENT.php

Thirumalai-Shaktivel:
https://github.com/yuangao2/Python_Interpreter/blob/master/includes/scan.l
https://github.com/python/cpython/blob/main/Parser/token.c
https://github.com/python/cpython/tree/main/Parser

Thirumalai-Shaktivel:
https://github.com/python/cpython/blob/main/Include/token.h

certik:
After #310 is merged, here are the next steps:

* [ ]    Emit `indent` and `dedent` tokens properly (i.e., handle indentation in the tokenizer)
* [ ]    Access the tokens from CPython somehow (figure out how to do that), and then compare against it, that we are getting exactly those tokens and nothing more or less.
* [ ]    Remove all the Fortran stuff from the tokenizer

It doesn't have to be perfect, it won't be perfect, but I would definitely do the above three things. Then we can move to the parser and start parsing it. And then we'll iterate on the design of the tokenizer as needed.

certik:
Here is how to access Python's tokenization:
```console
$ cat a.py 
def f():
    "aabcd"
    return 5

f()
$ python -m tokenize a.py   
0,0-0,0:            ENCODING       'utf-8'        
1,0-1,3:            NAME           'def'          
1,4-1,5:            NAME           'f'            
1,5-1,6:            OP             '('            
1,6-1,7:            OP             ')'            
1,7-1,8:            OP             ':'            
1,8-1,9:            NEWLINE        '\n'           
2,0-2,4:            INDENT         '    '         
2,4-2,11:           STRING         '"aabcd"'      
2,11-2,12:          NEWLINE        '\n'           
3,4-3,10:           NAME           'return'       
3,11-3,12:          NUMBER         '5'            
3,12-3,13:          NEWLINE        '\n'           
4,0-4,1:            NL             '\n'           
5,0-5,0:            DEDENT         ''             
5,0-5,1:            NAME           'f'            
5,1-5,2:            OP             '('            
5,2-5,3:            OP             ')'            
[...]

PR #310, opened 2022-03-31
Add initial tokenizer for Python

certik:
One can use it using `lpython --show-tokens somefile`. Currently the tokenizer is a Fortran tokenizer, so we need to adapt it to become a Python tokenizer.

Towards #298.


Thirumalai-Shaktivel (approved):
LGTM! Let's merge this once we fix the CI

Thirumalai-Shaktivel:
Hi @certik, I updated the tokenizer to recognize Comment, DocString, Symbols and also made the CI pass.
Can you review this??

Thirumalai-Shaktivel:
I tested this with many Python files. I got some errors, which I fixed and will push here.
The rest, `--show-tokens` provides output without any errors!
Now I think we can move to the parser (Bison).

certik:
Can you get the CI passing again please?

run_tests.py
```diff
@@ -78,6 +78,10 @@ def main():
             run_test("cpp", "lpython --no-color --show-cpp {infile}",
                     filename, update_reference, extra_args)
 
+        if tokens:
+            run_test("cpp", "lpython --no-color --show-tokens {infile} -o {outfile}",
```
certik, line 82: I think you forgot to update the prefix:
```suggestion
            run_test("tokens", "lpython --no-color --show-tokens {infile} -o {outfile}",
```

certik: (You'll have to update the tests --- remove everything in `tests/reference/*` and rerun `./run_tests.py -u`.

certik:
I left a fix above, after the update I think this is good to go in.

After it is merged, we should do two more steps before moving to the parser:

* Emit `indent` and `dedent` tokens properly (i.e., handle indentation in the tokenizer)
* Access the tokens from CPython somehow (figure out how to do that), and then compare against it, that we are getting exactly those tokens and nothing more or less.
* Remove all the Fortran stuff from the tokenizer

It doesn't have to be perfect, it won't be perfect, but I would definitely do the above three things. Then we can move to the parser and start parsing it. And then we'll iterate on the design of the tokenizer as needed.

run_tests.py
Thirumalai-Shaktivel: Good catch, Thanks!

Thirumalai-Shaktivel:
Perfect, I will work on `indent` and `dedent` and send a PR.
Thanks for the review!

certik:
Thanks!  I think this looks good, now we just have to iteratively improve it, the TODO list is written up at: https://github.com/lcompilers/lpython/issues/298#issuecomment-1086156155

PR #311, opened 2022-04-01
Fixed inline function calls pass to inline simple functions

czgdp1807:
Reference - https://gitlab.com/lfortran/lfortran/-/issues/673


czgdp1807:
Corresponding `lfortran` PR - https://gitlab.com/lfortran/lfortran/-/merge_requests/1694

certik:
We should create a pass manager, and move all passes handling into libasr, so that adding a new pass does not require modifying the driver (lpython or lfortran). 

Regarding syncing libasr, I do it from time to time, so we don't need to send PRs twice. Just send it to whatever project needs it.

PR #312, opened 2022-04-02
Do not recognize docstrings in the tokenizer

certik:
In Python, docstrings seem to be the first item in the body of a
function. They are not specifically recognized in the parser, and not in
the tokenizer either. Thus we should follow the same approach.

Tests were updated.

This also fixed the warning:

    tokenizer.re:684:20: warning: unreachable rule (shadowed by rule at line 677) [-Wunreachable-rules]


certik:
I am going to merge it, I think this is the correct fix.

PR #314, opened 2022-04-03
Remove all the Fortran related items from tokenizer

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
Here are list of Keywords in CPython: https://docs.python.org/3/reference/lexical_analysis.html#keywords

Thirumalai-Shaktivel:
This PR is ready...

src/lpython/parser/tokenizer.re
```diff
@@ -565,102 +205,37 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
...
-            // built-in or custom defined operator, such as: `.eq.`, `.not.`,
-            // or `.custom.`.
-            integer / defop {
-                lex_int_large(al, tok, cur,
-                    yylval.int_suffix.int_n,
-                    yylval.int_suffix.int_kind);
-                RET(TK_INTEGER)
-            }
-
+            'None' { RET(TK_NONE) }
```
certik, line 234: Shouldn't this be a keyword `KW_NONE`?

src/lpython/parser/tokenizer.re
```diff
@@ -23,133 +20,12 @@ void Tokenizer::set_string(const std::string &str)
...
-
-bool lex_dec(const unsigned char *s, const unsigned char *e, uint64_t &u)
-{
-    for (u = 0; s < e; ++s) {
-        if (!adddgt<10>(u, *s - 0x30u)) {
-            return false;
-        }
-    }
-    return true;
-}
```
certik, line 44: You want to keep these, and use them to convert ints and floats.

src/lpython/parser/tokenizer.re
```diff
@@ -565,102 +205,37 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
...
-            [oO] '"' [0-7]+ '"' { token(yylval.string); RET(TK_BOZ_CONSTANT) }
-            [oO] "'" [0-7]+ "'" { token(yylval.string); RET(TK_BOZ_CONSTANT) }
-            [zZ] '"' [0-9a-fA-F]+ '"' { token(yylval.string); RET(TK_BOZ_CONSTANT) }
-            [zZ] "'" [0-9a-fA-F]+ "'" { token(yylval.string); RET(TK_BOZ_CONSTANT) }
-
-            "&" ws_comment+ whitespace? "&"? {
-                line_num++; cur_line=cur; continue;
-            }
+            integer { token(yylval.string);  RET(TK_INTEGER) }
+            imag_number { token(yylval.string);  RET(TK_IMAG_NUM) }
```
certik, line 238: This returns integers and reals as strings. Rather, we should immediately convert integers, as before, just remove the support for Fortran specifix `_dp` style suffix. That is for performance reasons: we have to convert them anyway at some point, so if we do it right away, we are never even constructing the string representation of them.

certik:
I left some comments. Otherwise this looks great.

src/lpython/parser/tokenizer.re
Thirumalai-Shaktivel: Perfect, I will do it.

certik: Awesome. Ping me once you do and I'll review again.

certik (approved):
I think this is good enough to merge.

In subsequent PRs I think it might be easier to separate the integer conversions base on decimal/octal/hexadecimal/... since the re2c knows which one it is, so we can call the appropriate routine right away.

Also I think it makes sense to reintroduce the BOZ constant, to keep track of the original "encoding" of the integer. We can brainstorm and possibly just note this information together with the regular integer somehow.

certik:
There are some CI failures that need to be fixed.

Thirumalai-Shaktivel:
Thanks for the review and suggestions @certik. Going to merge when CI passes

PR #315, opened 2022-04-03
Add `interface` decorator

### LINKED ISSUE #239: Add `@interface` decorator
Currently we define a function interface using:
```python
def complex(x: f64, y: f64) -> c64:
    pass
```
But that is the same syntax as for an empty function. I think we should add an `@interface` decorator into `ltypes`, which would probably raise an exception if the function was ever being called (by CPython). In LPython, that would declare an interface, meaning the function with that name must be provided in global scope at some point before compiling to LLVM.

Smit-create:
Fixes #239

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -1230,7 +1230,7 @@ public:
                         current_procedure_abi_type = ASR::abiType::BindC;
                     } else if (name == "overload") {
                         overload = true;
-                    } else {
+                    } else if (name != "interface") {
                         throw SemanticError("Decorator: " + name + " is not supported",
                             x.base.base.loc);
                     }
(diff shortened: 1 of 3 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1230,7 +1230,7 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                         current_procedure_abi_type = ASR::abiType::BindC;
                     } else if (name == "overload") {
                         overload = true;
-                    } else {
+                    } else if (name != "interface") {
```
certik, line 1233: I think this is a cleaner way to do this:
```suggestion
                    } else if (name == "interface") {
                        // TODO: Implement @interface
                    } else {                    
```

src/runtime/ltypes/ltypes.py
```diff
@@ -92,6 +92,9 @@ def overload(f):
     return overloaded_f
 
 
+def interface(f):
+    raise Exception("Unexpected to be called by CPython")
```
certik, line 96: I think this will raise the exception already when you type `@interface`. But rather, it should raise this only if you actually call it, not just declare it.

certik (approved):

PR #316, opened 2022-04-04
Add mod operator support in ASR

### LINKED ISSUE #304: Needs Implementation: Modulo operator in Lpython
Example:
```
def f():
    print(23%3)

f()
```
Error:
```
semantic error: Binary operator type not supported
 --> integration_tests/test_math1.py:2:11
  |
2 |     print(23%3)
  |           ^^^^ 
```

Although I see `mod` function implemented in math.py.

Smit-create:
Fixes #304


namannimmo10 (commented):
Nice workaround! 

One thing I noticed,
```python
    a = -123282374
    b = 32771
    print(test_mod(a, b))
```
LPython and CPython give different outputs: 
```
Running: integration_tests/tmp/expr_05
2164
Running: PYTHONPATH=src/runtime/ltypes python integration_tests/expr_05.py
2128
```

src/runtime/math.py
```diff
@@ -154,13 +154,6 @@ def pow(x: f64, y: f64) -> f64:
     return x**y
 
 
-def mod(a: i32, b: i32) -> i32:
-    """
-    Returns a%b
-    """
-    return a - (a//b)*b
-
-
```
certik, line 163: I would revert this.

src/runtime/math.py
```diff
@@ -171,7 +164,7 @@ def gcd(a: i32, b: i32) -> i32:
     if b < 0:
         b = -b
     while b != 0:
-        a = mod(a, b)
+        a = a%b
```
certik, line 167: Possibly we should revert this also.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -859,18 +859,28 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
                 throw SemanticError("Binary operator type not supported",
                     x.base.base.loc);
             }
         }
+        if (is_mod) {
+            make_BinOp_helper(left, right, ASR::binopType::Div, x.base.base.loc, true);
+            ASR::expr_t *fd = ASRUtils::EXPR(tmp); /* left//right */
+            make_BinOp_helper(fd, right, ASR::binopType::Mul, x.base.base.loc, false);
+            right = ASRUtils::EXPR(tmp); /* (left//right)*right */
+            make_BinOp_helper(left, right, ASR::binopType::Sub, x.base.base.loc, false);
```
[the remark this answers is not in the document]
namannimmo10: I think there's no such intrinsic `mod` function in Python, so this PR does the job, without making `Mod` a part of,
```
binop = Add | Sub | Mul | Div | Pow
```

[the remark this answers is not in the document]
certik: `mod` has to be an intrinsic function, as in LFortran. Then our ASR optimizers can work with it.

certik:
A general design is that ASR never lowers. This PR lowers `mod` to `a - (a//b)*b`, i.e., it represents `mod` as `a - (a//b)*b`, and that is not the best design because we have lost semantic information that it is a mod. And that has practical consequences, such as our ASR optimizer that can optimize `mod` in certain cases can't do it, because it won't see a `mod`, but `a - (a//b)*b`, so we lose performance of the generated code. For this reason we always strive to represent the original code *exactly* in ASR. Then we apply optimizations using this full information. Only then we lower to LLVM or other representations.

Smit-create:
> LPython and CPython give different outputs:

Fixed it. It was having a floor division issue.

src/runtime/math.py
Smit-create: Done

certik (approved):

PR #319, opened 2022-04-05
Automatically calculated value member of expr nodes in asr

### LINKED ISSUE #294: Create a function to automatically compute the `value` member for an `expr` node
Let's start with `make_ImplicitCast_t`. Let's create a new function `make_ImplicitCast_t_value`, that will not accept a value argument, but rather compute it automatically based on the other arguments, and return an `ImplicitCast` node.

We should probably do this for all expr "make" functions and put this code into libasr, that way it can be shared with LFortran.

I think that will clean up both LFortran and LPython quite a bit, as you wouldn't need to worry about value anymore. LFortran already has all kinds of code to fill in the `value` attribute, and LPython has replicated quite a bit of this. I think this can be extracted into libasr using the above design, thus simplifying both codes.

Oshanath:
- Created the function `make_ImplicitCast_t_value()` in `ASRUtils` that will create an implicit cast node and calculate the `value` attribute if possible. If not possible, function is identical to `make_ImplicitCast_t()`.
- Replaced the `make_ImplicitCast_t()` calls in `python_ast_to_asr.cpp` with `make_ImplicitCast_t_value()`.
- Made the function `make_ImplicitCast_t_value()` non-inline since it will be large.

- `IntegerToReal` cast fails test cases in [integration_tests/test_builtin_pow.py](integration_tests/test_builtin_pow.py). When both arguments of the pow() are integer and right arg is negative, IntegerToReal cast node gets created with a constantReal_t as the argument.
- `IntegerToInteger` cast fails test cases in [integration_tests/test_types_01.py](integration_tests/test_types_01.py). 
- So both of the above casts along with `IntegerToLogical` remain unimplemented for now.

- Could not write any tests. Help needed for that. The purpose of this PR is for all implicitCast_t nodes to have a value in the 'value' attribute if it can be computed. But by the time integration tests are run, both compile time and runtime evaluations are complete. So there's no way to distinguish between the two without print statements. Any help regarding this would be appreciated.

Fixes #294


Oshanath:
@certik and @czgdp1807 I mistakenly deleted the fork I made from my github account. That closed the previous pull request. This is identical to the previous one. Sorry about the mess.

src/libasr/asr_utils.cpp
```diff
@@ -713,6 +714,68 @@ ASR::asr_t* symbol_resolve_external_generic_procedure_without_eval(
     }
 }
 
+ASR::asr_t* make_ImplicitCast_t_value(Allocator &al, const Location &a_loc, ASR::expr_t* a_arg, ASR::cast_kindType a_kind, ASR::ttype_t* a_type, ASR::expr_t* value) {
```
certik, line 717: Wouldn't a better API be to remove the `value` argument, and the idea being that you call this function if you want to compute the `value` automatically? But if you know `value` ahead of time, then you simply call the `make_ImplicitCast_t` directly.

src/libasr/asr_utils.cpp
```diff
@@ -713,6 +714,68 @@ ASR::asr_t* symbol_resolve_external_generic_procedure_without_eval(
...
+    ASR::ImplicitCast_t *n = (ASR::ImplicitCast_t *)ASR::make_ImplicitCast_t(
+        al, a_loc, a_arg, a_kind, a_type, value);
```
certik, line 719: Instead of creating this ahead of time, you should create it at the very end. The reason is that you first create the node, and then override some element of it, which can potentially be inefficient. We want to compute everything we need ahead of time, and then create the node.

certik:
I left some comments above, otherwise this looks good.

src/libasr/asr_utils.cpp
Oshanath: The reason I included the `value` argument is because now we can call only the `make_ImplicitCast_t_value()` and if there is a value, it can be included, and if the value is not provided, it will be `nullptr` by default, and the function will calculate the value of `value`.

So now we can stop using `make_ImplicitCast_t()` completely and just use `make_ImplicitCast_t_value()` for all purposes, since it will again call `make_ImplicitCast_t()` inside it.

If this is bad I will change it like you said.

src/libasr/asr_utils.cpp
Oshanath: You are correct. I will make the changes.

src/libasr/asr_utils.cpp
Oshanath: @certik I made a commit to the second suggestion you made. I will wait until you confirm this one again to make the other necessary changes.

certik:
Yes, update the tests using `./run_tests.py -u`.

src/libasr/asr_utils.cpp
certik: I see. It seems that you would almost always call this with "value=nullptr" as the argument, wouldn't you? In that case, I would just remove the argument altogether.

czgdp1807:
I think reference tests should be updated again after merging master into your branch. Ignore conflicts though, because the conflicts in reference tests will be overwritten (and hence resolved automatically) once you run, `python run_tests.py -u`.

src/libasr/asr_utils.cpp
czgdp1807: So, there are two cases,

1. `value` arg is not `nullptr` - In this case you are just creating the `ImplicitCast` node and returning it.
2. `value` arg is `nullptr` - In this case you are doing the computation.

The second case makes sense. Is there any example when `value` is not `nullptr`. Because if its not `nullptr` then its already computed for `ImplicitCast` node and if that's the case then why should we even call this function?

So, the bottom line is if you know that `value` is not `nullptr` then directly call, `make_ImplicitCast_t` and at places where you are passing `nullptr` as `value` argument of your function just don't do that and instead update the API here to not accept any `value` parameter. That should do the job.

Oshanath: I see the design now and have corrected it in the new commit. Thank you @certik and @czgdp1807 for explaining.

Oshanath:
@certik @czgdp1807 Even though I always merge with the main branch before pushing, I always end up with merge requests. It wasn't there when I pushed it though. It occured later. What should I do?

src/libasr/asr_utils.cpp
```diff
@@ -713,7 +714,69 @@ ASR::asr_t* symbol_resolve_external_generic_procedure_without_eval(
     }
 }
 
-    } // namespace ASRUtils
+ASR::asr_t* make_ImplicitCast_t_value(Allocator &al, const Location &a_loc, ASR::expr_t* a_arg, ASR::cast_kindType a_kind, ASR::ttype_t* a_type) {
+
+    ASR::expr_t* value = nullptr;
+
+    if (ASRUtils::expr_value(a_arg) != nullptr && value == nullptr){
```
certik, line 721: `value` is always `nullptr` here:
```suggestion
    if (ASRUtils::expr_value(a_arg)) {
```

src/libasr/asr_utils.cpp
```diff
@@ -713,7 +714,69 @@ ASR::asr_t* symbol_resolve_external_generic_procedure_without_eval(
...
+        // calculate value
+        if(a_kind == ASR::cast_kindType::RealToInteger){
```
certik, line 723:
```suggestion
        if (a_kind == ASR::cast_kindType::RealToInteger) {
```

src/libasr/asr_utils.cpp
```diff
@@ -713,7 +714,69 @@ ASR::asr_t* symbol_resolve_external_generic_procedure_without_eval(
...
+            int64_t v = ASR::down_cast<ASR::ConstantReal_t>(ASRUtils::expr_value(a_arg))->m_r;
+            value = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantInteger_t(al, a_loc, v, a_type));
+        }
+        else if(a_kind == ASR::cast_kindType::RealToReal){
```
certik, line 727:
```suggestion
        } else if (a_kind == ASR::cast_kindType::RealToReal) {
```

certik: And the same formatting change below.

src/libasr/asr_utils.cpp
```diff
@@ -713,7 +714,69 @@ ASR::asr_t* symbol_resolve_external_generic_procedure_without_eval(
...
+            double v = ASR::down_cast<ASR::ConstantReal_t>(ASRUtils::expr_value(a_arg))->m_r;
+            value = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantReal_t(al, a_loc, v, a_type));
+        }
+        else if(a_kind == ASR::cast_kindType::RealToComplex){
+            double double_value = ASR::down_cast<ASR::ConstantReal_t>(ASRUtils::expr_value(a_arg))->m_r;
+            std::complex<double> v(double_value, 0);
+            value = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantComplex_t(al, a_loc,
+                        std::real(v), std::imag(v), a_type));
```
certik, line 735:
```suggestion
            value = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantComplex_t(al, a_loc,
                        double_value, 0, a_type));
```

src/libasr/asr_utils.cpp
```diff
@@ -1,5 +1,6 @@
 #include <unordered_set>
 #include <map>
+#include <complex>
```
certik, line 3: I think it's not needed anymore after my changes below:
```suggestion
```

src/libasr/asr_utils.cpp
```diff
@@ -713,7 +714,69 @@ ASR::asr_t* symbol_resolve_external_generic_procedure_without_eval(
...
+        else if(a_kind == ASR::cast_kindType::IntegerToReal){
+            // TODO clashes with the pow functions
+            // int64_t value = ASR::down_cast<ASR::ConstantInteger_t>(ASRUtils::expr_value(a_arg))->m_n;
+            // value = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantReal_t(al, a_loc, (double)v, a_type));
+        }
+        else if(a_kind == ASR::cast_kindType::IntegerToComplex){
+            int64_t int_value = ASR::down_cast<ASR::ConstantInteger_t>(ASRUtils::expr_value(a_arg))->m_n;
+            std::complex<double> v(int_value, 0);
+            value = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantComplex_t(al, a_loc,
+                        std::real(v), std::imag(v), a_type));
```
certik, line 747:
```suggestion
            value = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantComplex_t(al, a_loc,
                        (double)int_value, 0, a_type));
```

src/libasr/asr_utils.cpp
```diff
@@ -713,7 +714,69 @@ ASR::asr_t* symbol_resolve_external_generic_procedure_without_eval(
...
+            double real = value_complex->m_re;
+            value = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantReal_t(al, a_loc, real, a_type));
+        }
+
+    }
+
+    ASR::ImplicitCast_t *n = (ASR::ImplicitCast_t *)ASR::make_ImplicitCast_t(
+        al, a_loc, a_arg, a_kind, a_type, value);
+
+    return (ASR::asr_t*)n;
```
certik, line 776: You can just do:
```suggestion
    return ASR::make_ImplicitCast_t(al, a_loc, a_arg, a_kind, a_type, value);
```

src/libasr/asr_utils.h
```diff
@@ -880,6 +880,8 @@ ASR::asr_t* symbol_resolve_external_generic_procedure_without_eval(
             SymbolTable* current_scope, Allocator& al,
             const std::function<void (const std::string &, const Location &)> err);
```
certik, line 882:
```suggestion
// Creates an ImplicitCast node and automatically computes the `value` if it can be computed at compile time
```

certik:
I finally got to review this. This is very good, I think the change is correct. I left some minor formatting changes and other simplifications. After you fix those (you can just click on the buttons to commit the changes and just follow the same style).

I noticed some of the compile time computations are commented out --- I assume some tests fail with them? If so, we should fix it later.

Overall it looks good. Please ping me once you fix it. I'll do a final review and we can merge.

src/libasr/asr_utils.cpp
```diff
@@ -713,7 +713,63 @@ ASR::asr_t* symbol_resolve_external_generic_procedure_without_eval(
...
+            ASR::ConstantComplex_t* value_complex = ASR::down_cast<ASR::ConstantComplex_t>(ASRUtils::expr_value(a_arg));
+            double real = value_complex->m_re;
+            double imag = value_complex->m_im;
+            value = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantComplex_t(al, a_loc, real, imag, a_type));
+        }
+        else if(a_kind == ASR::cast_kindType::ComplexToReal){
+            ASR::ConstantComplex_t* value_complex = ASR::down_cast<ASR::ConstantComplex_t>(ASRUtils::expr_value(a_arg));
+            double real = value_complex->m_re;
+            value = ASR::down_cast<ASR::expr_t>(ASR::make_ConstantReal_t(al, a_loc, real, a_type));
+        }
```
certik, line 765: If you can fix the formatting of these just like I suggested above, that would be great. I think it's ready after that.

src/libasr/asr_utils.cpp
```diff
@@ -713,7 +713,63 @@ ASR::asr_t* symbol_resolve_external_generic_procedure_without_eval(
...
+
+    }
+
+    return ASR::make_ImplicitCast_t(al, a_loc, a_arg, a_kind, a_type, value);
```
namannimmo10, line 769: `ImplicitCast` is now `Cast` in ASR -- see #362.

Thirumalai-Shaktivel:
Hi @Oshanath, I resolved the conflicts and formatting parts. Please review once and make sure the changes are correct as you wanted them to be in this PR.
If everything looks good, ping `certik` to review the implementation.

Oshanath:
> Hi @Oshanath, I resolved the conflicts and formatting parts. Please review once and make sure the changes are correct as you wanted them to be in this PR. If everything looks good, ping `certik` to review the implementation.

@Thirumalai-Shaktivel Thank you so much for this. I read through everything and it is exactly how it should be. @certik

certik:
I think the patch now looks good. @Oshanath will you be able to polish the history? I would do one commit that implements the new function, another commit that updates the rest of the code. And one commit that updates the tests.

Let's keep @Thirumalai-Shaktivel's commit, to preserve credits. 

@Oshanath if you don't know how to do it, please let me know and I can help.

Oshanath:
> I think the patch now looks good. @Oshanath will you be able to polish the history? I would do one commit that implements the new function, another commit that updates the rest of the code. And one commit that updates the tests.
> 
> Let's keep @Thirumalai-Shaktivel's commit, to preserve credits.
> 
> @Oshanath if you don't know how to do it, please let me know and I can help.

Could you please direct me to some learning resource on cleaning the git history?

certik:
The keyword is "git interactive rebase", I found, e.g.: https://www.sitepoint.com/git-interactive-rebase-guide/

The alternative is to use `git reset master` and `git add -p` and `git commit` and create a new set of patches. 

Important: make sure you make a backup of your branch before you do any of this. If you make a mistake, just use the backup and start over.

If you get stuck, just ask for help, we can help.

Oshanath:
@certik I will create a backup. If something goes wrong what I should do is close this PR, create another PR from the backup. Is that correct?

certik:
No, just keep this PR. Fix up your local branch, make sure it looks good (`git log --stat`) and then force push into this PR.

namannimmo10:
@Oshanath -- let us know if you need any help.

certik:
This looks good to merge, we just have to rebase.

certik (approved):
I think this is good to go in. We have thanked @Thirumalai-Shaktivel in the commit, as it was hard to preserve the commits due to the many changes needed.

Issue #306, opened 2022-03-31
Implement `max` and `min` intrinsic functions


namannimmo10:

purna135:
I'd like to work on this issue, but I'm not sure where to put the code for the `min` and `max` functions if I were to implement them.

purna135:
should it be `runtime/math.py` ?

certik:
@purna135 yes, you can work on any issue you want, simply send a PR, you don't have to ask for permission. Yes in `src/runtime` (although in `lpython_builtin` instead of `math`), this particular issue was fixed in #335, so you can look there for the details.

purna135:
Thank you @certik

PR #326, opened 2022-04-06
Add `max()`

Abdelrahman-Kh-Fouad:
#306 
i added  max only with i32 for three or two arguments


Abdelrahman-Kh-Fouad:
i have changed extension of `tokenizer.re` by mistake i didn't see `build0.sh ` ,i will return it.

src/runtime/lpython_builtin.py
```diff
@@ -212,7 +214,7 @@ def pow(x: i32, y: f64) -> f64:
 def pow(x: f64, y: i32) -> f64:
     return x**y
 
-
+#def max()
```
Thirumalai-Shaktivel, line 217:
```suggestion
```

tests/tests.toml
```diff
@@ -229,3 +229,8 @@ tokens = true
 [[test]]
 filename = "tokens/indent1.py"
 tokens = true
+
+[[test]]
+filename ="../integration_tests/test_max.py"
```
Thirumalai-Shaktivel, line 234: I think this file does not exist in the PR. Did you forget to include it?

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -162,6 +163,49 @@ struct PythonIntrinsicProcedures {
                 ASRUtils::type_to_str(arg_type) + "'", loc);
         }
     }
+    //max()
+    static ASR::expr_t * eval_max(Allocator &al , const Location &loc , Vec<ASR::expr_t*>&args){
+        LFORTRAN_ASSERT(ASRUtils::all_args_evaluated(args));
```
Thirumalai-Shaktivel, line 168: `all_args_evaluated()` makes sure that you provide only the `Constant value`; so here the input for the `max` function would be Constant Integer, real, etc...

Thirumalai-Shaktivel:
Resolve the conflicts, as suggested by Gagan (https://lfortran.zulipchat.com/#narrow/stream/311866-LPython/topic/GSoC.202022/near/278048279)

tests/tests.toml
Abdelrahman-Kh-Fouad: i included but i made conflict that remove things.

src/lpython/semantics/python_comptime_eval.h
Abdelrahman-Kh-Fouad: ok ,i got it

PR #327, opened 2022-04-06
adding conda-forge channel in the installation

Oshanath:
Multiple times when I install, I had to manually insert the channel `conda-forge` to create the environment. 

Even if this channel already exists in the environment, nothing bad will happen. But having this step in the instructions might make life easier for newcomers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ https://github.com/conda-forge/miniforge/#download
 
 Create a Conda environment:
 
+    conda config --append channels conda-forge
     conda create -n lp llvmdev=11.0.1 bison=3.4 re2c python cmake make toml numpy
     conda activate lp
 


Thirumalai-Shaktivel (approved):

README.md
```diff
@@ -32,6 +32,7 @@ https://github.com/conda-forge/miniforge/#download
 
 Create a Conda environment:
 
+    conda config --append channels conda-forge
     conda create -n lp llvmdev=11.0.1 bison=3.4 re2c python cmake make toml numpy
```
certik, line 36: I think this is better:
```suggestion
    conda create -c conda-forge -n lp llvmdev=11.0.1 bison=3.4 re2c python cmake make toml numpy
```
Note that this is not needed if you install conda from conda-forge, which is what I do. But it does not hurt to have it there explicitly.

PR #336, opened 2022-04-07
Overload pow() builtin function

namannimmo10:


integration_tests/test_builtin_pow.py
```diff
@@ -22,6 +21,31 @@ def test_pow():
...
     print(pow(a, b))
 
+    i1: i64
+    i2: i64
+    i1 = 2
+    i2 = 5
+    assert pow(i1, i2) == 32
+    i1 = 6
+    i2 = -3
+    print(pow(i1, i2))
```
Smit-create, line 31: Can we avoid printing and just use `assert`, because I think it does not make any sense to print them unless they aren't working properly(unnecessarily logs output which is hard to find if it is actually correct or not)?

namannimmo10: The issue is that when both the arguments are integers, the result should be an integer, except when the second argument is negative, the output is a floating-point number. 
Once we implement generics, we can deal with this.

certik (approved):
This looks good. Why cannot we test the result (even if it is a floating point)?

PR #340, opened 2022-04-08
Convert smaller integers directly to small ints

certik:


Thirumalai-Shaktivel:
Thanks for this PR, sir. I forgot to do this!

PR #341, opened 2022-04-08
Expose the new parser using the --new-parser flag

certik:


certik:
It's still WIP.

Thirumalai-Shaktivel:
I'm working on this now, I will make the changes and push it here.

Thirumalai-Shaktivel:
I'm working on this now, I will make the changes and push it here.

Thirumalai-Shaktivel:
Test 1:
```console
syntax error: Token 'def' is unexpected here
 --> examples/expr2.py:1:1
  |
1 | def main0():
  | ^^^ 


```
Test 2: 
```console
(Module [(Expr (BinOp (BinOp (BinOp (BinOp (BinOp (BinOp (Name a Load) Mult (Name z Load)) Add (ConstantInt 3 ())) Add (BinOp (ConstantInt 2 ()) Mult (Name x Load))) Add (BinOp (ConstantInt 3 ()) Mult (Name y Load))) Sub (BinOp (Name x Load) Div (BinOp (BinOp (Name z Load) Pow (ConstantInt 2 ())) Sub (ConstantInt 4 ())))) Sub (BinOp (Name x Load) Pow (BinOp (Name y Load) Pow (Name z Load)))))] [])
```

Thirumalai-Shaktivel:
Any other changes to be made? 
cc @certik

src/bin/lpython.cpp
```diff
@@ -138,18 +140,19 @@ int emit_asr(const std::string &infile,
...
     }
     LFortran::LPython::AST::ast_t* ast = r1.result;
 
     LFortran::LocationManager lm;
     lm.in_filename = infile;
-    std::string input = read_file(infile);
+    std::string input = LFortran::read_file(infile);
     lm.init_simple(input);
-    LFortran::diag::Diagnostics diagnostics;
+    diagnostics.diagnostics.clear();
```
certik, line 155: Do we want to clear it? I think we want to accumulate all messages, such as warnings from the parser, wouldn't we?

src/lpython/parser/parser.cpp
```diff
@@ -86,16 +86,18 @@ void Parser::handle_yyerror(const Location &loc, const std::string &msg)
...
     if (new_parser) {
         std::string input = read_file(infile);
-        Result<LPython::AST::Module_t*> res = parse(
-            al, input, diagnostics);
+        Result<LPython::AST::Module_t*> res = parse(al, input, diagnostics);
         LocationManager lm;
         lm.in_filename = infile;
         lm.init_simple(input);
+        CompilerOptions compiler_options;
+        std::cerr << diagnostics.render(input, lm, compiler_options);
```
certik, line 100: I thought about this, and I think we should not print the errors here, but rather return them to the caller, and the caller can decide when and how to visualize them. That way this parser function is fully general and can be used in all circumstances.

certik:
Thanks!! Let me test it, and fix up a few things (see above) and merge it. This is super helpful.

src/bin/lpython.cpp
certik: I left this in, as currently we print the messages from the parser, so that we clear it, so that we don't print it again.

certik:
I fixed it up, I think it works great now. We can now iteratively improve the parser.

src/lpython/parser/parser.cpp
certik: I fixed it.

Thirumalai-Shaktivel: > I thought about this, and I think we should not print the errors here, but rather return them to the caller, and the caller can decide when and how to visualize them. That way this parser function is fully general and can be used in all circumstances.

Yup, I agree. Thanks for fixing this.

PR #342, opened 2022-04-08
Remove unnecessary print statements

Smit-create:
We can avoid using `print` since the argument is already tested using `assert`. Moreover, it also doesn't help in verifying if the test is actually correct or not.


certik:
Perfect, thanks! I agree. Once tests pass, ping me and I'll review it once more.

certik (approved):

PR #343, opened 2022-04-08
Implement some basic numpy functions

namannimmo10:


namannimmo10:
These NumPy functions will work for `array_like` arguments in no time, once we have functions returning arrays in LPython. :-)

certik (approved):
Perfect. Btw, functions already can return arrays.

PR #345, opened 2022-04-08
Adding validation error msg for non integer str indexing

Abdullahjavednesar:
Adding Error msg for non integer indexing for strings.


tests/reference/asr-test_str_indexing-b200a4e.json
```diff
@@ -0,0 +1,13 @@
+{
+    "basename": "asr-test_str_indexing-b200a4e",
+    "cmd": "lpython --show-asr --no-color {infile} -o {outfile}",
+    "infile": "tests/errors/test_str_indexing.py",
+    "infile_hash": "d237030704fd7f04d4cad59083ca0d663561f9be6f3407e3ed04ed78",
+    "outfile": null,
+    "outfile_hash": null,
+    "stdout": null,
+    "stdout_hash": null,
+    "stderr": "asr-test_str_indexing-b200a4e.stderr",
```
namannimmo10, line 10: The change here seems correct but do we need to commit the error file?

certik: I think every error message should be eventually tested, and this shows how to do it. Tests like these will allow us to work on good test messages, and ensuring they keep working and look how they should.

certik (approved):

PR #346, opened 2022-04-09
Add bitwise operators support in ASR

### LINKED ISSUE #282: Implement Bitwise operations for integer type
Only the Logical type is supported in Bitwise operations. We have to extend this to integral values.
```python
from ltypes import i32

def main0():
    a: i32 = 60  # 60 = 0011 1100 
    b: i32 = 13  # 13 = 0000 1101  
    print(a | b) # 61 = 0011 1101

main0()
```
```console
$ lpython examples/expr2.py; ./a.out
semantic error: Binary operator type not supported
 --> examples/expr2.py:6:11
  |
6 |     print(a | b) # 61 = 0011 1101
  |           ^^^^^ 
```

namannimmo10:
Fixes #282 

- [x] compile-time bitwise OR (|)
- [ ] runtime bitwise OR
- [x] compile-time bitwise AND (&)
- [ ] runtime bitwise AND
- [x] compile-time bitwise XOR (^)
- [ ] runtime bitwise XOR
- [x] compile-time left shift (<<)
- [x] runtime left shift 
- [x] compile-time right shift (>>)
- [x] runtime right shift


integration_tests/expr_05.py
```diff
@@ -28,5 +27,11 @@ def main0():
     b = 32771
     assert test_mod(a, b) == 2128
 
+    assert 10 | 4 == 14
+    assert -105346 | -32771 == -32769
+    assert 10 & 4 == 0
+    assert -105346 & -32771 == -105348
+    assert 10 ^ 4 == 14
+    assert -105346 ^ -32771 == 72579
```
czgdp1807, line 35: Why are we not testing on the cases where operands are non-constant variables? For example,

```
a = 10
b = 4
assert a | b == 14
a = -105346
b = -32771
assert a & b == -32769
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1040,27 +1043,46 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+        if (is_bitwise_and) {
+            ASR::symbol_t *fn_mod = resolve_intrinsic_function(x.base.base.loc, "_bitwise_and");
+            tmp = make_call_helper(al, fn_mod, current_scope, args, "_bitwise_and", x.base.base.loc);
+            return;
+        }
+        if (is_bitwise_xor) {
+            ASR::symbol_t *fn_mod = resolve_intrinsic_function(x.base.base.loc, "_bitwise_xor");
+            tmp = make_call_helper(al, fn_mod, current_scope, args, "_bitwise_xor", x.base.base.loc);
+            return;
+        }
```
czgdp1807, line 1085:
```suggestion
            ASR::symbol_t *fn_mod = resolve_intrinsic_function(x.base.base.loc, bitwise_op_name);
            tmp = make_call_helper(al, fn_mod, current_scope, args, bitwise_op_name, x.base.base.loc);
            return;
```

namannimmo10: Done, thanks!

czgdp1807: You will have to add a check to not return when `bitwise_op_name` is different from the three cases considered.

integration_tests/expr_05.py
namannimmo10: I will address this comment soon.  Thanks.

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -242,6 +247,108 @@ struct PythonIntrinsicProcedures {
...
+        LFORTRAN_ASSERT(ASRUtils::check_equal_type(ASRUtils::expr_type(arg1),
+                                    ASRUtils::expr_type(arg2)));
+        ASR::ttype_t* type = ASRUtils::expr_type(arg1);
+        if (ASRUtils::is_integer(*type)) {
+            int64_t a = ASR::down_cast<ASR::ConstantInteger_t>(arg1)->m_n;
+            int64_t b = ASR::down_cast<ASR::ConstantInteger_t>(arg2)->m_n;
+            if (b < 0) {
+                throw SemanticError("Negative shift count not allowed.", loc);
+            }
+            std::cout << (a << b) << std::endl;
```
namannimmo10, line 322:
```suggestion
```

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -242,6 +247,107 @@ struct PythonIntrinsicProcedures {
...
+            if (b < 0) {
+                throw SemanticError("Negative shift count not allowed.", loc);
+            }
+            return ASR::down_cast<ASR::expr_t>(
+                ASR::make_ConstantInteger_t(al, loc, (a>>b), type));
+        } else {
+            throw SemanticError("_bitwise_rshift() must have both integer arguments.", loc);
+        }
+    }
+
```
Thirumalai-Shaktivel, line 350: Since the operations or process repeat the same in each function,  we can replace it with marcos, something like this:
```suggestion
#define BITWISE(X) \
    static ASR::expr_t *X(Allocator &al, const Location &loc, Vec<ASR::expr_t*> &args) { \
        return eval_bitwise(al, loc, args, &X); \
    }

BITWISE(eval__bitwise_or)
BITWISE(eval__bitwise_and)
BITWISE(eval__bitwise_xor)
BITWISE(eval__bitwise_lshift)
BITWISE(eval__bitwise_rshift)

    static ASR::expr_t *eval_bitwise(Allocator &al, const Location &loc,
                Vec<ASR::expr_t*> &args, comptime_eval_callback X) {
        LFORTRAN_ASSERT(ASRUtils::all_args_evaluated(args));
        if (args.size() != 2) {
            throw SemanticError("Bitwise Operation must have two integer arguments.", loc);
        }
        ASR::expr_t* arg1 = args[0], *arg2 = args[1];
        LFORTRAN_ASSERT(ASRUtils::check_equal_type(ASRUtils::expr_type(arg1),
                                    ASRUtils::expr_type(arg2)));
        ASR::ttype_t* type = ASRUtils::expr_type(arg1);
        if (ASRUtils::is_integer(*type)) {
            int64_t a = ASR::down_cast<ASR::ConstantInteger_t>(arg1)->m_n;
            int64_t b = ASR::down_cast<ASR::ConstantInteger_t>(arg2)->m_n;
            int64_t result;
            if(X == eval__bitwise_or) {
                result = a | b;
            } else if(X == eval__bitwise_and) {
                result = a & b;
            } else if(X == eval__bitwise_xor) {
                result = a ^ b;
            } else if(X == eval__bitwise_lshift) {
                result = a << b;
            } else if(X == eval__bitwise_rshift) {
                result = a >> b;
            }
            return ASR::down_cast<ASR::expr_t>(ASR::make_ConstantInteger_t(al, loc, result, type));;
        } else {
            throw SemanticError("Bitwise Operation must have both integer arguments.", loc);
        }
    }

```

integration_tests/expr_05.py
namannimmo10: So far these guys work:
```python
def main0():
  a: i32 = 10
  b: i32 = 4
  assert a & b == 0
  assert a | b == 14
  assert a ^ b == 14

main0()
```
Now we just have to deal with negative, non-constant variables, and other annotation types like `i64`, `i8`, etc.

src/lpython/semantics/python_comptime_eval.h
namannimmo10: Excellent, thanks!

Thirumalai-Shaktivel: Perfect, this looks good now.

integration_tests/expr_05.py
```diff
@@ -28,5 +27,20 @@ def main0():
...
+    assert 10 >> 1 == 5
+    assert 5 << 1 == 10
+    i1: i32 = 10
+    i2: i32 = 4
+    assert i1 << i2 == 160
+    assert i1 >> i2 == 0
+    assert i1 & i2 == 0
+    assert i1 | i2 == 14
+    assert i1 ^ i2 == 14
```
czgdp1807, line 44:
```suggestion
    assert i1 << i2 == 160
    assert i1 >> i2 == 0
    assert i1 & i2 == 0
    assert i1 | i2 == 14
    assert i1 ^ i2 == 14
    assert -i1 ^ -i2 == 10
```

certik:
Thanks for working on these, we definitely need all those.

Couple notes: I think it's a good design to call these as functions, as then we don't need to change ASR at all, we can just keep adding these as needed. I think these only accept integer arguments, correct? So initially we'll implement for i32 and i64. Typically they are of the type `(i32, i32) -> i32`, so all three integer arguments of the same type, so we only need to implement `(i32, i32) -> i32` and `(f64, f64) -> f64`.  We will insert implicit casts if needed.

Should these functions be implemented as interfaces only? Well, actually maybe it's a good idea to implement those that we can using multiplication, because then the LLVM backend doesn't even have to provide an implementation (initially!). Obviously for performance we want to implement all these in the backends directly.

Things bring me to the last point: we want very efficient way to "catch" these functions in the backend. So no string comparisons based on name (as we currently do), but rather using a very fast integer ID. So each function like `_bitwise_xor` would have an associated integer ID with it, part of ASR. For example: intrinsic functions have a positive ID, and all other functions have ID equal to 0 (non-intrinsic). And we would have a table of these, documented in libasr. The LPython/LFortran frontend would insert these, and the backends would then implement them.

certik:
We'll merge this for now, but I think we will refactor this to represent integer bitwise operations in the ASR itself, see #346.

PR #347, opened 2022-04-09
Error Handling: non integer indices for string slicing

Abdullahjavednesar:
Handling Exceptions when str slicing index in non integer

Example:
```
def f():
  s: str
  s = "01234"
  print(s[1:2.2])
```
Now throws error while calling `f()`
```
semantic error: slice indices must be integers or None
  |
4 |     print(s[1:2.2])
  |               ^^^ 
```


certik (approved):

certik:

PR #353, opened 2022-04-10
Create `environment.yml` file.

### LINKED ISSUE #350: Setup environment.yml for creating a development Conda environment
Reported by @czgdp1807 at https://github.com/lcompilers/lpython/pull/234#issuecomment-1093855868:

IMO, we can also an environment.yml file separately under the project root. Reason being, it helps new devs to quickly setup by just executing, conda create env -f environment.yml, instead of going through a bunch links on the web to figure how to do it all. In any case having such a file would be really helpful.

akshanshbhatt:
Closes #350

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ https://github.com/conda-forge/miniforge/#download
 Create a Conda environment using the preexisting environment.yml file:
 
     conda env create -f environment.yml
+    conda activate lp
 
 Install required packages (Linux - 64 bit):
 
--- a/README.md
+++ b/README.md
@@ -30,10 +30,9 @@ https://github.com/conda-forge/miniforge/#download
 
 ## Compile LPython
 
-Create a Conda environment:
+Create a Conda environment using the preexisting environment.yml file:
 
-    conda create -c conda-forge -n lp llvmdev=11.0.1 bison=3.4 re2c python cmake make toml numpy
-    conda activate lp
+    conda env create -f environment.yml
 
 Install required packages (Linux - 64 bit):
 
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,15 @@
+name: lp
+channels:
+  - conda-forge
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -30,10 +30,9 @@ https://github.com/conda-forge/miniforge/#download
 
 ## Compile LPython
 
-Create a Conda environment:
+Create a Conda environment using the preexisting environment.yml file:
 
-    conda create -c conda-forge -n lp llvmdev=11.0.1 bison=3.4 re2c python cmake make toml numpy
-    conda activate lp
```
certik, line 36: You still need to activate the environment, so leave the `conda activate lp` line here.

certik:
I tested this, and it works for me (on Apple M1). So I think this is good. I left a comment above, otherwise this is good to go in. Thanks!

README.md
akshanshbhatt: Removed it by mistake. Thanks.

environment.yml
```diff
@@ -0,0 +1,15 @@
...
+  - conda-forge
+dependencies:
+  - bison=3.4
+  - cmake
+  - flake8
+  - llvmdev=11.0.1
+  - make
+  - numpy
+  - pip
+  - python=3.10.2
```
czgdp1807, line 12: I don't think its a good idea to write a specific python version here. `conda` should be able to figure out on its own what version of python should be installed on a specific system.

Smit-create: I don't think we need python in `environment.yml`

czgdp1807: I see. So, how will someone update the reference tests i.e., `./run_tests.py -u`? By using system Python? I don't usually prefer using system Python when I am already creating a conda environment. Not sure about others.

Smit-create: I think `conda` automatically adds `python` while creating an environment.

Issue #331, opened 2022-04-06
Add runtime support for lists and dictionaries


certik:
Just like we have runtime support for arrays (size, ubound, lbound, ...), we need runtime support for lists and dictionaries. Those should be added by adding the interface to the `_lpython_builtin` and the ast2asr frontend calls them based on the syntax, and then the LLVM backend has to implement them.

Smit-create:
I wish to work on this

Smit-create:
Currently, we don't have an implementation of List in `llvm` backend. Should we consider it as an `array` for `llvm` backend or how we should plan to support its codegen?

certik:
Yes, that would be a good issue to work on. We need to add "interface only" functions for lists into `lpython_builtin.py`. Then call them from ASR on things like `a = []; a.append("x")`. Let's implement these functions for int, float and str. Nothing else for now. That is the first step.

The second step is to then implement these in the LLVM backend. We have to choose how lists should be implemented. Since we know the type of all elements, and they are 1D, they should probably be vectors in LLVM. I don't know if vector can have a runtime length, such as the "append" function would create a new vector and copy the data. In fact, for efficiency, we really want to double the allocation, so we need to store the "capacity" (allocated length) and "size" (the actual size). The append would keep filling the preallocated memory, until "capacity" is hit, then double the allocation and copy the data. So just like the `Vec` in libasr. So there will have to be a descriptor for lists, just like we have for arrays. The descriptor will have a pointer to the data, "size" and "capacity".

certik:
So I would say, let's start with just ASR, don't worry about LLVM for now. And let's compile the following types of codes to ASR properly:
```
a: list[i32]
a = [1, 2, 3]
a.append(5)
del a[2]
```
Once this compiles, then let's work on the LLVM backend and get it working. Then we'll extend it for i64, f32, f64 and str.

certik:
No, using `ListAppend` new `stmt` ASR node. Just add it to `ASR.asdl` and return it as a statement.

Smit-create:
I think we might want to wait on this until this is finalized: https://gitlab.com/lfortran/lfortran/-/merge_requests/1700

Thirumalai-Shaktivel:
@czgdp1807, is there anything else to be implemented to complete this issue?

PR #354, opened 2022-04-11
Implement list.append

Smit-create:
#331


certik:
Thanks. I think we should generate a new `ListAppend` ASR `stmt` node. Then implement it in the backend.

certik:
Thanks. I think we should generate a new `ListAppend` ASR `stmt` node. Then implement it in the backend.

certik (approved):
Excellent. Looks really simple. Consistent with https://gitlab.com/lfortran/lfortran/-/merge_requests/1700.

namannimmo10:
Thanks, @Smit-create ! I think we must also make sure that  ` l: list[i32];  l = [1, 2, 3];  l.append('a')` does not generate the ASR. See #365 .

PR #357, opened 2022-04-11
Handle printing of logical types in LLVM backend

### LINKED ISSUE #322: Handle printing of logical types
Reference code:
```python
b: bool
b = True
print(b)
```
```
% lpython t.py && ./a.out
 --> o.py:3:1
  |
3 | print(b)
  | ^^^^^^^^


```

kinshukdua:
fixes #322

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -3578,6 +3578,22 @@ public:
             if (ASRUtils::is_integer(*t) ||
                 ASR::is_a<ASR::Logical_t>(*ASRUtils::type_get_past_pointer(t))) {
                 switch( a_kind ) {
+                    case 1 : {
+                        if (llvm::ConstantInt* CI = llvm::dyn_cast<llvm::ConstantInt>(tmp)) {
+                            if (CI->getBitWidth() <= 8) {
+                                if (CI->isZero())
+                                {
+                                    tmp = builder->CreateGlobalStringPtr("False");
+                                }
+                                else if (CI->isOne())
+                                {
+                                    tmp = builder->CreateGlobalStringPtr("True");
+                                }
+                            }
+                        }
+                        fmt.push_back("%s");
+                        break;
+                    } 
                     case 4 : {
                         fmt.push_back("%d");
                         break;


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3578,6 +3578,22 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                                }
+                                else if (CI->isOne())
+                                {
+                                    tmp = builder->CreateGlobalStringPtr("True");
+                                }
+                            }
+                        }
+                        fmt.push_back("%s");
+                        break;
+                    }
```
certik, line 3596: This will not work for `i: i8`, which is an integer of `kind=1`. What we need here is to separate integer printing form logical printing, and integer printing should print for all kinds (1, 2, 4, 8) as a number, while logical printing should print True/False.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3578,6 +3578,22 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
             if (ASRUtils::is_integer(*t) ||
                 ASR::is_a<ASR::Logical_t>(*ASRUtils::type_get_past_pointer(t))) {
                 switch( a_kind ) {
+                    case 1 : {
+                        if (llvm::ConstantInt* CI = llvm::dyn_cast<llvm::ConstantInt>(tmp)) {
+                            if (CI->getBitWidth() <= 8) {
+                                if (CI->isZero())
```
certik, line 3584: The other issue is that this assumes a constant (compile time?) integer. I think this will not work for printing a variable. One has to do this "if" at runtime.

namannimmo10: +1

kinshukdua: @certik How do I access the print function in runtime? The only reference I see to a print function is the `lfortran_printf` function.

certik: We either have to create a runtime function that prints this correctly, or we have to insert this logic using LLVM commands. Writing a function seems easier. There might even be a way to make `printf` print True/False based on an integer value 1,0.

kinshukdua: Can you guide me regarding how and where to create the function. I'm guessing inside the `runtime` folder?

certik: Do `git grep lfortran_printf`, find the function, then add another one that creates the proper True/False based on an int, and then call it from llvm to create True/False at runtime.

czgdp1807: What about creating an array ([ArrayType](https://llvm.org/doxygen/classllvm_1_1ArrayType.html)) of `i8*` type in LLVM, `{"False", "True"}`. Then use, `CreateGEP` along with `tmp` to access 0-index or 1-index from the above value at run time.

certik: That should also work.

namannimmo10: @kinshukdua -- are you willing to complete this? Or, may I take over?

namannimmo10: @kinshukdua -- are you willing to complete this? Or, may I take over?

namannimmo10: I am not experienced either.  Let's try https://github.com/lcompilers/lpython/pull/357#discussion_r848264865 ?

namannimmo10:
First off, we should rebase with `main`.

namannimmo10:
Thanks for your contribution, @kinshukdua. This was merged in #504.

Issue #344, opened 2022-04-08
Add a parser benchmark


certik:
Here is what I tried so far on Apple M1 max:
```python
N = 50000
s1 = "(a*z+3+2*x + 3*y - x/(z**2-4) - x**(y**z))"
s = s1
for n in range(N):
    s = s + " * " + s1
open("a.py", "w").write(s)
```
Then compile in optimized mode:
```console
$ cmake -DCMAKE_CXX_FLAGS_RELEASE="-Wall -Wextra -O3 -funroll-loops -DNDEBUG" .
$ make -j
```
and benchmark:
```console
$ time lpython --show-ast --new-parser a.py --no-color > a.txt
lpython --show-ast --new-parser a.py --no-color > a.txt  0.11s user 0.03s system 95% cpu 0.139 total
```
So the full tokenizing, parsing, creating AST and printing AST using our printer (no colors) takes roughly 0.139s. Now we need to check how long it takes in CPython.

certik:
To test CPython, I applied the following patch:
```diff
--- a/src/runtime/lpython_parser.py
+++ b/src/runtime/lpython_parser.py
@@ -1,13 +1,18 @@
 import sys
 import python_ast
 import ast
+from timeit import default_timer as clock
 
 filename = sys.argv[1]
 #filename = "doconcurrentloop_01.py"
 #filename = "expr2.py"
 
 input = open(filename).read()
+t1 = clock()
 a = ast.parse(input, type_comments=True)
+t2 = clock()
+print(t2-t1)
+stop
 #print(ast.unparse(a))
 #print()
 #print(ast.dump(a))
```
Then it prints:
```console
$ python src/runtime/lpython_parser.py a.py
1.719355999986874
```
So it seems it takes CPython 1.719s to parse into Python AST.

Abdelrahman-Kh-Fouad:
@certik i think i can make a benchmark, my thought is adding `--parser-time` option for lpython and we can call as a new feature in `run_tests.py` it will run all files and compares parser times and output will be small blots graph that show time difference, or we can make it one file instead of feature on `run_tests.py`.

certik:
As of 27b89891e94020ebf8def532d8af7189b4833c00, the speed is now:
```console
$ time lpython --show-ast --new-parser a.py --no-color > a.txt
lpython --show-ast --new-parser a.py --no-color > a.txt  0.12s user 0.02s system 97% cpu 0.144 total
```

certik:
[...]

PR #359, opened 2022-04-11
Implementing `--time-report`

Abdelrahman-Kh-Fouad:
About #344 
Now, we can see parsering time.
```bash
$ ./src/bin/lpython --parser-time a.py
Parsing time: 5.32848 ms
```
I think we can add afile for comparing some files, or add feature on `run_test.py` to make this.


certik:
I think this is a good idea. Why not to simply add the timing information for the "--show-ast" code as an option if the `--time-report` option is passed? LFortran does it that way for some backends. So we can eventually add it to all stages of the compiler.

Abdelrahman-Kh-Fouad:
> I think this is a good idea. Why not to simply add the timing information for the "--show-ast" code as an option if the `--time-report` option is passed? LFortran does it that way for some backends. So we can eventually add it to all stages of the compiler.

ok, and if comilation fall in any stage report will show times of stages before this point.

Abdelrahman-Kh-Fouad:
@certik 

- i implemented `--time-report` like LFortran and it called from `run_benchmark.py` to find parsing time.
- i made `run_benchmark.py` to measure time of `benchmark` dir files.
  * in `benchmarks.toml` we can customize every file to add it any kind of comparison we want {parse or src -> ast, etc .. }  
  * and if want to add a comparison we should add class that represent times for files, now there is just parser comparison.
  * after comparison there is a graph of data will be printed, and it shows us the time of every file in millisecond like : 
   img

src/bin/lpython.cpp
```diff
@@ -222,7 +221,7 @@ int emit_cpp(const std::string &infile,
     std::cout << res.result;
     return 0;
 }
-
+#define HAVE_LFORTRAN_LLVM 1
```
certik, line 224: I think this should not be here, this is defined by cmake in `config.h`.

certik:
Interesting idea to run benchmarks like that in the `benchmarks` subdirectory --- we should probably do that. But let's submit that as a separate PR. In this PR, let's just implement the `--time-report`.

src/bin/lpython.cpp
```diff
@@ -269,55 +268,84 @@ int emit_llvm(const std::string &infile,
...
+    times.push_back(std::make_pair("Reading_file", std::chrono::duration<double, std::milli>(end_file - start_file).count()));
+
+    auto start_parser = std::chrono::high_resolution_clock::now();
+    LFortran::Result < LFortran::LPython::AST::ast_t * > r = parse_python_file(
+            al, runtime_library_dir, infile, diagnostics, compiler_options.new_parser);
+    auto end_parser = std::chrono::high_resolution_clock::now();
+    times.push_back(std::make_pair("parsing", std::chrono::duration<double, std::milli>(end_parser - start_parser).count()));
     std::cerr << diagnostics.render(input, lm, compiler_options);
     if (!r.ok) {
+        print_report(times, time_report);
```
certik, line 305: I have to test it, it seems it would always print the report if the parsing fails?

src/bin/lpython.cpp
```diff
@@ -269,55 +268,84 @@ int emit_llvm(const std::string &infile,
...
     // ASR -> LLVM
+    auto start_asr_to_llvm = std::chrono::high_resolution_clock::now();
     LFortran::PythonCompiler fe(compiler_options);
     LFortran::LLVMEvaluator e(compiler_options.target);
-    std::unique_ptr<LFortran::LLVMModule> m;
-    LFortran::Result<std::unique_ptr<LFortran::LLVMModule>>
-        res = fe.get_llvm3(*asr, diagnostics);
+    std::unique_ptr <LFortran::LLVMModule> m;
+    LFortran::Result <std::unique_ptr<LFortran::LLVMModule>>
+            res = fe.get_llvm3(*asr, diagnostics);
```
certik, line 333: The above seem like just formatting changes. I would do those as either separate PR, or at least separate commits. I think it looks better if written as `std::unique_ptr<LFortran::LLVMModule>` rather than `std::unique_ptr <LFortran::LLVMModule>`, doesn't it?

src/bin/lpython.cpp
Abdelrahman-Kh-Fouad: yes, and even if ast to asr  or asr to llvm, it will print last stage even if it faild.
i thought it will be usefull to know time even if stage faild, but i  will remove it.

Abdelrahman-Kh-Fouad:
> Interesting idea to run benchmarks like that in the `benchmarks` subdirectory --- we should probably do that. But let's submit that as a separate PR. In this PR, let's just implement the `--time-report`.

ok, i will make for `--show-time`.

src/bin/lpython.cpp
Abdelrahman-Kh-Fouad: yes, you're right, i just put it there just to test something and i forget to erasr it,sorry.

src/bin/lpython.cpp
Abdelrahman-Kh-Fouad: i made this formate by ide formatter, but i change it no problem.

certik: Which formatter do you use? We should probably setup a `clang-format` config file for the project.

Abdelrahman-Kh-Fouad: it just a formatter with clion ide with no standard i think, but i will add clang to be the formatter.

src/bin/lpython.cpp
certik: I think it would be useful even if it fails.

Abdelrahman-Kh-Fouad: ok, thats fine i will bring it back.

src/bin/lpython.cpp
```diff
@@ -269,55 +267,84 @@ int emit_llvm(const std::string &infile,
...
         const std::string &infile,
         const std::string &outfile,
         const std::string &runtime_library_dir,
-        CompilerOptions &compiler_options)
-{
-    Allocator al(4*1024);
+        const bool time_report,
+        CompilerOptions &compiler_options) {
+
+    Allocator al(4 * 1024);
```
certik, line 285: If you want to make formatting changes, you can, but please send them as a separate PR. In a PR where you are implementing new functionality, please do not change other code by adding white spaces, as it makes it harder to review, and also it seems to me these formatting changes are not necessarily making it more readable, but we should have that discussion on a separate PR, so that we can focus here on the actual functionality.

certik:
If you wouldn't mind extracting all pure formatting changes into a separate PR, I would really appreciate it. It I suggested a few above, but there are more.

PR #361, opened 2022-04-12
Raise Error for negative shift count

namannimmo10:


certik (approved):
That's fine. Once https://gitlab.com/lfortran/lfortran/-/merge_requests/1700 is in, these will be implemented in ASR itself and such an error will be raised at runtime if needed by the backend, possibly only in Debug mode for performance reasons, like indexing an array.

certik:
Also note that we currently do not support exceptions.

PR #363, opened 2022-04-12
Fix the `dedent` token generation in new tokenizer

### LINKED ISSUE #358: New tokenizer fails when there are multiple indentations
```py
class Test:
    def foo(a):
        if a == 'foo':
            return 'foo'
        return 'bar'
```

This code, when run with lpython binary, gives an error when I pass `--show-tokens` flag :

```sh
❯ src/bin/lpython --show-tokens examples/test_file.py
  File "/Users/akshansh/Documents/GitHub/lpython/src/bin/lpython.cpp", line 749
    return emit_tokens(arg_file, true, compiler_options);
  File "/Users/akshansh/Documents/GitHub/lpython/src/bin/lpython.cpp", line 92
    auto res = LFortran::tokens(al, input, diagnostics, &stypes, &locations);
  File "tokenizer.re", line 557
  File "tokenizer.re", line 151
  Binary file "/usr/lib/system/libsystem_platform.dylib", local address: 0x1803414c3
```

Python's tokenizer, on the other hand, gives this output :

```sh
❯ python -m tokenize examples/test_file.py
0,0-0,0:            ENCODING       'utf-8'
1,0-1,5:            NAME           'class'
1,6-1,10:           NAME           'Test'
1,10-1,11:          OP             ':'
1,11-1,12:          NEWLINE        '\n'
2,0-2,4:            INDENT         '    '
2,4-2,7:            NAME           'def'
2,8-2,11:           NAME           'foo'
2,11-2,12:          OP             '('
2,12-2,13:          NAME           'a'
2,13-2,14:          OP             ')'
2,14-2,15:          OP             ':'
2,15-2,16:          NEWLINE        '\n'
3,0-3,8:            INDENT         '        '
3,8-3,10:           NAME           'if'
3,11-3,12:          NAME           'a'
3,13-3,15:          OP             '=='
3,16-3,21:          STRING         "'foo'"
3,21-3,22:          OP             ':'
3,22-3,23:          NEWLINE        '\n'
4,0-4,12:           INDENT         '            '
4,12-4,18:          NAME           'return'
4,19-4,24:          STRING         "'foo'"
4,24-4,25:          NEWLINE        '\n'
5,8-5,8:            DEDENT         ''
5,8-5,14:           NAME           'return'
5,15-5,20:          STRING         "'bar'"
5,20-5,21:          NEWLINE        '\n'
6,0-6,0:            DEDENT         ''
6,0-6,0:            DEDENT         ''
6,0-6,0:            ENDMARKER      ''
```

How can I debug it and find the cause? The error message isn't giving any insight into the problem.
I am willing to fix it.

akshanshbhatt:
The tokenizer was throwing an error when the last line of the source code was dedented by one level less than the previous level. I debugged and fixed it.

Closes #358

### DIFF
--- a/src/lpython/parser/tokenizer.re
+++ b/src/lpython/parser/tokenizer.re
@@ -148,7 +148,8 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
     } else if(dedent == 2) {
         // Reduce the indent to `last_indent_length`
         long int sum = 0;
-        for (auto& n : indent_length) sum += n;
+        // for (auto& n : indent_length) sum += n;
+        sum = indent_length.back();
         if(sum != last_indent_length) {
             indent_length.pop_back();
             loc.first = loc.last;
--- a/src/lpython/parser/tokenizer.re
+++ b/src/lpython/parser/tokenizer.re
@@ -147,10 +147,7 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
         }
     } else if(dedent == 2) {
         // Reduce the indent to `last_indent_length`
-        long int sum = 0;
-        // for (auto& n : indent_length) sum += n;
-        sum = indent_length.back();
-        if(sum != last_indent_length) {
+        if((long int)indent_length.back() != last_indent_length) {
             indent_length.pop_back();
             loc.first = loc.last;
             return yytokentype::TK_DEDENT;
--- a/tests/reference/tokens-indent2-e702789.json
+++ b/tests/reference/tokens-indent2-e702789.json
@@ -1,0 +1,13 @@
+{
+    "basename": "tokens-indent2-e702789",
+    "cmd": "lpython --no-color --show-tokens {infile} -o {outfile}",
+    "infile": "tests/tokens/indent2.py",
+    "infile_hash": "8f79ea4e1f8564b61e3dfc63e2a3e9b44e9ae161753ea8a8710de206",
+    "outfile": null,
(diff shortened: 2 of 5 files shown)


certik:
@akshanshbhatt thanks for the PR! @Thirumalai-Shaktivel would you mind reviewing it please?

src/lpython/parser/tokenizer.re
```diff
@@ -148,7 +148,8 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
     } else if(dedent == 2) {
         // Reduce the indent to `last_indent_length`
         long int sum = 0;
-        for (auto& n : indent_length) sum += n;
+        // for (auto& n : indent_length) sum += n;
+        sum = indent_length.back();
         if(sum != last_indent_length) {
```
Thirumalai-Shaktivel, line 153:
```suggestion
        if((long int)indent_length.back() != last_indent_length) {
```

Thirumalai-Shaktivel: Also, Remove `long int sum = 0`.

Thirumalai-Shaktivel (commented):
Thanks for this PR @akshanshbhatt.
Yup, I just noticed that there is no need to sum up all the element's values in the `indent_length`, as the last element in the `indent_length` has the summed indent value.
I left some changes.

Thirumalai-Shaktivel (approved):
LGTM, Thanks!

certik:
Thanks @akshanshbhatt for the PR and thanks @Thirumalai-Shaktivel for the review!

PR #367, opened 2022-04-13
Improve chr ord impl

Shaikh-Ubaid:
Updated the implementation of `chr` and `ord` functions as discussed in [this comment](https://github.com/lcompilers/lpython/issues/1#issuecomment-1097514630) of #1 .

Both the `chr` and `ord` functions now support unicode values from 32 to 126 (including 32 and 126).


Shaikh-Ubaid:
Please review and please share if there are any updates/modifications needed

src/runtime/lpython_builtin.py
```diff
@@ -2,30 +2,29 @@
...
-    elif s == '1':
-        return 49
-#    else:
-#        exit(1)
+    if len(s) != 1:
+        return -1 # not a character
+    i: i32
+    for i in range(32, 127):
+        if chr(i) == s:
+            return i
```
certik, line 15: For now I think this is fine. This can be improved by casting `s` to an integer, once we support it in LPython (maybe we already do).

certik (approved):
Great job. I tested it locally as well and it works!

I left a comment how it can be improved in the future, but I think this is good to be merged.

Thanks for the PR.

src/runtime/lpython_builtin.py
Shaikh-Ubaid: Got it. Thank you for the feedback.

PR #369, opened 2022-04-14
Register the type mismatch error

namannimmo10:


src/libasr/asr_utils.h
```diff
@@ -138,7 +138,7 @@ static inline std::string type_to_str(const ASR::ttype_t *t)
             return "logical";
         }
         case ASR::ttypeType::Character: {
-            return "character";
+            return "string";
```
namannimmo10, line 141: In Fortran, there's an intrinsic data type `character` that stores characters and strings, while in Python, `str` is used everywhere.  Since the `libasr` part is shared by LPython and LFortran, I am skeptical about choosing one out of these two.  The error message looks pretty good here, though.
```console
semantic error: Type mismatch in assignment, the types must be compatible
 --> tests/errors/test_assign_type_mismatch.py:5:5
  |
5 |     a = "string"
  |     ^   ^^^^^^^^ type mismatch (integer and string)
```

tests/reference/asr-test_assign_type_mismatch-f9cd94e.stderr
```diff
@@ -1,4 +1,4 @@
-semantic error: Type mismatch in assignment, the types must be compatible
+semantic error: Type mismatch for an assignment. In LPython, the types must be compatible.
```
certik, line 1: I think the types must be compatible not just in LPython, but in Python in general. If you try "mypy", what does it say for this case?

certik: I actually like the previous message better I think. What we could do is to add multiple spans here with labels and improve the printing of the error message, then we can add more information, such as `In Python the types must be compatible" or something like that.

namannimmo10: _mypy_ won't give an error.
```console
$ cat a.py
def f():
  a: int
  a = "string"
  print(a)

f()
```
```console
$ python a.py
string
```
```console
$ mypy a.py
Success: no issues found in 1 source file
```

certik: That seems like a bug in mypy, isn't it? The type annotation for `a` is `int`, so how can it be a "string"? Unless mypy allows changing the type? Ok, try this:
```
a: int = "string"
```

namannimmo10: This isn't a bug :) Changing the type is actually common, so we can use` --strict`:
```console
% mypy a.py --strict
a.py:1: error: Function is missing a return type annotation
a.py:1: note: Use "-> None" if function does not return a value
a.py:2: error: Incompatible types in assignment (expression has type "str", variable has type "int")
a.py:5: error: Call to untyped function "f" in typed context
Found 3 errors in 1 file (checked 1 source file)
```

certik: I don't know what the default invocation of `mypy` without the `--strict` flag does, but MyPy indeed does not by default allow redefinitions of a type as documented here: https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-allow-redefinition, and even if you enable redefinitions with the `--allow-redefinition` flag, it will still not work in the example above, as documented at that page. You have to use the variable first.

All in all, the types must be compatible not just in LPython, but in CPython also, as enforced by MyPy.

src/libasr/asr_utils.h
certik: As mentioned in the other issue, we have to actually check more than just the type, we have to check dimension (if it is an array) and length (if it is just one character or a string).

So I would for now keep "character" in libasr, then in LFortran we check the len and return an error with either character or string; in LPython we always say string. Finally, we check dimensions in both, and if it is an array, we say array of strings, or something like that. Or possibly just write the type as one would write it in LPython, so the type is `str`, or possibly `str[5]` for an array (or `list[str]` for a list).

certik: We are planning to add Python and Fortran printing from ASR, into libasr. So we'll have to have a function that takes an ASR type, and returns `list[str]` (for LPython) or `character(len=5)` (for LFortran) etc.

PR #370, opened 2022-04-14
Raise error for a type mismatch in annassign

namannimmo10:


certik:
Can you please add a test for this error?

tests/reference/asr-test_annassign_type_mismatch-7dac7be.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Type mismatch in annotation-assignment, the types must be compatible
+ --> tests/errors/test_annassign_type_mismatch.py:4:5
+  |
+4 |     x: i32[4] = [1, 2, 3, 4]
+  |     ^           ^^^^^^^^^^^^ type mismatch (integer and list)
```
certik, line 5: So the type on the left is not just an integer, but an array of integers. So we should improve the error message to say it is an array.

certik: An idea what this can print is at https://github.com/lcompilers/lpython/issues/398#issuecomment-1104212520, but that's longer term. As part of this PR or a next PR, we change just print it as:
```

4 |     x: i32[4] = [1, 2, 3, 4]
  |     ^           ^^^^^^^^^^^^ type mismatch (i32[4] and list[i32])
```
We can do this by changing `ASRUtils::type_to_str` to `ASRUtils::type_to_str_python` and this function would print the type as `i32[4]`.

namannimmo10: I'll open a new PR for that soon. Thanks. We definitely need excellent and self-explanatory error messages.

PR #372, opened 2022-04-14
Overload complex functions

Smit-create:


certik:
We have a `ComplexConstructor` ASR node. Isn't that what this Python's builtin `complex` function does?

certik (approved):
It's consistent with what is already in `main`, so it's good to merge I think.

namannimmo10 (approved):

namannimmo10:
> We have a `ComplexConstructor` ASR node. Isn't that what this Python's builtin `complex` function does?

Yes, it seems. We could use a `ComplexConstructor` node here:

https://github.com/lcompilers/lpython/blob/22cc049ad41cf465e94bd2c8e625bb54eb4bac63/src/lpython/semantics/python_comptime_eval.h#L599

namannimmo10:
I take it back.   No, because we could have runtime values in the `ComplexConstructor`, while this file handles compile-time evaluation. That's what I thought when I was implementing this.

certik:
Yes, the `comptime` is compile time only, so there you have to use `ConstantComplex`. But the AST to ASR should just take the builtin `complex` function and return ComplexConstructor (and yes, provide a compile time value if it exists). Then we don't need any of the `complex` functions in the `builtin` module.

PR #373, opened 2022-04-14
Fix type mismatch error message for binary operators

ankitaS11:
This PR attempts to fix the error message which is raised when mismatched types are operated on a binary operator.

To reproduce:

Run a following script using lpython: (or modify `examples/expr2.py`)

```python
x = (2+3)*5+”s”
```

This, in the current state of the library, will throw the following error (observe `Not Implemented`):

```python
semantic error: Not Implemented: type mismatch in binary operator; only Integer, Real, Complex, Logical combinations and string concatenation/repetition are implemented for now.
 --> examples/expr2.py:3:9
  |
3 |     x = (2+3)*5+"s"
  |         ^^^^^^^ ^^^ type mismatch (integer and character)
```

Ideally, this should be just a `semantic error` with the corresponding error message. After this PR, the error message would look like:

```python
semantic error: type mismatch in binary operator; only Integer, Real, Complex, Logical combinations and string concatenation/repetition are implemented for now.
 --> examples/expr2.py:3:9
  |
3 |     x = (2+3)*5+"s"
  |         ^^^^^^^ ^^^ type mismatch (integer and character)
```

Thank you @certik for finding this bug, and also helping me with how to reproduce it locally! :)

cc: @certik

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -818,7 +818,7 @@ public:
             std::string ltype = ASRUtils::type_to_str(ASRUtils::expr_type(left));
             std::string rtype = ASRUtils::type_to_str(ASRUtils::expr_type(right));
             diag.add(diag::Diagnostic(
-                "Not Implemented: type mismatch in binary operator; only Integer, Real, Complex,"
+                "Type mismatch in binary operator; only Integer, Real, Complex,"
                 " Logical combinations and string concatenation/repetition are implemented for now.",
                 diag::Level::Error, diag::Stage::Semantic, {
                     diag::Label("type mismatch (" + ltype + " and " + rtype + ")",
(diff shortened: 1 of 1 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -818,7 +818,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
             std::string ltype = ASRUtils::type_to_str(ASRUtils::expr_type(left));
             std::string rtype = ASRUtils::type_to_str(ASRUtils::expr_type(right));
             diag.add(diag::Diagnostic(
-                "Not Implemented: type mismatch in binary operator; only Integer, Real, Complex,"
+                "Type mismatch in binary operator; only Integer, Real, Complex,"
                 " Logical combinations and string concatenation/repetition are implemented for now.",
```
certik, line 822: How about just this?
```suggestion
                "Type mismatch in binary operator; the types must be compatible",
```
I think we now implement enough combinations, that if something is not supported, we should simply fix it, or provide an error message just for the "not implemented" stuff, but a general message should be simple I think.

certik (commented):
Thanks for the PR! Looks great. I left a minor comment.

src/lpython/semantics/python_ast_to_asr.cpp
ankitaS11: That's a great suggestion @certik. The suggested error message is more concise and simple.

certik (approved):

certik:
Thanks for the fix!

PR #374, opened 2022-04-14
 Add `limits` include to `asr_utils.h` (g++ 11.2.0 compilation fix)

ankitaS11:
On g++ 11.2.0, the local build failed for me with an error of `numeric_limits` not being in the scope for `std`. The reason might be a missing `#include <limits>` in the `src/libasr/asr_utils.h` file.

This PR attempts to fix the above mentioned issue.

cc: @certik


certik (approved):
Thanks!

PR #375, opened 2022-04-14
Refactor README (reorder installation steps + color-formatting for commands)

ankitaS11:
This PR updates the README, notably following changes have been made:

1. Re-formatted the codes to have bash like color-formatting.
2. Reordered installation to include linux installation command -> git clone repo -> conda environment creation -> cmake building -> run tests.
3. Use `$nproc` instead of hard-coding `16` in the `cmake –build` command.

cc: @certik


README.md
```diff
@@ -30,51 +30,67 @@ https://github.com/conda-forge/miniforge/#download
...
-Create a Conda environment using the preexisting environment.yml file:
-
-    conda env create -f environment.yml
-    conda activate lp
-
 Install required packages (Linux - 64 bit):
 
-    sudo apt install binutils-dev zlib1g-dev
+```bash
+sudo apt install binutils-dev zlib1g-dev
```
certik, line 36: We should add `zlib` into the Conda environment, that should take care of it.

The `binutils` are currently needed for stacktraces on Linux --- I'd like to get rid of it and move fully to the same approach as macOS is using which does not require binutils, but I didn't have time to finish it yet.

README.md
```diff
@@ -30,51 +30,67 @@ https://github.com/conda-forge/miniforge/#download
...
+call build0.bat  # Windows
+```
 
 Compile LPython:
 
-    cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_LLVM=yes -DWITH_STACKTRACE=yes -DWITH_LFORTRAN_BINARY_MODFILES=no .
-    cmake --build . -j16
+```bash
+cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_LLVM=yes -DWITH_STACKTRACE=yes -DWITH_LFORTRAN_BINARY_MODFILES=no .
+cmake --build . -j$nproc
```
certik, line 64: I don't have `$nproc` defined. Is the `-j16` confusing? 

We could switch to ninja, which runs in parallel by default.

certik:
Thanks! Very useful.

README.md
ankitaS11: > We should add zlib into the Conda environment, that should take care of it.

Good point @certik, fixed it.

> The binutils are currently needed for stacktraces on Linux --- I'd like to get rid of it and move fully to the same approach as macOS is using which does not require binutils, but I didn't have time to finish it yet.

That makes sense, I can work on this and create a separate PR if it is okay with you. :)

README.md
ankitaS11: May be `$nproc` doesnot work on Mac, the reason it was confusing to me was that I only had 8 cores in my CPU but it had hardcoded 16 in the readme, but I guess it wouldn't matter much as it will always take 8 cores if that's the maximum count. 

I can revert this change, may be switching to ninja is an option but IMO not something thats needed right now.

certik (approved):

certik:
Thanks!

PR #377, opened 2022-04-15
math.trunc function implemented

Madhav2310:


integration_tests/test_math.py
```diff
@@ -125,6 +125,13 @@ def test_log1p():
 def test_expm1():
     assert expm1(1.0) - 1.71828182845904509 < eps
 
+def test_trunc():
+   i: f64
+   i = trunc(3.5)
+   assert i == 3
+   i = trunc(-4.5)
+   assert i == -4
```
Thirumalai-Shaktivel, line 133:
```suggestion
    i: f64
    i = trunc(3.5)
    assert i == 3
    i = trunc(-4.5)
    assert i == -4
```

src/runtime/math.py
```diff
@@ -217,6 +217,16 @@ def hypot(x: i32, y: i32) -> f64:
...
     return sqrt(1.0*(x**2 + y**2))
 
+def trunc(x: f64) -> f64:
+   """
+   Return x with the fractional part removed, leaving the integer part.
+   """
+   if x>0:
+       return floor(x)
+   else:
+       return ceil(x)
```
Thirumalai-Shaktivel, line 227:
```suggestion
    """
    Return x with the fractional part removed, leaving the integer part.
    """
    if x>0:
        return floor(x)
    else:
        return ceil(x)
```
Please use four spaces for the `tab`

src/runtime/math.py
```diff
@@ -217,6 +217,16 @@ def hypot(x: i32, y: i32) -> f64:
     """
     return sqrt(1.0*(x**2 + y**2))
 
+def trunc(x: f64) -> f64:
```
Thirumalai-Shaktivel, line 220: Here the return type should be `i64`, right?
See:
https://github.com/lcompilers/lpython/blob/9ac009e68fd9139957de7385fa00f86afb8672f2/src/runtime/math.py#L34-L48
https://github.com/lcompilers/lpython/blob/9ac009e68fd9139957de7385fa00f86afb8672f2/src/runtime/math.py#L60-L73

integration_tests/test_math.py
```diff
@@ -125,6 +125,13 @@ def test_log1p():
 def test_expm1():
     assert expm1(1.0) - 1.71828182845904509 < eps
 
+def test_trunc():
+   i: f64
+   i = trunc(3.5)
```
Thirumalai-Shaktivel, line 130: As suggested above, here the return type will be `i64` so change the declaration with `i: i64`.

Thirumalai-Shaktivel:
Can you also import `trunc` in `test_math.py`?

integration_tests/test_math.py
Madhav2310: Thanks, I'll implement these corrections right away!

Thirumalai-Shaktivel:
Perfect! Looks good to me, Can you please implement the same for `f32`?

src/runtime/math.py
```diff
@@ -1,4 +1,5 @@
-from ltypes import i32, f64, ccall
+from typing import overload
```
Thirumalai-Shaktivel, line 1:
```suggestion
```

Thirumalai-Shaktivel:
Very good, Thanks for this PR! LGTM

Smit-create (approved):
Thanks! @Madhav2310

Smit-create:
This looks as of now. Will fix if something needs

Smit-create:
@Madhav2310 Please rebase the PR and resolve conflicts

Madhav2310:
@Smit-create Done!

certik (commented):
Looks good, thanks!

HaoZeke:
This is good, but is an incomplete implementation in some sense. The [definition of](https://docs.python.org/3/library/math.html?highlight=trunc#math.trunc) `trunc` is:

> math.trunc(x)
Return x with the fractional part removed, leaving the integer part. This rounds toward 0: trunc() is equivalent to [floor()](https://docs.python.org/3/library/math.html?highlight=trunc#math.floor) for positive x, and equivalent to [ceil()](https://docs.python.org/3/library/math.html?highlight=trunc#math.ceil) for negative x. If x is not a float, delegates to [x.__trunc__](https://docs.python.org/3/reference/datamodel.html#object.__trunc__), which should return an [Integral](https://docs.python.org/3/library/numbers.html#numbers.Integral) value.

Do we have a path forward now for handling non-float values (i.e. objects)?

certik:
You can open up an issue to discuss this. This is an old PR that is already merged.

PR #378, opened 2022-04-15
Handle integer printing in the llvm backend

namannimmo10:
```python
def f():
  i1: i8
  i1 = 4
  print(i1)
  i2: i16
  i2 = 2
  print(i2)


f()
```
Before this PR it gives,
```console
% lpython a.py && ./a.out
 --> a.py:4:3
  |
4 |   print(i1)
  |   ^^^^^^^^^ 



```
Now it prints correctly,
```console
% lpython a.py && ./a.out
4
2
```


certik (approved):
That looks good, thanks!

PR #380, opened 2022-04-15
Remove fixed TODO

Smit-create:


integration_tests/test_max_min.py
```diff
@@ -1,4 +1,4 @@
-from ltypes import i32 ,f32 ,f64
+from ltypes import i32, f32, f64
```
namannimmo10, line 1: unused import
```suggestion
from ltypes import i32, f64
```

namannimmo10:
Sorry, you'd have to update the tests again.

PR #381, opened 2022-04-15
use nproc to utilise max processing units

Aasthaengg:
Using nproc so that there is better compatibility with different build machines and different CPUs

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Compile LPython:
 
 ```bash
 cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_LLVM=yes -DWITH_STACKTRACE=yes -DWITH_LFORTRAN_BINARY_MODFILES=no .
-cmake --build . -j16
+cmake --build . -j$(nproc)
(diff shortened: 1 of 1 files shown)


Smit-create:
Hi, Thanks for the PR. This seems to be working okay on a Linux machine but does not work on my Mac. Can you please verify?

Thirumalai-Shaktivel:
See: https://github.com/lcompilers/lpython/pull/375#discussion_r850784082

Aasthaengg:
@Smit-create Thanks for pointing it out. I just added an alternate command for Mac users.

certik:
What about Windows?

I would recommend not to complicate this, and just keep things simple. We can reduce to `-j8`. That way you can increase it if you have more processors.

Aasthaengg:
For windows we can write `%NUMBER_OF_PROCESSORS%`. 
Thank you for the recommendation @certik, sure `-j8` sounds good. 
I intended to restructure README for multiple operating systems in the same way that build is written.

PR #382, opened 2022-04-16
Implement comparison of logical types in the LLVM backend

namannimmo10:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2662,6 +2662,9 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 }
             }
         } else if (optype == ASR::ttypeType::Logical) {
+            // Convert True to 1 and False to 0 and then compare
```
namannimmo10, line 2665:
```suggestion
            // i1 -> i32
```

certik (approved):

PR #384, opened 2022-04-16
Allow message printing with `assert` in the LLVM backend

namannimmo10:


certik (approved):

PR #385, opened 2022-04-17
Implement missing cast conversions in cpp codegen

Smit-create:


tests/reference/cpp-test_builtin_pow-56b3f92.stdout
```diff
@@ -0,0 +1,199 @@
...
+        return _lpython_return_variable;
+    }
+    _lpython_return_variable = 1;
+    return _lpython_return_variable;
+}
+
+std::complex<float> __lpython_overloaded_9__complex(int x, int y)
+{
+    std::complex<float> _lpython_return_variable;
+    _lpython_return_variable = std::complex<double>(x) + std::complex<double>(y)*{0.000000, 1.000000};
```
namannimmo10, line 175: Will it compile? We should improve the C++ codegen for `complex` types.

Smit-create: Does not compile because of `{.., ..}`. Fixed that.

certik (approved):
Thanks for the fixes!

PR #387, opened 2022-04-17
Degrees radians overload to support more ltypes

Aasthaengg:
This PR has changes that overloads `degrees` and `radians` in `math.py` to support `i8, i16, i32, i64, f32, f64` ltypes. Releavant test code is added to `test_math.py`.


Aasthaengg:
Closing this due to unnecessary commits being present. Will make separate PR.

PR #389, opened 2022-04-18
Overload math degrees and radians to support more ltypes

Aasthaengg:
This PR includes changes to overload `degrees` and `radians` functions to support `i8, i16, i32, i64, f32, f64` ltypes. Relevant test code is added to `test_math.py`.

### DIFF
--- a/integration_tests/test_math.py
+++ b/integration_tests/test_math.py
@@ -30,13 +30,25 @@ def test_isqrt():
 
 
 def test_degrees():
+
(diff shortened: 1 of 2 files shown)


Aasthaengg:
@Smit-create thanks for your suggestion. Could you have a look at it now, please?

Smit-create (approved):
Thanks! LGTM.

certik (approved):

Issue #330, opened 2022-04-06
Implement Classes


certik:
We should implement classes. ASR has a basic support for them. So at least things that are already supported we should add to the ast to asr translation.

PR #391, opened 2022-04-18
Implement class

Smit-create:
#330

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -1487,6 +1487,21 @@ public:
         }
     }
 
+    void visit_ClassDef(const AST::ClassDef_t &x) {
+        SymbolTable *parent_scope = current_scope;
+        current_scope = al.make_new<SymbolTable>(parent_scope);
+        for (size_t i=0; i<x.n_body; i++) {
+            visit_stmt(*x.m_body[i]);
+        }
+        tmp = ASR::make_ClassType_t(al, x.base.base.loc, current_scope,
+                x.m_name, ASR::abiType::Source, ASR::accessType::Public);
+        tmp = ASR::make_Class_t(al, x.base.base.loc,
+                ASR::down_cast<ASR::symbol_t>(tmp), nullptr, 0);
+        ASR::symbol_t * t = ASR::down_cast<ASR::symbol_t>(tmp);
+        parent_scope->scope[x.m_name] = t;
+        current_scope = parent_scope;
+    }
+
     void create_GenericProcedure(const Location &loc) {
         for(auto &p: overload_defs) {
             std::string def_name = p.first;


certik:
Thanks for working on this.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1487,6 +1487,21 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+        current_scope = al.make_new<SymbolTable>(parent_scope);
+        for (size_t i=0; i<x.n_body; i++) {
+            visit_stmt(*x.m_body[i]);
+        }
+        tmp = ASR::make_ClassType_t(al, x.base.base.loc, current_scope,
+                x.m_name, ASR::abiType::Source, ASR::accessType::Public);
+        tmp = ASR::make_Class_t(al, x.base.base.loc,
+                ASR::down_cast<ASR::symbol_t>(tmp), nullptr, 0);
+        ASR::symbol_t * t = ASR::down_cast<ASR::symbol_t>(tmp);
+        parent_scope->scope[x.m_name] = t;
```
Smit-create, line 1501: @certik Does this look good?

On Python code:
```py
class Foo:
    def __init__(self, p):
        return p
```

ASR:
```
  File "/Users/thebigbool/repos/lpython/src/bin/lpython.cpp", line 755
    return emit_asr(arg_file, runtime_library_dir,
  File "/Users/thebigbool/repos/lpython/src/bin/lpython.cpp", line 169
  File "/Users/thebigbool/repos/lpython/src/libasr/diagnostics.cpp", line 71
    out += render_diagnostic(d, input, lm,
  File "/Users/thebigbool/repos/lpython/src/libasr/diagnostics.cpp", line 134
    populate_spans(d, lm, input);
  File "/Users/thebigbool/repos/lpython/src/libasr/diagnostics.cpp", line 121
    populate_span(s, lm, input);
  File "/Users/thebigbool/repos/lpython/src/libasr/diagnostics.cpp", line 113
    LFORTRAN_ASSERT(s.source_code.size() > 0)
AssertFailed: s.source_code.size() > 0
```

Also, what is the difference between `Class` and `ClassProcedure` Nodes in ASR?

certik: That's a bug in printing an error message, I just created an issue #398 to fix that.

Yes, I think that is a constructor. Fortran does not have constructors in this sense, so I would tackle those a little bit later. We should add whatever we need into ASR, but first let's get something like this working:

```
class Foo:
    def __init__(self: Foo, a: i32, b: i32, c: i32):
        self.a: i32 = a
        self.b: i32 = b
        self.c: i32 = c

    def get_a(self: Foo) -> i32:
        return self.a
```

I am not quite sure where you should annotate the variables inside the class, whether how I did it, or some other way.

The other thing we need are structs, which we can represent already in ASR, we just need some syntax in Python, possibly the "data classes" or whatever it is called:
```
@dataclass # or whatever it is called
class MyStruct:
    a: i32
    b: f64
    c: str
```
And it works like a struct.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1487,6 +1487,21 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+    void visit_ClassDef(const AST::ClassDef_t &x) {
+        SymbolTable *parent_scope = current_scope;
+        current_scope = al.make_new<SymbolTable>(parent_scope);
+        for (size_t i=0; i<x.n_body; i++) {
+            visit_stmt(*x.m_body[i]);
+        }
+        tmp = ASR::make_ClassType_t(al, x.base.base.loc, current_scope,
+                x.m_name, ASR::abiType::Source, ASR::accessType::Public);
+        tmp = ASR::make_Class_t(al, x.base.base.loc,
+                ASR::down_cast<ASR::symbol_t>(tmp), nullptr, 0);
```
certik, line 1499: The `Class` ASR node is used for a variable type, so in here we should just remove it.

src/lpython/semantics/python_ast_to_asr.cpp
czgdp1807: > I am not quite sure where you should annotate the variables inside the class, whether how I did it, or some other way.

The way you did it seems fine.

czgdp1807:
I implemented `dataclass` recently in #607. So I think this can be closed.

certik:
Well, dataclasses are implemented, but not regular classes. But we can do it later, that's a lower priority.

PR #397, opened 2022-04-19
Capitalize function of String.py implemented

Madhav2310:


Smit-create (commented):
Remove the unnecessary changes present in the diff

Madhav2310 (commented):
Removed commented code

integration_tests/test_string.py
```diff
@@ -0,0 +1,10 @@
+from string import (capitalize)
+from ltypes import i32, f64, f32, i64, c32, c64
+
+def test_capitalize():
```
namannimmo10, line 4: Thanks for the PR. You'd need to call this function -- and add this file to `run_tests.py`.

namannimmo10, line 4: Thanks for the PR. You'd need to call this function -- and add this file to `run_tests.py`.

src/runtime/string.py
```diff
@@ -0,0 +1,17 @@
...
+hexdigits: str = digits + 'abcdef' + 'ABCDEF'
+octdigits: str = '01234567'
+punctuation: str = "#$%&'()*+,-./:;<=>?@[\]^_`{|}~"
+
+def capitalize(s: str) -> str:
+    """
+    Return a copy of the string with its first character capitalized and the rest lowercased.
+    """
+    result: str
+    result = s[0].upper() + s[1:]
```
namannimmo10, line 16: Right now, `.upper()` won't work. We should implement this in the library.

namannimmo10, line 16: Right now, `.upper()` won't work. We should implement this in the library.

Madhav2310: Shall I wait for the implementation of .upper() or proceed via other techniques like adding to its ASCII value to convert to capital?

certik: You can implement upper as a function in `string.py` for now.

faze-geek: @certik @Madhav2310 I came across this while surfing another `string` issue . I have implemented upper as a function 
```
def upper1(s):        # plain python 
    alphabet = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
    result = ''
    for x in s:
        for pos in range(52):
            if alphabet[pos] == x:
                i = pos
        if x not in alphabet or i>=26:
            result += x
        else:
            result += alphabet[i+26]
    return result

def upper2(s):      # using some inbuilt functions like ord() , chr()
    result = ''
    for char in s:
        if ord(char) >= 97 and ord(char) <=122 :
            result += chr(ord(char) - 32)
        else :
            result += char
    return result
```
Adding related functions like `lower`,`islower` ,`isupper` wouldn't be difficult either. Should I go ahead with this ? I think these can be updated in `lpython_builtin.py` or simply make a new file `runtime/string.py` which would also support this PR !

certik: @faze-geek yes, go ahead and implement those, and add tests.

Madhav2310: @faze-geek Thank you for your inputs! I'll go ahead and commit these functions with this PR. Do tell if you come up with further additions. 
cc: @certik @czgdp1807

Madhav2310:
@namannimmo10 @czgdp1807 Could you please assist me with the following error?
```
(lp) madhav@madhav-ThinkPad-E590:~/lpython$ lpython test_string.py
  File "/home/madhav/lpython/src/bin/../runtime/lpython_parser.py", line 6, in <module>
    filename_out = sys.argv[2]
IndexError: list index out of range
The command 'python /home/madhav/lpython/src/bin/../runtime/lpython_parser.py test_string.py' failed.
```

Madhav2310:
Once this is properly implemented, I'll add more tests

faze-geek:
@Madhav2310 have you pushed from your main ? I think it was not in sync with the current master, so you should have fetched and pulled it before . Anyways It is better to open a new branch for a pr. I have tackled this issue previously-

You can restore your master branch like this:
```
# make sure the working directory is clean and you are on master
git status
git checkout master

# completely delete the last x wrongfully pushed commits
git reset --hard HEAD~x

# update local master from the main repo master
git remote add upstream <url of lcompilers/lpython>
git fetch upstream
git rebase upstream/master
```
That will restore your master branch to reflect the main one. To make a PR use
```
# first update your main
git checkout main
git fetch upstream
git rebase upstream/main

# Now create a new branch from your updated main
git checkout -b pr_mybranchname

# Now make changes, commit and push etc.
```

Madhav2310:
Yep, I'll make a new PR for this, thank you!

PR #402, opened 2022-04-20
fixed iterator syntax

yashsain:
here, in the if statement the syntax was incorrect. I have modified it to make it correct.


certik (approved):
Thanks for the fix. To be precise, the syntax only seems to be supported by some compilers, but not all.

Issue #398, opened 2022-04-19
Fixing bugs and improving printing of diagnostic messages


certik:
Now when we can test error messages (see `tests/errors`), we should work on improving how they are printed, and if we need any more information stored in the `Diagnostics` class, we can do that too.

Some bugs to fix:

* [ ] https://github.com/lcompilers/lpython/pull/391#discussion_r853102454

Urgent improvements:

* [ ] When two error labels are on the same line, we have to draw an ascii-art from the underline (both on the same line) to the label (on two separate lines)
* [ ] When error labels point to different files, we draw both files, make sure everything works with multiple files
* [ ] Error labels in a single file should be merged, and lines sorted (currently a later line in a file can be printed first)

Once these are fixed and tests added, we should work on good error messages, for example when you call a function `f(1, 2)` and some of the arguments don't agree, to highlight the declaration (possibly in different file), show where it was imported in the current file, show the expected type, show the current type, and so on.

Other improvements:

* [ ] Optionally using unicode box drawing characters
* [ ] Optionally experiment with highlighting the text itself (in red or blue, possibly in the background) instead of using the `~~~` and `----` underlines, this should save vertical space. This must be optional, as you need to be able to see colors. With `--no-color` we should still print underlines.
* [ ] We can also experiment with some nice html output, possibly printing the source code (syntax highlighted) and then highlighting errors/warnings/hints somehow, and when you hover over it with a mouse, show more information.

Thirumalai-Shaktivel:
Thanks for pointing it out!
This issue is related to `--indent` which was also reported [here](https://gitlab.com/lfortran/lfortran/-/issues/533)! Yup, we need to fix this...

certik:
Here is another one: https://github.com/lcompilers/lpython/pull/412#discussion_r856247040

Smit-create:
```
4 |     x: i32[4] = [1, 2, 3, 4]
  |     ^           ^^^^^^^^^^^^
  |     +-- i32[4]           |
  |                          +---- list[i32]
```
 Does this design seems general? because this is looking specifically to the assignment. Moreover, looking into this, I see that we might also need to pass additional information in semantic errors to show this kind of error.

certik:
Yes, the design is general (almost any message can be encoded), but the visualization of the diagnostic messages is not.

certik:
[...]

PR #408, opened 2022-04-21
Add support for list insert and remove

Smit-create:


certik (approved):
Excellent, very good. Do you want to add tests for the errors as well? I think we can start testing all error messages. That will help later on once we make changes as part of #398, to see how the messages change and that we didn't forget about some corner case. So we need a lot of tests for our error messages. Also from the tests we can see how the messages look like and try to improve them.

Smit-create:
> Do you want to add tests for the errors as well? I think we can start testing all error messages. That will help later on once we make changes as part of #398, to see how the messages change and that we didn't forget about some corner case

Sure, that makes sense.

PR #411, opened 2022-04-22
Fixes overload issue with subroutines

### LINKED ISSUE #399: Different behaviour of overload with return and without return
On this python code, `overload` works fine and prints `3` as the result:
```py
@overload
def add(a: i32) -> i32:
    return a

@overload
def add(a: i32, b: i32) -> i32:
    return a+b

print(add(1, 2))
```

But the following one fails:
```py
@overload
def add(a: i32):
    print(a)

@overload
def add(a: i32, b: i32):
    print(a+b)

add(1, 2)
```

ASR:
```
    throw LFortranException("ASR verify failed: " + error_msg);
LFortranException: ASR verify failed: 92:100: SubroutineCall::m_name cannot point outside of its symbol table
```

Smit-create:
Fixes #399

### DIFF
--- a/integration_tests/test_generics_01.py
+++ b/integration_tests/test_generics_01.py
@@ -1,6 +1,7 @@
 from overload_testing import foo, test
-from ltypes import overload, i32
+from ltypes import overload, i32, i64
 import overload_testing2
+
 
 @overload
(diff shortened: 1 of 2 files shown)


certik (approved):
Excellent, thanks!

PR #412, opened 2022-04-22
Test error messages

Smit-create:
https://github.com/lcompilers/lpython/pull/408#pullrequestreview-949218985

### DIFF
--- a/tests/reference/asr-test_list2-10ffdd7.stderr
+++ b/tests/reference/asr-test_list2-10ffdd7.stderr
@@ -1,0 +1,5 @@
+semantic error: append() takes exactly one argument
+ --> tests/errors/test_list2.py:6:5
+  |
+6 |     a.append(4, 5)
+  |     ^^^^^^^^^^^^^^ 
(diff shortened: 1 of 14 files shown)


tests/reference/asr-test_list2-10ffdd7.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: append() takes exactly one argument
+ --> tests/errors/test_list2.py:6:5
+  |
+6 |     a.append(4, 5)
+  |     ^^^^^^^^^^^^^^
```
certik, line 5: In the future, we can improve these messages, here is an idea:
```suggestion
6 |     a.append(4, 5)
  |       ^^^^^^ ┬  ┬                       
  |          │   └──┴── two arguments provided    
  |          │                              
  |        expected one argument       
```

This is how it looks like in a terminal (without colors):

Screen Shot 2022-04-22 at 7 46 34 AM

And of course there would be colors.

certik (approved):
This looks perfect, thank you!

PR #414, opened 2022-04-22
Add `ListConcat` node to the ASR

namannimmo10:


tests/list1.py
```diff
@@ -20,3 +20,4 @@ def test_List():
     a.append(10)
     a.remove(1)
     a.insert(2, 13)
+    a += [4, 5]
```
certik, line 23: Let's also add a test like this to make sure it works too:
```suggestion
    a += [4, 5]
    a = [6, 7] + a
```

certik (approved):
I think this looks good, I left a comment about another test.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -817,11 +817,25 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+            if (!ASRUtils::check_equal_type(left_type2, right_type2)) {
+                diag.add(diag::Diagnostic(
+                    "Both the lists should be of the same type for concatenation.",
+                    diag::Level::Error, diag::Stage::Semantic, {
+                        diag::Label("type mismatch (list[" + ltype + "] and list[" + rtype + "])",
+                                {left->base.loc, right->base.loc})
+                    })
+                );
+                throw SemanticAbort();
+            }
```
certik, line 833: We should upgrade `ASRUtils::type_to_str_python` to return `list[i32]` for this:
```suggestion
            std::string ltype = ASRUtils::type_to_str_python(left_type);
            std::string rtype = ASRUtils::type_to_str_python(right_type);
            if (!ASRUtils::check_equal_type(left_type, right_type)) {
                diag.add(diag::Diagnostic(
                    "Both the lists should be of the same type for concatenation.",
                    diag::Level::Error, diag::Stage::Semantic, {
                        diag::Label("type mismatch ('" + ltype + "' and '" + rtype + "')",
                                {left->base.loc, right->base.loc})
                    })
                );
                throw SemanticAbort();
            }
```
As well as `check_equal_type` to work for lists.

tests/list1.py
certik: Don't forget about this test.

certik:
I would suggest to split the upgraded type handling into a separate PR. Let's finish the list PR so that we can merge it, and then we can work on the error messages and type printing and clean that up.

namannimmo10:
Good idea. I'll open up one [WIP] PR to improve all the diagnostic messages.

tests/list1.py
namannimmo10: We need to improve `check_equal_type` in ASR utils to make this test work.

namannimmo10: I realized that the existing test worked because currently the `augassign` visitor doesn't really check the types. We should fix that as well.

namannimmo10: Done!

namannimmo10:
I have added the test as per your suggestion. Let me know if this looks good to merge since the CI is green. Will further improve the error messages in a subsequent PR.

certik (approved):
The list part is good now I think. Thanks!

Ok, let's merge it and improve upon it in subsequent PRs.

PR #415, opened 2022-04-23
BUG: Fix list/set mismatch type assign

Smit-create:
#398


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1845,6 +1847,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            std::string obj = "";
+            if (ASR::is_a<ASR::List_t>(*value_type)) {
+                target_type = ASR::down_cast<ASR::List_t>(target_type)->m_type;
+                value_type = ASR::down_cast<ASR::List_t>(value_type)->m_type;
+                obj = "list";
+            } else if (ASR::is_a<ASR::Set_t>(*value_type)) {
+                target_type = ASR::down_cast<ASR::Set_t>(target_type)->m_type;
+                value_type = ASR::down_cast<ASR::Set_t>(value_type)->m_type;
+                obj = "set";
+            }
```
certik, line 1859: Since you only use `obj` in the error below, you should move that inside the `if` statement.

I think we need to move the `List` downcast inside the `ASRUtils::check_equal_type`, not doing it before hand.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1845,6 +1847,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            if (!ASRUtils::check_equal_type(target_type, value_type)) {
```
certik, line 1860: Can this if statement every trigger? There is an identical check above. If it only triggers for List/Set, then making sure `check_equal_type` works for List and Set also should fix it.

Smit-create: Actually, for `List` we need to check `List[type]`. So, yes, it will trigger in the cases such as `List[i32]` and `List[str]`.

PR #418, opened 2022-04-24
Improve the error messages

namannimmo10:


certik:
Yes, this is good.

The next step is then to return `list[i32]` directly from the function if the type is a list.

certik (approved):
That looks good to me, thanks!

Issue #355, opened 2022-04-11
Allow unary operation on logical types


namannimmo10:
```console
$ cat t.py
def main0():
  b: bool
  b = True
  print(-b)


main0()
```
```console
$ lpython t.py && ./a.out
code generation error: Unary type not implemented yet in Logical


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).
```
```console
$ python t.py
-1
```

kinshukdua:
We need to somehow allow changing types of variables when using unary operators. I think this will be easier after we implement generic types.
For example the following code is valid in Cpython but not in lpython 
```
  x: bool
  x = True
  y: i32
  y = -x
  print(y)
```
```
5 |     y = -x
  |     ^   ^^ type mismatch (integer and logical)
```
Even if we fix mixing the types right now, we can't do the following without dynamic typing.
```
x: bool
x = -x
```

namannimmo10:
This works in the current `main` branch.

PR #419, opened 2022-04-24
Allow unary operation on non-constant variables

namannimmo10:
fixes one part of #355.
We also need to make this work at the backend.


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2196,7 +2222,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             throw SemanticAbort();
         }
         ASR::ttype_t *type = ASRUtils::TYPE(
-            ASR::make_Logical_t(al, x.base.base.loc, 4, nullptr, 0));
+            ASR::make_Logical_t(al, x.base.base.loc, 1, nullptr, 0));
```
certik, line 2225: Currently we use logical with kind=4 everywhere, so let's stick to that. If we are going to change it from 4 to 1, then let's do it as a separate PR and do it everywhere. I currently don't know the pros/cons.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1182,24 +1213,19 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
                     }
                     value = ASR::down_cast<ASR::expr_t>(
                         ASR::make_ComplexConstant_t(al, x.base.base.loc,
                         std::real(result), std::imag(result), operand_type));
                 }
             }
         }
-        tmp = ASR::make_UnaryOp_t(al, x.base.base.loc, op, operand, operand_type,
+        tmp = ASR::make_UnaryOp_t(al, x.base.base.loc, op, operand, result_type,
                               value);
```
certik, line 1229: I think that this will take `not 5` as an Integer operand and return a Logical. That is not the way to do it.

We have to treat UnaryOp just like we treat BinOp. It accepts and returns exactly the same type, always. And we use `Cast` to cast things correctly. So `not 5` should be represented as `UnaryOp(Cast(5, IntToLogical))`. 

Also, in our ASR refactoring, we split UnaryOp into `UnaryMinus` (accepts integer, float, complex and returns the same type) and `LogicalNot` (accepts and returns logical) and `IntegerBitNot` (I think that is the "Invert" thing). So that is how it should be implemented.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1102,13 +1102,50 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+            if (op == ASR::unaryopType::Not) {
+                result_type = logical_type;
+            } else if (op == ASR::unaryopType::Invert) {
+                throw SemanticError("Bad operand type for unary " +
+                    ASRUtils::unop_to_str(op) + ": " + ASRUtils::type_to_str_python(operand_type),
+                    x.base.base.loc);
+            } else {
+                result_type = operand_type;
+            }
+        }
```
czgdp1807, line 1147: Is it possible to merge `is_real` and `is_complex` in a single `if` branch? I see that their bodies are same.

src/lpython/semantics/python_ast_to_asr.cpp
czgdp1807: Shifting from 4 to 1 will help in using lesser memory in the backend. That's one advantage. However, I agree that we should do this separately otherwise the diff gets confusing here.

tests/expr10.py
```diff
@@ -7,6 +7,8 @@ def test_UnaryOp():
     b = not 5
     b = not -1
     b = not 0
+    b = True
+    a = -b
```
czgdp1807, line 11: I think changes also include integer, real and complex types? Can we add tests for these types as well and try to cover as many kinds of each of these types as possible?

src/lpython/semantics/python_ast_to_asr.cpp
certik: See https://github.com/lcompilers/lpython/pull/476.

certik: Note that how the backend actually implements `logical` with `kind=4` is a separate thing, and it can still implement it using just one byte --- in fact that's exactly what we do: https://github.com/lcompilers/lpython/blob/e5af88ae1454236105c416e08dcd9a0ae69b6f27/src/libasr/codegen/asr_to_llvm.cpp#L3375.

namannimmo10:
Unary operation on logical types works in the current `main` branch. Thanks for your suggestions.

PR #420, opened 2022-04-24
Add  `--time-report`

Abdelrahman-Kh-Fouad:
I deleted older pull requests, and make new one without any changes in formatting.
output example: 
```bash
File reading: 0.153125ms
Parsing: 62.0069ms
AST to ASR: 130.276ms
ASR to LLVM: 22.1845ms

```
and if any stage doesn't pass, it will print its time also.
@certik


certik:
Thanks a lot for the PR! I tested it locally and it seems to work.

PR #421, opened 2022-04-25
Add `ComplexRe` and `ComplexIm` in ASR

Smit-create:
https://gitlab.com/lfortran/lfortran/-/merge_requests/1700#note_921724331

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -3343,6 +3343,39 @@ public:
         arg_kind = ASRUtils::extract_kind_from_ttype_t(curr_type);
     }
 
+    void visit_ComplexRe(const ASR::ComplexRe_t &x) {
+        this->visit_expr_wrapper(x.m_arg, true);
+        ASR::ttype_t* curr_type = extract_ttype_t_from_expr(x.m_arg);
+        int arg_kind = ASRUtils::extract_kind_from_ttype_t(curr_type);
+        int dest_kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);
+        llvm::Value *re;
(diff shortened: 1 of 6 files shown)


certik:
Very cool, much simpler!

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3343,6 +3343,78 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+        }
+    }
+
+    void visit_ComplexIm(const ASR::ComplexIm_t &x) {
+        ASR::ttype_t* curr_type = extract_ttype_t_from_expr(x.m_arg);
+        int arg_kind = ASRUtils::extract_kind_from_ttype_t(curr_type);
+        llvm::Function *fn = nullptr;
+        llvm::Type *ret_type = nullptr;
+        llvm::AllocaInst *arg = nullptr;
+
```
Smit-create, line 3381: Does this look correct? It fails with llvm:
Python:
```py
def test_complex():
    x: c64
    x = 2j
    eps: f64
    eps = 1e-12
    assert abs(x.imag - 2) < eps
```
LLVM:
```
Store operand must be a pointer.
  store %complex_8* %4, %complex_8 %5, align 8
  %4 = alloca %complex_8, align 8
 %complex_8 = type { double, double }  %6 = call double @_lfortran_zaimag(%complex_8* %4)
  %8 = fsub double* %7, double 2.000000e+00
  store double* %8, double* %call_arg_value, align 8
 double
```

Smit-create: Fixed it.

certik (approved):

PR #423, opened 2022-04-25
Adding str slicing

### LINKED ISSUE #273: Fix string indexing
The following code:
```python
def f():
    s: str
    s = "abcdefg"
    print(s[1]) # prints "bcdefg", but should print "b"
    print(s[1:2]) # prints "cdefg", but should print "b"
    print(s[1:3]) # prints "defg", but should print "bc"

f()
```
Prints (correctly) using CPython:
```console
$ python a.py 
b
b
bc
```
But with LPython it prints:
```console
$ lpython a.py && ./a.out 
bcdefg
cdefg
defg
```

We have to fix it.

Abdullahjavednesar:
Fixes #273


src/runtime/impure/lfortran_intrinsics.c
```diff
@@ -606,7 +606,10 @@ LFORTRAN_API void _lfortran_strcat(char** s1, char** s2, char** dest)
 
 // idx1 and idx2 both start from 1
 LFORTRAN_API char* _lfortran_str_copy(char* s, int32_t idx1, int32_t idx2) {
-    char* dest_char = (char*)malloc(idx2-idx1+2);
+    if (idx2 == idx1) {
+        idx2 = idx1 + 1;
+    }
+    char* dest_char = (char*)malloc(idx2-idx1+1);
     for (int i=idx1; i <= idx2; i++)
```
namannimmo10, line 613:
```suggestion
    for (int i=idx1; i < idx2; i++)
```
With this change, all the tests in https://github.com/lcompilers/lpython/issues/273#issue-1176977053 will work. Right now they don't -- and the build is a success because string comparison is not handled properly - 

https://github.com/lcompilers/lpython/blob/c251f6427cd2671353d8264a6b82c087e17981ae/src/libasr/codegen/asr_to_llvm.cpp#L2646

With the latest commit, I get -
```console
def main():
  s: str
  s = "012345"
  print(s[2:4])
  print(s[1:6])
  print(s[0:6])
  s = "abcdefg"
  print(s[1])
  print(s[0])
  print(s[1:2])
  print(s[1:3])

main()

234
12345
012345
bc
ab
bc
bcd
```

Abdullahjavednesar: Good catch! I think I missed that. Seems like string comparison doesn't work well.

src/runtime/impure/lfortran_intrinsics.c
```diff
@@ -606,8 +606,11 @@ LFORTRAN_API void _lfortran_strcat(char** s1, char** s2, char** dest)
 
 // idx1 and idx2 both start from 1
 LFORTRAN_API char* _lfortran_str_copy(char* s, int32_t idx1, int32_t idx2) {
-    char* dest_char = (char*)malloc(idx2-idx1+2);
-    for (int i=idx1; i <= idx2; i++)
+    if (idx2 == idx1) {
+        idx2 = idx1 + 1;
+    }
+    char* dest_char = (char*)malloc(idx2-idx1+1);
+    for (int i=idx1; i < idx2; i++)
```
certik, line 613: I think this assumes that `idx1:idx2` is inclusive of both ends if `idx1==idx2`, but exclusive of the right end if `idx1 < idx2`?

That doesn't seem like a good design.

In ASR, we should do inclusive of both ends (if we don't do it for strings, we should). Then here in this functions, it seems easiest to assume inclusive of both ends, I believe that's how the previous logic was.

We might need to "add 1" in AST to ASR for string slicing for the right end, to convert from Python convention to ASR convention.

certik: Btw, things will also simplify once we split ArrayRef for string into StringItem (just one index) and StringSection (`idx1`, `idx2`).

Abdullahjavednesar: > I think this assumes that `idx1:idx2` is inclusive of both ends if `idx1==idx2`, but exclusive of the right end if `idx1 < idx2`?

Yes, I think that's how it works in cpython. Here, when `idx1==idx2` we are incrementing idx2, then the loop should be `i < idx2`, isn't that correct? 

in other case i.e, `idx1<idx2` which is string slicing, it should return substring from position `idx1` to `idx2-1`. So the loop should be `i<idx2`.

Cpython example:
```
>>> a = "0123456"
>>> a[2:3]
'2'
```

Abdullahjavednesar: Sure, I'll keep it as per current design.

certik (approved):
Very cool, thanks for fixing this!

PR #424, opened 2022-04-26
`pow` warning message

ankitaS11:
This will raise a warning when `pow` is being used instead of '**'


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2639,6 +2639,16 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                if (call_name == "pow") {
+                     diag.add(diag::Diagnostic(
+                    "Could have used '**' instead of 'pow'",
+                     diag::Level::Style, diag::Stage::Semantic, {
+                        diag::Label("'**' could be used instead" ,
+                                {x.base.base.loc}),
+                    })
+                );
+                }
+
```
certik, line 2651: I would indent this differently:
```suggestion
                     diag.add(diag::Diagnostic(
                         "Could have used '**' instead of 'pow'",
                         diag::Level::Style, diag::Stage::Semantic, {
                            diag::Label("'**' could be used instead",
                                    {x.base.base.loc}),
                        })
                    );
                }
```

tests/reference/asr-expr7-480ba2f.stderr
```diff
@@ -0,0 +1,11 @@
+warning: Could have used '**' instead of 'pow'
```
certik, line 1: I think this should say `style`, not `warning`. You might need to recompile and update tests.

certik:
The other thing to do is to update to the latest main:
```
git remote rename origin ankita
git remote add origin https://github.com/lcompilers/lpython.git
git fetch origin main:main
git rebase main
```
Now resolve the conflicts, so open `tests.toml` and reconcile the changes, then `git add tests/tests.toml`. Then compile and `./run_tests.py -u` again, then `git add tests/reference` and `git rebase --continue` and that should work. Then:
```
git push ankita +pow_warning
```

src/lpython/semantics/python_ast_to_asr.cpp
ankitaS11: Thanks @certik. It should be fixed now.

tests/reference/asr-expr7-480ba2f.stderr
ankitaS11: Done.  :)

ankitaS11:
> The other thing to do is to update to the latest main:
> 
> ```
> git remote rename origin ankita
> git remote add origin https://github.com/lcompilers/lpython.git
> git fetch origin main:main
> git rebase main
> ```
> 
> Now resolve the conflicts, so open `tests.toml` and reconcile the changes, then `git add tests/tests.toml`. Then compile and `./run_tests.py -u` again, then `git add tests/reference` and `git rebase --continue` and that should work. Then:
> 
> ```
> git push ankita +pow_warning
> ```

Thanks @certik, the PR is updated now, please take a look whenever you have the time :)

certik:
Thanks. I squashed the commits into just one, since the individual commits were not polished.

Thirumalai-Shaktivel:
I think this is not an error, this is something similar to the following:
```console
$ lfortran --show-ast tests/return1.f90
style suggestion: Use '==' instead of '.eq.'
 --> tests/return1.f90:8:11
  |
8 |     if (i .eq. a) return 1
  |           ^^^^ help: write this as '=='
```

ankitaS11:
> Why should this raise an error? CPython supports this:
> 
> ```python
> >>> pow(2, 2)
> 4
> ```

Hi @Smit-create,

We are not raising a warning/error here (I'll have to update the PR title and description, will do it by tomorrow) - but we are just raising a style hint suggestion that one can use an operator `**` instead of using the function `pow`.

cc: @certik for his views.

certik:
Yes, we are all in agreement now, I discussed this with Smit in the morning.

The intrinsic `pow` has two versions. One with two arguments, one with three. The version with two arguments seems exactly equivalent to `**` semantically, and for CPython the `**` version might even be faster. For LPython it wouldn't matter, as we would eventually improve our AST2ASR semantics to generate identical ASR for both.

The Zen of Python contains:

> There should be one-- and preferably only one --obvious way to do it.

And so we should use such "style hints" to help users write "canonical" Python code. In this case, instead of `pow(a,b)` they should write `a**b`. Fortran has a lot more of this, and LFortran follows a similar approach with "style suggestions" to hep users write modern, canonical, Fortran code, where there is just one obvious way to do it.

Of course, we need ways to be able to turn it off with a compiler flag eventually.

PR #425, opened 2022-04-26
Refactor the ASR

namannimmo10:


tests/reference/asr-test_builtin_len-55b0dec.json
```diff
@@ -5,9 +5,9 @@
...
-    "stdout": "asr-test_builtin_len-55b0dec.stdout",
-    "stdout_hash": "3bb11c9908c48b2a469c36cac5cc7219163ac291a697bd1a0ad1bd30",
-    "stderr": null,
-    "stderr_hash": null,
-    "returncode": 0
+    "stdout": null,
+    "stdout_hash": null,
+    "stderr": "asr-test_builtin_len-55b0dec.stderr",
+    "stderr_hash": "7a3316e0d950e6cd4ac0f1cb958d3f3c36d6dbba1e1392d711fd43fd",
+    "returncode": 1
```
certik, line 12: Not sure what the issue is here, but this might be two bugs:

* The CI should fail, because `asr-test_builtin_len-55b0dec.stderr` is not committed, but it succeeded
* The test should not fail

certik: I see, this test is commented out below, so that explains it.

src/libasr/ASR.asdl
```diff
@@ -227,13 +227,22 @@ expr
     | LogicalConstant(bool value, ttype type)
 
     | ListConstant(expr* args, ttype type)
+    | ListLen(expr arg, ttype type)
```
certik, line 230:
```suggestion
    | ListLen(expr arg, ttype type, expr? value)
```

src/libasr/ASR.asdl
```diff
@@ -227,13 +227,22 @@ expr
...
     | ListConcat(expr left, expr right, ttype type, expr? value)
 
     | ArrayConstant(expr* args, ttype type)
+
     | SetConstant(expr* elements, ttype type)
+    | SetLen(expr arg, ttype type)
```
certik, line 236:
```suggestion
    | SetLen(expr arg, ttype type, expr? value)
```

src/libasr/ASR.asdl
```diff
@@ -227,13 +227,22 @@ expr
...
+
     | TupleConstant(expr* elements, ttype type)
+    | TupleLen(expr arg, ttype type)
```
certik, line 239:
```suggestion
    | TupleLen(expr arg, ttype type, expr? value)
```

src/libasr/ASR.asdl
```diff
@@ -227,13 +227,22 @@ expr
...
+
     | StringConstant(string s, ttype type)
+    | StringLen(expr arg, ttype type)
```
certik, line 242:
```suggestion
    | StringLen(expr arg, ttype type, expr? value)
```

src/libasr/ASR.asdl
```diff
@@ -227,13 +227,22 @@ expr
...
+
     | DictConstant(expr* keys, expr* values, ttype type)
+    | DictLen(expr arg, ttype type)
```
certik, line 245:
```suggestion
    | DictLen(expr arg, ttype type, expr? value)
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2681,6 +2681,26 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         return nullptr;
     }
 
+    ASR::asr_t* handle_intrinsic_len(Allocator &al, Vec<ASR::call_arg_t> args,
+                                        const Location &loc) {
+        ASR::expr_t *arg = args[0].m_value;
+        ASR::ttype_t *type = ASRUtils::expr_type(arg);
+        ASR::ttype_t *to_type = ASRUtils::TYPE(ASR::make_Integer_t(al, loc,
+                                    4, nullptr, 0));
+        if (ASRUtils::is_character(*type)) {
```
certik, line 2690: In here you check if `arg` is a `StringConstant` and if so, set `value` to `std::string(down_cast<ASR::StringConstant_t>(arg)->m_s).size()` (converted to `IntegerConstant`).

integration_tests/test_builtin_bool.py
```diff
@@ -33,13 +33,13 @@ def test_bool():
...
-    assert bool(s)
-    assert not bool('')
-    assert bool("str")
+    # s: str
+    # s = ""
+    # assert not bool(s)
+    # s = "str"
+    # assert bool(s)
+    # assert not bool('')
+    # assert bool("str")
```
czgdp1807, line 42: Why are these tests commented out?

namannimmo10: It was a work in progress. Ready now. Can you take a look once again, please?

integration_tests/run_tests.py
```diff
@@ -26,9 +26,9 @@
     "test_builtin_bool.py",
     "test_builtin_pow.py",
     "test_builtin_int.py",
-    "test_builtin_len.py",
+    # "test_builtin_len.py",
     "test_builtin_float.py",
-    "test_builtin_str_02.py",
+    # "test_builtin_str_02.py",
```
czgdp1807, line 31: Why these two test files are commented out?

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2808,6 +2865,13 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     tmp = handle_intrinsic_float(al, args, x.base.base.loc);
                 }
                 return;
+            } else if (call_name == "len") {
+                if (args.size() != 1) {
+                    throw SemanticError("len() takes exactly one argument (" +
+                        std::to_string(args.size()) + " given)", x.base.base.loc);
+                }
```
czgdp1807, line 2872: We should put these two checks in `handle_intrinsic_len` IMO. The reason is simple if we are already passing `args` to `handle_intrinsic_len` then why keep a portion of `args` handling outside `handle_intrinsic_len`.

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -35,7 +35,7 @@ struct PythonIntrinsicProcedures {
             {"bool", {m_builtin, &eval_bool}},
             {"chr", {m_builtin, &eval_chr}},
             {"ord", {m_builtin, &eval_ord}},
-            {"len", {m_builtin, &eval_len}},
+            // {"len", {m_builtin, &eval_len}},
```
czgdp1807, line 38: Remove the dead code in a separate PR IMO. It gets confusing otherwise.

src/runtime/lpython_builtin.py
```diff
@@ -177,6 +170,12 @@ def bool(c: c32) -> bool:
 def bool(c: c64) -> bool:
     return c.real != 0.0 or _lfortran_zaimag(c) != 0.0
 
+@overload
+def bool(s: str) -> bool:
+    """
+    Return False when the argument `s` is an empty string, True otherwise.
+    """
+    return len(s) > 0
```
czgdp1807, line 178: I think the function is moved from one location to another in the same file. Can we avoid this, as it doesn't seem to produce any consequence except a diff in this PR.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2681,6 +2681,63 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            if (ASRUtils::expr_value(arg) != nullptr) {
+                int64_t ival = (int64_t)ASR::down_cast<ASR::DictConstant_t>(
+                                        ASRUtils::expr_value(arg))->n_keys;
+                value = ASR::down_cast<ASR::expr_t>(make_IntegerConstant_t(al,
+                                loc, ival, to_type));
+            }
+            return (ASR::asr_t *)ASR::down_cast<ASR::expr_t>(
+                    ASR::make_DictLen_t(al, loc, arg, to_type, value));
+        }
+        return nullptr;
```
czgdp1807, line 2738: An error message would be better here. It should say that `len` is only supported for `Set`, `Character/String`, `Dict`, `List`. Returning `nullptr` now would create problems in future because then most probably we will receive a Segmentation fault while compiling a big library using Lpython. So, debugging process will involve first figuring out why segmentation fault is there and then getting to know that Ah!, the support for `len` isn't there for the input type. A nice error message would help in avoiding that hustle.

tests/constants1.py
```diff
@@ -35,13 +35,12 @@ def test_len():
...
-    # a = len([1, 2, 3])
-    # a = len([[-4, -5, -6], [-1, -2, -3]])
-    # a = len({1, 2, 3})
-    # a = len({1: "c", 2: "b", 3: "c"})
+    a = len((1, 2, 3))
+    a = len((("c", "b", 3.4), ("c", 3, 5.6)))
+    a = len([1, 2, 3])
+    a = len([[-4, -5, -6], [-1, -2, -3]])
+    a = len({1, 2, 3})
+    a = len({1: "c", 2: "b", 3: "c"})
```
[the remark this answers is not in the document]
Smit-create: `l.append()` won't work with LLVM at present. It's only supported at ASR level.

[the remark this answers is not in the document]
czgdp1807: We can test only till ASR level in that case. If `constants1.py` is tested on LLVM, then may be create a new test file and add the tests using `l.append` and test it till ASR level only.

integration_tests/run_tests.py
certik: We fixed this in b07a4b31e998ca4f8a6081c221cd985367df9e1d.

src/lpython/semantics/python_ast_to_asr.cpp
certik: Fixed in 593b8c448a62816d8ab323481a1918c18e70694f.

czgdp1807: Great. Thanks for doing this.

tests/constants1.py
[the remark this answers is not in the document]
namannimmo10: I have added this in `constants1.py` which is not tested on LLVM, only on ASR. Thanks!

PR #426, opened 2022-04-27
Parse Subscription and Dictionary

Thirumalai-Shaktivel:


src/lpython/parser/tokenizer.re
```diff
@@ -262,6 +262,9 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
             }
             end { RET(END_OF_FILE); }
             whitespace {
+                if(cur[0] == '#') {
+                    continue;
+                }
```
Thirumalai-Shaktivel, line 267: Previously, the following was recognized as `TK_EOLCOMMENT`
```python
     # Comment
```
After making the above change, it is recognized as `TK_COMMENT`
This change was made to recognize the comment after the colon 
```python 
for i in range(N): # type: parallel
```

src/lpython/parser/tokenizer.re
```diff
@@ -315,6 +318,12 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
 
             // Tokens
             newline {
+                if (last_token == yytokentype::TK_COMMA
+                    || last_token == yytokentype::TK_LPAREN
+                    || (last_token == yytokentype::TK_NAME && cur[0] == ')')) {
+                    continue;
+                }
```
Thirumalai-Shaktivel, line 325:
```python
from x import (
    a,
    b
)
```
But the following doesn't work:
from x import ( # Comment
    a, # Comment
    b # Comment
)

src/lpython/parser/semantics.h
```diff
@@ -273,18 +274,47 @@ Vec<ast_t*> MERGE_EXPR(Allocator &al, ast_t *x, ast_t *y) {
...
 #define INTEGER(x, l) make_ConstantInt_t(p.m_a, l, \
-        std::stoi(x.int_n.str()), nullptr)
+        std::stol(x.int_n.str()), nullptr)
 #define STRING(x, l) make_ConstantStr_t(p.m_a, l, x.c_str(p.m_a), nullptr)
 #define FLOAT(x, l) make_ConstantFloat_t(p.m_a, l, \
-        std::stof(x.c_str(p.m_a)), nullptr)
+        std::stod(x.c_str(p.m_a)), nullptr)
 #define COMPLEX(x, l) make_ConstantComplex_t(p.m_a, l, \
-        0, std::stof(x.int_n.str()), nullptr)
+        0, std::stod(x.int_n.str()), nullptr)
```
certik, line 282: These conversions should be done in the tokenizer, not here.

src/lpython/parser/tokenizer.re
certik: I think we have to track parentheses, and if we are inside a parentheses, we do not emit indents. Here is how CPython tracks parentheses: https://github.com/python/cpython/blob/b156578bd63a94fe3d80a4c6a8dbc067feba5d06/Parser/tokenizer.c#L2012

certik: After implementing
```
MAX_PAREN_LEVEL=200
char parenstack[MAX_PAREN_LEVEL]
size_t parenlevel
```
Then replace the above with: `if (parenlevel > 0) continue` or something like that.

certik:
I would:

* Split dict parts into a separate PR, only add tests that pass (possibly single line only), add AST test and we can merge that soon
* Add the int/float conversion into a separate PR, we can merge that
* Then implement the parentheses stack (see the above comment how to do it), then use the `parenlevel` to not emit indentation inside them and write tokenizer tests for all combinations. Also add tests into `tests/errors` and test that mismatched parentheses give an error message (`([)]`), as well as unmatched (`[`).

src/lpython/parser/semantics.h
```diff
@@ -287,4 +288,40 @@ Vec<ast_t*> MERGE_EXPR(Allocator &al, ast_t *x, ast_t *y) {
 #define ATTRIBUTE_REF(val, attr, l) make_Attribute_t(p.m_a, l, \
         EXPR(val), name2char(attr), expr_contextType::Load)
 
+expr_t* CHECK_TUPLE(expr_t *x) {
+    if(is_a<Tuple_t>(*x) && down_cast<Tuple_t>(x)->n_elts == 1) {
+        return down_cast<Tuple_t>(x)->m_elts[0];
+    } else {
+        return x;
+    }
+}
```
Thirumalai-Shaktivel, line 297: This function makes sure that the `Tuple` node is not created for a single element.
For ex:
```python
    z = x[1]
```

certik:
Looks good.

certik:
You might need to update the tests, or there is some error in tests.

certik (approved):
Looks great, all straightforward.

Thirumalai-Shaktivel:
Thanks for reviewing and merging this PR...!

PR #429, opened 2022-04-27
Implement `StringLen` visitor in the CPP backend

namannimmo10:


src/libasr/codegen/asr_to_cpp.cpp
```diff
@@ -941,6 +937,17 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
...
 
+    void visit_StringLen(const ASR::StringLen_t &x) {
+        std::string out = "";
+        this->visit_expr(*x.m_arg);
+        if (x.m_value) {
+            out += "std::string(" + src + ").length()";
+        } else {
+            out += src + ".length()";
+        }
+        src = out;
```
certik, line 948: I didn't know `std::string` also has a `length()` method. It seems equivalent to `.size()`, which seems what is used in other containers as well, why not use it also?

I would recommend to just do exactly what the original implementation did. First of all, you have to put parentheses around it, since the `src` can be things like `a+b` where `a` and `b` are strings.

Second, inserting `std::string` can be superfluous sometimes. Rather, we should always make sure that every string is an instance of `std::string`, so `StringConstant` should just return `std::string("something")`, then just doing `"(" + src + ").size()"` will work.

namannimmo10: > I didn't know `std::string` also has a `length()` method. It seems equivalent to `.size()`, which seems what is used in other containers as well, why not use it also?

+1 to that.

I tried your other suggestion but for this code,
```python
def main():
    x: str
    x = "abc"
    print(x)
    print(len(x))
    y: str
    y = x + "def"
    z: str
    z = x + y
    print(len(z))
    print(len(y))
    print(len("abc" + "def"))

    print("separate")
    print("abcd")
    print(x)

main()
```
I get,
```cpp
    std::string x;
    std::string y;
    std::string z;
    x = std::string("abc");
    std::cout << x << std::endl;
    std::cout << (x).size() << std::endl;
    y = std::string(x) + std::string(std::string("def"));
    z = std::string(x) + std::string(y);
    std::cout << (z).size() << std::endl;
    std::cout << (y).size() << std::endl;
    std::cout << (std::string(std::string("abc")) + std::string(std::string("def"))).size() << std::endl;
    std::cout << std::string("separate") << std::endl;
    std::cout << std::string("abcd") << std::endl;
    std::cout << x << std::endl;
```
But with the current implementation (+ changing `length()` to `size()`), I get-
```cpp
    std::string x;
    std::string y;
    std::string z;
    x = "abc";
    std::cout << x << std::endl;
    std::cout << x.size() << std::endl;
    y = std::string(x) + std::string("def");
    z = std::string(x) + std::string(y);
    std::cout << z.size() << std::endl;
    std::cout << y.size() << std::endl;
    std::cout << std::string(std::string("abc") + std::string("def")).size() << std::endl;
    std::cout << "separate" << std::endl;
    std::cout << "abcd" << std::endl;
    std::cout << x << std::endl;
```
It seems to me that the latter one is neat because
> `std::string` can be superfluous sometimes

integration_tests/test_builtin_len.py
```diff
@@ -6,5 +6,13 @@ def test_len():
...
     assert len("abcd") == 4
     assert len("") == 0
+    t: str
+    t = "efg"
+    s = "abcd" + t
+    i: i32
+    i = len(s)
+    assert i == 7
+    i = len("abc" + "def")
+    assert i == 6
```
certik, line 16: Add a test:
```
assert len(t + s) == 10
```
It seems based on the code that the result will be:
```
assert t + s.size() == 0;
```
But rather it should be:
```
assert (t+s).size() == 0;
```
```

(We also have to implement the `assert` node correctly, probably using a C++ macro, but that's a separate issue.)

File: src/libasr/codegen/asr_to_cpp.cpp [hunk 3]
```diff
@@ -941,6 +937,17 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
         src = out;
     }
 
+    void visit_StringLen(const ASR::StringLen_t &x) {
+        std::string out = "";
+        this->visit_expr(*x.m_arg);
+        if (x.m_value) {
+            out += "std::string(" + src + ").size()";
```
certik, line 944:
```suggestion
            out += "(" + src + ").size()";
```

src/libasr/codegen/asr_to_cpp.cpp
certik: Apply the diff:
```diff
--- a/src/libasr/codegen/asr_to_cpp.cpp
+++ b/src/libasr/codegen/asr_to_cpp.cpp
@@ -577,7 +577,7 @@ Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
     }
 
     void visit_StringConstant(const ASR::StringConstant_t &x) {
-        src = "\"" + std::string(x.m_s) + "\"";
+        src = "std::string(\"" + std::string(x.m_s) + "\")";
         last_expr_precedence = 2;
     }
 
```

certik: Also add here:
```c++
last_expr_precedence = 2
```

certik:
I am going to close this PR and delete the branch, as it was pushed into the main lpython repository by a mistake.

PR #434, opened 2022-04-30
Implementation for negative indexing in str

Abdullahjavednesar:


certik (approved):

src/runtime/impure/lfortran_intrinsics.c
```diff
@@ -616,6 +616,13 @@ LFORTRAN_API void _lfortran_strcat(char** s1, char** s2, char** dest)
 
 // idx1 and idx2 both start from 1
 LFORTRAN_API char* _lfortran_str_copy(char* s, int32_t idx1, int32_t idx2) {
+    int s_len = strlen(s);
+    if(idx1 <= 0) {
+        idx1 = s_len + idx1;
```
Smit-create, line 621: This fails for:
```py
def f():
    x: str
    x = "abcde"
    y: str
    y = x[-4:-1]
    print(y)

f()
```

I get `e`, whereas CPython gives `bcd`. Is this related to this PR?

Abdullahjavednesar: No, its just negative indexing thats been worked on here, #423 covers slicing, which is in progress.

PR #437, opened 2022-05-02
Raise error for zero division

namannimmo10:


certik (approved):
Perfect, thanks! Generally comparing a floating point directly to 0.0 is problematic, but in this case I think it is fine: if it is exactly equal to 0.0, then I think the operation would fail otherwise, so it is good to just give an error message. It can also fail if it is close to 0.0, and that would not get caught here, but it would give an error (eventually) at runtime, so I think this PR is good to go.

namannimmo10:
Thanks!

Issue #433, opened 2022-04-30
Add strict type checking in tuple


namannimmo10:
```python
def main():
  t: tuple[i32, str]
  t = (1, 2)

main()
```
This will generate an ASR, but actually should give an error that the second element is expected to be of `str` type.

namannimmo10:
Same goes for dictionaries:
```python
def main():
  d: dict[i32, i32]
  d = {1: "a", 2: "b"}

main()
```

yashsain:
hey, I'd like to work on this. I am relatively new to open source, so can you please guide me a bit as to how to proceed. Thanks

certik:
See #439 which fixes something similar for lists. You have to implement a similar logic for tuples.

PR #438, opened 2022-05-02
Add strict type checking for dictionary

namannimmo10:
#433


certik (approved):
Awesome, thanks for this!

PR #440, opened 2022-05-03
Create new file to handle attributes

### LINKED ISSUE #431: Proposal to create separate files for attribute
As we have started supporting some of the basic Python data types such as `list`, `dict`, `set`, etc, we also plan to allow operations on these data types such as `append`, `remove`, `add`, `insert`, etc. This in python is done with the help of attributes. So, in order to keep the design more compact and cleaner, I am thinking to separate out the attributes part in a different file(much like `compile_time_eval`) and use a map to access them. Let me know your views and suggestions.

Presently, we are just writing everything here:
https://github.com/lcompilers/lpython/blob/a1d726f9e2f740c09a1fe7123e3a7d215578da66/src/lpython/semantics/python_ast_to_asr.cpp#L2424

Smit-create:
Fixes #431

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -359,7 +359,7 @@ public:
 
     void handle_attribute(ASR::symbol_t *s, std::string attr_name,
                 const Location &loc, Vec<ASR::expr_t*> &args) {
-        tmp = attr_handler.get_attribute(s, attr_name, al, loc, args);
+        tmp = attr_handler.get_attribute(s, attr_name, al, loc, args, diag);
         return;
     }
 
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -2762,49 +2762,13 @@ public:
                         if (current_scope->scope.find(mod_name) != current_scope->scope.end()) {
                             // this case when we have variable and attribute
                             st = current_scope->scope[mod_name];
-                            ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(st);
-                            ASR::ttype_t *type = v->m_type;
-                            if (ASR::is_a<ASR::Dict_t>(*type)) {
-                                if (call_name == "get") {
-                                    ASR::expr_t *def = nullptr;
-                                    if (args.size() > 2 || args.size() < 1) {
-                                        throw SemanticError("'get' takes atleast 1 and atmost 2 arguments",
-                                            x.base.base.loc);
-                                    }
-                                    ASR::ttype_t *key_type = ASR::down_cast<ASR::Dict_t>(type)->m_key_type;
-                                    ASR::ttype_t *value_type = ASR::down_cast<ASR::Dict_t>(type)->m_value_type;
-                                    if (args.size() == 2) {
-                                        def = args[1].m_value;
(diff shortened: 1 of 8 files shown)


Smit-create:
This works fine with `ListAppend` for now. Will update the PR to shift the complete support.

tests/reference/asr-test_dict3-d28f38f.stderr
```diff
@@ -1,5 +1,5 @@
-semantic error: Type mismatch in get's key value, the types must be compatible
+semantic error: Found type mismatch in 'get'.Type mismatch (found: 'i32', expected: 'str')
  --> tests/errors/test_dict3.py:5:15
   |
 5 |     x = y.get(1)
-  |               ^ type mismatch (found: 'i32', expected: 'str')
+  |               ^
```
certik, line 5: This seems like a change for the worse? We want to go the other direction and highlight `x` as well, and "attach" labels to both parts showing what the types are.

I would recommend not to change errors in this PR, which should only do refactoring and not change tests.

certik: I see --- the errors were possibly consolidated. That is fine to change some tests then, but I would still try to restore the previous error, and while doing it, even highlight (underline) the `x` as well.

certik: My bad --- in this error, we do not want to highlight `x`, as the mismatch is in the argument of the `get` method. So just one highlight, but I would still add some meaningful label, such as the one that was there before.

Smit-create: Previously, it was because of a `Diagnostic` error. I'll pass that as the argument to restore the previous error.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2846,49 +2762,13 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
-                                    tmp = make_DictItem_t(al, x.base.base.loc, st, args[0].m_value, def,
-                                                          value_type);
-                                    return;
-                                }
+                            Vec<ASR::expr_t*> eles;
+                            eles.reserve(al, x.n_args);
+                            for (size_t i=0; i<x.n_args; i++) {
+                                eles.push_back(al, args[i].m_value);
                             }
+                            handle_attribute(st, at->m_attr, x.base.base.loc, eles);
```
certik, line 2770: You might need to pass in `diag` as an argument, so that the handler can report more complicated errors.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -355,6 +357,12 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         return v;
     }
 
+    void handle_attribute(ASR::symbol_t *s, std::string attr_name,
+                const Location &loc, Vec<ASR::expr_t*> &args) {
+        tmp = attr_handler.get_attribute(s, attr_name, al, loc, args);
```
certik, line 362: To be precise, since `handle_attribute` is a method, no need to pass `diag` in it, but pass it in here into `get_attribute`.

certik:
This looks really good, much cleaner! I left a few comments, but it's really just one comment: pass in `diag`, and then emit more richer error messages by putting the errors into `diag` directly.

certik (approved):

PR #442, opened 2022-05-03
Support for string repeat

Abdullahjavednesar:


src/runtime/impure/lfortran_intrinsics.c
```diff
@@ -614,6 +614,24 @@ LFORTRAN_API void _lfortran_strcat(char** s1, char** s2, char** dest)
...
+//repeat str for n time
+LFORTRAN_API void _lfortran_strrepeat(char** s, int32_t n, char** dest)
+{
+    int cntr = 0;
+    char trmn = '\0';
+    int s_len = strlen(*s);
+    int trmn_size = sizeof(trmn);
+    char* dest_char = (char*)malloc(s_len*n+trmn_size);
+    for (int i = 0; i < n; i++) {
+        for (int j =0; j < s_len; j++) {
```
certik, line 626:
```suggestion
        for (int j = 0; j < s_len; j++) {
```

integration_tests/test_str_01.py
```diff
@@ -33,10 +33,16 @@ def test_str_slice():
     assert a[2:4] == "23"
     assert a[2:3] == a[2]
 
+def test_str_repeat():
+    a: str
+    a = "Xyz"
+    assert a*3 == "XyzXyzXyz"
+
```
namannimmo10, line 40: Thanks for the PR! :) Can we add these:
```diff
diff --git a/integration_tests/test_str_01.py b/integration_tests/test_str_01.py
index 348fdc66a..39e4ba91a 100644
--- a/integration_tests/test_str_01.py
+++ b/integration_tests/test_str_01.py
@@ -37,6 +37,8 @@ def test_str_repeat():
     a: str
     a = "Xyz"
     assert a*3 == "XyzXyzXyz"
+    assert 3*a*3 == "XyzXyzXyzXyzXyzXyzXyzXyzXyz"
+    assert a*-1 == ""
 
 def check():
     f()
```
Again, the string comparison doesn't work properly yet, so we should work on that too.

Abdullahjavednesar: Good catch, we need to cover those cases as well. Thanks for the review!

src/runtime/impure/lfortran_intrinsics.c
certik: Note that next time you can just click on the "Commit suggestion" button.

certik (approved):

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2821,7 +2845,12 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         llvm::Value *left_val = tmp;
         this->visit_expr_wrapper(x.m_right, true);
         llvm::Value *right_val = tmp;
-        tmp = lfortran_strop(left_val, right_val, "_lfortran_strrepeat");
+        if (ASRUtils::expr_type(x.m_right)->type == ASR::ttypeType::Integer) {
+            tmp = lfortran_strrepeat(left_val, right_val);
+        }
+        else if (ASRUtils::expr_type(x.m_left)->type == ASR::ttypeType::Integer) {
+            tmp = lfortran_strrepeat(right_val, left_val);
+        }
```
certik, line 2853: P.S. Once we create a `StringRepeat` node, we should swap the arguments in AST->ASR phase, so that the first argument is always a string, and a second argument an integer.

Abdullahjavednesar: Noted, I'll implement that out in another PR. I presume it should be similar to #447.

certik: Yes, something like that. If you don't get to it, just open up an issue please, so that we don't forget.

Abdullahjavednesar: Sure, here it is #462.

Issue #242, opened 2022-03-16
How to implement sqrt(x) for x an array


certik:
For a 1D array:
```
from ltypes import f64, TypeVar
n = TypeVar('n')

@overload
def sqrt(x: f64[n]) -> f64[n]:
    r: f64[n] = empty(n)
    i: i32
    for i in range(size(x)):
        r[i] = sqrt(x[i])
    return r
```

For a 2d array:
```
from ltypes import f64, TypeVar
n = TypeVar('n')
m = TypeVar('m')

@overload
def sqrt(x: f64[n,m]) -> f64[n,m]:
    r: f64[n,m] = empty((n,m))
    i: i32
    j: i32
    for i in range(size(x)):
        for j in range(size(x)):
            r[i,j] = sqrt(x[i,j])
    return r
```

certik:
For now, let's pass `n` as an argument.

PR #443, opened 2022-05-04
Add fsum() and prod() to math.py

Aasthaengg:
This PR includes overloaded functions to handle different `ltypes` for finding the sum and product of arrays. 
The arguments necessary are length of array and the array itself in variable representation -> `def fsum(n: i32, arr: f64[:]) -> f64`

I used #242 as reference.

### DIFF
--- a/integration_tests/test_math.py
+++ b/integration_tests/test_math.py
@@ -1,6 +1,7 @@
 from math import (factorial, isqrt, perm, comb, degrees, radians, exp, pow,
-                  ldexp, fabs, gcd, lcm, floor, ceil, remainder, expm1, fmod, log1p, trunc)
-from ltypes import i32, f64, i64
+                  ldexp, fabs, gcd, lcm, floor, ceil, remainder, expm1, fmod, log1p, trunc, fsum, prod)
+from ltypes import i8, i16, i32, i64, f32, f64
+from numpy import empty
 
 eps: f64
 eps = 1e-12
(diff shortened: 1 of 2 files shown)


Aasthaengg:
Removed integration tests for now as pythonic `math.fsum()` and `math.prod()` expect only 1 argument but here 2 arguments are being passed.

certik:
I think the Python functions accept a list? You can try to create one with a list, and create an issue if it fails, so that we can fix it.

There are similar functions in NumPy, those would accept arrays.

Madhav2310:
@Aasthaengg are you working on this?

Smit-create:
@czgdp1807 This looks ready now

integration_tests/test_math.py
```diff
@@ -183,6 +284,8 @@ def check():
     test_expm1()
     test_log1p()
     test_trunc()
+    # test_fsum()
+    # test_prod()
```
certik, line 288: Should this be uncommented?

certik: I think we should not be adding any more functions that do not work with LPython.

Smit-create: Ahh, I missed that. I'll check that again. Thanks for pointing!

src/runtime/math.py
```diff
@@ -87,6 +86,167 @@ def ceil(x: f32) -> i32:
...
+    for i in range(n):
+        sum += arr[i]
+    return sum
+
+@overload
+def fsum(n: i32, arr: f32[:]) -> f64:
+    """
+    Sum of the elements of `arr`.
+    """
+    sum: f64
```
Smit-create, line 149: The API differs here: https://docs.python.org/3/library/math.html. We have `size` as the argument which is not available in CPython's `math.py`

src/runtime/math.py
```diff
@@ -87,6 +86,167 @@ def ceil(x: f32) -> i32:
         return r
     return r + 1
 
+# fsum
+# supported data types: i8, i16, i32, i64, f32, f64
+
+@overload
+def fsum(n: i32, arr: i8[:]) -> f64:
```
certik, line 93: All these functions should follow CPython's API: https://docs.python.org/3/library/math.html#math.fsum

```suggestion
def fsum(iterable: list[i8]) -> f64:
```

We do not support "iterable" yet, so at first just getting it working for lists would be helpful. Later we might support "iterable", so it would start working for other types, such as NumPy arrays as well. In the meantime, one can use NumPy to do such operations to NumPy arrays in LPython.

certik (approved):
Looks great. Why not keeping one of the commits with `git commit --amend --author="..."` with the original author? Let's say the second.

PR #444, opened 2022-05-04
refactor: AST->ASR

namannimmo10:


namannimmo10:
This block of code is repeated many times:

https://github.com/lcompilers/lpython/blob/c772d72bfbc7190653f6609662313b53346a739d/src/lpython/semantics/python_ast_to_asr.cpp#L892-L899

A template function can be used here, too.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -960,6 +952,23 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+        }
+        else if (ASRUtils::is_real(*type)) {
+            return ASR::down_cast<ASR::RealConstant_t>(ASRUtils::expr_value(expr))->m_r;
+        }
+        else if (ASRUtils::is_logical(*type)) {
+            return ASR::down_cast<ASR::LogicalConstant_t>(ASRUtils::expr_value(expr))->m_value;
+        }
+        else {
+            throw SemanticError("Cannot get constant value of " + ASRUtils::type_to_str_python(type), loc);
+        }
```
certik, line 969: This is slower however, as previously it simply down casted, which is a free operation in Release mode, while now you have to do a linear search for the right type using the if statements. So we don't want to do that.

However, there might be a way to achieve the same speed by somehow using templates and casting right away without the linear search.

Smit-create: May be we can use enumeration and then switch operator to avoid searching.

namannimmo10: Thanks for pointing!  @certik @Smit-create 

> May be we can use enumeration and then switch operator to avoid searching.

Let me try this.

certik: Even switch is slower. Compilers often do a jump table instead of a linear search for switches, but it is still slower than the essentially free downcast. 

The way to do this is with templates --- you can specialize the templated function for `double` and in the specialization cast to `RealConstant_t` directly. Then it should be as fast. It does seems like quite a lot of complication to save a few characters, so you might as well just create a function like `get_integer_value` for all types at that point. Or just leave the original code as is. :)

certik: No worries! I am very happy that you are thinking about how to simplify the code. We definitely want to extract as much stuff as we can into libasr, and share with LFortran (and other future frontends down the road), which will greatly simplify the number of unique code we need to maintain as part of LPython. We just have to be careful about performance.

PR #231, opened 2022-03-14
Move numpy functions into a standalone module

certik:
TODO:
* [ ] There are some bugs with ExternalSymbol that we need to fix

PR #446, opened 2022-05-04
Use handle_return_type() for FunctionCall

certik:
This ensures that FunctionCall's expr type only uses variables from the
local scope, composed from the arguments of the function call.

Copy handle_return_type() from LFortran, this function should be moved
to libasr.

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -363,12 +363,205 @@ public:
         return;
     }
 
+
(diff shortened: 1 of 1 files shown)


certik:
I am going to merge this. It fixes an important bug, now the type of a FunctionCall is correct when it returns an array of size `n` where `n` is an argument to the function.

There are still more bugs to fix towards #231.

certik:
I am going to merge this. It fixes an important bug, now the type of a FunctionCall is correct when it returns an array of size `n` where `n` is an argument to the function.

There are still more bugs to fix towards #231.

Issue #405, opened 2022-04-20
ASR refactoring


certik:
Based on:

https://gitlab.com/lfortran/lfortran/-/merge_requests/1700

PR #447, opened 2022-05-05
Add `StringConcat` node to the ASR

namannimmo10:
#405


certik (approved):
Very good, thanks for doing that. I think this looks good.

#442 will have to be rebased, but that is easy.

PR #451, opened 2022-05-06
Fix args and return type in CPP codegen

Smit-create:
Also covers: https://gitlab.com/lfortran/lfortran/-/merge_requests/1728

### DIFF
--- a/tests/tests.toml
+++ b/tests/tests.toml
@@ -125,6 +125,10 @@ filename = "expr14.py"
 llvm = true
 
 [[test]]
+filename = "expr15.py"
+cpp = true
+
+[[test]]
 filename = "expr_01.py"
 ast = true
 asr = true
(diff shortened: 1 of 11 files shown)


tests/tests.toml
```diff
@@ -124,6 +124,12 @@ asr = true
 filename = "expr14.py"
 llvm = true
 
+
+[[test]]
+filename = "expr15.py"
+cpp = true
+
+
```
namannimmo10, line 132:
```suggestion
```

certik (approved):
Perfect, thanks! I noticed the complex was always using `float`. 

You can also port this change: https://gitlab.com/lfortran/lfortran/-/merge_requests/1728, I haven't done it yet.

certik (approved):
Thanks!

PR #453, opened 2022-05-07
Add `StringItem` node to the ASR

namannimmo10:


namannimmo10:
See the generated ASR:
```python
def main():
  s: str
  s = "abcd"
  print(s[2])

main()
```
```console
% asr t.py
(TranslationUnit 
   (SymbolTable 
      1 
      {
         _lpython_main_program: 
            (Subroutine 
               (SymbolTable 
                  4 
                  {
                  }) 
               _lpython_main_program [] [
               (SubroutineCall 1 main () [] ())] 
               Source 
               Public 
         main: 
            (Subroutine 
               (SymbolTable 
                  2 
                  {
                     s: 
                        (Variable 
                           2 
                           s 
                           Local () () 
                           Default 
                           (Character 1 -2 () []) 
                           Source 
                           Public 
                           Required .false.)
                  }) 
               main [] [
               (= 
                  (Var 2 s) 
                  (StringConstant "abcd" 
                     (Character 1 4 () [])) ()) 
               (Print () [
                  (StringItem 
                     (Var 2 s) 
                     (BinOp (IntegerConstant 2 
                        (Integer 4 [])) 
                        Add (IntegerConstant 1 
                        (Integer 4 [])) 
                        (Integer 4 []) () ()) 
                     (Character 1 -2 () []) ())])] 
               Source 
               Public 
         main_program: 
            (Program 
               (SymbolTable 
                  3 
                  {
                  }) 
               main_program [] [
      }) 
   [])

```
Is this fine, or do we need to stick to:

https://github.com/lcompilers/lpython/blob/71a2cb4997886b43923b194174f1b5c748a5ebed/src/lpython/semantics/python_ast_to_asr.cpp#L1523

certik:
Something like that. We have to update tests.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1105,6 +1105,21 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            return;
+        }
+        this->visit_expr_wrapper(x.m_idx, true);
+        llvm::Value *idx = tmp;
+        idx = builder->CreateSub(idx, llvm::ConstantInt::get(context, llvm::APInt(32, 1)));
+        std::vector<llvm::Value*> idx_vec = {idx};
+        this->visit_expr_wrapper(x.m_arg, true);
+        llvm::Value *p = builder->CreateGEP(tmp, idx_vec);
+        tmp = builder->CreateAlloca(character_type, nullptr);
+        builder->CreateStore(p, tmp);
```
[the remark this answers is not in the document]
namannimmo10: This didn't work as expected, but I improved it further. Thanks. :)

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1105,6 +1105,20 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+        if (x.m_value) {
+            this->visit_expr_wrapper(x.m_value, true);
+            return;
+        }
+        this->visit_expr_wrapper(x.m_idx, true);
+        llvm::Value *idx = tmp;
+        this->visit_expr_wrapper(x.m_arg, true);
+        llvm::Value *str = tmp;
+        llvm::AllocaInst *parg = builder->CreateAlloca(character_type, nullptr);
+        builder->CreateStore(tmp, parg);
```
certik, line 1118: What do these two lines do? They seem to move `tmp` to `parg`, but `parg` is never used?

namannimmo10: I used `AllocaInst` to create a pointer to allocate memory for the `character_type` on the stack. And we use `CreateStore` to take that to `tmp`. The same thing is done in b07a4b31e998ca4f8a6081c221cd985367df9e1d.

namannimmo10:
Let me know if I should go ahead and add `StringSection` node to try to clean up the `ArrayRef` visitor in the backend.

src/libasr/codegen/asr_to_llvm.cpp
certik: It's the other way round: it stores the content of `tmp` into `parg`. But you don't use `parg`, so it's not needed. in https://github.com/lcompilers/lpython/commit/b07a4b31e998ca4f8a6081c221cd985367df9e1d the `parg` is passed as an argument to `lfortran_str_len`.

certik: I think this `StringItem` is never used. I think the bug is that you have to replace `str` with `parg`.

I think you have to hook up `StringItem` in AST->ASR.

Issue #455, opened 2022-05-07
Unsupported type annotation: list


Aasthaengg:
I have implemented functions for `math.py` which work when running with `PYTHONPATH=src/runtime/ltypes python3 src/runtime/math.py` but give `Unsupported type annotation: list` when running with `./src/bin/lpython src/runtime/math.py -o output && ./output`. 

Here's the sample `fsum()` function implementation. 

```
def fsum(data: list) -> f64:
    """
    Sum of the elements of `data`.
    """
    sum: f64
    sum = 0.0

    i: i32
    for i in range(len(data)):
        sum += data[i]
    return sum
```

certik:
Thanks for reporting this. Let's fix that.

namannimmo10:
Yes, we need to add `list` as a supported type annotation, so that we can create an empty list like this:
```console
(lp) namannimmo@srf lpython % cat t.py
def main():
  l: list
  l = []
  print(l)

main()

(lp) namannimmo@srf lpython % lpython t.py
semantic error: Unsupported type annotation: list
 --> t.py:2:3
  |
2 |   l: list
  |   ^^^^^^^ 

(lp) namannimmo@srf lpython % python t.py
[]
```

namannimmo10:
In your implementation of `fsum`, `list[f64]` can be used.

Smit-create:
> Yes, we need to add `list` as a supported type annotation, so that we can create an empty list like this:

I don't think that's appropriate. In LPython we have everything with strict type notation. And just adding `list` annotation for creating an empty list does not seem to be useful since that won't have any lifeline in code as it won't perform any operations (neither we support `list` to `list[anytype]` conversion).

czgdp1807:
@Aasthaengg Have you tried using `list[f64]` instead of just `list`? What's the output of that?

certik:
Yes, `list` must specify the type of the elements, such as `list[i32]`, or `list[list[f64]]`. Just `list` could be a shortcut for `list[any]`, for a runtime dynamic type (each element a different type), but I would tackle this much later, if at all, for now let's focus on static types.

Aasthaengg:
@czgdp1807 I still get `semantic error: Unsupported type annotation: list` when I make the argument as `list[f64]`

Here's the modified snippet - 
```
def fsum(data: list[f64]) -> f64:
    """
    Sum of the elements of `data`.
    """
    sum: f64
    sum = 0.0

    i: i32
    for i in range(len(data)):
[...]

PR #456, opened 2022-05-07
Add dist() function in math.py

Aasthaengg:
This PR consists of `fsum()` `prod()` and `dist()` functionality for `math` module. 

These functions are passing the integration test with `PYTHONPATH=src/runtime/ltypes python integration_tests/test_math.py` but failing with `./src/bin/lpython src/runtime/math.py -o output && ./output` as pointed in #455.


certik:
Thanks!

Smit-create:
@czgdp1807 This is ready to go in.

certik (approved):
Looks great!

Issue #457, opened 2022-05-07
Internal Compiler Error: Unhandled exception with tuple return type


Aasthaengg:
Here's my sample program - 
```py
def modf(x: f64) -> tuple[f64, i32]:
    """
    Return fractional and integral parts of `x` as a pair.
    """
    return (x - int(x), int(x))

res: tuple[f64, i32]
res = modf(3.14)
```

It raises the following error - 
```
Internal Compiler Error: Unhandled exception
BFD: DWARF error: section .debug_info is larger than its filesize! (0x5b4757 vs 0x429c68)
BFD: DWARF error: section .debug_info is larger than its filesize! (0x5b4757 vs 0x429c68)
Traceback (most recent call last):
  Binary file "/home/kn1ght/lpython/lpython_personal/src/bin/lpython", in _start()
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", in __libc_start_main()
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", in __libc_init_first()
  File "/home/kn1ght/lpython/lpython_personal/src/bin/lpython.cpp", line 833, in ??
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir, compiler_options, time_report);
  File "/home/kn1ght/lpython/lpython_personal/src/bin/lpython.cpp", line 332, in ??
    res = fe.get_llvm3(*asr, diagnostics);
  File "/home/kn1ght/lpython/lpython_personal/src/lpython/python_evaluator.cpp", line 57, in LFortran::PythonCompiler::get_llvm3(LFortran::ASR::TranslationUnit_t&, LFortran::diag::Diagnostics&)
    run_fn);
  File "/home/kn1ght/lpython/lpython_personal/src/libasr/codegen/asr_to_llvm.cpp", line 4446, in LFortran::asr_to_llvm(LFortran::ASR::TranslationUnit_t&, LFortran::diag::Diagnostics&, llvm::LLVMContext&, Allocator&, LFortran::Platform, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    v.visit_asr((ASR::asr_t&)asr);
[...]

PR #458, opened 2022-05-07
Add modf() in math module

Aasthaengg:
This PR adds functionality to use `modf()` function in `math` module. It compiles fine with `python` but gives a compilation error with `lpython` as pointed out in #457.

### DIFF
--- a/integration_tests/test_math.py
+++ b/integration_tests/test_math.py
@@ -1,7 +1,7 @@
 from math import (factorial, isqrt, perm, comb, degrees, radians, exp, pow,
-                  ldexp, fabs, gcd, lcm, floor, ceil, remainder, expm1, fmod, log1p, trunc, fsum, prod)
-from ltypes import i8, i16, i32, i64, f32, f64
-from numpy import empty
+                  ldexp, fabs, gcd, lcm, floor, ceil, remainder, expm1, fmod, log1p, trunc,
+                  modf, fsum, prod)
+from ltypes import i32, i64, f32, f64
 
 eps: f64
 eps = 1e-12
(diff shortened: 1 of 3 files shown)


src/runtime/math.py
```diff
@@ -1,4 +1,4 @@
-from ltypes import i8, i16, i32, f32, f64, ccall
+from ltypes import i8, i16, i32, i64, f32, f64, ccall, overload
```
czgdp1807, line 1: `overload` imported but not used.

integration_tests/test_math.py
```diff
@@ -149,6 +150,14 @@ def test_trunc():
     i = trunc(-4.5)
     assert i == -4
 
+def test_modf():
+    i: f64
+    i = 3.14
+
+    res: tuple[f64, i32]
+    res = modf(i)
+    assert res[0] == 0.14000000000000012
```
czgdp1807, line 159: I think we should tests like, `assert res[0] - 0.14 <= 1e-6`

czgdp1807: Seems like we don't have ability to index tuples. :-(

```bash
```

Smit-create: EDIT: `assert abs(res[0] - 0.14) <= 1e-6`

namannimmo10:
@Aasthaengg -- we can complete this now. Would you like to?

namannimmo10:
Right, I wanted to do that. Did a squash and merge -- added @Aasthaengg as a co-author.

certik:
That works, perfect! Thanks for finishing this.

PR #466, opened 2022-05-09
Set Logical default `kind` to be 4 instead of 1

namannimmo10:


certik:
Can you explain a little bit the motivation for this change?

namannimmo10:
We use `kind=1` here:

 https://github.com/lcompilers/lpython/blob/e593ecd7b35ba3f6030dabff71052c3e4dac3422/src/lpython/semantics/python_ast_to_asr.cpp#L738-L740

And here too:

https://github.com/lcompilers/lpython/blob/e593ecd7b35ba3f6030dabff71052c3e4dac3422/src/lpython/semantics/python_ast_to_asr.cpp#L1244-L1249

So I advocate for the use of `kind=1` everywhere in the codebase.

certik:
Given that "Booleans in Python are implemented as a subclass of integers": https://docs.python.org/3/c-api/bool.html, and our default integer (for constants like `5`) is currently `integer(kind=4)`, that is, `i32`, then I would also keep the default Logical kind as 4. That is consistent with Python. It is also consistent with Fortran, where the standard seems to require the logical to have the same default kinda as default integer (currently we use kind=4 for default integers). From https://gcc.gnu.org/onlinedocs/gfortran/Internal-representation-of-LOGICAL-variables.html, it says "The Fortran standard does not specify how variables of LOGICAL type are represented, beyond requiring that LOGICAL variables of default kind have the same storage size as default INTEGER and REAL variables."

All in all, it seems the most consistent to use default kind=4 for logicals in ASR, at least for now.

certik:
We should document this design in `ASR.asdl`.

namannimmo10:
I have applied the suggestion.  We can document this design in a separate PR.

certik:
I documented it in #476.

PR #471, opened 2022-05-10
Add ListItem and ListSection in ASR

Smit-create:
https://gitlab.com/lfortran/lfortran/-/merge_requests/1700/diffs#ab61192f5842e28452c28e60bed1dae0d9c5bf49_213_271


src/libasr/ASR.asdl
```diff
@@ -259,7 +259,8 @@ expr
     | ComplexRe(expr arg, ttype type, expr? value)
     | ComplexIm(expr arg, ttype type, expr? value)
     | DictItem(symbol a, expr key, expr? default, ttype type)
-
+    | ListItem(symbol a, expr pos, ttype type, expr? value)
+    | ListSection(symbol a, array_index section, ttype type, expr? value)
```
certik, line 263: I wonder if this should be an `expr` instead of `symbol`:
```suggestion
    | ListSection(expr a, array_index section, ttype type, expr? value)
```
So that we can represent things like `(a+b)[2:3]` or indexing/slicing of the result of a function such as `f()[2:3]`.

I think we can do this improvement in subsequent PRs, not here.

certik (approved):
I think this looks excellent. Thanks for adding tests for error messages, looks beautiful.

src/libasr/ASR.asdl
certik: I opened up an issue for this at #475.

Smit-create: Sure, will do that

Issue #475, opened 2022-05-10
Make all *Item/*Section ASR nodes accept `expr` instead of `symbol`


certik:
More details:

https://github.com/lcompilers/lpython/pull/471#discussion_r869587954

PR #477, opened 2022-05-11
Use `expr` in ListSection instead of `symbol`

Smit-create:
#475


certik (approved):
Looks good. We should also add a test for something like `(a+b)[2:3]`.

Smit-create:
> We should also add a test for something like `(a+b)[2:3]`.

We can add this improvement test later because it will require restructuring `visit_Subscript` as there is the simultaneous usage of `symbol` and `expr`.

certik (approved):

PR #480, opened 2022-05-11
Add `StringSection` node to the ASR

namannimmo10:


src/libasr/ASR.asdl
```diff
@@ -247,6 +247,7 @@ expr
     | StringRepeat(expr left, expr right, ttype type, expr? value)
     | StringLen(expr arg, ttype type, expr? value)
     | StringItem(expr arg, expr idx, ttype type, expr? value)
+    | StringSection(expr arg, array_index section, ttype type, expr? value)
```
certik, line 250: I would probably do it like this:
```suggestion
    | StringSection(expr arg, expr start, expr end, expr step, ttype type, expr? value)
```
And `step` is always present, and equal to 1 if it is not specified in AST explicitly.

certik:
This looks much cleaner! Is this ready for review?

namannimmo10:
Yes, this is ready!

certik (approved):
I think that this is good, to merge. Thanks for doing this!

PR #484, opened 2022-05-11
LLVM: initial list[i32] implementation

certik:


Smit-create:
I was thinking that this design seems to be okay just temporarily, but we can't build much over these because we can create an infinite number of different data types using the combination of basic data types. We might need to think of some alternative?

certik:
Open up an issue to discuss this. I think probably we need to operate on any "value type" that can simply be copied. We just need to know how many bytes. For some types like `str`, one also needs to deallocate/free them.

certik:
Probably the easiest might end up being doing this using LLVM directly --- we always know the actual types at compile time, so we can generate the proper LLVM code to handle it. So we can create an LLVM equivalent of `_lcompilers_list_append_i32`, by creating a function in `asr_to_llvm`, say `create_list_append` which accepts the actual type as an argument, and internally it generates the LLVM code that appends to the list, of this actual argument. I think this will not be complicated, definitely doable for a list. For a dictionary that might be more difficult, we should look how Python handles dictionaries of any type.

namannimmo10:
Yes, we should definitely see how other compilers handle this part.

PR #487, opened 2022-05-12
LLVM-list: Realloc dynamically

Smit-create:


src/runtime/impure/lfortran_intrinsics.c
```diff
@@ -688,20 +688,19 @@ LFORTRAN_API int8_t* _lcompilers_list_init_i32() {
     l = (struct _lcompilers_list_i32*)malloc(
             sizeof(struct _lcompilers_list_i32));
     l->n = 0;
-    // TODO:
-    l->capacity = 1024;
-    l->p = (int32_t*)malloc(1024*sizeof(int32_t));
+    l->capacity = 1;
+    l->p = (int32_t*)malloc(sizeof(int32_t));
```
certik, line 692: I would do it like this:
```suggestion
    l->capacity = 4;
    l->p = (int32_t*)malloc(l->capacity*sizeof(int32_t));
```
The initial capacity is 4, that way, you can do 4 appends for free, then it will reallocate. Typically if you are going to use a list, you most likely will put it more than just 1 element. The only thing you are losing is some memory, so you don't want to have the initial capacity too big, and we can experiment with this by running some benchmarks, but 4 seems like a nice number. I experimented with the initial capacity of the `Vec` class in the parser, and there using 4 seemed to be the fastest. Using 1 was slower, and using 16 or more was also slower (I think because the linear allocator that we use then must skip a lot of unused memory and things get slower, as you start missing cache lines, and so on).

Smit-create: Yeah, makes sense.

certik (approved):

PR #491, opened 2022-05-12
Try parsing comma separated expressions

certik:


certik:
This makes `expr` equal to `expr_list`, which will cause the whole grammar to change and we would need to rework everything. It is not clear if there is another way or not. For now I recommend not allowing parsing tuples without parentheses.

certik:
I think we can close this now, we have found a solution, already implemented in main.

PR #492, opened 2022-05-12
Remove target

certik:
Just use expr. This shows how to parse:
```
a = 3, x
```


Thirumalai-Shaktivel:
Thanks for this PR!
I problem is that `expr_context` for `Name`, `Attribute` or Other `expr` Nodes will be `Store` in some cases and `Load` in some cases. That is reason I separated them as `target` and `expr` in the parser.
No issues, I will figure out some way to parse the above changes and make all the tests passes

certik:
It seems you can't determine if it is a Store or Load until you encounter Assignment, then you know. So it is based on semantics, not local syntax. As such, I would simply construct the expression and set the `expr_context` to Load. Then in Assignment, go to the expression in LHS and change it to Store. If this simplifies the grammar (which it does greatly it seems), then I think it's worth it.

Thirumalai-Shaktivel:
> It seems you can't determine if it is a Store or Load until you encounter Assignment, then you know. So it is based on semantics, not local syntax. As such, I would simply construct the expression and set the `expr_context` to Load. Then in Assignment, go to the expression in LHS and change it to Store. If this simplifies the grammar (which it does greatly it seems), then I think it's worth it.

Perfect, I had the same Idea and implemented it! Can you please review this?

certik:
I think this looks good to merge, thanks! In subsequent PRs, can you add tests and try to parse the following tuples:
```
a, b, c = x
a, = x
a, b, = x
return ()
return (a,)
return a,
return a, b,
```

Thirumalai-Shaktivel:
Thanks for the review! Yup, I will try to parse the above example in subsequent PR's

certik:
Also on the right hand side:
```
a, b, = b, a
a, b, = b, a,
```
Etc.

PR #493, opened 2022-05-13
Handle list as arguments

Smit-create:
#455


certik (approved):
Thanks! In subsequent PRs, we also need to get this working in ASR->LLVM and add it as an integration_test.

PR #495, opened 2022-05-14
Fix tuple type checking

### LINKED ISSUE #433: Add strict type checking in tuple

Smit-create:
Overrides #452 

Closes #433


namannimmo10 (approved):

certik (approved):
Excellent, thanks!

PR #499, opened 2022-05-16
Add Set pop in ASR

namannimmo10:


src/lpython/semantics/python_attribute_eval.h
```diff
@@ -177,6 +178,17 @@ struct AttributeHandler {
...
+    static ASR::asr_t* eval_set_pop(ASR::symbol_t *s, Allocator &al, const Location &loc,
+            Vec<ASR::expr_t*> &args, diag::Diagnostics &/*diag*/) {
+        if (args.size() != 0) {
+            throw SemanticError("pop() takes no arguments (" + std::to_string(args.size()) + " given)", loc);
+        }
+        ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(s);
+        ASR::ttype_t *type = v->m_type;
+        ASR::ttype_t *set_type = ASR::down_cast<ASR::Set_t>(type)->m_type;
+        return make_SetPop_t(al, loc, s, set_type, nullptr);
+    }
```
czgdp1807, line 190:
```suggestion
        if (args.size() != 0) {
            throw SemanticError("pop() takes no arguments (" + std::to_string(args.size()) + " given)", loc);
        }

        ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(s);

        ASR::ttype_t *type = v->m_type;
        ASR::ttype_t *set_type = ASR::down_cast<ASR::Set_t>(type)->m_type;

        return make_SetPop_t(al, loc, s, set_type, nullptr);
    }
```

tests/errors/test_set3.py
```diff
@@ -0,0 +1,5 @@
+def f():
+    s: set[i32]
+    s = {1, 2, 3, 4}
+    i: i32
+    i = s.pop(3)
```
czgdp1807, line 5: Can we test a use case like the following?

```python
>>> def return_set():
            return {1, 2, 3, 4}
>>> def f():
            i = return_set().pop()
           return i 
>>> f()
1
```

src/libasr/ASR.asdl
```diff
@@ -265,6 +265,7 @@ expr
     | ListSection(expr a, array_index section, ttype type, expr? value)
     | ListPop(symbol a, expr? index, ttype type, expr? value)
     | DictPop(symbol a, expr key, ttype type, expr? value)
+    | SetPop(symbol a, ttype type, expr? value)
```
certik, line 268: This needs to be:
```suggestion
    | SetPop(expr a, ttype type, expr? value)
```

certik: Yes, they all have to be changed to `expr`.

namannimmo10: > I think the nodes in 267 and 266 also need a similar change. `ListPop(expr a, ...)`, `DictPop(expr a, ...)`

Please see: https://github.com/lcompilers/lpython/pull/500#discussion_r873772682.

tests/errors/test_set3.py
namannimmo10: See #501.  We have to make it work and register this as a test.

PR #500, opened 2022-05-16
ASR: Add `TupleItem` node

namannimmo10:


src/libasr/ASR.asdl
```diff
@@ -262,6 +262,7 @@ expr
     | ComplexIm(expr arg, ttype type, expr? value)
     | DictItem(symbol a, expr key, expr? default, ttype type)
     | ListItem(symbol a, expr pos, ttype type, expr? value)
+    | TupleItem(symbol a, expr pos, ttype type, expr? value)
```
certik, line 265:
```suggestion
    | TupleItem(expr a, expr pos, ttype type, expr? value)
```

namannimmo10: Yes, let me change `symbol` to `expr` for all these nodes in a separate PR.

certik (approved):
Ok, then it is fine to merge this PR as is.

PR #503, opened 2022-05-17
Parse ellipsis (...) in the old parser

certik:


grammar/Python.asdl
```diff
@@ -83,6 +83,8 @@ module LPython
          | ConstantBool(bool value, string? kind)
          | ConstantFloat(float value, string? kind)
          | ConstantComplex(float re, float im, string? kind)
+         | ConstantEllipsis(string? kind)
```
namannimmo10, line 86: How are we going to support this in the ASR?

certik: I think `...` is similar to `pass`, so that's just an empty subroutine, so you leave the `body` argument empty in ASR.

namannimmo10: What I've noticed is that people use `Ellipsis` to indicate something they want to fill in later (like a todo empty block) and `pass` stmt to mean a block is intended to have _no_ code. 

`...` is used in NumPy code, too, as a placeholder in slice notation. 
See more uses here: https://python.land/python-ellipsis

certik: It has different uses. For the "empty block" use, we can just represent that with an empty block in ASR.

PR #504, opened 2022-05-18
Handle printing of logical types in the LLVM backend

### LINKED ISSUE #322: Handle printing of logical types

namannimmo10:
fixes #322

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -4317,8 +4317,7 @@ public:
                 fmt.push_back("%lld");
                 llvm::Value* d = builder->CreatePtrToInt(tmp, getIntType(8, false));
                 args.push_back(d);
-            } else if (ASRUtils::is_integer(*t) ||
-                ASR::is_a<ASR::Logical_t>(*ASRUtils::type_get_past_pointer(t))) {
+            } else if (ASRUtils::is_integer(*t)) {
                 switch( a_kind ) {
                     case 1 : {
                         fmt.push_back("%d");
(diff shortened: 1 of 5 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3833,6 +3832,17 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            } else if (ASRUtils::is_logical(*t)) {
+                int ivalue = down_cast<ASR::LogicalConstant_t>(
+                                       ASRUtils::expr_value(v))->m_value;
+                std::string logical_str;
+                if (ivalue == 1) {
+                    logical_str = "True";
+                } else {
+                    logical_str = "False";
+                }
+                llvm::Value *fmt_ptr = builder->CreateGlobalStringPtr(logical_str);
```
czgdp1807, line 3844: I think this will work only if `v` is a constant? What if `v` is a variable? I think we should generate the `if-else` check in LLVM (should be easy to do) to select a LLVM string to be stored in `fmt_ptr`.

czgdp1807:
I reviewed. Thanks for requesting review.

src/libasr/codegen/asr_to_llvm.cpp
certik: Yes, this will only work at compile time. One has to pass in the strings `True`/`False` for Python and `.true.`/`.false.` for Fortran at compile time, and then at runtime select true or false, using either a C function or some logic in LLVM.

czgdp1807: @namannimmo10 Are you still working on this?

namannimmo10: Thanks for the ping and for reviewing the PR.  Please feel free to move with this; I'm focusing more on refactoring the ASR as of now. If you don't have time, I can come back to it this week.

Smit-create:
I think this is already fixed in `main`. 
```
print(True, False)
```

Gives,
```
1 0
```

Handling booleans as integers in the LLVM backend is good in my opinion because ASR is language independent and the LLVM backend is used for different frontend languages, we might have some confits in printing booleans explicitly like "True" in some front-end languages and "true" in some others.

certik:
The issue is that LPython should print it like CPython does, and LFortran like other Fortran compilers do. But this can be easily configurable, say as an option to `asr_to_llvm`, so not a big deal in practice.

namannimmo10:
I am sorry for the long delay in addressing the comments. This can go in now.

certik (approved):
I think this works. Thanks!

PR #515, opened 2022-05-18
Allow to parse negative integers in serialization

certik:


certik:
I am going to close this --- the int is unsigned, so rather it must be saved properly.

certik:
Proper fix is in #516.

PR #522, opened 2022-05-20
Use expr instead of symbol in new added ASR nodes

Smit-create:


Smit-create:
@certik Any views on how to move forward with this?

certik:
Just do one ASR node change at a time, and fix up the AST->ASR to compile. I don't think there is any fundamental issue. If there is, you can describe the issue here and let's have a look.

namannimmo10:
Thanks for working on this, Smit.

certik (commented):
I think this looks good, thanks Smit and Naman! In future PRs, let's also add tests for this, such as `(a+b)[3]` for lists.

Smit-create:
Thanks! I would still recommend avoiding `Merge main` commits that unnecessarily populate git history, instead use rebase and force push.

certik:
I agree. Only merge main into a PR if it is a difficult PR that would be hard to rebase, usually for a big PR. Or if we want to preserve test commits, if it was very hard to get things passing. For simple PRs like this one we should always rebase, as it is cleaner.

PR #110, opened 2022-02-07
Make `str()` built-in function work for basic types

namannimmo10:

PR #524, opened 2022-05-21
Add cast handling in function calls

### LINKED ISSUE #518: LLVM: parameter type does not match function signature
```py
from ltypes import i64

def fib(n: i64) -> i64:
    if n < 2:
        return n
    else:
        return fib(n - 1) + fib(n - 2)

def main():
    ans: i64
    ans = fib(35)
    print(ans)

main()
```
Compiling gives:
```console
  %call_arg_value = alloca i32, align 4
 i64*  %0 = call i64 @fib(i32* %call_arg_value)
```

Smit-create:
Fixes #518

### DIFF
--- a/integration_tests/test_issue_518.py
+++ b/integration_tests/test_issue_518.py
@@ -6,9 +6,9 @@ def fib(n: i64) -> i64:
     else:
         return fib(n - 1) + fib(n - 2)
 
-def main():
+def main0():
     ans: i64
-    ans = fib(35)
-    print(ans)
+    ans = fib(15)
+    assert ans == 610
 
-main()
+main0()
--- a/integration_tests/test_issue_518.py
+++ b/integration_tests/test_issue_518.py
@@ -1,0 +1,14 @@
+from ltypes import i64
+
+def fib(n: i64) -> i64:
+    if n < 2:
+        return n
+    else:
+        return fib(n - 1) + fib(n - 2)
+
+def main():
+    ans: i64
+    ans = fib(35)
+    print(ans)
+
+main()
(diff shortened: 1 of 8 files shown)


tests/constants1.py
```diff
@@ -65,7 +65,9 @@ def test_str():
     s = str()
     s = str(5)
     s = str(-4)
-    s = str(5.6)
+    # TODO: This test was had hidden failure and was noticed during type
+    # checking.
```
Smit-create, line 69: This test was not tested properly. There were no runtime functions matching the function signature. Introduced here: https://github.com/lcompilers/lpython/pull/110/commits/a1da515dac03aedade79d90a0c735a4e24181927

integration_tests/test_issue_518.py
```diff
@@ -0,0 +1,14 @@
...
+
+def fib(n: i64) -> i64:
+    if n < 2:
+        return n
+    else:
+        return fib(n - 1) + fib(n - 2)
+
+def main():
+    ans: i64
+    ans = fib(35)
```
certik, line 11: This makes it run "immediately":
```suggestion
    ans = fib(15)
```
The previous number was fast with LPython, but slow with CPython.

certik:
This looks good to me. There is a failure at the CI for macOS, but I tested on my Mac and everything seems to work, so I can't reproduce it. We have to see what is going on.

Smit-create:
It also passes on my mac

certik:
The failure is real. We have to figure out why it fails.

Are the `str` changes needed for this PR? Maybe they can be split into a separate MR to reduce the number of changes that could break.

Smit-create:
> Maybe they can be split into a separate PR to reduce the number of changes that could break.

#526

Smit-create:
I'm still not sure what's the issue with mac failure. It does pass for me locally. Might be some updates on GA macOS?

certik:
The CI failure is due to https://github.com/lcompilers/lpython/issues/529.

certik (approved):
I think it works now.

PR #525, opened 2022-05-22
Fix pow in C backend

Smit-create:


src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -77,6 +77,7 @@ class BaseCCPPVisitor : public ASR::BaseVisitor<Derived>
     bool gen_stdstring;
     // Use std::complex<float/double> or float/double complex
     bool gen_stdcomplex;
+    bool is_c = false;
 
     BaseCCPPVisitor(diag::Diagnostics &diag,
             bool gen_stdstring, bool gen_stdcomplex) : diag{diag},
```
certik, line 83: I think that this is cleaner:
```suggestion
    bool is_c;

    BaseCCPPVisitor(diag::Diagnostics &diag,
            bool gen_stdstring, bool gen_stdcomplex, bool is_c) : diag{diag},
```
And add `is_c{is_c}` at the end of the initializer list.

certik:
Otherwise this looks good.

PR #526, opened 2022-05-22
Add some missing str builtin functions

Smit-create:


src/runtime/lpython_builtin.py
```diff
@@ -99,6 +99,21 @@ def abs(c: c64) -> f64:
...
+@overload
+def str() -> str:
+    return ''
+
+@overload
+def str(x: str) -> str:
+    return x
+
+@overload
+def str(x: bool) -> str:
```
namannimmo10, line 111: We do not have tests for these runtime `str` functions. Can you add a few?

certik (approved):
Looks good to me, thanks!

PR #528, opened 2022-05-25
Handing out of bounds indexing in String

Abdullahjavednesar:


src/runtime/impure/lfortran_intrinsics.c
```diff
@@ -637,7 +637,12 @@ LFORTRAN_API void _lfortran_strrepeat(char** s, int32_t n, char** dest)
 
 // idx1 and idx2 both start from 1
 LFORTRAN_API char* _lfortran_str_copy(char* s, int32_t idx1, int32_t idx2) {
+
     int s_len = strlen(s);
+    if(idx1 > s_len || idx1 <= (-1*s_len)){
+        printf("String index out of Bounds\n");
+        exit(1);
```
Smit-create, line 644: We should have similar checks for `idx2`?

Abdullahjavednesar: I initially thought the same, seems currently in python we do not check range for `idx2`:
```python
>>> s='012345'
>>> s[1:10]
'12345'
```

certik: I also didn't know that, but apparently the upper slice index behaves that way: https://stackoverflow.com/questions/9490058/why-does-substring-slicing-with-index-out-of-range-work

I wonder if in AST->ASR we should replace `s[1:10]` with `s[1:min(10,len(s))]`. That way in ASR the start/end is always valid and correct, and in the backend we don't need to do any kind of rearrangements, and simply use the start/end which is assumed correct. And we add bounds checking for both ends, just in case. The upper end should never fail, with the `min(10,len(s))` condition.

certik:
I think it's fine to merge. We'll eventually add tests for runtime errors (like this one).

PR #530, opened 2022-05-26
Rename main to handle backend failures in llvm/c/cpp

### LINKED ISSUE #529: Fix an issue with a function called `main`
This returns a non-zero exit code (but only sometimes, such as on macOS):
```python
from ltypes import i64

def main():
    x: i64
    x = 15

main()
```
The current workaround is to rename the function, such as:
```python
from ltypes import i64

def main0():
    x: i64
    x = 15

main0()
```
And the fix has to happen in the LLVM backend to rename the function automatically under the hood.

Smit-create:
Fixes #529

### DIFF
--- a/integration_tests/test_issue_518.py
+++ b/integration_tests/test_issue_518.py
@@ -11,4 +11,12 @@ def main0():
     ans = fib(15)
     assert ans == 610
 
+def main():
+    # test of issue-529
+    ans: i64
+    ans = fib(10)
+    assert ans == 55
+
+
 main0()
+main()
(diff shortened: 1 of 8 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1772,6 +1772,9 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                 var)));
         }
         std::string sym_name = x.m_name;
+        if (sym_name == "main") {
+            sym_name = "_xx_lpython_defined_main_xx";
+        }
```
certik, line 1777: I think we should not do it here. In ASR I think we should allow any names.

Rather, in the LLVM and C backends we should rename.

Smit-create: Yeah, that might be good.

certik (approved):
I think this looks good now.  Thanks!

PR #531, opened 2022-05-26
Temporary Workaround till --fast fibonacci is fixed.

### LINKED ISSUE #519: --fast miscompiles fibonacci
```
from ltypes import i64

def fib(n: i64) -> i64:
    if n < 2:
        return n
    else:
        return fib(n - 1) + fib(n - 2)

def main():
    ans: i64
    x: i64
    x = 40
    ans = fib(x)
    print(ans)

main()
```
It works in Debug mode but not in Release mode:
```console
$ lpython b.py && ./a.out
102334155
$ lpython --fast b.py && ./a.out
$
```

faze-geek:
#519
Commenting out the lines leading to the bug. 

P.S - This is my first PR after setting up lpython just a couple of hours ago, would request for some guidance and patience .Thanks !


certik:
Congrats on your first PR!

I think we should fix this properly. @czgdp1807 what do you think the best way forward is? I think we should try fixing the inline pass first, only if we can't do it easily, we should disable it.

faze-geek:
Closing as completed in #715

PR #535, opened 2022-05-28
Fix exit codegen in C/CPP

Smit-create:


tests/reference/c-loop2-ce7de51.stdout
```diff
@@ -0,0 +1,40 @@
...
+}
+
+void exit(int32_t error_code)
+{
+    exit(error_code);
+}
+
+int main(int argc, char* argv[])
+{
+        _lpython_main_program();
```
namannimmo10, line 38: Can we fix the indentation here?

certik: Yes. But I think that's unrelated to this PR.

certik (approved):

PR #536, opened 2022-05-29
Add `UnaryMinus` node to the ASR

namannimmo10:
#405


certik (approved):
I think this is a good start and can be merged.

In subsequent PRs we should then remove UnaryOp completely, and also we can consider whether we should further split into IntegerUnaryOp, RealUnaryOp, ComplexUnaryOp.

PR #540, opened 2022-05-30
ASR: Split `UnaryMinus`

namannimmo10:
https://gitlab.com/lfortran/lfortran/-/merge_requests/1700#note_964643463

PR #549, opened 2022-06-04
Update ASR from LFortran

certik:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1162,39 +1165,69 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-        this->visit_expr_wrapper(x.m_arg, true);
-        llvm::Value *str = tmp;
-        tmp = lfortran_str_copy(str, idx, idx);
-    }
-
-    void visit_StringSection(const ASR::StringSection_t& x) {
-        if (x.m_value) {
-            this->visit_expr_wrapper(x.m_value, true);
-            return;
+        if (is_a<ASR::Character_t>(*x.m_type)
```
certik, line 1168: This part can be removed once LFortran is updated to use the new `StringItem`.

certik:
Also I had to add back some UnaryMinus handling, until LFortran is updated. LPython also needs to be updated to fully remove UnaryMinus.

Otherwise I think this is good.

certik:
Corresponding LFortran sync MR: https://gitlab.com/lfortran/lfortran/-/merge_requests/1764

certik:
@namannimmo10, @Smit-create let me know if you see anything missing or any changed reverted by a mistake. I did my best to merge additions from both LPython and LFortran.

certik:
I am going to go ahead and merge this, so that ASR is now synchronized with LFortran. Then we can submit PRs to both if needed.

src/libasr/codegen/asr_to_llvm.cpp
certik: Fixed in #550.

namannimmo10:
It seems like while making changes to the `libasr` (which LFortran and LPython share,) we should submit PRs to both the projects. Otherwise, we're repeating the steps.

namannimmo10:
> LPython also needs to be updated to fully remove UnaryMinus.

I don't see the occurrence of `UnaryMinus` in the current `main` after it is synchronized. I added that node in #536 but then split that based on different types in #540. 

I guess we just need to update the frontend in LFortran to support `IntegerUnaryMinus`, `RealUnaryMinus`, ... etc.

certik:
I mean UnaryOp, we need to remove UnaryOp.

certik:
Yes, I think when modifying libasr, it's good to synchronize it soon and submit a PR to both projects. It's not a big deal for small changes, but for large refactoring changes we should do it, otherwise it takes me a lot of time to synchronize later.

src/libasr/asr_utils.h
```diff
@@ -289,57 +245,7 @@ static inline std::string boolop_to_str(const ASR::boolopType t) {
...
-        case ASR::exprType::TupleConstant: // Drop through
-        case ASR::exprType::DictConstant: // Drop through
-        case ASR::exprType::SetConstant: // Drop through
-        case ASR::exprType::ListConstant: // Drop through
-        case ASR::exprType::StringConstant:{ // For all Constants
-            return f;
-        }
-        default : throw LFortranException("Not implemented");
-    }
+    return ASR::expr_value0(f);
```
namannimmo10, line 248: Happy about this change. :)

certik: I figured you would be. :) It's been bothering me too, so I finally fixed it.

PR #553, opened 2022-06-04
ASR: Remove `UnaryOp`

namannimmo10:


certik:
I thought about Unary Plus. I don't think we have to have it in ASR. Is there any reason in any backend why we should keep it? 

The only reason I can think of is to preserve it in the C++ translation. But we are rewriting expressions anyway, such as `1/2` will get rewritten depending on the frontend language. So I would only store things that are truly needed for the backends, and not losing any semantic information. It seems `+f` and `f` is exactly equivalent, so I would not store it in ASR at all.

namannimmo10:
@certik -- now, for this to compile, the only parts left to be modified are in `libasr/pass` (`array_op.cpp`, `param_to_const.cpp`). Can you take a look, please?

certik:
Great. Is this ready for review?

certik:
Can you submit an equivalent change to LFortran? To make sure that this will work.

namannimmo10:
Yes, this is ready for review.

I'm not fully confident about the changes I made in `library/pass` so once we make an equivalent change in LFortran, we can possibly identify some bugs because of the bigger test suite.

certik:
I think this PR is mostly fine, but do not merge this. First let's submit to LFortran, and make sure everything passes there as well. Then we'll merge both at once.

namannimmo10:
I am not so familiar with Fortran syntax, so I need some time to make correct changes in the frontend, i.e., `ast_common_visitor.h` to remove `UnaryOp`.

certik:
No worries. I think the operations are quite similar to LPython, it should feel quite familiar. Just more advanced/complete (more corner cases covered and more features present).

certik:
LFortran MR: https://gitlab.com/lfortran/lfortran/-/merge_requests/1778

certik (approved):
This looks good. Can you please resolve conflicts with main?

certik:
Use the new integration_tests runner (https://github.com/lcompilers/lpython/pull/573).

Issue #511, opened 2022-05-18
[New] Parse IN and IS in the expression


Thirumalai-Shaktivel:
Example:
```python
if field in ["vararg", "kwarg"]:
    if value is None:
	pass
```
Error:
```console
$ lpython --show-ast --new-parser examples/expr2.py
syntax error: Token 'in' is unexpected here
 --> examples/expr2.py:1:10
  |
1 | if field in ["vararg", "kwarg"]:
  |          ^^ 


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).
```

PR #557, opened 2022-06-06
[Parser] Rules for expressions containing `is` and `is not`

akshanshbhatt:
Related Issue: #511


src/lpython/parser/parser.yy
```diff
@@ -730,6 +730,8 @@ expr
     | expr "<=" expr { $$ = COMPARE($1, LtE, $3, @$); }
     | expr ">" expr { $$ = COMPARE($1, Gt, $3, @$); }
     | expr ">=" expr { $$ = COMPARE($1, GtE, $3, @$); }
+    | expr KW_IS expr { $$ = COMPARE($1, Is, $3, @$); }
+    /* | expr KW_IS TK_NOT expr { $$ = COMPARE($1, IsNot, $4, @$); } */
```
Thirumalai-Shaktivel, line 734: I had the same problem and was unable to resolve this.
cc @certik 
The problem here was that I didn't know how to assign precedence for "is not" when we uncomment the above code. We get:
 ```console
$ ./build0.sh
+ ci/version.sh
++ git describe --tags --dirty
+ version=v0.2.0-756-g1ae99e23c-dirty
+ version=0.2.0-756-g1ae99e23c-dirty
+ echo 0.2.0-756-g1ae99e23c-dirty
+ python grammar/asdl_py.py
Assuming default values of Python.asdl and python_ast.py
+ python grammar/asdl_cpp.py grammar/Python.asdl src/lpython/python_ast.h
+ python grammar/asdl_cpp.py src/libasr/ASR.asdl src/libasr/asr.h
+ python src/libasr/wasm_instructions_visitor.py
Assuming default values of wasm_instructions.txt and wasm_visitor.h
+ cd src/lpython/parser
+ cd src/lpython/parser
+ bison -Wall -d -r all parser.yy
parser.yy: error: reduce/reduce conflicts: 27 found, 0 expected

```

Comment (Thirumalai-Shaktivel) [reply src/lpython/parser/parser.yy:734]: That works, Thank you!
```diff
diff --git a/src/lpython/parser/parser.yy b/src/lpython/parser/parser.yy
index f3a7d7797..0954e071b 100644
--- a/src/lpython/parser/parser.yy
+++ b/src/lpython/parser/parser.yy
@@ -131,6 +131,7 @@ void yyerror(YYLTYPE *yyloc, LFortran::Parser &p, const std::string &msg)
 %token TK_GT ">"
 %token TK_GE ">="
 %token TK_NOT "not"
+%token TK_IS_NOT "is not"
 %token TK_AND "and"
 %token TK_OR "or"
 %token TK_TRUE "True"
@@ -243,7 +244,7 @@ void yyerror(YYLTYPE *yyloc, LFortran::Parser &p, const std::string &msg)
 %left "or"
 %left "and"
 %precedence "not"
-%left "==" "!=" ">=" ">" "<=" "<" KW_IS //"is not" "is" "not in" "in"
+%left "==" "!=" ">=" ">" "<=" "<" KW_IS "is not" // "is" "not in" "in"
 %left "|"
 %left "^"
 %left "&"
@@ -731,7 +732,7 @@ expr
     | expr ">" expr { $$ = COMPARE($1, Gt, $3, @$); }
     | expr ">=" expr { $$ = COMPARE($1, GtE, $3, @$); }
     | expr KW_IS expr { $$ = COMPARE($1, Is, $3, @$); }
-    /* | expr KW_IS TK_NOT expr { $$ = COMPARE($1, IsNot, $4, @$); } */
+    | expr TK_IS_NOT expr { $$ = COMPARE($1, IsNot, $3, @$); }
 
     | expr "and" expr { $$ = BOOLOP($1, And, $3, @$); }
     | expr "or" expr { $$ = BOOLOP($1, Or, $3, @$); }
diff --git a/src/lpython/parser/tokenizer.re b/src/lpython/parser/tokenizer.re
index f2edf0d2d..032df4462 100644
--- a/src/lpython/parser/tokenizer.re
+++ b/src/lpython/parser/tokenizer.re
@@ -426,6 +426,7 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
             'not'  { RET(TK_NOT) }
             'and'  { RET(TK_AND) }
             'or'   { RET(TK_OR) }
+            'is not'  { RET(TK_IS_NOT) }
```
But, more than one whitespace between `is` and `not` does not work like
```python
if type(x) is  not int:
    pass
```

Thirumalai-Shaktivel: @akshanshbhatt Apply the above diff. This should work

akshanshbhatt: > I think we might need to parse `is not` as a token in the tokenizer.

This is exactly was I was going to suggest. Will implement it ASAP!

src/lpython/parser/tokenizer.re
```diff
@@ -426,6 +426,7 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
             'not'  { RET(TK_NOT) }
             'and'  { RET(TK_AND) }
             'or'   { RET(TK_OR) }
+            'is' whitespace 'not'  { RET(TK_IS_NOT) }
```
akshanshbhatt, line 429: I think instead of using whitespace; we should use something generic.

Here is the issue with using white space -

```
❯ cat examples/test1.py
x = 10
if x is \
    not type(int):
    pass
```
Old Parser
```
❯ lpyast examples/test1.py --indent
(Module [
   (Assign [
      (Name
         x
         Store)]
      (ConstantInt 10 ()) ())
   (If
      (Compare
         (Name
            x
            Load)
         IsNot [
         (Call
            (Name
               type
               Load) [
            (Name
               int
               Load)] [])]) [
      (Pass)] [])] [])
```
New Parser
```
❯ lpynewpar examples/test1.py --indent
(Module [
   (Assign [
      (Name
         x
         Store)]
      (ConstantInt 10 ()) ())
   (If
      (Compare
         (Name
            x
            Load)
         Is [
         (UnaryOp
            Not
            (Call
               (Name
                  type
                  Load) [
               (Name
                  int
                  Load)] []))]) [
      (Pass)] [])] [])
```
Diff
```diff
❯ diff old.txt new.txt
12,19c12,21
<          IsNot [
<          (Call
<             (Name
<                type
<                Load) [
<             (Name
<                int
<                Load)] [])]) [
---
>          Is [
>          (UnaryOp
>             Not
>             (Call
>                (Name
>                   type
>                   Load) [
>                (Name
>                   int
>                   Load)] []))]) [
```

Thirumalai-Shaktivel: Yup, We have to make it. But, for now we can do something like this:
```diff
diff --git a/src/lpython/parser/tokenizer.re b/src/lpython/parser/tokenizer.
re
index 54449d338..bda807106 100644
--- a/src/lpython/parser/tokenizer.re
+++ b/src/lpython/parser/tokenizer.re
@@ -367,7 +367,6 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Locat
ion &loc, diag::Diagnost
                 RET(TK_NEWLINE);
             }
 
-            "\\" newline { continue; }
 
             // Single character symbols
             "(" { token_loc(loc); record_paren(loc, '('); RET(TK_LPAREN) }
@@ -427,6 +426,9 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Locat
ion &loc, diag::Diagnost
             'and'  { RET(TK_AND) }
             'or'   { RET(TK_OR) }
             'is' whitespace 'not'  { RET(TK_IS_NOT) }
+            'is' whitespace? "\\" newline whitespace? 'not'  { RET(TK_IS_NO
T) }
+
+            "\\" newline { continue; }
 
             // True/False
```
@certik, what are your thoughts?

certik: I would start with your latest diff, add tests. That should cover 99% of cases. Then later we can revisit this. For example, once we can parse all of Python, we can see if there is a way to parse this in Bison. It's too early now, as we will be adding more rules, and so it's safer to do it from the tokenizer. We can also later investigate if there is a way to make this 100% robust in the tokenizer.

src/lpython/parser/parser.yy
```diff
@@ -243,7 +244,7 @@ void yyerror(YYLTYPE *yyloc, LFortran::Parser &p, const std::string &msg)
 %left "or"
 %left "and"
 %precedence "not"
-%left "==" "!=" ">=" ">" "<=" "<" //"is not" "is" "not in" "in"
+%left "==" "!=" ">=" ">" "<=" "<" TK_IS_NOT KW_IS //"is not" "is" "not in" "in"
```
Thirumalai-Shaktivel, line 247:
```suggestion
%left "==" "!=" ">=" ">" "<=" "<" TK_IS_NOT KW_IS // "not in" "in"
```

Thirumalai-Shaktivel:
Can you please apply this diff as well:
```diff
diff --git a/src/lpython/parser/tokenizer.re b/src/lpython/parser/tokenizer
.re
index b116b82cb..0ab70eb1c 100644
--- a/src/lpython/parser/tokenizer.re
+++ b/src/lpython/parser/tokenizer.re
@@ -550,6 +550,7 @@ std::string token2text(const int token)
         T(TK_NOT, "not")
         T(TK_AND, "and")
         T(TK_OR, "or")
+        T(TK_IS_NOT, "is not")
 
         T(TK_TRUE, "True")
         T(TK_FALSE, "False")
```
so that it resolves the following error:
```console
$ cat examples/expr2.py                  
if x is not type(float):
    pass
$ lpython --show-tokens examples/expr2.py
(KEYWORD "if") 0:1
(TOKEN "identifier" x) 3:3
TOKEN: 316
  Binary file "/home/thirumalai/Open_Source/lpython/src/bin/lpython", in _start()
  File "./csu/../csu/libc-start.c", line 392, in __libc_start_main_impl()
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main()
  File "/home/thirumalai/Open_Source/lpython/src/bin/lpython.cpp", line 821, in ??
    return emit_tokens(arg_file, true, compiler_options);
  File "/home/thirumalai/Open_Source/lpython/src/bin/lpython.cpp", line 106, in ??
    std::cout << LFortran::pickle_token(toks[i], stypes[i]);
  File "/home/thirumalai/Open_Source/lpython/src/lpython/tokenizer.re", line 638, in LFortran::pickle_token[abi:cxx11](int, LFortran::YYSTYPE const&)
LFortranException: Token conversion not implemented yet.
```

Thirumalai-Shaktivel:
Other than above changes, this PR looks great. Thank you.

Thirumalai-Shaktivel (approved):
LGTM! Thanks for this PR.

certik (approved):
Looks great, thanks!

Issue #546, opened 2022-06-01
Manipulating C pointers


certik:
Here is an example of how pointers should be manipulated, in Fortran, Python, both should generate the same ASR, and the C backend should then generate the output below:

Fortran:

    type(c_ptr) :: queries
    integer(2), pointer :: x(:)         
    c_f_pointer(queries, x)
    c_loc(x(idx))       

Python:

    from ltypes import c_loc, c_p_pointer, CPtr, Pointer
    from typing import Annotate
    queries: CPtr
    x: Annotate[i16[:], Pointer]
    x = c_p_pointer(queries, i16)                      
    c_loc(x[idx])           
    
C output:

    void *queries;
    int16_t *x;
    x = (int16_t*)queries;
    &x[idx];

Also allow passing `void *` pointers as arguments to functions.

certik:
To pass `void *` as function arguments, in Python:
```python
@ccaller
def f(A_void: CPtr, idx: i32) -> None:
    A: Annotate[i16[:], Pointer]   
    A = c_p_pointer(A_void, i16)
    g(c_loc(A[idx])                    
```
translates to C:
```c
void f(void *A_void, int32_t idx) {
    int16_t *A;
    A = (int16_t *) A_void;
    g(&A[idx])
```

certik:
Note: I think `c_loc(x[idx])` will not work in CPython, since `x[idx]` will return an integer as a value, so we can't take pointer to it from inside `c_loc`. We'll have to use some syntax from https://docs.python.org/3/library/ctypes.html#pointers.

certik:
Yes, that's fine, if it can be done.

certik:
Most of these are fixed. The remaining issue is #609.

Shaikh-Ubaid:
>can we do, c_p_pointer(queries, x)

Can this be done @certik , @czgdp1807 ? Probably not. I think `Python` does not allow/support updating a variable passed as argument.

certik:
Right, Python does not support it.

PR #558, opened 2022-06-07
Adding support for C-interop in Python

czgdp1807:
https://github.com/lcompilers/lpython/issues/546


czgdp1807:
cc: @certik

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -745,46 +800,23 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                AST::Name_t *n = AST::down_cast<AST::Name_t>(c->m_func);
+                var_annotation = n->m_id;
+                m_args = c->m_args;
+                n_args = c->n_args;
+            } else {
+                throw SemanticError("Only Name in Call supported for now in annotation",
+                    loc);
+            }
         } else {
             throw SemanticError("Only Name or Subscript supported for now in annotation of annotated assignment.",
```
namannimmo10, line 815:
```suggestion
            throw SemanticError("Only Name, Subscript, and Call supported for now in annotation of annotated assignment.",
```

certik:
I think this looks good in general, but I think we need to unify `CFPointer` and `CPPointer` at the ASR level. What is the difference between them from the ASR's perspective?

certik:
It seems that both convert a C void pointer (`CPtr`) into a pointer type in ASR (such as `Pointer(Integer)`). So I would create one `expr` node, called `CPtrToPointer` or something like that, which accepts an argument of type `CPtr` and returns a `Pointer(...)` to something. Then both LFortran and LPython convert their surface level syntax into `CPtrToPointer`.

czgdp1807:
> I think this looks good in general, but I think we need to unify `CFPointer` and `CPPointer` at the ASR level.

Done.



> What is the difference between them from the ASR's perspective?

I don't think there is any. Both do the same thing.

certik (approved):
I think that looks good to merge.

czgdp1807:
Well I have some more work to do here.

czgdp1807:
@certik This is ready. Are you okay if I merge this after cleaning up the git history here?

czgdp1807:
Git history cleaned up. Feel free to merge now.

Smit-create:
The test in integration bindc gives me segmentation fault.
```
% lpython integration_tests/bindc_01.py && ./a.out 
ld: warning: dylib (/Users/thebigbool/repos/lpython/src/runtime/liblpython_runtime.dylib) was built for newer macOS version (11.5) than being linked (11.0)
zsh: segmentation fault  ./a.out

```

czgdp1807:
Try checking on https://github.com/lcompilers/lpython/pull/559. In `main` I also get a segmentation fault but not with gh-559. 

```llvm
; ModuleID = 'LFortran'
source_filename = "LFortran"

@queries = global void* null
@x = global i16* null
@0 = private unnamed_addr constant [9 x i8] c"%lld %d\0A\00", align 1

define void @_lpython_main_program() {
.entry:
  %0 = load void*, void** @queries, align 8
  %1 = bitcast void* %0 to i16*
  store i16* %1, i16** @x, align 8
  %2 = load void*, void** @queries, align 8
  %3 = ptrtoint void* %2 to i64
  %4 = load i16*, i16** @x, align 8
  %5 = load i16, i16* %4, align 2
  call void (i8*, ...) @_lfortran_printf(i8* getelementptr inbounds ([9 x i8], [9 x i8]* @0, i32 0, i32 0), i64 %3, i16 %5)
  br label %return

  ret void
}

define i32 @main() {
.entry:
  call void @_lpython_main_program()
  ret i32 0
}
```

The problem is with the following part,

```llvm
%4 = load i16*, i16** @x, align 8
%5 = load i16, i16* %4, align 2
```

`@x` is points to a NULL, so `%4` is a NULL. Now the very next line, loads a NULL which obviously will lead to a segmentation fault.

With gh-559 the LLVM code is as follows,

```llvm
; ModuleID = 'LFortran'
source_filename = "LFortran"

@queries = global void* null
@x = global i16* null
@0 = private unnamed_addr constant [11 x i8] c"%lld %lld\0A\00", align 1

define void @_lpython_main_program() {
.entry:
  %0 = load void*, void** @queries, align 8
  %1 = bitcast void* %0 to i16*
  store i16* %1, i16** @x, align 8
  %2 = load void*, void** @queries, align 8
  %3 = ptrtoint void* %2 to i64
  %4 = load i16*, i16** @x, align 8
  %5 = ptrtoint i16* %4 to i64
  call void (i8*, ...) @_lfortran_printf(i8* getelementptr inbounds ([11 x i8], [11 x i8]* @0, i32 0, i32 0), i64 %3, i64 %5)
  br label %return

  ret void
}

define i32 @main() {
.entry:
  call void @_lpython_main_program()
  ret i32 0
}
```

The problematic is now updated,

```llvm
%4 = load i16*, i16** @x, align 8
%5 = ptrtoint i16* %4 to i64
```

The actual address where `@x` points to is loaded in `%4` and then converted to `i64` for printing, which is what we needed. So I think its all fixed in gh-559.

PR #559, opened 2022-06-08
Supporting array pointers in ``c_p_pointer``

czgdp1807:
#546


integration_tests/bindc_02.py
```diff
@@ -0,0 +1,13 @@
...
+# x: pointer(i16[:])
+# c_p_pointer(queries, x)
+# print(queries, pointer(x))
+
+def f():
+    yptr: pointer(i16[:])
+    y: i16[2]
+    y[0] = 1
+    y[1] = 2
+    y <= yptr
```
czgdp1807, line 13: Not sure which syntax to use the above, `y <= yptr` or `yptr = pointer(y)` to allow assigning arrays to pointers in Python. Thoughts @certik ?

certik: In Fortran:
```fortran
yptr => y
```
In C++:
```c++
yptr = y // (or &y if `y` was not already a pointer)
```

certik: So it seems `yptr = pointer(y)` should be transformed to `Associate(yptr, y)` ASR node. If on the RHS is anything else than just `pointer(y)`, then we can give an error message for now.

src/libasr/ASR.asdl
```diff
@@ -272,6 +272,7 @@ expr
     | ComplexIm(expr arg, ttype type, expr? value)
     | DictItem(symbol a, expr key, expr? default, ttype type, expr? value)
     | CLoc(expr arg, ttype type, expr? value)
+    | GetPointer(expr arg, ttype type, expr? value)
```
czgdp1807, line 275: This is needed to represent, `pointer(x)` call in ASR.

czgdp1807:
Next and final step: Make `c_p_pointer` work for arrays i.e., the address should be same after assignment.

integration_tests/bindc_02.py
czgdp1807: Done. `yptr = pointer(y)` works now.

czgdp1807:
@certik This is ready for review and possibly merge if everything looks good to you. Thanks.

src/libasr/ASR.asdl
czgdp1807: No. `Associate` associates a variable to a pointer. `GetPointer` is used for representing the operation of obtaining the pointer to a variable at lower level. Basically it represents `pointer(x)` call in ASR. Its different from associate.

certik:
I am always trying to work hard on keeping the ASR small and consistent. It seems the operations here are:

* `CLoc`: Get a C pointer to an ASR entity
* `Associate`: Get an ASR pointer to an ASR entity, and assign it to a pointer variable
* `GetPointer`: Get an ASR pointer to an ASR entity

So it seems we should unify this into just:

* `GetPointer(expr arg, ttype type, expr? value)`: Get an ASR  to an ASR entity

The `arg` is any ASR expression (with appropriate restrictions) and `type` is the type of `expr` but with the `Pointer` "wrapper" in the type, so if `expr` is `Integer` then `GetPointer` is `Pointer(Integer)`.

Then let's remove `Associate` and instead implement it using `Assignment(x, GetPointer(y))`.

Finally, there are two ways to implement `CLoc`. Either via `GetPointer` and set the type to `CPtr`. However, it seems that would require quite a bit of special logic and checking in all backends for this case. An alternative is to add another node `PointerToCPtr`, to the already existing `CPtrToPointer` node:

```ml
    | CPtrToPointer(expr cptr, expr ptr, expr? shape)
    | PointerToCPtr(expr ptr, expr cptr, expr? shape)
```

Then we implement `CLoc(x)` as `PointerToCPtr(GetPointer(x))`.

This makes it nicely symmetric and consistent. ASR is not "polluted" with C pointers, our `GetPointer` always returns a typed ASR pointer, no void. The `PointerToCPtr` and `CPtrToPointer` are then used to convert between an ASR pointer and a C pointer.

Another idea: should `CPtr` be renamed to `CPointer` (`CPointerToPointer`)? It's fine to use `CPtr` for now, it's shorter.

src/libasr/ASR.asdl
czgdp1807: I think removal should happen in a separate PR. What do you say?

certik: Yes, let's keep this PR manageable. It looks like instead of Assignment, we should have `GetPointer`.

czgdp1807:
`CPtrToPointer` is a statement but `CLoc` is an expression. So, `PointerToCPtr(expr ptr, expr cptr, expr? shape)` should be modified as, `PointerToCPtr(expr ptr)`.

czgdp1807:
> `CPtrToPointer` is a statement but `CLoc` is an expression. So, `PointerToCPtr(expr ptr, expr cptr, expr? shape)` should be modified as, `PointerToCPtr(expr ptr)`.

> Then we implement CLoc(x) as PointerToCPtr(GetPointer(x))

Not necessarily. `PointerToCPtr` can just be implemented as `PointerToCPtr(x)` as pointers are already supported at syntax level. However you need `GetPointer` to represent `pointer(x)` call because pointers are not supported already in Python at syntax level.

czgdp1807:
Corresponding LFortran MR - https://gitlab.com/lfortran/lfortran/-/merge_requests/1783

czgdp1807:
Associate will be removed in a separate PR. While I will remove Associate from LFortran at least. That should make it easy to do here.

czgdp1807:
Okay. This is ready. Please feel free to review @certik . I will clean up the git history once you approve.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -65,6 +65,8 @@
 #include <libasr/codegen/llvm_utils.h>
 #include <libasr/codegen/llvm_array_utils.h>
 
+#include <lpython/pickle.h>
```
certik, line 68: This has to be removed, as it won't compile with LFortran.

certik:
Otherwise I think this is ok.

src/libasr/codegen/asr_to_llvm.cpp
czgdp1807: Ah Yes.

czgdp1807:
Cleaned up the Git history. @certik

czgdp1807:
Tests pass. Thanks.

certik:
I am going to merge this. Let's improve upon it later as needed.

PR #564, opened 2022-06-09
cbrt and exp2 functions implemented

Madhav2310:


certik:
@Madhav2310 you should try to setup LPython locally and simply test this locally before pushing. You can then iterate very quickly, instead of waiting for the CI to finish.

integration_tests/test_math_02.py
```diff
@@ -1,7 +1,16 @@
-from math import (sin, cos, tan, pi, sqrt, log, log10, log2, erf, erfc, gamma,
+from math import (sin, cos, tan, pi, sqrt, cbrt, log, log10, log2, erf, erfc, gamma,
                   lgamma, asin, acos, atan, atan2, asinh, acosh, atanh,
                   tanh, sinh, cosh, hypot, copysign)
 
+from ltypes import i32, f64, i64
+
+pi: f64 = 3.141592653589793238462643383279502884197
+e: f64 = 2.718281828459045235360287471352662497757
+tau: f64 = 6.283185307179586
```
czgdp1807, line 9: Do we need this level of precision for testing? I think 4-5 decimal places is fine.

src/runtime/math.py
```diff
@@ -312,8 +317,18 @@ def trunc(x: f32) -> i32:
...
+    Returns square root of a non negative x
+    """
+
     return x**(1/2)
 
+def cbrt(x: f64) -> f64:
+    """
+    Returns cube root of a number x
+    """
+    return x**(1/3)
```
czgdp1807, line 330: Do these expressions actually work with Lpython?

Madhav2310:
@Smit-create Could you please see why it is failing to import the functions in the test files?

namannimmo10:
> @Smit-create Could you please see why it is failing to import the functions in the test files?

@Madhav2310 -- you should merge this with the latest `main`.

integration_tests/test_math_02.py
certik: Ultimately these constants should be imported from the builtin math module.

Madhav2310:
> > @Smit-create Could you please see why it is failing to import the functions in the test files?
> 
> @Madhav2310 -- you should merge this with the latest `main`.

@namannimmo10  My local repository is up to date, could you elaborate on what to do exactly?

src/runtime/math.py
Smit-create: yeah, they should.

integration_tests/test_math_02.py
Smit-create, line 9: Can you please try removing these global variables and try running tests with local variables?

Smit-create:
Are these functions actually available in CPython's math? If not then it will definitely raise an error.

Smit-create:
I don't see these functions here: https://docs.python.org/3/library/math.html

integration_tests/test_math_02.py
Madhav2310: Sure I'll do that

Madhav2310:
> I don't see these functions here: https://docs.python.org/3/library/math.html

Right, I just saw it. It is very weird, they have tests for these functions, but not the functions themselves https://github.com/python/cpython/blob/2d787971c65b005d0cce219399b9a8e2b70d4ef4/Lib/test/test_math.py

namannimmo10:
Assuming that you've added remote `origin` as github.com/Madhav2310/lpython.git and `upstream` as github.com/lcompilers/lpython.git, this is what you can do:
```console
$ git checkout main
$ git pull upstream main
$ git checkout madhav
$ git rebase main
```
Now, if you run the integration tests, they'll pass.

Smit-create:
> Now, if you run the integration tests, they'll pass.

Well, that won't pass for sure.

```py
>>> from math import exp2
ImportError: cannot import name 'exp2' from 'math' (/Users/thebigbool/opt/anaconda3/envs/lp/lib/python3.10/lib-dynload/math.cpython-310-darwin.so)
```

Smit-create:
> It is very weird, they have tests for these functions

See the test carefully, they have tested for `TypeError`.
```
def testCbrt(self):
        self.assertRaises(TypeError, math.cbrt)
```

namannimmo10:
> > Now, if you run the integration tests, they'll pass.
> 
> Well, that won't pass for sure.
> 
> ```python
> >>> from math import exp2
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> ImportError: cannot import name 'exp2' from 'math' (/Users/thebigbool/opt/anaconda3/envs/lp/lib/python3.10/lib-dynload/math.cpython-310-darwin.so)
> ```

My bad. Yes, they won't pass because of this.

Madhav2310:
> > It is very weird, they have tests for these functions
> 
> See the test carefully, they have tested for `TypeError`.
> 
> ```
> def testCbrt(self):
>         self.assertRaises(TypeError, math.cbrt)
> ```

I see, thank you! I'll take care of it in the future

PR #565, opened 2022-06-09
Add a CMake based system for integration tests

certik:
It tests CPython, LPython/LLVM and LPython/C. One invokes it using:

* CPython: cmake .
* LPython/LLVM: cmake -DLPYTHON_BACKEND=llvm .
* LPython/C: cmake -DLPYTHON_BACKEND=c .

Then one does `make` and `ctest`.

It should be equivalent to the previous testing script.


certik:
@czgdp1807 After this is merged, simply add `c` to a given test and you'll see failures, the most common is that in C we need to forward declare functions (unlike C++). I haven't tested adding extra C files to link, but it will not be difficult.

If you want, you can try to get as many tests working with `c` as possible. It will require fixes in the C backend. I added one test that works with `c`, to ensure at least simple things work.

certik:
@czgdp1807 regarding this PR, would you mind please testing locally by hand as well? This is a big change in our infrastructure, so to ensure it works.

czgdp1807:
> If you want, you can try to get as many tests working with `c` as possible. It will require fixes in the C backend. I added one test that works with `c`, to ensure at least simple things work.

Cool. Sounds good.



> would you mind please testing locally by hand as well? This is a big change in our infrastructure, so to ensure it works.

Okay. On it right now.

certik:
The failure on Linux and macOS at the CI is:
```
[  1%] Generating exit_01.o
python: can't open file '/usr/share/miniconda/envs/test/bin/../share/lpython/lib/lpython_parser.py': [Errno 2] No such file or directory
The command 'python /usr/share/miniconda/envs/test/bin/../share/lpython/lib/lpython_parser.py exit_01.py' failed.
```

Not sure if this path is not there at the CI anymore, or what the problem could be.

czgdp1807:
Works on my mac as well. All tests pass.

czgdp1807:
I check the CI on my debug commit and found out the following,

Inside, `/usr/local/miniconda/envs/test/bin/../share/` `lpython` is in `~/work/lpython/lpython/integration_tests` but we are executing `python` command over, `/usr/local/miniconda/envs/test/bin/../share/lpython/lib/lpython_parser.py` but we should be doing, `/usr/local/miniconda/envs/test/bin/../share/~work/lpython/lib/lpython_parser.py` (see the extra `~work`).

Issue #410, opened 2022-04-21
Don't overload the `bool` intrinsic function


namannimmo10:
When I run this:
```python
def f():
  i: i32
  i = 3
  b: bool
  b = bool(i)
  print(b)


f()
```
```console
% lpython --show-cpp c.py
#include <iostream>
#include <string>
#include <vector>
#include <cassert>
#include <cmath>
#include <complex>
#include <Kokkos_Core.hpp>
#include <lfortran_intrinsics.h>

template <typename T>
Kokkos::View<T*> from_std_vector(const std::vector<T> &v)
{
    Kokkos::View<T*> r("r", v.size());
    for (size_t i=0; i < v.size(); i++) {
        r(i) = v[i];
    }
    return r;
}

void _lpython_main_program()
{
    f();
}

void f()
{
    bool b;
    int i;
    i = 3;
    b = __lpython_overloaded_0__bool(i);
    std::cout << b << std::endl;
}

bool __lpython_overloaded_0__bool(int x)
{
    bool _lpython_return_variable;
    _lpython_return_variable = x != 0;
    return _lpython_return_variable;
}

namespace {

void main2() {
        _lpython_main_program();
}

}
int main(int argc, char* argv[])
{
    Kokkos::initialize(argc, argv);
    main2();
    Kokkos::finalize();
    return 0;
}

```
I think in this case, `f()` function could be as simple as:
```c++
void f()
{
    bool b;
    int i;
    i = 3;
    b = (bool)i;
    std::cout << b << std::endl;
}
```
That can happen if we don't overload the `bool` intrinsic function, and use `IntegerToLogical` casting in AST->ASR.
So, in other words, all we have to do is implement `handle_intrinsic_bool`, similar to

https://github.com/lcompilers/lpython/blob/2739a8596a4bad5528a75f4aa290af4e7b15443d/src/lpython/semantics/python_ast_to_asr.cpp#L2572

Comment out the current implementation of `bool` in `lpython_builtin.py` and `python_comptime_eval.h`,

And modify the C++ backend (`Cast` visitor):
```diff
diff --git a/src/libasr/codegen/asr_to_cpp.cpp b/src/libasr/codegen/asr_to_cpp.cpp
index 9dd4732cd..3af93c340 100644
--- a/src/libasr/codegen/asr_to_cpp.cpp
+++ b/src/libasr/codegen/asr_to_cpp.cpp
[...]

PR #566, opened 2022-06-09
ASR: split `Compare` based on different types

namannimmo10:


namannimmo10:
This is almost ready. I'll raise a merge request in LFortran just to make sure that every test passes there. Then we'll merge both at once. For now, a review would be helpful!

namannimmo10:
MR opened: https://gitlab.com/lfortran/lfortran/-/merge_requests/1784

namannimmo10:
MR opened: https://gitlab.com/lfortran/lfortran/-/merge_requests/1784

src/libasr/ASR.asdl
```diff
@@ -217,7 +217,6 @@ expr
         -- Such as: (x, y+z), (3.0, 2.0) generally not known at compile time
     | ComplexConstructor(expr re, expr im, ttype type, expr? value)
     | NamedExpr(expr target, expr value, ttype type)
-    | Compare(expr left, cmpop op, expr right, ttype type, expr? value, expr? overloaded)
```
certik, line 220: Also you need to put in `OverloadedCompare` node to represent the `expr? overloaded`.

certik:
I think this change is in the right direction, I think we should do it. The comparison operator becomes simple/more explicit.

namannimmo10:
@certik -
```python
from ltypes import c32, c64

def main():
  c: c32
  c = complex(3, 2)
  print(c)

  c2: c64
  c2 = complex(4.5, 6.7)
  print(c2)


main()
```
This code seems to produce the following ASR on the `main` branch:
```
(TranslationUnit 
   (SymbolTable 
      1 
      {
         _lpython_main_program: 
            (Subroutine 
               (SymbolTable 
                  100 
                  {
                  }) 
               _lpython_main_program [] [
               (SubroutineCall 1 main () [] ())] 
               Source 
               Public 
         complex@__lpython_overloaded_5__complex: 
            (ExternalSymbol 
               1 
               complex@__lpython_overloaded_5__complex 4 __lpython_overloaded_5__complex 
               lpython_builtin [] 
               __lpython_overloaded_5__complex 
               Public), 
      complex@__lpython_overloaded_9__complex: 
         (ExternalSymbol 
            1 
            complex@__lpython_overloaded_9__complex 4 __lpython_overloaded_9__complex 
            lpython_builtin [] 
            __lpython_overloaded_9__complex 
            Public), 
   lpython_builtin: (IntrinsicModule lpython_builtin), 
main: 
   (Subroutine 
      (SymbolTable 
         2 
         {
            c: 
               (Variable 
                  2 
                  c 
                  Local () () 
                  Default 
                  (Complex 4 []) 
                  Source 
                  Public 
                  Required .false.), 
            c2: 
               (Variable 
                  2 
                  c2 
                  Local () () 
                  Default 
                  (Complex 8 []) 
                  Source 
                  Public 
                  Required .false.), 
            complex: 
               (ExternalSymbol 
                  2 
                  complex 4 complex 
                  lpython_builtin [] 
                  complex 
                  Private)
         }) 
      main [] [
      (= 
         (Var 2 c) 
         (Cast 
            (FunctionCall 1 complex@__lpython_overloaded_9__complex 2 complex [
               ((IntegerConstant 3 
                  (Integer 4 []))) 
               ((IntegerConstant 2 
                  (Integer 4 [])))] 
               (Complex 8 []) 
               (ComplexConstant   3.00000000000000000e+00   2.00000000000000000e+00 
                  (Complex 8 [])) ()) 
            ComplexToComplex 
            (Complex 4 []) 
            (ComplexConstant   3.00000000000000000e+00   2.00000000000000000e+00 
               (Complex 4 []))) ()) 
      (Print () [
         (Var 2 c)]) 
      (= 
         (Var 2 c2) 
         (FunctionCall 1 complex@__lpython_overloaded_5__complex 2 complex [
            (
               (RealConstant   4.50000000000000000e+00 
                  (Real 8 []))) 
            (
               (RealConstant   6.70000000000000018e+00 
                  (Real 8 [])))] 
            (Complex 8 []) 
            (ComplexConstant   4.50000000000000000e+00   6.70000000000000018e+00 
               (Complex 8 [])) ()) ()) 
      (Print () [
         (Var 2 c2)])] 
      Source 
      Public 
main_program: 
   (Program 
      (SymbolTable 
         99 
         {
         }) 
      main_program [] [
      }) 
   [])

```
It doesn't look correct to me -- there's some issue with casting.  And because of this, the tests don't pass.

certik:
LPython should translate `x = complex(a, b)` into `Assign(ComplexConstructor(a, b))` and `complex(1.0, 3.0)` into `ComplexConstant(1.0, 3.0)`, it should not call a function `_lpython_overloaded_9__complex`, as it currently does.

certik:
But that issue is unrelated --- we can make that change later.

For this PR we should split Compare, but otherwise not do any other modifications.

Smit-create:
> It doesn't look correct to me -- there's some issue with casting. And because of this, the tests don't pass.

What's the issue here? I see that it works correctly. Any pointers?

> LPython should translate `x = complex(a, b)` into `Assign(ComplexConstructor(a, b))` and `complex(1.0, 3.0)` into `ComplexConstant(1.0, 3.0)`, it should not call a function `_lpython_overloaded_9__complex`, as it currently does.

Yeah, it uses `complex` from `lpython_builtin` but doesn't create any bug I suppose?

src/libasr/ASR.asdl
```diff
@@ -266,6 +270,7 @@ expr
     | ArrayPack(expr array, expr mask, expr? vector, ttype type, expr? value)
     | Transfer(expr source, expr mold, expr? size, ttype type, expr? value)
     | DerivedRef(expr v, symbol m, ttype type, expr? value)
+    | OverloadedCompare(expr left, cmpop op, expr right, ttype type, expr? value, expr overloaded)
```
czgdp1807, line 273: I am -1 on putting overloaded compare as a separate node. An extra attribute in `Compare` was enough I think.

czgdp1807:
I feel this is a very big change. May be first we should add new nodes and if it works correctly then remove the older nodes. Doing everything all at once makes reviewing harder and the predicting the long term effect of these changes is tricky. @namannimmo10 We should discuss a step by step strategy to achieve this kind of ASR refactoring.

src/libasr/ASR.asdl
certik: We don't have Compare anymore.

So which of the IntegerCompare, RealCompare etc should have the extra member? Note that neither Python nor Fortran allows you to overload Integer or Real comparison. It allows you to overload a user defined type compare, which is what the OverloadedCompare node does.

czgdp1807: No issues. Let’s have it then.

czgdp1807:
@namannimmo10 Could you please resolve the conflicts here anyways?

certik:
It's a huge change, although perhaps smaller than the BinOp split change.

One could do it step by step, although one issue is that often errors only show after we actually remove the BinOp node, otherwise you don't easily notice which places in the code need updating.

The best workflow with these big changes is to get them done relatively quickly, with as little changes as possible, and make tests pass, and just get it done and get it in. Then we'll polish it as we go.

Alternatively, we can just keep BinOp and only split IntegerBinOp, but as I said, it might actually make it more difficult, as we will not see all the changes that need fixing.

namannimmo10:
The CI is green. I suggest that we merge this PR as it is. Sorry, this was a big change! But this kind of refactoring had to be done by removing/adding the ASR nodes altogether.

certik:
I agree to just merge. I am reviewing it now to see if there are any blockers.

certik (approved):
I think that this is fine. No major issues. Let's merge this.

certik:
@namannimmo10 would you mind please sending a similar PR to LFortran?

certik:
Thanks a lot for doing this. Big change, a lot of work to get everything done. I think the ASR looks a lot cleaner, so does the LLVM backend (our main backend). When LLVM looks clean, all other backends are typically also clean, by proxy.

namannimmo10:
> @namannimmo10 would you mind please sending a similar PR to LFortran?

[!1784](https://gitlab.com/lfortran/lfortran/-/merge_requests/1784) was merged. ✌️

certik:
Excellent! Thank you again for doing that. It's a lot of work.

In terms of ASR design, what else should we do, if anything, right now?

namannimmo10:
My pleasure!

Some random ideas:

We should verify the ASR optimization passes since the significant splitting of nodes is done ([!709](https://gitlab.com/lfortran/lfortran/-/issues/709)).

Since we are on our way to having a self-sufficient ASR layer, we should try to avoid overloading functions, and have ASR nodes do the job. For example, see https://github.com/lcompilers/lpython/issues/410. As a result, the ASR looks a lot cleaner -- it is one of LCompilers' major selling points.

I was reading about the [Intermediate representations in `rustc`](https://rustc-dev-guide.rust-lang.org/overview.html#intermediate-representations). It looks like they have three IRs between AST to LLVM IR, each having a different purpose. We have only one such IR and we even do some high-level optimizations in it. I'm sure this is a great design as noticed from the benchmarks. But I'm curious, why didn't you decide to have more than one intermediate representation?

czgdp1807:
> We should verify the ASR optimization passes since the significant splitting of nodes is done ([!709](https://gitlab.com/lfortran/lfortran/-/issues/709)).

I will start doing this.

czgdp1807:
I think in `rustc` they have HIR, Typed HIR and Mid level HIR. ASR fulfills the purposes of each of these simultaneously i.e, type checking is easily possible in ASR and it is sort of a forest of trees (similar to control flow graph). Each stmt follows the previous and each stmt has expression trees. So I don’t feel the need to separate out ASR more or in other words I feel it is self sufficient concept. @certik Do you have plans to split out ASR into more intermediate representations?

certik:
Yes, things like bool should be done in ASR itself I think.

With each new IR we add, there is a very likely compilation slow down (you can only avoid it if somehow it simplifies the backend so much that it actually becomes faster), as well as more code to maintain, etc. So adding a new IR must pay for itself. Adding ASR seems to pay for itself, some compilation slowdown might be there, it's hard to measure, but it greatly simplified the whole design.

We can sit down and brainstorm what other alternative designs there are and if any of them would be better. For example we can see how HIR, Typed HIR and Mid level HIR corresponds to ASR. For example, we will have a lot of optimization passes. What is the best IR to do optimization passes fast?

It seems the current ASR is fulfilling these roles very well, but I am always interested in seeing if there is a way to do it even better.

PR #574, opened 2022-06-10
Enabling ``c`` backend testing for ``integration_tests/test_c_interop_01.py``

czgdp1807:


src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -298,6 +298,9 @@ R"(#include <stdio.h>
     std::string declare_all_functions(const SymbolTable &scope) {
         std::string code;
         for (auto &item : scope.get_scope()) {
+            if( startswith(item.first, "_lfortran_") ) {
+                continue ;
+            }
```
czgdp1807, line 303: Functions like, `_lfortran_bgt32`, `_lfortran_bgt64` i.e., all those in `lfortran_intrinsics.h` are already declared. Since we are including the header file at the top so we shouldn't re-declare those functions in the generated `.c` file. This checks helps in doing the same. Very broad but works for now.

certik:
I think it's fine, but didn't have time to study carefully. Tests should be fixed first.

czgdp1807:
> Tests should be fixed first.

Will do that once I convert this to ready mode. Meanwhile if you have any thoughts/comments/objections on the changes, then feel free to let me know.

src/libasr/codegen/asr_to_c_cpp.h
certik: How did they get declared in the first place? We could use something like the `is_function_intrinsic` (not sure the exact spelling) in asr utils.

certik:
I think the changes are good, thanks!

src/libasr/codegen/asr_to_c_cpp.h
certik: If there is no other way right now, then at least add `// TODO: refactor this properly later`, to make clear that this is a temporary hack for a "prototype", that we need to "harden" later.

czgdp1807: Sure I can add one TODO. I agree its a very broad check but at least we can live with it until we find a corner case in future. I think I will wrap it inside a utility function and then use it everywhere else.

certik: The actual check should check the "intrinsic" attribute on a module, not string based like this. We already have a function that does that in utils.

czgdp1807: I see. Will try to do this on Monday. If functions in `lfortran_intrinsics.h` make their way to ASR then it must be easy to check for `intrinsic` attribute on a symbol. Otherwise we will have to satisfy ourselves with a string based check like this.

certik: I think they only make their way in because the LPython intrinsic modules call them, e.g., here:

https://github.com/lcompilers/lpython/blob/f14ce2f69aee550648c0a187b304ebb552cc995f/src/runtime/math.py#L317

For now the above is fine as a "prototype" (that we can merge into master). On Monday you can implement a proper fix.

Regarding the test failures, do you know what needs to be done? I'll try to finish this PR, so that I can work more on the C interop testing.

czgdp1807: We just need to update the reference tests I think. Nothing much. I will continue working on https://github.com/lcompilers/lpython/pull/574#issuecomment-1152415451 on Monday. Do you want me to pass the tests of this PR quickly? I can do that after dinner. Just let me know.

czgdp1807:
Tests should pass. I hope. 🤞.

czgdp1807:
Tests pass. This is ready.

certik:
I don't like commenting out tests, as we need them for LLVM and pure Python too. But fortunately I fixed most of these issues in main already, so I got the asserts working in this PR.

src/libasr/codegen/asr_to_c_cpp.h
certik: I got rid of this. I think we really need to put the declaration in there, for the C compiler to know what to compile. The `lfortran_intrinsics.h` is a special case that should be handled separately.

certik: After I did it, it uncovered a bug: https://github.com/lcompilers/lpython/pull/574/commits/ae8fdc11a9ea6097babe1ca59a6d7bdbecb6efc2

Issue #556, opened 2022-06-05
Optimising C generated code


Smit-create:
I was looking at the C backend code and found one way to optimize the generated code (not sure if that is true). At present we just include all the headers at the top of the file without actually verifying if they are used or not. Some compilers may warn about unused includes. Should we enable this verifying of includes?

certik:
We can keep track of what features are used, and "enable" the header include as needed.

czgdp1807:
I think we can close this. Optimisation of code is not a concern since we do machine independent optimisations via ASR to ASR passes. We can call those passes in the below function,

https://github.com/lcompilers/lpython/blob/70fef49dbbb6cbb0447f7013231171e5c8b8e5df/src/libasr/codegen/asr_to_c.cpp#L377-L391

See the LLVM code generation for example (we call optimisation passes before starting the visitor),

https://github.com/lcompilers/lpython/blob/70fef49dbbb6cbb0447f7013231171e5c8b8e5df/src/libasr/codegen/asr_to_llvm.cpp#L4924-L4990

certik:
Yes, we'll be able to do all high level optimizations in ASR. ASR is platform independent. The optimizations that are left for the ASR->LLVM or ASR->C backends to do are low level, such as promoting variables to registers, or vectorizing loops. Both LLVM and C compilers can do these. There might be some "collaboration" between ASR optimizations and LLVM / C optimizations for vectorization, such as vectorizing the outer loop, but it might be enough to just do the operation in ASR and just lower. So the ASR->something phase might not need any optimizations on its own.

PR #575, opened 2022-06-11
C: Include only required headers

Smit-create:
#556

### DIFF
--- a/src/libasr/codegen/asr_to_c_cpp.h
+++ b/src/libasr/codegen/asr_to_c_cpp.h
@@ -13,6 +13,7 @@
 
 #include <iostream>
 #include <memory>
+#include <set>
 
 #include <libasr/asr.h>
 #include <libasr/containers.h>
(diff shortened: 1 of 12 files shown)


src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -561,43 +563,6 @@ R"(#include <stdio.h>
...
-            }
-            case (ASR::cast_kindType::LogicalToInteger) : {
-                src = "(int)(" + src + ")";
-                break;
-            }
-            default : throw CodeGenError("Cast kind " + std::to_string(x.m_kind) + " not implemented");
-        }
-        last_expr_precedence = 2;
-    }
-
```
certik, line 600: Why did you remove these? 

Is it needed for this PR? If not, let's do it separately. I am hoping to have as much in the common visitor as possible, but it looks like currently they are overridden in the subclasses (C and C++ visitors)?

certik:
This is a great idea, thanks for doing it. We can later put the ASSERT macros into our runtime library header file, and also only include the runtime library if it is being used.

In general, I want both the C and C++ backend to eventually produce code that you would write by hand if you translated Python or Fortran into C or C++ manually. So this change goes in that direction.

src/libasr/codegen/asr_to_c_cpp.h
Smit-create: Okay, will change that in new PR.

certik:
I think that this is good, thanks! Can you please resolve conflicts?

certik (approved):
I think this loos good, thanks!

PR #578, opened 2022-06-11
[Parser] Rules for `not in`

akshanshbhatt:
Related Issue: #511


src/lpython/parser/parser.yy
```diff
@@ -244,7 +245,7 @@ void yyerror(YYLTYPE *yyloc, LFortran::Parser &p, const std::string &msg)
 %left "or"
 %left "and"
 %precedence "not"
-%left "==" "!=" ">=" ">" "<=" "<" TK_IS_NOT KW_IS // "not in" "in"
+%left "==" "!=" ">=" ">" "<=" "<" TK_IS_NOT KW_IS TK_NOT_IN // "in"
```
certik, line 248: One should be also able to do:
```suggestion
%left "==" "!=" ">=" ">" "<=" "<" "is not" "is" "not in" // "in"
```
But it's not clear to me which one is more readable, so just keep it as it is.

certik (approved):
I think that this looks good, thanks!

src/lpython/parser/parser.yy
Thirumalai-Shaktivel: I think the suggested one is more readable, but I feel it is ok.

Thirumalai-Shaktivel:
Thanks for this PR, @akshanshbhatt

src/lpython/parser/parser.yy
certik: We can change it later in another PR.

akshanshbhatt: > We can change it later in another PR.

@certik, I also prefer the suggested one. In fact, I tried it first, but the build failed.

Verbose:
```
❯ lpbuild
+ ci/version.sh
++ git describe --tags --dirty
+ version=v0.2.0-835-g9e3cef7a1-dirty
+ version=0.2.0-835-g9e3cef7a1-dirty
+ echo 0.2.0-835-g9e3cef7a1-dirty
+ python grammar/asdl_py.py
Assuming default values of Python.asdl and python_ast.py
+ python grammar/asdl_cpp.py grammar/Python.asdl src/lpython/python_ast.h
+ python grammar/asdl_cpp.py src/libasr/ASR.asdl src/libasr/asr.h
+ python src/libasr/wasm_instructions_visitor.py
Assuming default values of wasm_instructions.txt and wasm_visitor.h
+ cd src/lpython/parser
+ cd src/lpython/parser
+ bison -Wall -d -r all parser.yy
parser.yy: error: shift/reduce conflicts: 55 found, 0 expected
parser.yy:248.1-5: warning: useless precedence and associativity for "is" [-Wprecedence]
  248 | %left "==" "!=" ">=" ">" "<=" "<" "is not" "is" "not in" // "in"
      | ^~~~~
```
What can be the possible cause?

certik: You have to make sure the "is not" and others are assigned next to the token definitions. I see "is not" is, "not in" is, but "is" is not. 
```diff
--- a/src/lpython/parser/parser.yy
+++ b/src/lpython/parser/parser.yy
@@ -159,7 +159,7 @@ void yyerror(YYLTYPE *yyloc, LFortran::Parser &p, const std::string &msg)
 %token KW_IF
 %token KW_IMPORT
 %token KW_IN
-%token KW_IS
+%token KW_IS "is"
 %token KW_LAMBDA
 %token KW_NONE
 %token KW_NONLOCAL
```

certik: Alternatively, since we do not do this for keywords, just keep `KW_IS`, but use "is not" and "not in", that should work now.

PR #579, opened 2022-06-12
Expose Default values in Function definition to AST

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
ping @akshanshbhatt

certik:
I would need to review this one more carefully, it seems a bit complex, but probably we have to do it this way?

akshanshbhatt:
> ping @akshanshbhatt

Awesome! I tried it locally, and it seems to work as intended.
Any idea why the MacOS CI is failing?

certik:
I tested locally on macOS, the test passes. At the CI the failure seems to be
```
TEST: parser/function_def1.py
    * ast_new The JSON metadata differs against reference results
Reference JSON: tests/reference/ast_new-function_def1-1a872df.json
Output JSON:    tests/output/ast_new-function_def1-1a872df.json
Omitting 8 identical items
Differing items:
{'returncode': -11} != {'returncode': 0}
{'stdout': None} != {'stdout': 'ast_new-function_def1-1a872df.stdout'}
{'stdout_hash': None} != {'stdout_hash': 'ad5ea65f2ef1994819a5c48936f936ea955303c08253419d72aa450b'}
```
Which is one of the tests modified in this PR, it seems at the CI LPython failed with a returncode "-1" and did not produce any output nor errors.

This could happen if lpython segfaulted, possibly if some pointer handling in this PR is incorrect and just happened to be exposed at the CI.

src/lpython/parser/semantics.h
```diff
@@ -283,76 +284,131 @@ static inline withitem_t *WITH_ITEM(Allocator &al, Location &l,
...
-        expr_t** m_defaults, size_t n_defaults) {
+        Arg* /*m_posonlyargs*/, size_t /*n_posonlyargs*/,
+        Arg** m_args, size_t n_args,
+        Arg* m_vararg, size_t n_vararg,
+        Arg** m_kwonlyargs, size_t n_kwonlyargs,
+        expr_t** /*m_kw_defaults*/, size_t /*n_kw_defaults*/,
+        Arg* m_kwarg, size_t n_kwarg) {
     Args *r = al.allocate<Args>();
+    Vec<arg_t> _m_args;
+    _m_args.from_pointer_n(nullptr, 0);
```
certik, line 310: I don't think you need this line?

src/lpython/parser/semantics.h
```diff
@@ -283,76 +284,131 @@ static inline withitem_t *WITH_ITEM(Allocator &al, Location &l,
...
+        _m_vararg.reserve(al, n_vararg);
+        defaults.reserve(al, n_vararg);
+        _m_vararg.push_back(al, m_vararg->_arg);
+        if(m_vararg->def) {
+            defaults.push_back(al, m_vararg->defaults);
+        }
+    }
+    if(n_kwonlyargs > 0) {
+        _m_kwonlyargs.reserve(al, n_kwonlyargs);
+        defaults.reserve(al, n_kwonlyargs);
```
certik, line 339: You seem to be calling "reserve" couple times on `defaults`. That might be the issue.

certik: Replace this:
```

    Vec<arg_t> _m_args;
    _m_args.from_pointer_n(nullptr, 0);
```
with:
```
    Vec<arg_t> _m_args;
    _m_args.reserve(al, 4);
```
The same for all the other ones. Remove all calls to reserve beyond that. I don't know if this fixes the issue at the CI, but that's one issue I can see.

Thirumalai-Shaktivel:
> I would need to review this one more carefully, it seems a bit complex, but probably we have to do it this way?

Yup, I don't think we have other option!

src/lpython/parser/semantics.h
Thirumalai-Shaktivel: Yup, I will do it.

certik:
One possible issue is eliminated. But the bug is still there.

You can now try to split this PR into smaller pieces and "bisect" it. You can open a separate PR for the debugging.

Thirumalai-Shaktivel:
This PR is ready to go in!

certik (approved):

Issue #379, opened 2022-04-15
Introduce multiple return type


Smit-create:
We should support `lpython` to allow multiple return types. One example of this is the `pow` function. We know that `pow(x: i32, y: i32)` will return `i32` if `y>=0` else it should return `f64`. So, I am thinking about how should we design the return type notation in this case:

Some of the designs are:

```py
def pow(x: i32, y: i32) -> Union[i32, f32]:
    pass
```
or

```py
def pow(x: i32, y: i32) -> i32 | f32:
    pass
```

Let me know you views and suggestions.

Madhav2310:
I like the latter more, looks elegant. Also, it's more predominantly used so will be easier to understand by developers.

certik:
This is more tricky than it might seem. To allow `Union` in a performant manner, it would be a generic (template) feature. The code would get instantiated once it is called, and specialized for the given types. This can be done for things like `def f(x: i32 | f32)`. However, as a return type that gets tough --- one could infer that from where it is assigned to, but that might not always be possible if you for example have another generic function `def g(y: i32 | f32)` and you do `g(pow(..))`, then you do not know the types, and thus what function to call.

The other way to implement `Union` is via runtime generics, for example you could represent the `i32 | f32` as a struct (actual C style union) with a tag to know the type. That will work, but would not be performing well. I think we should not go that route. 

The good news is that I think the builtin function `pow` is rarely used. So we could simply always return `f32`. And one can do something like `x: i32 = int(pow(2, 3))`, which will work in both CPython and LPython and behave the same way. And be performing.

More important is what to do about regular `**` operator which behaves the same way:

```
>>> 3**2
9
>>> 3**(-2)
0.1111111111111111
```

Well, if we know the values at compile time, we can figure out the type. But what about this:
```
>>> x = 3
>>> y = 2
>>> x**y
9
>>> y = -2
>>> x**y
0.1111111111111111
```
The type truly depends on the runtime values of `y`, which could be read from an input file, thus we do not know at compile time what the type will be.

Here is one way out:

* For integer x,y, we always return an integer for positive `y`, and for negative `y` we return a runtime exception in Debug mode. That way you would know that the feature is not supported. But in the code you can assume the result is an integer. You test your code and ensure you get no runtime exceptions in Debug mode. Then you recompile in Release mode and you will not have any checks and thus things will be fast. The contract of LPython is that if it compiles and you get no runtime errors in Debug mode, the behavior will be equivalent as CPython.

[...]

PR #581, opened 2022-06-12
Implemented multiple datatypes for pow and fabs functions

Madhav2310:


src/runtime/math.py
```diff
@@ -215,25 +227,68 @@ def fabs(x: f64) -> f64:
...
+@overload
+def pow(x: f32, y: f32) -> f64:
+    """
+    Return `x` raised to the power `y`.
+    """
+    return x**y
 
-def pow(x: f64, y: f64) -> f64:
+@overload
+def pow(x: i32, y: i32) -> i64:
```
namannimmo10, line 273: See #379; We either raise an exception if `y` is less than 0 or always return a float.

certik: We do not have exceptions yet. We could implement them by simply terminating the program with the message of the exception right away. So you can't catch them, but you can use them to report errors, and it would "work" both in LPython and CPython (assuming you don't catch it).

Madhav2310: Shall I raise a `ValueError('`y` should be nonnegative')` for the same?

integration_tests/test_math.py
```diff
@@ -72,13 +72,22 @@ def test_ldexp():
...
 def test_fabs():
-    i: f64
-    j: f64
+
+
+    # i: f64
+    # j: f64
+    # i = fabs(10.3)
+    # j = fabs(-10.3)
+    # assert abs(i - j) < eps
```
czgdp1807, line 81: Are you planning uncomment these tests?

integration_tests/test_math.py
```diff
@@ -72,13 +72,22 @@ def test_ldexp():
...
     eps: f64
     eps = 1e-12
-    i = fabs(10.3)
-    j = fabs(-10.3)
-    assert abs(i - j) < eps
+
+    assert abs(fabs(10) - fabs(-10)) < eps
+    assert abs(fabs(10.3) - fabs(-10.3)) < eps
+    assert fabs(0.5) == 0.5
+    assert fabs(-5) == 5
```
czgdp1807, line 88: I see only constant values being passed as inputs. Can we pass things like variables, a complicated expression of variables, expressions like, `fabs(fabs(a + fabs(b)))`?

src/runtime/math.py
```diff
@@ -215,25 +227,68 @@ def fabs(x: f64) -> f64:
...
     """
     Return `x` raised to the power `y`.
     """
     return x**y
```
czgdp1807, line 277:
```suggestion
    assert y >= 0
    return x**y
```

Use only if `ValueError` fails to work.

src/runtime/math.py
```diff
@@ -215,25 +227,68 @@ def fabs(x: f64) -> f64:
...
     return x
 
+@overload
+def fabs(x: i64) -> i64:
+    """
+    Return the absolute value of `x`.
+    """
+    if x < 0:
+        return -x
+    return x
```
czgdp1807, line 237: Might not be the best way to implement but fine for now.

integration_tests/test_math.py
Madhav2310: No, I was going to remove them in the next commit after your review. I'll do that now

integration_tests/test_math.py
Madhav2310: Lets try this too!

src/runtime/math.py
Madhav2310: I just noticed the mistake, correcting it it in the next commit

Madhav2310:
However, I wasn't able to check if the value error works... Got the following style suggestion/error. 
```
 --> /home/madhav/lpython/src/runtime/try.py:2:7
  |
2 | print(pow(10,-4))
  |       ^^^^^^^^^^ '**' could be used instead


```

Madhav2310:
@czgdp1807 @namannimmo10 Is this good to be merged? How should I improve upon it?

src/runtime/math.py
```diff
@@ -215,25 +227,76 @@ def fabs(x: f64) -> f64:
...
     return x
 
+@overload
+def fabs(x: i64) -> i64:
+    """
+    Return the absolute value of `x`.
+    """
+    if x < 0.0:
+        return -x
+    return x
```
namannimmo10, line 237: This should return a float, no?
```python
>>> import math
>>> math.fabs(4)
4.0
>>> import numpy as np
>>> np.fabs(4)
4.0
```

integration_tests/test_math.py
```diff
@@ -72,13 +72,21 @@ def test_ldexp():
...
+    i = -10.5
+    j = -7.0
+    assert fabs(i + fabs(j)) == 3.5
+    assert abs(fabs(10) - fabs(-10)) < eps
+    assert abs(fabs(10.3) - fabs(-10.3)) < eps
+    assert fabs(0.5) == 0.5
+    assert fabs(-5) == 5.0
+
+
+
```
namannimmo10, line 89:
```suggestion
```

src/runtime/math.py
Madhav2310: Yes, it was returning me float value only when I ran it locally

integration_tests/test_math.py
```diff
@@ -72,13 +72,18 @@ def test_ldexp():
...
-    assert abs(i - j) < eps
+    i: f64
+    j: f64
+    i = -10.5
+    j = -7.0
+    assert fabs(i + fabs(j)) == 3.5
+    assert abs(fabs(10) - fabs(-10)) < eps
+    assert abs(fabs(10.3) - fabs(-10.3)) < eps
+    assert fabs(0.5) == 0.5
+    assert fabs(-5) == 5.0
```
namannimmo10, line 86: It seems to me that you are not testing these functions properly. `fabs(x: i64)` according to your implementation below returns an integer. You should try the runtime values (non-constant values):

```diff
diff --git a/integration_tests/test_math.py b/integration_tests/test_math.py
index 6e17c5bd1..970c569cb 100644
--- a/integration_tests/test_math.py
+++ b/integration_tests/test_math.py
@@ -83,7 +83,9 @@ def test_fabs():
     assert abs(fabs(10) - fabs(-10)) < eps
     assert abs(fabs(10.3) - fabs(-10.3)) < eps
     assert fabs(0.5) == 0.5
-    assert fabs(-5) == 5.0
+    _i: i64
+    _i = 5
+    assert fabs(i) == 5.0
```
This test will not pass.

src/runtime/math.py
namannimmo10: This is what I get with this PR:
```console
from math import fabs

def main():
  i: i32
  i = 4
  print(fabs(i))
  print(fabs(4))
  print(fabs(-4))

main()
4.0
4.0
4.0
4
4
4
```

src/runtime/math.py
```diff
@@ -215,25 +227,76 @@ def fabs(x: f64) -> f64:
         return -x
     return x
 
+@overload
+def fabs(x: i64) -> i64:
```
czgdp1807, line 231:
```suggestion
def fabs(x: i64) -> f64:
```

Shoudln't we change the return type to `f64`.

src/runtime/math.py
czgdp1807: Have you tried `float(x)` instead of `x`?

src/runtime/math.py
```diff
@@ -215,25 +227,76 @@ def fabs(x: f64) -> f64:
...
+    return x**y
 
-def pow(x: f64, y: f64) -> f64:
+@overload
+def pow(x: i32, y: i32) -> i64:
     """
     Return `x` raised to the power `y`.
     """
+    if y < 0:
+        raise ValueError('y should be nonnegative')
```
czgdp1807, line 284: Add a test for this error in https://github.com/lcompilers/lpython/tree/main/tests/errors

src/runtime/math.py
```diff
@@ -215,25 +227,76 @@ def fabs(x: f64) -> f64:
...
+    return x**y
 
+@overload
+def pow(x: f32, y: f32) -> f64:
+    """
+    Return `x` raised to the power `y`.
+    """
+    if y < 0:
+        raise ValueError('y should be nonnegative')
```
czgdp1807, line 275: Add a test for this error in https://github.com/lcompilers/lpython/tree/main/tests/errors

src/runtime/math.py
```diff
@@ -215,25 +227,76 @@ def fabs(x: f64) -> f64:
...
-def exp(x: f64) -> f64:
+@overload
+def pow(x: i64, y: i64) -> i64:
     """
-    Return `e` raised to the power `x`.
+    Return `x` raised to the power `y`.
     """
-    return e**x
+    if y < 0:
+        raise ValueError('y should be nonnegative')
```
czgdp1807, line 266: Add a test for this error in https://github.com/lcompilers/lpython/tree/main/tests/errors

src/runtime/math.py
```diff
@@ -215,25 +227,76 @@ def fabs(x: f64) -> f64:
...
-    return x * (2**i)
+# pow
+# supported data types: i32, i64, f32, f64
+@overload
+def pow(x: f64, y: f64) -> f64:
+    """
+    Return `x` raised to the power  `y`.
+    """
+    if y < 0:
+        raise ValueError('y should be nonnegative')
```
czgdp1807, line 256: Add a test for this error in https://github.com/lcompilers/lpython/tree/main/tests/errors

src/runtime/math.py
Madhav2310: Please see #599. I was getting errors when doing this.

czgdp1807: You should return `float(x)` instead of `x`. Since you are returning an integer from a function which is supposed to return a floating point, the error in #599 is bound to happen.

src/runtime/math.py
```diff
@@ -207,33 +207,114 @@ def radians(x: f64) -> f64:
...
+    return float(x)
 
+@overload
+def fabs(x: i32) -> f32:
+    """
+    Return the absolute value of `x`.
+    """
+    if x < 0.0:
+        return -float(x)
+    return float(x)
```
czgdp1807, line 246: Can `float(x)` return `f32` or `f64` depending on the kind of `x`? Or does it always return `f64`? @certik

czgdp1807: Well seems like it always returns `f64`,

https://github.com/lcompilers/lpython/blob/4163cd7d4fc3dcca1dc3eb8132ac6d4ce0761744/src/lpython/semantics/python_ast_to_asr.cpp#L3282-L3290

Madhav2310: Shall I report it as an issue? @czgdp1807

czgdp1807: @certik What do you think is the best way to do casting using `float`?

1. Retain the kind but change the type. That would mean `i32` will be converted to `f32`, `i64` to `f64` and like that. 
2. Always convert to `f64`. Then to convert `i32` to `f32` we will be compelled to use tricks like,

```python
result: f32 = float(x)
return result
```

certik: I think `float` has to always return f64, as default floating point numbers in CPython (in fact the only floating point numbers) are f64.

Then we already have automatic casting if you do:
```
a: f64
b: f32
a = 3.4444
b = a
```
Then `a` is f64, and in `b = a` it gets cast from f64 to f32.

So if you do: `result: f32 = float(x)`, then `float(x)` returns f64, and then it gets cast to f32.


In ASR `float` will also be a cast, `IntegerToReal 64`, then there will be another cast `RealToReal 64->32`. We can add an optimization pass that will "fuse" these two into just one "IntegerToReal 32". So that in Release mode there is no performance hit with the above design.

czgdp1807: I see. @Madhav2310 Go with the idea in point 2 of [my comment](https://github.com/lcompilers/lpython/pull/581#discussion_r903316579) and update this PR accordingly.

Madhav2310:
To add the reference json files I run the command `./run_tests.py -u`
Now running the git status, it is showing to have modified a whole lot of reference files(as shown below).

Screenshot from 2022-06-24 05-01-38

Screenshot from 2022-06-24 05-01-42

Screenshot from 2022-06-24 05-01-46

Is this normal? should I commit and push?

Madhav2310:
@czgdp1807 I've made all improvements and corrections as we discussed, could you please look into why GitHub is not checking for conflicts

faze-geek:
> could you please look into why GitHub is not checking for conflicts

The branch has merge conflicts which need to be resolved, it seems that commits from other pull requests have been pushed here ?

czgdp1807:
> I've made all improvements and corrections as we discussed, could you please look into why GitHub is not checking for conflicts

I don't think I can see all the changes that we discussed. `tests/reference/asr-test_factorial-9983e4f.stderr` still has wrong error. We overloaded `factorial` for `i64` as well I think. See `madhav2` branch in my fork for hints.

PR #584, opened 2022-06-12
Get CPython C interop working with custom C code

certik:
We build it as a shared library and pass in the name using the
LPYTHON_PY_MOD environment variable. If it is empty, we will use the
LPython runtime library.

It discovered a bug in our test, we have to test single precision only
up to eps 1e-6, not 1e-12. We fixed the test.

Depends on #583.


certik:
A little tricky to get working on all platforms, but it looks like that this will do the job for now. We can simplify later.

PR #586, opened 2022-06-13
Add C interop test for pointers

certik:
Initial code to test C interoperability of pointers.

### DIFF
--- a/integration_tests/test_c_interop_03.py
+++ b/integration_tests/test_c_interop_03.py
@@ -1,14 +1,16 @@
-from ltypes import ccall, f32, f64, i32, i64, CPtr, pointer, c_pp
+from ltypes import ccall, f32, f64, i32, i64, CPtr, pointer, Pointer, p_c_pointer
 
 @ccall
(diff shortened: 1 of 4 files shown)


integration_tests/test_c_interop_03.py
```diff
@@ -0,0 +1,14 @@
...
+@ccall
+def f_pi32_i32(x: CPtr) -> i32:
+    pass
+
+def test_c_callbacks():
+    xi32: i32
+    xi32 = 3
+    p: CPtr
+#    p_c_pointer(p, xi32)
+#    assert f_pi32_i32(p) == 4
```
certik, line 12: @czgdp1807 I think we need to get this one working. The `p_c_pointer` would behave like CLoc. Then we'll be able to call C functions that accept a pointer, using CPtr as the declaration.

certik:
Tests pass, so I am going to go ahead and merge it, as a start.

integration_tests/test_c_interop_03.py
certik: Fixed in #587.

PR #590, opened 2022-06-13
Support passing arrays to C code

czgdp1807:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2627,7 +2627,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
             target_type, nullptr, "call_arg_value_ptr");
         builder->CreateStore(llvm_tmp, target);
         llvm_tmp = target;
-        return llvm_tmp;
+        return builder->CreateLoad(llvm_tmp);
```
czgdp1807, line 2630: This load is necessary because if you don't do it then the address of `call_arg_value_ptr` is passed and not the address of the variable/array which we are interested in. I am not sure why we needed creation of a variable `call_arg_value_ptr`, putting the address of our variable of interest into it then passing it to the function. Related - https://github.com/lcompilers/lpython/pull/587/files#r895409003

Bottomline - I think `call_arg_value_ptr` is not needed and we can remove the same.

@certik

integration_tests/test_c_interop_03.py
```diff
@@ -20,7 +20,8 @@ def test_c_callbacks():
     xi32: i32
     xi32 = 3
     p: CPtr
-    p_c_pointer(xi32, p)
+    p_c_pointer(pointer(xi32), p)
+    print(pointer(xi32), p)
```
czgdp1807, line 24: Its output verifies whether the addresses were correctly stored. Having these prints helps a lot to verify our implementations.

src/libasr/codegen/asr_to_llvm.cpp
certik: Yes, that's better. Let's implement few more cases and tests. Then we can meet and discuss, I think there must be a way to clean this up.

certik:
Go ahead and merge it.

src/libasr/codegen/asr_to_llvm.cpp
czgdp1807: Commented the logic for `call_arg_value_ptr` for now. If needed in future we can uncomment it.

PR #591, opened 2022-06-13
Fix integer casting of strings.

### LINKED ISSUE #582: Casting string to int type in Lpython
Does Lpython have a separate implementation of string to integer in any of the builtin.py files ? It seems to support this but is only paritally correct. 
```
print(int('10'))
10
(lp) anurag-3:lpython user$ ./src/bin/lpython try.py
(lp) anurag-3:lpython user$ ./a.out
10
```

```
print(int('abc'))
  File "/Users/user/lpython/try.py", line 2, in <module>
    print(int('abc'))
ValueError: invalid literal for int() with base 10: 'abc'
(lp) anurag-3:lpython user$ ./src/bin/lpython try.py
(lp) anurag-3:lpython user$ ./a.out
0    # Deserves a sematic error I guess !
```

faze-geek:
Fix #582

```
print(int('1'))
1
1

print(int('1.2'))
ValueError: invalid literal for int() with base 10: '1.2'
semantic error: invalid literal for int() with base 10: '1.2'          # Returns 0 on master

print(int('1abc'))
ValueError: invalid literal for int() with base 10: '1abc'
semantic error: invalid literal for int() with base 10: '1abc'         # Returns 0 on master
 
```


certik:
Thanks for the fix. Let's add some tests that test this error message --- just put them into `tests/error`.

After that I might have some ideas how to improve the code, but let's do that after we have tests.

faze-geek:
I have added a new file in `tests/errors` for now , I'll change the location if needed.

certik:
Perfect, now add it to `tests/tests.toml`, and enable just the `asr` test. Then run all tests with `./run_tests.py -u` and commit the files into git.

tests/errors/test_str_to_int.py
```diff
@@ -0,0 +1,4 @@
+def test_e1():
+    print(int('3abc'))
```
namannimmo10, line 2: Please also add a separate file under `integration_tests`. And make this work too:
```python
assert int("-423") == -423
```

faze-geek: Thanks for pointing this case out !

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3037,8 +3037,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                            default :
+                                if (chr < '0' || chr > '9')
+                                    throw std::string(""); break;
+                        }
+                        ch++;
+                    }
+                    ival = std::stoi(c);
+                } catch(...) {
+                    throw SemanticError("invalid literal for int() with base 10: '"+ std::string(c) + "'", arg->base.loc);
+                }
```
certik, line 3060: Overall this PR looks good, tests are good, it fixes a real issue. My last comment is to the code here --- we should not use `catch(...)` similar as in Python you should not use a catch all `except`. We should not use exceptions at all in a code like this. Rather let's rewrite this. You can use `break` to exit a while loop.

Would you please mind improving the code above? 

After that I think we can merge this.

namannimmo10:
You need to register `test_str_to_int` in https://github.com/lcompilers/lpython/blob/main/integration_tests/CMakeLists.txt.

faze-geek:
> You need to register `test_str_to_int` in https://github.com/lcompilers/lpython/blob/main/integration_tests/CMakeLists.txt.

`RUN(NAME test_str_to_int    LABELS cpython llvm)` is this as expected ?

certik:
Yes. Just enable the backends that work.

src/lpython/semantics/python_ast_to_asr.cpp
czgdp1807: You can use `if` condition to raise a `SemanticError`.

faze-geek: I have used `if` statements now and it passes all tests locally ,kindly check it out .

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3184,8 +3184,24 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             if (ASRUtils::expr_value(arg) != nullptr) {
                 char *c = ASR::down_cast<ASR::StringConstant_t>(
                                     ASRUtils::expr_value(arg))->m_s;
+                int ival = 0;
+                char *ch = c;
+                if ((ch)[0] == '-') {
```
certik, line 3189:
```suggestion
                if (*ch == '-') {
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3184,8 +3184,24 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                    ch++;
+                }
+                while (*ch) {
+                    char chr = *ch;
+                    if (chr == '.') {
```
certik, line 3194:
```suggestion
                    if (*ch == '.') {
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3184,8 +3184,24 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                        throw SemanticError("invalid literal for int() with base 10: '"+ std::string(c) + "'", arg->base.loc);
+                    }
+                    if (chr < '0' || chr > '9') {
```
certik, line 3197:
```suggestion
                    if (*ch < '0' || *ch > '9') {
```

certik:
Much simpler, than you. I left a few suggestions how to further improve it.

faze-geek:
> Much simpler, than you. I left a few suggestions how to further improve it.

Thanks !

certik:
I think this looks pretty good now, thanks!

Before we merge, do you think you could please clean up the git history? Just create a few nice commits, and do not merge master, just rebase on top of master (main). If you don't know how to do it, let me know, I can help.

faze-geek:
>  If you don't know how to do it, let me know, I can help.

I went through some online resources to squash commits into 1 and did that for the last 3 commits . I can't figure out why for the prior commits when I use `git rebase -i HEAD~11` I get all commits (even from other Prs). Could you help how to work around this or I could `reset --hard` and just make 1 commit if that is suitable .

certik:
You typically do `git rebase -i main`. It's possible that `git rebase -i HEAD~11` was too deep. If it is ok with you, I'll just squash these commits into one while merging.

PR #603, opened 2022-06-14
Implement cpython support in test_c_interop_03

certik:


certik:
This gets it done, in a "prototype" mode. As we get more experience actually using this, we'll "harden" it.

PR #604, opened 2022-06-14
Enabling C backend for integration_tests

czgdp1807:


certik (approved):
I think this looks good. Thanks!

PR #611, opened 2022-06-15
Initial support for structs at ASR, LLVM levels.

czgdp1807:
**Side notes**

While implementing structs I noticed that `python_ast_to_asr.cpp` is not well implemented. Instead of using recursion and visiting the child nodes, they are handled manually. We should improve upon this later.


namannimmo10:
@czgdp1807 -- thanks for adding this. 🎉

integration_tests/structs_01.py
```diff
@@ -0,0 +1,27 @@
...
+    print(a.y)
+
+def g():
+    x: A
+    x = A(3, 3.3)
+    f(x)
+    # TODO: the above constructor does not initialize `A` in LPython yet, so
+    # the below does not work:
+    #assert x.x == 3
+    #assert x.y == 3.3
```
certik, line 19: You can fix this in a subsequent PR, for now I commented it out.

certik (approved):
Thank you for implementing this!

I pushed in commits that make it work for CPython as well, and commented out asserts that do not work yet. Apparently dataclasses by default require to provide initialization.

integration_tests/structs_01.py
czgdp1807: I will look into this tomorrow.

Issue #608, opened 2022-06-15
Pointer to structs


certik:
After #606 is fixed, we need to support the following use case:
```python
@dataclass
class A:
    x: i32
    y: f32

def f(pa: Pointer[A]):
    print(pa.x)
    print(pa.y)

def g():
    x: A = A()
    px: Pointer[A] = pointer(x)
    px.x = 5
    px.y = 5.5
    f(px)

g()
```

czgdp1807:
Initialisation at declaration is not supported yet. So, instead of the following,

```python
x: A = A()
px: Pointer[A] = pointer(x)
```

you should use the following,

```python
x: A
x = A()
px: Pointer[A]
px = pointer(x)
```

Rest everything works. Should we prioritise supporting initialisation at declaration for structs?

certik:
I think we can, I opened up #626 for the initialization, I've been hitting bugs in this a lot, and our users will as well, so we might as well sit down and fix it properly and robustly.

Issue #607, opened 2022-06-15
Add structs using dataclasses


certik:
We need to support structs using the following syntax:
```python
@dataclass
class A:
    x: i32
    y: f32

def f(a: A):
    print(a.x)
    print(a.y)

def g():
    x: A = A()
    x.x = 5
    x.y = 5.5
    f(x)

g()
```
I think this already works at the ASR level and in LLVM, so the tasks are:

* [x] Implement support for this at the AST->ASR level
* [x] LLVM tests
* [x] CPython tests
* [x] C tests

certik:
The basic features now work and are tested in CPython, LLVM and C.

PR #614, opened 2022-06-16
Support for pointer to structs

czgdp1807:
https://github.com/lcompilers/lpython/issues/608
https://github.com/lcompilers/lpython/issues/607


integration_tests/structs_02.py
```diff
@@ -0,0 +1,15 @@
+from ltypes import i32, f32, dataclass, CPtr, Pointer, c_p_pointer
+
+@dataclass
+class A:
+    x: i32
+    y: f32
+
+@ccallable
```
czgdp1807, line 8: @certik How will we support `ccallable` in CPython?

czgdp1807: I think its not supported in `LPython` as well. What do we mean by `ccallable`? Anything that can be called from `C` code?

certik: Yes. LPython supports it, it works with both LLVM and C. It can't work with CPython, since we can't build the C library that expects this function. The only way to support it from CPython would be to somehow expose LPython as a Python library and use it to generate C code, and then link it, etc. Too complex, we can maybe do that later, but for now we do not support `@ccallable` in CPython.

czgdp1807: I get the following error,

```bash
semantic error: Decorator: ccallable is not supported
  --> integration_tests/structs_02.py:10:1 - 17:12
   |
10 |    def f(a: CPtr) -> None:
   |    ^^^^^^^^^^^^^^^^^^^^^^^...
...
   |
17 |        y = a2.y
   | ...^^^^^^^^^^^^ 


```

czgdp1807: Note the update - https://github.com/lcompilers/lpython/pull/614#discussion_r899118569

integration_tests/CMakeLists.txt
```diff
@@ -169,6 +169,8 @@ RUN(NAME test_unary_minus    LABELS cpython llvm)
 RUN(NAME test_unary_plus     LABELS cpython llvm)
 RUN(NAME test_issue_518      LABELS cpython llvm)
 RUN(NAME structs_01          LABELS cpython llvm)
+RUN(NAME structs_02          LABELS llvm)
+RUN(NAME structs_03          LABELS llvm)
```
czgdp1807, line 173: As far as I can figure out, `structs_02.py` isn't possible with CPython because `ccallable` isn't supported there. For `struct_03.py` I get the following error,

```bash
  File "/Users/czgdp1807/lpython_project/lpython/integration_tests/structs_03.py", line 8, in <module>
    def f(pa: Pointer[A]):
```

@certik Do you know how to fix the above?

certik (approved):
I think this is fine to merge, thanks!!

integration_tests/CMakeLists.txt
certik: CPython can't use `ccallable`, at least in the near future. The only way to make it work is by wrapping LPython itself into Python and calling it to take care of this. Which might not be a bad idea, but it's heavyweight and we should do that later. For now we won't test `ccallable` using cpython.

Regarding the above error, I think `Pointer[A]` does not work in CPython yet. That one we should get working.

Issue #609, opened 2022-06-15
Indexing a C array


certik:
We need to be able to generate the following code in the C backend:
```c
void f(uint16_t *q) {                                                  
    for (size_t i = 0; i < 10; i++) {
        g(&q[i])
    }
}
```
How should the Python code look like?

One idea:
```python
@ccall
def g(a: CPtr) -> None:
    pass

@ccallable
def f(q_void: CPtr) -> None:
    q: Pointer[i16[:]]
    c_p_pointer(q_void, q)
    for i in range(10):
        q2: CPtr
        p_c_pointer(pointer(q[i]), q2)
        g(q2)
```
In LPython we can support this, but in CPython it might not be possible, so maybe we need to change `pointer(q[i])` to `pointer_index(q, i)` or something like that.

czgdp1807:
I will look into this later today.

PR #622, opened 2022-06-17
Supporting C-interop with array indices

czgdp1807:
https://github.com/lcompilers/lpython/issues/609


czgdp1807:
@certik This is ready as well.

certik (approved):
I think this is fine.

integration_tests/bindc_03.py
```diff
@@ -0,0 +1,33 @@
...
+
+@ccall
+def g(a: CPtr, value: i32) -> None:
+    pass
+
+@ccall
+def get_array(size: i32) -> CPtr:
+    pass
+
+@ccallable
```
certik, line 11: Since this is not called from C, it doesn't have to be `@ccallable` I think.

czgdp1807: Yes. I think I just took the example as it is from #609.

czgdp1807: Since its merged do you want `ccallable` to be removed in the next PR.

certik:
Sure, if things work. Those are small changes, right now it's important for me to get these features in.

PR #623, opened 2022-06-17
Add support for more binary operations in the ASR

namannimmo10:
@certik, @czgdp1807  --

These newly added nodes work with mostly integers only. 
It would complicate the code to have a separate `enum` (`integerbinopType`, `realbinopType`, `complexbinopType`) for different types, so one simple way is to only have a `binopType` `enum` and raise an error if the operands are not integers (for bitwise operations). 

Let me know your opinions, and I'll update the rest of the code.


certik:
I would first separate the BinOp, then extend it.

certik:
I think that BinOp is already separated, isn't it? It seems both LFortran and LPython has separate BinOps.

certik:
Regarding separate integerbinopType, realbinopType, complexbinopType vs just one for all, I don't have a strong opinion. It seems the integer one many more operations (all kinds of bit stuff), while the real one and complex one can share it? So it would be just two?

namannimmo10:
> Regarding separate integerbinopType, realbinopType, complexbinopType vs just one for all, I don't have a strong opinion. It seems the integer one has many more operations (all kinds of bit stuff), while the real one and complex one can share it? So it would be just two?

Can you please look into the changes I made with this PR? I feel like it's going to get complex by having separate types -  integerbinopType, realbinopType, complexbinopType, etc. So we can have just one enum, i.e. `binopType` and raise a `SemanticError` if, for instance, the bitwise operation is done on real or complex.

namannimmo10:
It won't be apparent with the code itself, I will try to register some errors messages in `tests/errors`.

certik:
I think that's fine. The PR looks good so far. We can always separate the enum later if we determine that would be better. The current design is to share the enum, so I think your approach is the minimal extension of the current design, so it seems fine.

src/libasr/asr_utils.cpp
```diff
@@ -388,6 +388,15 @@ bool is_op_overloaded(ASR::binopType op, std::string& intrinsic_op_name,
...
             break;
         }
+        case ASR::binopType::Mod: {
+            if(intrinsic_op_name != "~mod") {
+                result = false;
+            }
+            break;
+        }
+        default: {
+            throw LFortranException("Binary operation not supported yet");
```
czgdp1807, line 398:
```suggestion
            throw LFortranException("Binary operation " + std::to_string(op) + " not supported yet");
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3257,6 +3257,9 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 tmp = builder->CreateFPToSI(tmp, type);
                 break;
             };
+            default: {
+                throw CodeGenError("Binary operation not supported yet", x.base.base.loc);
```
czgdp1807, line 3261:
```suggestion
                throw CodeGenError("Binary operation not " + std::to_string(x.m_op) + " supported yet", x.base.base.loc);
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3304,6 +3307,9 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 tmp = builder->CreateCall(fn_pow, {left_val, right_val});
                 break;
             };
+            default: {
+                throw CodeGenError("Binary operation not supported yet", x.base.base.loc);
```
czgdp1807, line 3311:
```suggestion
                throw CodeGenError("Binary operation " + std::to_string(x.m_op) + " not supported yet", x.base.base.loc);
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3369,6 +3375,9 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 }
                 break;
             };
+            default: {
+                throw CodeGenError("Binary operation not supported yet", x.base.base.loc);
```
czgdp1807, line 3379:
```suggestion
                throw CodeGenError("Binary operation " + std::to_string(x.m_op) + " not supported yet", x.base.base.loc);
```

src/libasr/codegen/asr_to_x86.cpp
```diff
@@ -328,10 +328,9 @@ class ASRToX86Visitor : public ASR::BaseVisitor<ASRToX86Visitor>
                 m_a.asm_div_r32(X86Reg::ecx);
                 break;
             };
-            case ASR::binopType::Pow: {
-                throw CodeGenError("Pow not implemented yet.");
-                break;
-            };
+            default: {
+                throw CodeGenError("Binary operation not supported yet");
```
czgdp1807, line 332:
```suggestion
                throw CodeGenError("Binary operation " + std::to_string(x.m_op) + " not supported yet");
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1505,41 +1534,26 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
-        arg1.m_value = left;
-        args.push_back(al, arg1);
-        arg2.loc = right->base.loc;
-        arg2.m_value = right;
-        args.push_back(al, arg2);
-        if (op_name != "") {
-            ASR::symbol_t *fn_mod = resolve_intrinsic_function(x.base.base.loc, op_name);
-            tmp = make_call_helper(al, fn_mod, current_scope, args, op_name, x.base.base.loc);
-            return;
-        }
```
czgdp1807, line 1542: Why is this removed?

namannimmo10: Because now we don't use these implementations: 

https://github.com/lcompilers/lpython/blob/f7a7e6b4383e43f89039e8ae8b83e85e6b1b4c9c/src/lpython/semantics/python_comptime_eval.h#L53-L57

I will remove those in a separate PR.

namannimmo10:
@certik -- this is ready.  I commented out some tests for now because LLVM support for modulo and bitwise left/right shift needs to be added. We can do that in subsequent PRs.

certik:
I would suggest to either implement the missing backend features now, or keep the original functions for now, until the backend is fixed. That way things keep working --- I need the `main` to keep working.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3304,6 +3320,10 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 tmp = builder->CreateCall(fn_pow, {left_val, right_val});
                 break;
             };
+            default: {
+                throw CodeGenError("Binary operator '" + ASRUtils::binop_to_str(x.m_op) + "' not supported yet",
+                    x.base.base.loc);
+            }
```
certik, line 3326: This is the downside of the current approach: I don't think we should support bit operations on floating point numbers. If you want to do that, you should cast it to integer and then do it. So we should remove the "yet", as we do not plan to support it.

In fact, we need to have two kinds of casts: the current `Cast` transforms the type, such as truncating the float to int. We also need a `Transfer` node, that keeps the bits, just reinterprets the type from float to int.

namannimmo10: I removed the "yet" part from the error messages. 

@czgdp1807 and I discussed that we should rename `Transfer` (it's confusing!) to something else... `BitCast` might be a good name.

certik:
I agree. `BitCast` seems a good name.

certik:
We can also rename `Cast` to `Convert` or something like that. I do still like `Cast`, it's just a bit confusing which is which. But if we document that, I think we can have:

* `Cast`: converts a value from one type to another, almost always changing the bit representation/value in the process
* `BitCast`: changes the type of the given value, while preserving the bits exactly

Or something like that.

integration_tests/expr_05.py
```diff
@@ -23,9 +23,12 @@ def main0():
...
     assert test_mod(a, b) == 30643
-    a = -123282374
-    b = 32771
-    assert test_mod(a, b) == 2128
+    a = -5345
+    b = -534
+    assert a % b == -5
+    # a = -123282374
+    # b = 32771
+    # assert test_mod(a, b) == 2128
```
namannimmo10, line 31: This PR is almost ready to go in. We just have to make this work.
`a % b` gives `-30643` in C++ and that is the same result we get from the current implementation. Meanwhile Python gives `2128`.

namannimmo10: So it seems like it depends on how various programming languages implement it.

namannimmo10: If you have any suggestions, please let me know.

certik: Yes, that's right. Python uses a certain variant, they all differ for negative numbers, but agree for positive numbers. It's actually tricky to agree on one in ASR. We could add two, say Mod1, Mod2 (there might be a better name), and use one for LFortran, another one for LPython. If we add a C++ frontend, it would use one of the two.

It might be the easiest to just keep calling the manual function that we already have in LPython, for now.

namannimmo10:
This is ready. I'll address https://github.com/lcompilers/lpython/pull/623#discussion_r901078875 in the next PR.

integration_tests/expr_05.py
```diff
@@ -35,19 +38,24 @@ def main0():
     assert -105346 ^ -32771 == 72579
     assert 10 >> 1 == 5
     assert 5 << 1 == 10
-    i1: i32 = 10
-    i2: i32 = 4
-    assert i1 << i2 == 160
-    assert i1 >> i2 == 0
```
certik, line 41: This does not work?

namannimmo10: Everything works, I just moved the tests below.

certik: I just tested it manually, it seems to work. Let's keep this test there, I want to make sure we do not break anything accidentally.

certik: Fixed in f169f16bc6645dfc6bff2c0f5b80b9e4785ba9f2.

certik: I see -- you moved the tests down. I restored the original order, just to make it easy to review, to see that nothing was lost by a mistake (see my updated fix in 0dffbd6fe7fdbd3810f1ce1d544b51dadb7343b6).

If you like the new order better, let's submit a new PR after this is merged.

certik (approved):
I think that this looks good now, thanks!

We can further polish things up in subsequent PRs.

certik:
It seems we are hitting some URL errors at the CI. Let me try to rerun it.

certik:
I forgot to update tests after modifying the file, it should work now.

PR #630, opened 2022-06-19
Add test for global pointer

Smit-create:


integration_tests/CMakeLists.txt
```diff
@@ -115,6 +115,7 @@ endmacro(RUN)
 # Test zero and non-zero exit code and assert statements
 RUN(NAME bindc_01            LABELS llvm c)
 RUN(NAME bindc_02            LABELS llvm c)
+RUN(NAME bindc_04            LABELS llvm)
```
czgdp1807, line 118: Let's remove it from here as `bindc_04` won't work with LLVM backend right now.

integration_tests/bindc_04.py
```diff
@@ -0,0 +1,21 @@
...
+x: Pointer[i16[:]]
+
+def f():
+    yptr1: Pointer[i16[:]]
+    y: i16[2]
+    y[0] = 1
+    y[1] = 2
+    yptr1 = pointer(y)
+    assert yptr1[0] == 1
+    assert yptr1[1] == 2
```
czgdp1807, line 13: Using `ArrayRef` directly in `assert` doesn't work yet. I will fix this next week.

integration_tests/bindc_04.py
```diff
@@ -0,0 +1,21 @@
...
+    x = pointer(y)
+
+def check():
+    f()
+    assert x[0] == 1
+    assert x[1] == 2
```
czgdp1807, line 19: +1.

P.S. No changes needed in the test.

czgdp1807:
`bindc_04` works with LLVM. Surprising.

integration_tests/CMakeLists.txt
czgdp1807: It works. Ignore.

czgdp1807 (approved):

czgdp1807:
@certik This looks good to me. Any thoughts from your end.

certik (approved):
I think this is fine.

Issue #577, opened 2022-06-11
Improving integrations tests


Smit-create:
Since we have recently added multiple backend integration testing, there are some improvements needed as following:
1. It should rebuild from scratch every time we run the script.
2. It should clean up all the generated files before exiting.

cc @certik @czgdp1807

certik:
I would probably leave the generated files, I found it very useful for debugging and to see what the C backend generates. But I would automatically clean it before running, currently I have to manually do `git clean -dfx` in the integration_test directory.

One option is to build it in a "build" directory, then all we need is to remove the build directory before a run. That will force to rebuild from scratch.

czgdp1807:
I think in LFortran we do integration testing outside the source directory. So it should be able to do in LPython as well. The scene is that right now CMake doesn't use relative path to integration test files,

```bash
(lp) 19:30:59:~/lpython_project/lpython % cd ../build-lpython-llvm
(lp) 19:31:04:~/lpython_project/build-lpython-llvm % cmake -DKIND=llvm ../lpython/i
ntegration_tests


Configuration results
---------------------
C compiler      : /Library/Developer/CommandLineTools/usr/bin/cc
Build type: Debug
C compiler flags      : -g
Installation prefix: /usr/local
KIND: llvm
LPYTHON_RTLIB_DIR: /Users/czgdp1807/lpython_project/lpython/src/runtime/impure
LPYTHON_RTLIB_LIBRARY: /Users/czgdp1807/lpython_project/lpython/src/runtime/liblpython_runtime_static.a
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/czgdp1807/lpython_project/build-lpython-llvm
(lp) 19:31:20:~/lpython_project/build-lpython-llvm % make
[  1%] Generating bindc_01.o
Traceback (most recent call last):
  File "/Users/czgdp1807/lpython_project/lpython/src/bin/../runtime/lpython_parser.py", line 8, in <module>
    input = open(filename).read()
FileNotFoundError: [Errno 2] No such file or directory: 'bindc_01.py'
The command 'python /Users/czgdp1807/lpython_project/lpython/src/bin/../runtime/lpython_parser.py bindc_01.py bindc_01.pyC57BB51' failed.
make[2]: *** [CMakeFiles/bindc_01.dir/build.make:74: bindc_01.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:166: CMakeFiles/bindc_01.dir/all] Error 2
[...]

PR #632, opened 2022-06-19
Allow to run integration tests outside source directory

Smit-create:
https://github.com/lcompilers/lpython/issues/577#issuecomment-1152933194

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -59,8 +59,8 @@ macro(RUN)
         if (KIND STREQUAL "llvm")
             add_custom_command(
                 OUTPUT ${name}.o
-                COMMAND lpython -c ${name}.py -o ${name}.o
-                DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
+                COMMAND lpython -c ${CMAKE_SOURCE_DIR}/${name}.py -o ${name}.o
+                DEPENDS ${CMAKE_SOURCE_DIR}/${name}.py
                 VERBATIM)
             add_executable(${name} ${name}.o ${RUN_EXTRAFILES})
             set_target_properties(${name} PROPERTIES LINKER_LANGUAGE C)
@@ -75,8 +75,8 @@ macro(RUN)
         elseif(KIND STREQUAL "c")
             add_custom_command(
                 OUTPUT ${name}.c
-                COMMAND lpython --show-c ${name}.py > ${name}.c
-                DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
+                COMMAND lpython --show-c ${CMAKE_SOURCE_DIR}/${name}.py > ${name}.c
+                DEPENDS ${CMAKE_SOURCE_DIR}/${name}.py
                 VERBATIM)
             add_executable(${name} ${name}.c ${RUN_EXTRAFILES})
             set_target_properties(${name} PROPERTIES LINKER_LANGUAGE C)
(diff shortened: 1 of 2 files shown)


integration_tests/CMakeLists.txt
```diff
@@ -59,8 +59,8 @@ macro(RUN)
         if (KIND STREQUAL "llvm")
             add_custom_command(
                 OUTPUT ${name}.o
-                COMMAND lpython -c ${name}.py -o ${name}.o
-                DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
+                COMMAND lpython -c ${CMAKE_SOURCE_DIR}/${name}.py -o ${name}.o
```
certik, line 62: Isn't `CMAKE_CURRENT_SOURCE_DIR` equivalent to `CMAKE_SOURCE_DIR` in this case, because we do not use `add_subdirectory`?

Smit-create: Yeah, they are same.

integration_tests/CMakeLists.txt
```diff
@@ -100,7 +100,7 @@ macro(RUN)
 
             add_test(${name} python ${CMAKE_CURRENT_BINARY_DIR}/${name}.py)
             set_tests_properties(${name} PROPERTIES
-                ENVIRONMENT "PYTHONPATH=${CMAKE_SOURCE_DIR}/../src/runtime/ltypes;LPYTHON_PY_MOD_NAME=${PY_MOD};LPYTHON_PY_MOD_PATH=${CMAKE_CURRENT_BINARY_DIR}")
+                ENVIRONMENT "PYTHONPATH=${CMAKE_SOURCE_DIR}/../src/runtime/ltypes;LPYTHON_PY_MOD_NAME=${PY_MOD};LPYTHON_PY_MOD_PATH=${CMAKE_SOURCE_DIR}")
```
certik, line 103: I think this does not seem correct. The `PY_MOD` is the shared library that is compiled from hand written C, and so it is compiled in the binary directory, not in the source directory. The `LPYTHON_PY_MOD_PATH` must point to it, thus it should be the `${CMAKE_CURRENT_BINARY_DIR}`.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -96,7 +98,13 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
     Result<std::string> rinfile = get_full_path(infile0, rl_path, ltypes,
         numpy);
     if (!rinfile.ok) {
-        err("Could not find the module '" + infile0 + "'", loc);
+        if (getenv("LPYTHON_PY_MOD_PATH") != NULL) {
+            rinfile = get_full_path(infile0, getenv("LPYTHON_PY_MOD_PATH"), ltypes,
+                numpy);
+        }
```
certik, line 104: The `LPYTHON_PY_MOD_PATH` is only used for CPython tests, and thus should not be needed here, as LPython does not use this environment variable.

Smit-create: Then, how should we find the modules that are present in the current directory of tests? CPython does that itself.

certik:
Here is how to get it working: #634.

src/lpython/semantics/python_ast_to_asr.cpp
certik: Please open up an issue for this --- we have to find all imported modules relative to the file they are imported from, not relative to the current directory.

Smit-create:
I see, thanks!! Should I close this one?

certik:
Sure. There are quite a few things we need to fix to get #634 working, but we can just keep fixing them until things pass.

PR #636, opened 2022-06-19
Do not generate the main program if it is empty

certik:


certik:
I am not sure if this is a good approach, because now this fails:
```console
$ cat a.py
def f():
    print("OK")
$ lpython a.py
Undefined symbols for architecture arm64:
  "_main", referenced from:
     implicit entry/start for main executable
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
The command 'cc -o a.out a.out.tmp.o  -L"/Users/ondrej/repos/lpython/src/bin/../runtime" -Wl,-rpath,"/Users/ondrej/repos/lpython/src/bin/../runtime" -llpython_runtime -lm' failed.
```
While previously it was working and producing an (empty) executable.

certik:
It seems even more messy than it was before.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -654,11 +654,21 @@ R"(#include <stdio.h>
                 }
                 break;
             }
+            case (ASR::cast_kindType::RealToComplex) : {
+                if (is_c) {
+                    headers.insert("complex");
+                    src = "complex(" + src + ")";
+                } else {
+                    src = "std::complex(" + src + ")";
+                }
```
namannimmo10, line 663:
```suggestion
                    headers.insert("complex");
                    src = "CMPLX(" + src + ", 0.0)";
                } else {
                    src = "std::complex<double>(" + src + ")";
                }
```

certik:
The #638 version is better/cleaner, so I need to port it there. I am still fixing some stuff.

certik:
Closing in favor of #638.

PR #653, opened 2022-06-20
Do not use <filesystem>, use a workaround

certik:
This makes LPython compile with gcc 7.5.0.

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -9,7 +9,6 @@
 #include <complex>
 #include <sstream>
 #include <iterator>
-#include <filesystem>
 
 #include <libasr/asr.h>
 #include <libasr/asr_utils.h>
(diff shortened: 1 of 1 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3638,13 +3637,18 @@ std::string pickle_tree_python(AST::ast_t &ast, bool colors) {
     return v.get_str();
 }
 
+std::string get_parent_dir(const std::string &path) {
+    int idx = path.size()-1;
+    while (idx >= 0 && path[idx] != '/' && path[idx] != '\\') idx--;
+    return path.substr(0,idx);
```
Smit-create, line 3643:
```suggestion
    if (idx == -1) return "";
    return path.substr(0,idx);
```

certik: Let's send this as a separate PR, I need a quick fix in now, and we can do this improved one right after.

Smit-create: This will raise an error if `path.substr(0, -1)`.

certik: Yes. Can you please submit a PR with a fix?

Issue #631, opened 2022-06-19
Global Pointer issue with C backend


Smit-create:
On the following python test:
```py
from ltypes import pointer, i16, Pointer

# Testing Global Pointers
x: Pointer[i16[:]]

def f():
    yptr1: Pointer[i16[:]]
    y: i16[2]
    y[0] = 1
    y[1] = 2
    yptr1 = pointer(y)
    assert yptr1[0] == 1
    assert yptr1[1] == 2
    x = pointer(y)

def check():
    f()
    assert x[0] == 1
    assert x[1] == 2

check()
```
The C backend fails because the global pointer gets wrapped inside the main_program function and raises the scope error.

czgdp1807:
Will fix this tomorrow.

PR #655, opened 2022-06-20
Declare global variables globally

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/631


certik:
I think that's fine, thanks!

PR #657, opened 2022-06-21
Implement __LPYTHON_VERSION__ 

### LINKED ISSUE #646: Implement `platform.python_version()`
The issue #645 introduced the `platform` module. This issue is to implement `platform.python_version()` in `src/runtime/platform.py` as follows:
```python
def python_version() -> str:
    return __LPYTHON_VERSION__
```
And we have to implement __LPYTHON_VERSION__ in LPython itself.

faze-geek:
Fix #646

On branch -
```
import platform
print(platform.python_implementation())
print(platform.python_version())

LPython
0.2.0-1041-g196214be3-dirty
```

### DIFF
--- a/integration_tests/test_version.py
+++ b/integration_tests/test_version.py
@@ -1,0 +1,5 @@
+def test_lpython_version():
+    print(__LPYTHON_VERSION__)
+    assert __LPYTHON_VERSION__ == "0.2.0-1041-g196214be3-dirty"
+
+test_lpython_version()
--- a/integration_tests/test_version.py
+++ b/integration_tests/test_version.py
@@ -1,5 +1,6 @@
+from platform import python_version
+
 def test_lpython_version():
-    print(__LPYTHON_VERSION__)
-    assert __LPYTHON_VERSION__ == "0.2.0-1041-g196214be3-dirty"
+    print(python_version())
 
 test_lpython_version()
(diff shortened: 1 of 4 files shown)


integration_tests/test_version.py
```diff
@@ -0,0 +1,5 @@
+def test_lpython_version():
+    print(__LPYTHON_VERSION__)
+    assert __LPYTHON_VERSION__ == "0.2.0-1041-g196214be3-dirty"
```
certik, line 3:
```suggestion
from platform import python_version

def test_lpython_version():
    print(python_version())
```

Since the version changes even between commits for LPython, and it is different for CPython and LPython, we can't test it directly. But as long as it prints, that should be fine.

faze-geek: > Since the version changes even between commits for LPython, and it is different for CPython and LPython, we can't test it directly. But as long as it prints, that should be fine.

Even I thought it was strange that when I was running  `./src/bin/lpython /Users/user/lpython/integration_tests/test_version.py && ./a.out ` and playing around with the assertion test the output was as expected , but it gives an error in the integration tests .

certik: With the new patch in, what error do you get?

The CI right now has some http errors, unrelated to this PR.

faze-geek:
I don't get any errors locally with this diff, the failing check is unrelated to this pr .I guess it could go in .

certik:
Thanks, I think everything works!

Issue #661, opened 2022-06-21
Incorrect results due to outdated compile time values


czgdp1807:
**Code**

```python
from ltypes import i32

def main0():
    i1: i32 = 10
    i2: i32 = 4
    i1 = 3
    i2 = 5
    print(-i1 ^ -i2)
    assert -i1 ^ -i2 == 6

main0()
```

**Output**

```bash
10
AssertionError
```

Reasons for failure,

1. `value` attribute (this is the attribute which stores compile time values) is being set for non-constant variables.
2. This leads to compile time value of `-i1` as `10` and `-i2` as `4`. Hence, `bitwise_xor` is being evaluated at compile time. 
3. If we set `value` attribute as `nullptr` then we get random outputs because there is no implementation of `bitwise_xor` provided which can be used at runtime.

See the original code where `value` attribute of `Variable` ASR node is being set,

https://github.com/lcompilers/lpython/blob/6ccd336621781603e3274441d813b5006f7acb66/src/lpython/semantics/python_ast_to_asr.cpp#L2082-L2144

You can clearly see, `init_expr` is `nullptr`,

https://github.com/lcompilers/lpython/blob/6ccd336621781603e3274441d813b5006f7acb66/src/lpython/semantics/python_ast_to_asr.cpp#L2129

and `value` is filled with initial expression,

https://github.com/lcompilers/lpython/blob/6ccd336621781603e3274441d813b5006f7acb66/src/lpython/semantics/python_ast_to_asr.cpp#L2115

I was compelled to follow this pattern in some of my PRs but now I think we should fix this because this is producing incorrect results.

czgdp1807:
I am fixing this in #662. The only thing left to be done is implementing `bitwise_xor` which can be used at runtime.

czgdp1807:
cc: @cer

Issue #649, opened 2022-06-20
Implement if at the top level


certik:
Currently `if` is implemented in the body visitor, but the global level `if` statement seems to be translated in the symbol table visitor:
```python
def main0():
    print("OK!")

if __name__ == "__main__":
    main0()
```
The error is:
```console
$ lpython a.py 
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/ondrej/repos/lpython/src/bin/lpython.cpp", line 880
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir, compiler_options, time_report);
  File "/Users/ondrej/repos/lpython/src/bin/lpython.cpp", line 354
    r1 = LFortran::LPython::python_ast_to_asr(al, *ast, diagnostics, true,
  File "/Users/ondrej/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 3630
    auto res = symbol_table_visitor(al, *ast_m, diagnostics, main_module,
  File "/Users/ondrej/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2251
    v.visit_Module(ast);
  File "/Users/ondrej/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 1994
    visit_stmt(*x.m_body[i]);
  File "/Users/ondrej/repos/lpython/src/lpython/python_ast.h", line 1862
    void visit_stmt(const stmt_t &b) { visit_stmt_t(b, self()); }
  File "/Users/ondrej/repos/lpython/src/lpython/python_ast.h", line 1741
    case stmtType::While: { v.visit_While((const While_t &)x); return; }
  File "/Users/ondrej/repos/lpython/src/lpython/python_ast.h", line 1874
    void visit_If(const If_t & /* x */) { throw LFortran::LFortranException("visit_If() not implemented"); }
LFortranException: visit_If() not implemented
```
We might be able to leave these to be parsed in the body visitor somehow, or we need to move these into the symbol table visitor.

(The `__name__` function works already, so that's not an issue.)

PR #662, opened 2022-06-21
Convert initialisation at declaration to assignments and fix Variable.value and Variable.m_symbolic_value settings

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/661


certik (approved):
I think that this is fine.

The global variables in Python are effectively just like local (and should be handled like local): see #649 for an example. But let's fix it one step at a time. First inside functions, only later let's tackle the code at the global level.

czgdp1807:
The issue is not with global and local variables. Its related to incorrect setting of `value` attribute of `Variable`. The `Variable` node is same for both global and local variables. Therefore it doesn’t matter. There are only two possible situations, either it is fixed completely or it is not fixed at all. In any case, I think its almost done. If I feel that the fix is getting messy then I will split it. Will let you know shortly.

czgdp1807:
@certik This is ready. I am going to keep this PR for the fixes (as they are clean and simple IMO). Please let me know if any change is required. Thanks.

czgdp1807:
Tests pass.

certik:
Looks good, thanks!

czgdp1807:
Great. Fixing #667 right now. Made some comments on your PR, #666

PR #663, opened 2022-06-21
Parse integers with underscore

certik:


namannimmo10:
Nice!  We can also do this for floats.

certik:
@namannimmo10 go ahead and do that, that would be cool.

PR #666, opened 2022-06-21
Add a test for ignoring a function result

certik:


integration_tests/expr_10.py
```diff
@@ -0,0 +1,10 @@
+from ltypes import i32
+
+def g() -> i32:
+    return 5
+
+def test_fn1():
+    i: i32 = g()
+    g()
```
czgdp1807, line 8: Simply convert it to a `SubroutineCall`?

czgdp1807: A bit tricky. Because using a `Function` in `SubroutineCall` might cause problems in the backend.

certik: I think ASR always has to be well formed. So I think we need to create a temporary variable, and assign to it in ASR. Then there is no problem. We can name the variable "__lpython_dummy", so the (future) ASR->Python backend can detect this and convert it back to the original "no assign" function call.

czgdp1807: Okay. I thought of this approach. It also needs special handling similar to allowing `Function` in `SubroutineCall`. So, either is fine for me. Just an extra variable in ASR and probably in LLVM backend. The only drawback I see is those extra `store` instructions after function calls (accumulates if done inside a loop). But anyways, if user wants to ignore the return value then the proper way is to write a "subroutine" in Python. I will go with dummy variable assignment approach then.

certik: No changes should be needed in any backends.

In Release mode, the LLVM optimizer completely eliminates such stores, so there is no performance overhead.

PR #669, opened 2022-06-21
capitalize and upper functions (String module)

Madhav2310:


Madhav2310:
@czgdp1807 could you please help me fix this error?

integration_tests/test_string.py
```diff
@@ -0,0 +1,22 @@
...
+    i: str
+    i = upper("lpython")
+    assert i == "LPYTHON"
+
+
+def check():
+    test_capitalize()
+    test_upper()
+
+check()
```
czgdp1807, line 22: Add final new line.

integration_tests/test_string.py
```diff
@@ -0,0 +1,22 @@
+from string import (capitalize, upper)
+from ltypes import i32, f64, f32, i64, c32, c64
+
+def test_capitalize():
+    i: str
+    i = capitalize("lpython")
+    assert i == "Lpython"
```
czgdp1807, line 7:
```
i = "lPythoN"
assert capitalize(i) == "Lpython"
```

integration_tests/test_string.py
```diff
@@ -0,0 +1,22 @@
...
+    i: str
+    i = capitalize("development")
+    assert i == "Development"
```
czgdp1807, line 10: Similar for this.

integration_tests/test_string.py
```diff
@@ -0,0 +1,22 @@
...
+
+def test_upper():
+    i: str
+    i = upper("lpython")
```
czgdp1807, line 14: Similar for this.

src/runtime/string.py
```diff
@@ -0,0 +1,29 @@
...
+hexdigits: str = digits + 'abcdef' + 'ABCDEF'
+octdigits: str = '01234567'
+punctuation: str = "#$%&'()*+,-./:;<=>?@[\]^_`{|}~"
+
+def capitalize(s: str) -> str:
+    """
+    Return a copy of the string with its first character capitalized and the rest lowercased.
+    """
+    result: str
+    result = upper(s[0]) + s[1:]
```
czgdp1807, line 16: The rest is not getting lower cased here I think. The rest is just picked up as is.

czgdp1807: May be fill `result` character wise.

czgdp1807:
I think you have to register `captialize`, `upper` and the module `string.py` in https://github.com/lcompilers/lpython/blob/main/src/lpython/semantics/python_comptime_eval.h file. Just look how other functions are registered.

namannimmo10:
You can also maybe include [string constants](https://docs.python.org/3/library/string.html#string-constants).

namannimmo10:
These functions are not available in CPython:
```python
>>> from string import upper
ImportError: cannot import name 'upper' from 'string' (/Users/namannimmo/miniconda3/envs/lp/lib/python3.10/string.py)
>>>
```
That's why the integration tests won't pass. This would work as a method:
```python
>>> "abc".upper()
'ABC'
>>> "abc".capitalize()
'Abc'
>>>
```

Madhav2310:
> I think you have to register `captialize`, `upper` and the module `string.py` in https://github.com/lcompilers/lpython/blob/main/src/lpython/semantics/python_comptime_eval.h file. Just look how other functions are registered.

Should I just add these functions to python_builtin.py? 
On looking more closely these string methods are all of str of python builtin functions: https://docs.python.org/3/library/stdtypes.html#string-methods

The string.py module has wholly different functions : https://docs.python.org/3/library/string.html

Madhav2310:
@czgdp1807 I've tried adding ```upper```, ```lower```, and ```capitalize``` in ```python_comptime_eval.h```. Please review and point me towards the next step

integration_tests/test_builtin_lower.py
```diff
@@ -0,0 +1,9 @@
+from lpython_builtin import (lower)
+from ltypes import i32, f64, f32, i64, c32, c64
+
+def test_lower():
+    i: str
+    i = lower("CoMpIlEr")
+    assert i == "compiler"
+
+test_lower()
```
czgdp1807, line 9:
```suggestion
test_lower()

```

integration_tests/test_builtin_upper.py
```diff
@@ -0,0 +1,9 @@
+from lpython_builtin import (upper)
+from ltypes import i32, f64, f32, i64, c32, c64
+
+def test_upper():
+    i: str
+    i = upper("lpython")
+    assert i == "LPYTHON"
+
+test_upper()
```
czgdp1807, line 9:
```suggestion
test_upper()

```

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -72,7 +75,7 @@ struct PythonIntrinsicProcedures {
         } else {
             throw SemanticError("Function '" + name
                 + "' not found among intrinsic procedures",
-                loc);
+                loc);/home/madhav/lpython/src/lpython/semantics/
```
czgdp1807, line 78:
```suggestion
                loc);
```

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +166,64 @@ struct PythonIntrinsicProcedures {
                 ASRUtils::type_to_str_python(arg_type) + "'", loc);
         }
     }
+    static ASR::expr_t *eval_upper(Allocator &al, const Location &loc, Vec<ASR::expr_t*> &args) {
+        LFORTRAN_ASSERT(ASRUtils::all_args_evaluated(args));
+        if (args.size() != 1) {
+            throw SemanticError("upper() takes exactly one argument (" +
+                std::to_string(args.size()) + " given)", loc);
+        }
+        ASR::ttype_t* str_type = ASRUtils::TYPE(ASR::make_Character_t(al,loc, 1, 1, nullptr, nullptr, 0));
```
czgdp1807, line 175: I think the return type should be same as input type which means you can just do,

```suggestion
        ASR::ttype_t* str_type = ASRUtils::expr_type(args[0]);
```

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +166,64 @@ struct PythonIntrinsicProcedures {
...
+        ASR::expr_t *arg = args[0];
+        ASR::ttype_t* arg_type = ASRUtils::expr_type(arg);
+
+        if (ASRUtils::is_character(*arg_type)) {
+            char* c = ASR::down_cast<ASR::StringConstant_t>(arg)->m_s;
+            std::string s = std::string(c);
+            return ASR::down_cast<ASR::expr_t>(ASR::make_StringConstant_t(al, loc, s2c(al, s), str_type));
```
czgdp1807, line 182: Well nothing happened here. The string just copy pasted from one place to another. I think you should convert all the lower case characters in `s` to upper case and then return `StringConstant` using this converted string.

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +166,64 @@ struct PythonIntrinsicProcedures {
...
+        }
+        else {
+            throw SemanticError("upper() only woks on strings", loc);
+        }
+    }
+        static ASR::expr_t *eval_lower(Allocator &al, const Location &loc, Vec<ASR::expr_t*> &args) {
+        LFORTRAN_ASSERT(ASRUtils::all_args_evaluated(args));
+        if (args.size() != 1) {
+            throw SemanticError("upper() takes exactly one argument (" +
```
czgdp1807, line 191:
```suggestion
            throw SemanticError("lower() takes exactly one argument (" +
```

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +166,64 @@ struct PythonIntrinsicProcedures {
...
+                std::to_string(args.size()) + " given)", loc);
+        }
+        ASR::ttype_t* str_type = ASRUtils::TYPE(ASR::make_Character_t(al,loc, 1, 1, nullptr, nullptr, 0));
```
czgdp1807, line 194: Same as in `upper`.

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +166,64 @@ struct PythonIntrinsicProcedures {
...
+        ASR::expr_t *arg = args[0];
+        ASR::ttype_t* arg_type = ASRUtils::expr_type(arg);
+
+        if (ASRUtils::is_character(*arg_type)) {
+            char* c = ASR::down_cast<ASR::StringConstant_t>(arg)->m_s;
+            std::string s = std::string(c);
+            return ASR::down_cast<ASR::expr_t>(ASR::make_StringConstant_t(al, loc, s2c(al, s), str_type));
+        }
+        else {
+            throw SemanticError("upper() only woks on strings", loc);
```
czgdp1807, line 204:
```suggestion
            throw SemanticError("lower() only woks on strings", loc);
```

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +166,64 @@ struct PythonIntrinsicProcedures {
...
+        }
+    }
+        static ASR::expr_t *eval_capitalize(Allocator &al, const Location &loc, Vec<ASR::expr_t*> &args) {
+        LFORTRAN_ASSERT(ASRUtils::all_args_evaluated(args));
+        if (args.size() != 1) {
+            throw SemanticError("upper() takes exactly one argument (" +
```
czgdp1807, line 210:
```suggestion
            throw SemanticError("capitalize() takes exactly one argument (" +
```

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +166,64 @@ struct PythonIntrinsicProcedures {
...
+                std::to_string(args.size()) + " given)", loc);
+        }
+        ASR::ttype_t* str_type = ASRUtils::TYPE(ASR::make_Character_t(al,loc, 1, 1, nullptr, nullptr, 0));
```
czgdp1807, line 213: Same as in `upper`.

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +166,64 @@ struct PythonIntrinsicProcedures {
...
+        ASR::expr_t *arg = args[0];
+        ASR::ttype_t* arg_type = ASRUtils::expr_type(arg);
+
+        if (ASRUtils::is_character(*arg_type)) {
+            char* c = ASR::down_cast<ASR::StringConstant_t>(arg)->m_s;
+            std::string s = std::string(c);
+            return ASR::down_cast<ASR::expr_t>(ASR::make_StringConstant_t(al, loc, s2c(al, s), str_type));
```
czgdp1807, line 220: Same as in `upper`. Nothing is happening just string getting copy pasted from one place to another.

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +166,64 @@ struct PythonIntrinsicProcedures {
...
+        }
+        else {
+            throw SemanticError("upper() only woks on strings", loc);
```
czgdp1807, line 223:
```suggestion
            throw SemanticError("capitalize() only woks on strings", loc);
```

src/runtime/lpython_builtin.py
```diff
@@ -145,6 +145,93 @@ def str(x: i32) -> str:
...
+    return b
+
+@overload
+def bool(c: c32) -> bool:
+    return c.real != 0.0 or _lfortran_caimag(c) != 0.0
+
+@overload
+def bool(c: c64) -> bool:
+    return c.real != 0.0 or _lfortran_zaimag(c) != 0.0
+
```
czgdp1807, line 234: I think @namannimmo10 recently removed `bool` from this file. I don't remember exactly.

src/runtime/string.py
```diff
@@ -0,0 +1,42 @@
...
+#     result : str
+#     result = ''
+#     char : str
+
+#     for char in s:
+#         if ord(char) >= 65 and ord(char) <=90 :
+#             result += chr(ord(char) + 32)
+#         else :
+#             result += char
+#     return result
```
czgdp1807, line 41: Is the commented code needed here? If not then please remove it.

czgdp1807 (changes requested):

src/lpython/semantics/python_comptime_eval.h
Madhav2310: I assumed we're focussing on type values of input and output. I didn't see any logic being implemented in any of the other eval functions... Could you please elaborate on what exactly should I write/ give a reference to where something similar is being implemented?

integration_tests/test_builtin_capitalize.py
```diff
@@ -0,0 +1,12 @@
+from lpython_builtin import (capitalize)
+from ltypes import i32, f64, f32, i64, c32, c64
+
+def test_capitalize():
+    i: str
+    i = capitalize("lpyThon")
```
namannimmo10, line 6: s: str
s = "lpyThon"
capitalize(s)

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +166,64 @@ struct PythonIntrinsicProcedures {
...
+            throw SemanticError("lower() takes exactly one argument (" +
+                std::to_string(args.size()) + " given)", loc);
+        }
+        ASR::ttype_t* str_type = ASRUtils::expr_type(args[0]);
+        ASR::expr_t *arg = args[0];
+        ASR::ttype_t* arg_type = ASRUtils::expr_type(arg);
+
+        if (ASRUtils::is_character(*arg_type)) {
+            char* c = ASR::down_cast<ASR::StringConstant_t>(arg)->m_s;
+            std::string s = std::string(c);
```
namannimmo10, line 200:
```suggestion
            std::string s = tolower(std::string(c));
```

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -178,7 +178,7 @@ struct PythonIntrinsicProcedures {
         ASR::ttype_t* arg_type = ASRUtils::expr_type(arg);
         if (ASRUtils::is_character(*arg_type)) {
             char* c = ASR::down_cast<ASR::StringConstant_t>(arg)->m_s;
-            std::string s = toupper(std::string(c));
+            std::string s = (int (*)(int))std::toupper(std::string(c));
```
certik, line 181: Why is this needed? I would prefer not to write code like this, unless there is no other way.

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +167,66 @@ struct PythonIntrinsicProcedures {
...
+        }
+        ASR::ttype_t* str_type = ASRUtils::expr_type(args[0]);
+        ASR::expr_t *arg = args[0];
+        ASR::ttype_t* arg_type = ASRUtils::expr_type(arg);
+
+        if (ASRUtils::is_character(*arg_type)) {
+            char* c = ASR::down_cast<ASR::StringConstant_t>(arg)->m_s;
+            std::string s = std::string(c);
+            s[0] = ::toupper(s[0]);
+            std::transform(s.begin()+1, s.end(), s.begin()+1, (int (*)(int))std::tolower);
```
certik, line 222: Let's not do `(int (*)(int))`. If you need a C string, let's write some wrapper callback.

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -163,6 +167,66 @@ struct PythonIntrinsicProcedures {
...
+            throw SemanticError("lower() takes exactly one argument (" +
+                std::to_string(args.size()) + " given)", loc);
+        }
+        ASR::ttype_t* str_type = ASRUtils::expr_type(args[0]);
+        ASR::expr_t *arg = args[0];
+        ASR::ttype_t* arg_type = ASRUtils::expr_type(arg);
+
+        if (ASRUtils::is_character(*arg_type)) {
+            char* c = ASR::down_cast<ASR::StringConstant_t>(arg)->m_s;
+            std::string s = (int (*)(int))std::tolower(std::string(c));
```
certik, line 201:
```suggestion
            std::string s = std::tolower(std::string(c));
```

src/runtime/lpython_builtin.py
certik: Yes, bool should be part of ASR.

certik:
If I understand correctly, this implements both runtime and compile time capitalization? If so, we need to also add tests for both use cases.

Madhav2310:
> If I understand correctly, this implements both runtime and compile time capitalization? If so, we need to also add tests for both use cases.

Yes, as soon as we get this working, I'll add the tests too

Madhav2310:
@czgdp1807 please check now, I think the python_comptime_eval.h code works now. I'm unsure why its shows an error in the complex1 test file, as I didn't make any changes whatsoever there. Thank you!

certik:
@czgdp1807 what is the status of this, was this implemented?

certik:
Obviously it is not finished, so I'll mark it as Draft for now.

czgdp1807:
Yes. I have added `please take over` label to it if someone wants to complete this.

Smit-create:
Just the `upper` method is needed now. Will rebase this.

Harsh-Upadhayay:
Hi, if this issue is still remaining then I'd like to take it over and solve it.
After reading the above conversation I have a vague idea that some of the bulitins of string library were implemented and the changes didn't passed integration tests.
Could someone please summaries **what needs to be done** and what was done.

PR #671, opened 2022-06-21
Get C strings passing from Python to C

certik:
In all CPython, LLVM, C backends, we need some way to pass strings into a function implemented in C.

I got stuck doing it in CPython already, I can't figure out a natural way to do it.

There might be a way via CPtr.

### DIFF
--- a/integration_tests/test_c_interop_02.py
+++ b/integration_tests/test_c_interop_02.py
@@ -53,4 +53,9 @@ def test_c_callbacks():
     xi8 = 3
     assert f_i8_i8(xi8) == 4
 
+    s: str
+    s = "Hello World!"
+    print(f_str_i32(s))
+    assert f_str_i32(s) == 3
+
 test_c_callbacks()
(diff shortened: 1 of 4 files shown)


czgdp1807:
Well I can try for LPython. Don't know if we should expect it to work for CPython. The reason being, CPython doesn't compile, rather it interprets so linking object files never comes into the picture.

czgdp1807:
I am opening a new PR for this. Will add you as a co-author there.

certik:
CPython calls into linked C codes, that already works, see our tests in `integration_tests`.  The `ccall` decorator does it, see `ltypes.py`.

It works for ints, pointers, etc. But not strings yet, that's what we need to get working. I am doing CPython first, so that we know what syntax will work, and then we implement it in LPython. It works in LFortran already, so ASR and backends have the feature, but we need to hook it up in AST->ASR.

czgdp1807:
Well if you want to continue here, then please feel free to do so.

czgdp1807:
Note that LPython already compiles the new test (with strings) and hence everything is already working there. What needs to be done is to make CPython work.

certik:
Yes, CPython is the hard work here. :)

certik:
Closing this in favor of #674.

Issue #656, opened 2022-06-20
ASR passes refactoring


certik:
* register passes and their order
* see ASR printout after every pass, as well as perhaps an equivalent C code or other backends applied
* The backends shouldn't call and apply passes. They should call something like, `pass_applier->apply_passes([list of passes])`. The user should be able to override the `[list of passes]`. If user doesn't provide the list of passes then just apply the list of default passes desired by the backend.

certik:
* Ability to select which passes to run, which order, even more than once.

PR #675, opened 2022-06-22
Initial support for automatic vectorisation of loops

czgdp1807:
I am starting off with a very simple test, where everything is a constant and vectors/arrays of only one dimension. In future we will add more complicated tests with multiple dimensions and variable sized arrays.


integration_tests/vec_01.py
```diff
@@ -0,0 +1,9 @@
+from ltypes import f64
+from numpy import empty
+
+a: f64[9216] = empty(9216)
+b: f64[9216] = empty(9216)
+b[:] = 5.0
+a[:] = b[:]
+
+print(a)
```
czgdp1807, line 9: Does this test make sense to you?

integration_tests/CMakeLists.txt
```diff
@@ -185,6 +185,7 @@ RUN(NAME test_str_to_int     LABELS cpython llvm)
 RUN(NAME test_platform       LABELS cpython llvm)
 RUN(NAME test_vars_01        LABELS cpython llvm)
 RUN(NAME test_version        LABELS cpython llvm)
+RUN(NAME vec_01        LABELS cpython)
```
certik, line 188:
```suggestion
RUN(NAME vec_01              LABELS cpython)
```

integration_tests/vec_01.py
certik: Another example:
```
for i in range(9216):
    A[i] = 4*B[i]+C[i]
```
Turns into:
```
for i in range(0,n,8):
    vector_fma(8, A[i*8:(i+1)*8], 4, B[i*8:(i+1)*8], C[i*8:(i+1)*8])
```

czgdp1807: Well, I avoided the loop because `visit_For` hasn't been implemented yet in AST to ASR transition. I can implement `visit_For` provided if its okay with you. We will just go out of context for a little while.

Other than implementing `visit_For` we will also need to implement a pass which converts `For` into `While` so that LLVM handles only one type of loop all the time.

czgdp1807: Or we can just start with implementing vectorisation in LFortran and then come to LPython. What do you say?

certik: I am remembering now. The next step that we also need is to then unroll the loop, so from:
```
for i in range(0,n,8):
    vector_copy(8, A[i*8:(i+1)*8], B[i*8:(i+1)*8])
```
into:
```
for i in range(...):
    vector_copy(8, A[...], B[...])
    vector_copy(8, A[...], B[...])
    vector_copy(8, A[...], B[...])
    vector_copy(8, A[...], B[...])
    vector_copy(8, A[...], B[...])
    vector_copy(8, A[...], B[...])
    vector_copy(8, A[...], B[...])
    vector_copy(8, A[...], B[...])
```
That is how speed is obtained. So in assembly the loop body is "thick", full of vector instructions.

certik: I think we have for loops already implemented in LPython, and we also have do loops -> while loop ASR pass.

czgdp1807: I am getting the following error,

```bash
  File "/Users/czgdp1807/lpython_project/lpython/src/bin/lpython.cpp", line 842
    return emit_asr(arg_file, runtime_library_dir,
  File "/Users/czgdp1807/lpython_project/lpython/src/bin/lpython.cpp", line 168
    r = LFortran::LPython::python_ast_to_asr(al, *ast, diagnostics, true,
  File "/Users/czgdp1807/lpython_project/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 3752
    auto res = symbol_table_visitor(al, *ast_m, diagnostics, main_module,
  File "/Users/czgdp1807/lpython_project/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2361
    SymbolTableVisitor v(al, nullptr, diagnostics, main_module, ast_overload, parent_dir);
  File "/Users/czgdp1807/lpython_project/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2104
    visit_stmt(*x.m_body[i]);
  File "/Users/czgdp1807/lpython_project/lpython/src/lpython/python_ast.h", line 1862
  File "/Users/czgdp1807/lpython_project/lpython/src/lpython/python_ast.h", line 1738
    case stmtType::AnnAssign: { v.visit_AnnAssign((const AnnAssign_t &)x); return; }
  File "/Users/czgdp1807/lpython_project/lpython/src/lpython/python_ast.h", line 1871
    void visit_For(const For_t & /* x */) { throw LFortran::LFortranException("visit_For() not implemented"); }
LFortranException: visit_For() not implemented
```

czgdp1807: Ah I see. Its implemented in the BodyVisitor.

certik: The error you are getting is due to #649, which is a separate issue.

certik: >  But since we don't have for loops -> while loop ASR pass.

We do: For loops are represented as ASR `DoLoop` node, which then gets transformed to `While` for the LLVM backend.

czgdp1807: Okay. Makes sense.

czgdp1807: Made some changes. Feel free to take a look and let me know if any changes are needed. I will implement a vectorisation pass in the next step.

certik: Looks good!

certik:
Looks good so far.

src/bin/lpython.cpp
```diff
@@ -173,6 +175,51 @@ int emit_asr(const std::string &infile,
...
+            }
+            case (ASRPass::unused_functions) : {
+              LFortran::pass_unused_functions(al, *asr, true);
+                break;
+            }
+            case (ASRPass::loop_vectorise) : {
+                LFortran::pass_loop_vectorise(al, *asr, runtime_library_dir);
+                break ;
+            }
+            default : throw LFortran::LFortranException("Pass not implemened");
```
certik, line 220: I think it's time to implement the ASR manager (https://github.com/lcompilers/lpython/issues/656) so that we don't have to maintain such code that is almost (but not quite) identical between LFortran / LPython.

czgdp1807: I will do that after this.

czgdp1807:
The tests should pass hopefully. @certik Please let me know if you have any comments. Otherwise I will clean up the git history and merge. Thanks.

certik (approved):
I think this looks great to merge. Once you cleanup / are happy with it, go ahead and merge.

certik:
Oops, the CI failed in master, I fixed it in #692.

PR #677, opened 2022-06-22
Integrating Language Server to LPython

### LINKED ISSUE #441: Add VSCode extension / language server
We need to create a VSCode extension in TypeScript that sets up the VSCode Language Server Protocol (LSP) (https://docs.microsoft.com/en-us/visualstudio/extensibility/language-server-protocol), here is how it is done for Fortran:

* https://github.com/fortran-lang/vscode-fortran-support

Then we need to create a server that communicates with this extension using the LSP. Here is an example for Fortran, the server is written in Python:

* https://github.com/gnikit/fortls

We should first create a simple server in Python, and start with linting, and figure out how to do linting over the LSP protocol. And then we'll write a server in LPython itself, in C++, that will communicate via the LSP with the VSCode extension directly.

ankitaS11:
This PR is an attempt to partially fix #441 by implementing a language server support for LPython. Currently, this language server supports two features `didSave` (Draws a line when a document is saved, [for now])  and `documentSymbol` (Symbol lookup with the hardcoded symbols). 

Note: All the meta-data is hard coded for now.

**Steps to reproduce the result:** 

```bash
git clone git@github.com:ankitaS11/LPython.git && cd LPython && git checkout lsp_support
```

Now build the LPython binary, libraries, along with LSP binary. Refer https://github.com/lcompilers/lpython#compile-lpython for instructions.

For the extension part, clone the extension repository and build the extension: (it’s expected that `nodejs` and `npm` are installed on your system)

```bash
git@github.com:ankitaS11/custom-vscode-linter.git && cd custom-vscode-linter
# Mention the path of the binary generated for lsp (lpython/src/libasr/lsp/lsp_bin) as an executable path in the file named `client.ts`.
# ^^^ This is a must!

# Now build the extension
npm install && npm run compile
```

Once installation is done, enter the custom-vscode-linter folder, open VSCode and launch the extension using ctrl + shift + D for debugging mode, else open `extension.ts` file and press ctrl + F5 to run the extension, and select VSCode Extension Development if a dropdown appears.

Now create a new “python” file, just make sure it has more than 10 lines, document symbol lookup is hardcoded from line number 1 to line number 10.  Save the file or try for document symbol lookup by pressing (Ctrl + Shift + O).

This PR is currently WIP. I am now working on taking the necessary information from the LPython compiler instead of hardcoding it but I am open for any kind of suggestions. Thanks to @certik for mentoring me throughout this :)

cc: @certik


certik:
@ankitaS11 awesome, thanks for this! I will try to set this up and play with this very soon and I'll help integrating this into LPython.

@namannimmo10, @czgdp1807 do one of you have time to test this PR out?

czgdp1807:
I might test it tonight on my machine.

czgdp1807:
Is `src/libasr/lsp/nlohmann/json.hpp` a generated file?

certik:
That's a `nlohmann` json, that we should replace with something else, since it is very slow to compile. It's a vendored dependency.

certik:
@namannimmo10, @czgdp1807 just a reminder to please try it out.

czgdp1807:
@Smit-create Could you please try this and let us know your feedback? Thanks.

Smit-create:
> @Smit-create Could you please try this and let us know your feedback? Thanks.

I am using atom editor. I'll install these things soon over the weekend and try to have a look.

namannimmo10:
Sorry for the delay. I will try to play with this soon.

namannimmo10:
Can you also please use a screen recorder in VS Code and show it to us?  It will hardly take 5 mins. Thanks!

certik:
@namannimmo10 the instructions in the PR description should work. Do they not work for you? If so, which step does not work?

Smit-create:
Hi @ankitaS11,
Thanks, this works for me. I believe you have hardcoded this value to `lpyth_symbol`. And that works for me on mac M1.
Screenshot 2022-07-02 at 11 15 11 AM

certik:
Thanks a lot @Smit-create for testing it!

Yes, it's hardcoded, just to get started.

certik:
@ankitaS11 so I think the next step is to integrate this with `lpython language_server` command, and making it use the ASR from LPython, and changing JSON to RapidJSON. Let me know if you know what to do, otherwise I am happy to help.

ankitaS11:
> Hi @ankitaS11, Thanks, this works for me. I believe you have hardcoded this value to `lpyth_symbol`. And that works for me on mac M1.
> 
> Screenshot 2022-07-02 at 11 15 11 AM

Thanks @Smit-create for testing it out. Glad to know it worked! Yes, @certik is right, the value is being hard coded for now and the features are yet to be implemented.

ankitaS11:
> @ankitaS11 so I think the next step is to integrate this with `lpython language_server` command, and making it use the ASR from LPython, and changing JSON to RapidJSON. Let me know if you know what to do, otherwise I am happy to help.

Thanks @certik. I am almost done with integrating the code with RapidJSON. It will hardly take a couple of days more to finish. Regarding the other two tasks, I will update you once I am done with RapidJSON integration..

gnikit:
@certik I don't know how realistic my comment is given the time already invested in this, but I would highly recommend using an existing LSP implementation rather than maintaining your own in C++. LSP implementations require constant upating from the Specification docs and getting the async functionality to work is very tricky. There are few LSP implementations that you can use, with the most robust and widely adopted being Microsoft's own https://github.com/microsoft/vscode-languageserver-node in TypeScript, the second one and my personal choice being https://github.com/openlawlibrary/pygls in Python.

ankitaS11:
> @certik I don't know how realistic my comment is given the time already invested in this, but I would highly recommend using an existing LSP implementation rather than maintaining your own in C++. LSP implementations require constant upating from the Specification docs and getting the async functionality to work is very tricky. There are few LSP implementations that you can use, with the most robust and widely adopted being Microsoft's own https://github.com/microsoft/vscode-languageserver-node in TypeScript, the second one and my personal choice being https://github.com/openlawlibrary/pygls in Python.

Hi, @gnikit - Thank you for taking a look at it. Even though a lot of it will be up to @certik and I am looking forward to his thoughts as well, I thought I'll add mine as well:

1. Most of the programming languages have dedicated language servers implemented which follow the LSP Protocol specifications. (example: `rust-analyzer` for Rust, `ccls`/clangd for C++, _several_ for python, `fortls` for Fortran etc.) The reason I mention this: is because it gives us more control over the library, what we want to offer, and if anything breaks.
2. LFortran/LPython don't have community versions for linters (like fortran has gfortran if I'm not wrong), and other capabilities like for symbol lookup. So these will mostly come from the ASR generated from libasr backend. In any case, we'll have to integrate those capabilities with the LSP protocol, and it's the easiest if LSP is written in C++. Also more efficient.
3. I've seen how `ccls` handles the asynchronous functionality, and that was one of the things I wanted to look at. Yes, it's tricky, but it's something that I had in my mind for future goals.

One con (I believe) on using external libraries like pygls for Language Server would mean, that we will have to have a separate library/server installed & running for a linter and other capabilities. So there are multiple end-points to look at and maintain.

On your concern about catching up with the constant updates on the specification docs, I believe that even makes our job easier if we have the control to fix things. Though, I'm very curious if the LSP protocol is in a stage where things that break all the servers will be introduced frequently.

Pardon me if I was incorrect at any place, and please feel free to correct me if I was. :) Just sharing my thoughts here.

gnikit:
>Most of the programming languages have dedicated language servers implemented

I think that is fairly inaccurate. The overwhelming majority of Language Servers (LS) are coded in TypeScript (46), followed by Java (20), Go (12), C#/C++(11) and Python/Rust (10). Moreover, most LSs use the existing LSP implementation, not defining their own. Here is the list: https://microsoft.github.io/language-server-protocol/implementors/servers/. 

However, all of this is besides the point. What matters is the languages used for the most common LSs. There you will find TypeScript, Java and C# being the main three. For C++ the only major open-source example available is from LLVM https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clangd/Protocol.h, which maybe you can use.

What in theory could be perceived as added flexibility, in practice it will be continuous maintenance costs, which unless you are able to offset by monetising your products like Microsoft, Facebook, IBM, RedHat, etc. you are likely to fall behind in the maintenance cycle, like most maintainers have in the LSP list.

With regards to the rest of the comments: the performance of the LSP implementation will never be an issue, the main bottleneck will be the parser. As for integration, Python, C and C++ can be quite easily integrated, numerous examples exist. 

>using external libraries like pygls for Language Server would mean, that we will have to have a separate library/server installed & running for a linter and other capabilities.

Not really. You register custom capabilities to the server, nothing to do with the LSP implementation. The client should then register that custom capability to listen to the server.

> I'm very curious if the LSP protocol is in a stage where things that break all the servers will be introduced frequently.

That doesn't happen. In short, developers of the servers and the clients are responsible for updating the version of LSP used. If your tests pass and your coverage is decent you are safe to upgrade. Microsoft can't break your server just because you use LSP v3.15 vs their latest release v3.17. If you fall behind in your maintenance cycle, what will happen is that clients will start dropping support for deprecated features, thus making your server unusable.

These are the monthly download stats of the TypeScript modules for LSP, with roughly ~50k users on GitHub. It's unlikely that anyone else will beat Microsoft to the punch when it comes to triage and general fixes.

image

ankitaS11:
Thanks @gnikit - for sharing the stats, however, I guess I was misunderstood there when I said. I’ll make an attempt to clarify this:

> Most of the programming languages have dedicated language servers implemented

What I meant was, LSs exist for those programming languages which do not use custom implementations (I listed a few examples above). Plus, they are used widely and have a fairly good amount of downloads. I know there will be LSs who use custom implementations, maybe more.

But! I pondered more over it today morning, and I’ve started to shift direction in my thoughts. It seems to be that you are on point with the suggestion. Today morning, I looked at this language server for Jakt programming language in the Serenity’s ecosystem: https://github.com/SerenityOS/jakt/blob/main/editors/vscode/server/src/server.ts - and they seem to be using `vscode-languageserver/node`. I agree, that as we move along, things will start looking more complicated than they are right now, especially converting the diagnostics to appropriate JSON format, parsing results properly, and printing to the output stream. I’ve already faced a few issues while porting RapidJSON for the JSON support in this existing language server, though they were solved, it will become another `ccls` in no time, which means it will require a lot of maintenance as people start using it. So I agree with your point. :)

I can understand where you are coming from, since you have had more experience with `fortls` and maybe other language servers that I may not be aware of, and experience counts. Thanks for the discussion, and for sharing your points with me in the open discussion. Even though a lot of time was spent on this, but probably, it will probably be worth porting it as soon as possible to use `vscode-languageserver/node` or pygls, whichever sounds better and more suitable for LPython/LFortran. What do you think, @certik?

certik:
Thanks @gnikit for the feedback. We are using a TypeScript VSCode extension here: https://github.com/ankitaS11/custom-vscode-linter (it's a prototype, we polish as we go) which implements the LSP, and this extension is communicating with LPython to do the actual symbol lookups etc. We currently communicate with LPython using JSON, but we don't need to, it's our own protocol. The TypeScript extension uses `vscode-languageclient`: https://github.com/ankitaS11/custom-vscode-linter/blob/6207d4ae5b8dc5f29004a2910ec5f16be09eb631/src/client.ts#L3 to do the actual LSP communication, which I think is exactly the client you recommended above (https://github.com/microsoft/vscode-languageserver-node).

So it seems we are already doing exactly what you recommended.

gnikit:
Hi @certik, the client side looks great so far and I think @ankitaS11 has made a good start for the client-side. What I was referring to was the LSP implementation for the server-side (LPythonServer.cpp, JSONRPC2Connection.cpp, and their dependencies). You don't have to write your own, they are massive pieces of work to undertake. What I suggested instead was to use an existing interface that offers the construction of the jsonrpc2.0 requests, the server connection and also an LSP requests API for the types and interfaces. Therefore, the only dev work that you will have to do is implement the backend to the server.

----

This is just a short demo of how annoyingly verbose and complicated things get if you have to also maintain the LSP requests and implementations API for the server. 

Both `textDocument/documentSymbol` and `textDocument/completion` requests need to map your ASR nodes to a jsonrpc2.0 request for their corresponding interfaces, be that a Function, a Class, a Variable, etc., however that mapping of kinds (enums) are slightly different, [completionKind](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemKind) and [symbolKind](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#symbolKind) and hence you have to define them separately.

You will have to repeat the above for ALL the parts of the Language Server Protocol you wish to implement. Also, not all LSP types have a straightforward ways of being defined in all programming languages, see the cyclical type definitions of `LSPAny`, `LSPObject` and `LSPArray`: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#lspAny

There are a few more "issues" with implementing your own server-frontend and LSP request interfaces, like hover having to use Markdown or requiring your server jsonrpc2.0 implementation to listen and respond to client responses (not sure if possible with the current design),  but I think my concerns are a bit clearer now.

certik:
TODO:
* Make RapidJSON optional, the language server as well then (similar to LLVM if it is not installed)

certik:
@gnikit Is your comment about the code in here: https://github.com/lcompilers/lpython/pull/677/files#diff-cdaa6118b7d0afb68eb5e0e95da768e2bd8f510423865dee21153603fe37dd3aR74, such as:
```c++

  void serve_document_symbol(rapidjson::Document &request, JSONRPC2Connection& obj, int rid) {
    std::string uri = request["params"]["textDocument"]["uri"].GetString();

    int start_character = 0;
    int start_line = 1;
    int end_character = 10;
    int end_line = 10;

    rapidjson::Document range_object(rapidjson::kObjectType);
    rapidjson::Document::AllocatorType &allocator = range_object.GetAllocator(); 
    range_object.SetObject();

    rapidjson::Document start_detail(rapidjson::kObjectType); 
    start_detail.SetObject();
    start_detail.AddMember("character", rapidjson::Value().SetInt(start_character), allocator);
    start_detail.AddMember("line", rapidjson::Value().SetInt(start_line), allocator);
    range_object.AddMember("start", start_detail, allocator);
...
```
Where we are constructing a custom JSON reply back to our TypeScript VSCode extension that then translates the result and hands it over to Microsoft's LSP? And you suggest instead to "use an existing interface that offers the construction of the jsonrpc2.0 requests"?

Since this code has to be in C++ (as we interact with the ASR from C++), do you know of a library that can make this part of the code easier?

(Note that this is not any LSP JSON, this is our custom JSON that we are free to design in any way we like, we don't even have to use JSON. It's just about communicating the results back to our TypeScript extension.)

gnikit:
> do you know of a library that can make this part of the code easier?

These are all the LSP SDKs available, https://microsoft.github.io/language-server-protocol/implementors/sdks/. That list is fairly small and you will note that most of the SDKs are lagging behind the Microsoft versions. Only those developed by funded organisations, like `pygls` (Open Law Library), `langium` (TypeFox) and `lsp4j` (Eclipse foundation and also one of the creators of LSP) are up to date.

For C++ your best bet is either ccls, see for example: https://github.com/MaskRay/ccls/blob/master/src/lsp.hh
or better yet what LLVM does (not in the list) for `clangd`, see: https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clangd/Protocol.h

Not sure if that is very helpful, since staying in C++ is a requirement, but at least you might be able to draw some inspiration from the projects.

As a side note, I think LLVM would be keen on making their LSP implementation as a more general LSP SDK for C++.

certik:
Ah I see: you are suggesting to simply implement the LSP itself in C++ (using one of the above libraries) instead of TypeScript, so we do not have to maintain the TypeScript extension at all (but don't all VSCode extensions have to be implemented in TypeScript anyway?). The C++ libraries seem to be lagging behind the Microsoft implementation, so I am worried we will run into problems. With the current approach here, we are using the latest Microsoft TypeScript implementation that we know will always work, and all we need is to just communicate the results from LPython into this extension.

So I think here is a good way forward:

* We keep the current approach, to get started. We implement a few basic things, get everything working. That is not too much code.
* Then we can try using some of the C++ libraries above and see if it simplifies the overall code: it might slightly simplify the C++ code we have to maintain, and it should remove the need to "translate" it in TypeScript. If it is overall simpler, and the 3rd library is in Conda (or we can put it there) and easy to install, we can use them. Otherwise if things are not overall simpler, we can keep using our current approach.

certik:
@czgdp1807 that's my understanding as well. We'll go with this design for now, and we can change it later if needed.

gnikit:
Apologies for getting back to this with a delay, just to provide some clarification on a couple of things that were mentioned

@certik 

>you are suggesting to simply implement the LSP itself in C++ (using one of the above libraries) instead of TypeScript...

So there are two sides that you need to think when it comes to LSP, the *server side* and the *client side*.

- the client is in your code editor, that would be your Typescript extension that integrates with VScode, neovim, etc..
- the server is the program that does all the work for the LSP, finding symbols, going to definitions, etc.

The way that these two communicate is through the LSP interface (alternatively known as an implementation) that uses a jsonrpc2.0 syntax. If you are to create a language server you need to serve your server responses into that specific LSP, jsonrpc2.0 syntax. 

My original point was to use a library that has an API for that syntax instead of you defining it, because it will take too much work to keep up to date with the latest LSP version(s). The reason why you need to remain relatively up to date is because the clients (editors and editor extensions) slowly upgrade their LSP versions, so if you stay too behind then your server will not be able to integrate with the editor. 
The other, more obvious, reason why you need to update your LSP implementation is that new functionality is added with every release and your server would benefit from that.

So maintaining your own LSP API **for the server** is perfectly fine, especially when it comes to prototyping, I just thought it would be important to share some of the less obvious problems you will face down the line.

---

@czgdp1807 

> If that's true then C++ is only being used for data transformation. It makes sense to me.

Essentially yes. That transformation (to jsonrpc2.0) can be quite costly to maintain for the entire LSP (or the parts that you end up using) and does not necessarily need to be in C++. I made a schematic to help make things clear

image


- The client implements the *Client Capability* sections from [LSP](https://microsoft.github.io/language-server-protocol/specifications/specification-current)
- The client API implements the *Request* sections from the [LSP](https://microsoft.github.io/language-server-protocol/specifications/specification-current)
- The server API implements the *Response* sections from the [LSP](https://microsoft.github.io/language-server-protocol/specifications/specification-current)
- The server implements the *Server Capability* sections from the [LSP](https://microsoft.github.io/language-server-protocol/specifications/specification-current)


For VS Code your *client* has to be written in Typescript which is convenient since Microsoft also developed all the LSP API (server and client) to be in Typescript. Your server can be in whatever language you want. It makes more sense if it is in Typescript since you would be able to use all the existing libraries, but realistically that's not always possible.

What a *Language Server* usually is, is just the last part of that diagram. That basically translates to 
- registering the *Server Capability* parts of LSP
- and **this is the main one**, generate the information needed for the server *Response*, which is what goes into the LSP API (server)

In theory, you can define your own *LSP API (server)* as long as the final product, that is received by the *client* is a JSONRPC2.0 LSP compliant message; this is what I did with `fortls` and that's the case with `lpython`. What I was suggesting is you use an existing *LSP API (server)* and/or consider doing the front-end of your server in another language like Python. This would allow you to use C++ for all the heavy lifting (parsing and ASR) but have a high level, easy to maintain and extend, interface to the server.

---

Not sure if I answered everything, or if I did a good job at explaining the chaos that is LSP, but if you want, you can open Discussions on the repo and move this conversation there to go into more detail. 

Please feel free to ping me if you need any help and congrats @ankitaS11 on this PR! I am very happy seeing people create better tooling that is open-source.

certik:
I think there is still a point of confusion. Currently this PR implemented the following design:

client (in TypeScript as the VSCode extension) <-> LSP via JSON <-> server (still in TypeScript as the VSCode extension, using Microsoft's LSP library in TypeScript) <-> custom communication <-> LPython C++ server

The "custom communication" part currently is in JSON, but doesn't have to be, and it is not LSP in general. So this design does not have the drawback that we need to keep LPython up-to-date with regards to changes in LSP. LPython can stay, we just need to maintain the TypeScript extension and update it if LSP changes.

The alternative design that you suggested to consider is:

client (in TypeScript as the VSCode extension) <-> LSP via JSON <-> LPython C++ server

Which is simpler, in a way, but now we need to maintain LSP in LPython in C++ itself, either by hand (with the downsides that you mentioned above), or by using some C++ library.

gnikit:
Do you think you can point me to the Typescript server extension, I would be curious to see how you interfaced with C++?

My comments are based on what I was able to find for this PR, so please correct me if I am missing something:
- the LSP [client](https://github.com/ankitaS11/custom-vscode-linter/blob/main/src/client.ts) in Typescript
- the LSP [server](https://github.com/lcompilers/lpython/blob/main/src/libasr/lsp/LPythonServer.cpp) in C++.

From what I can tell the LSP server is fully functional and does not need a Typescript layer.
For example [this](https://github.com/lcompilers/lpython/blob/c11c68ba6a8795afd8c19ea890556792dc378f19/src/libasr/lsp/LPythonServer.cpp#L17-L26) is the LSP server's initialisation. 

Another example is, typing "textDocument/documentSymbol", that is defining your own LSP server API. [This](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol:~:text=export%20interface%20SymbolInformation,string%3B%0A%7D) is the LSP reponse to "textdocument/documentSymbol" and [this](https://github.com/lcompilers/lpython/blob/c11c68ba6a8795afd8c19ea890556792dc378f19/src/libasr/lsp/LPythonServer.cpp#L108-L111) is LPython creating the same API. You can see how the two share `name`, `location` and `kind`, that is effectively forming the JSONRPC2.0 response manually. That is the part that you have to maintain! If you had chosen to use the `deprecated` field in `SymbolInformation` , when LSP released 3.16 you would then need to edit the calls to `test_capture` to account for that and create a JSONRPC2.0 message that has a `tags` of type array `SymbolTag` instead of `deprecated` which was type boolean.

I am not sure if the [`rapidjson::Document test_capture`](https://github.com/lcompilers/lpython/blob/c11c68ba6a8795afd8c19ea890556792dc378f19/src/libasr/lsp/LPythonServer.cpp#L106) is part of the "custom communication" using JSON, but you cannot remove or replace that with something that is not JSON. That JSON is what is read by the client.

If my reading of the files I linked at the beginning is correct, there is nothing standing between the LSP server JSON response and what your VS Code client receives. Specifically, the response gets piped through an IO stream via [your own implementation of the JSONRPC2.0 protocol](https://github.com/lcompilers/lpython/blob/c11c68ba6a8795afd8c19ea890556792dc378f19/src/libasr/lsp/JSONRPC2Connection.cpp#L88-L95) straight to the [client](https://github.com/ankitaS11/custom-vscode-linter/blob/6207d4ae5b8dc5f29004a2910ec5f16be09eb631/src/client.ts#L67-L70) that launches the server (compiled binary).

I don't see a JSON layer that could be replaced by some other representation, am I missing something?

certik:
> Do you think you can point me to the Typescript server extension, I would be curious to see how you interfaced with C++?

Yes:

https://github.com/ankitaS11/custom-vscode-linter/blob/6207d4ae5b8dc5f29004a2910ec5f16be09eb631/src/client.ts#L3

This is interfacing our prototype, we now need to update it to just use LPython.

You are right, I think this line here:

https://github.com/ankitaS11/custom-vscode-linter/blob/6207d4ae5b8dc5f29004a2910ec5f16be09eb631/src/client.ts#L67

Makes it simply use the C++ server as an LSP server, it does not do any translation. So we are using the second version:

> client (in TypeScript as the VSCode extension) <-> LSP via JSON <-> LPython C++ server

Sorry about that, the confusion was on my part. :)

I agree this might not be the best approach, since we do not want to maintain the complex (and changing) LSP protocol by hand. If there was a solid C++ library, we could use it, but I didn't find one. So it seems to me going the other direction (as we improve this work):

> client (in TypeScript as the VSCode extension) <-> LSP via JSON <-> server (still in TypeScript as the VSCode extension, using Microsoft's LSP library in TypeScript) <-> custom communication <-> LPython C++ server

Would be better.

Do you see a better approach? I want to use a well supported and maintained LSP library. And just maintain a minimum of code ourselves. It seems using the TypeScript Microsoft LSP library is a good bet. We just need to implement the translation from our custom (simple --- hopefully!) custom protocol that only does what we implement and maintain.

gnikit:
>Do you see a better approach? 

I can think of 3 things:

1. Go to clang and ask them if they can redistribute their LSP SDK as a separate library. I will open an issue and CC you.
2. Use Pygls in Python, it is relatively stable and has a steady source of funding and maintainers working on it part-time.
3. Interface C++ with Typescript. You can use [nbind](https://github.com/charto/nbind#readme), 1.9k stars but hasn't received an update since 2019. Maybe we can contact the authors and see what is going on. Maybe there is some other library out there but I couldn't find anything.

PR #678, opened 2022-06-23
Make error messages consistent

### LINKED ISSUE #673: Inconsistency in error messages ?
```
s = "abc"
s = s + 1          # Int type

 --> try.py:8:9
  |
8 |     s = s + 1
  |         ^   ^ type mismatch (str and i32)


s = "abc"
s = s + 1.0        # Float type

 --> try.py:8:9
  |
8 |     s = s + 1.0
  |         ^   ^^^ type mismatch (str and f64)


s = "abc"
s = s + complex(1,2)   # Complex type - error message not uniform with the above types

semantic error: Casting str to complex is not Implemented   # How is this str to complex casting anyways ?
 --> try.py:8:9
  |
8 |     s = s + complex(1,2)
  |         ^ 


```

faze-geek:
Remove unnecessary lines to fix #673

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -950,10 +950,6 @@ public:
                     al, right->base.loc, right, ASR::cast_kindType::LogicalToInteger, int_type));
                 return ASR::down_cast<ASR::expr_t>(ASRUtils::make_Cast_t_value(
                     al, right->base.loc, right, ASR::cast_kindType::IntegerToComplex, left_type));
-            } else {
-                std::string rtype = ASRUtils::type_to_str_python(right_type);
-                throw SemanticError("Casting " + rtype + " to complex is not Implemented",
-                        right->base.loc);
             }
         }
         if (!is_assign) { // This will only be used for BinOp
--- a/tests/reference/asr-test_str_mismatch-4aad88f.stderr
+++ b/tests/reference/asr-test_str_mismatch-4aad88f.stderr
@@ -1,0 +1,5 @@
+semantic error: Type mismatch in binary operator; the types must be compatible
+ --> tests/errors/test_str_mismatch.py:5:9
+  |
+5 |     s = "abc" + c
+  |         ^^^^^   ^ type mismatch (str and c32)
--- a/tests/errors/test_str_mismatch.py
+++ b/tests/errors/test_str_mismatch.py
@@ -1,0 +1,8 @@
+def main():
+    s : str
+    c : c32
+    c = 1+ 1j
+    s = "abc" + c
+    print(s)
+
+main()
--- a/tests/reference/asr-test_str_mismatch-4aad88f.json
+++ b/tests/reference/asr-test_str_mismatch-4aad88f.json
@@ -1,0 +1,13 @@
+{
+    "basename": "asr-test_str_mismatch-4aad88f",
+    "cmd": "lpython --show-asr --no-color {infile} -o {outfile}",
+    "infile": "tests/errors/test_str_mismatch.py",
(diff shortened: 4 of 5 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -950,10 +950,6 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     al, right->base.loc, right, ASR::cast_kindType::LogicalToInteger, int_type));
                 return ASR::down_cast<ASR::expr_t>(ASRUtils::make_Cast_t_value(
                     al, right->base.loc, right, ASR::cast_kindType::IntegerToComplex, left_type));
-            } else {
-                std::string rtype = ASRUtils::type_to_str_python(right_type);
-                throw SemanticError("Casting " + rtype + " to complex is not Implemented",
-                        right->base.loc);
```
certik, line 956: This change makes LPython return the correct message.

One thing that worries me is that this branch of the if statement structure in this function does not do anything after this change. Is that a good design of `cast_helper`? 

It seems that `cast_helper` gives an error message right away if things cannot be cast. In which case, maybe we should just reformulate what this error message says. Instead of saying "not implemented", we should say "cannot cast c32 to string". 

I do like the current error message (in this PR, as seen in the tests). Do we effectively have two places in LPython that return the error? In `cast_helper` and then later as well?

Should we just have one place, say `cast_helper`, that checks for errors, and then later we assume they are correct?

czgdp1807: Well IMO,

> Is that a good design of `cast_helper`?

`cast_helper` is not abstract enough. Instead of a long `if-else-if` ladder I would prefer a simple rule map based with a tuple of type as keys and the desired cast as the value. Just pass in the types to the rule map and get your cast type. But anyways doesn't matter much now as its already implemented so let's keep using it.

> Instead of saying "not implemented", we should say "cannot cast c32 to string".

Agreed. The second message is more transparent and helps in debugging.

> Should we just have one place, say cast_helper, that checks for errors,

In an abstract sense, yes there should be only one place which checks for type compatibilities and return errors. If no errors then assume correctness in rest of your implementation.

czgdp1807: @faze-geek Why have you removed this `SemanticError` anyways?

faze-geek: The current `SemanticError` is inconsistent (as pointed in #673) so I first I was changing the error inside the `else` itself . But then by `--show-stacktrace` I realised that by removing these lines , the `type mismatch` errors for 1. `str + int`, 2. `str + float ` and 3. `str + complex` were getting generated from the same lines in the code ( which I thought is better than error generated from different lines)

faze-geek: I thought this would be straightforward ,I was unaware of `cast_helper`.

tests/reference/asr-test_str_mismatch-4aad88f.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Type mismatch in binary operator; the types must be compatible
+ --> tests/errors/test_str_mismatch.py:5:9
+  |
+5 |     s = "abc" + c
+  |         ^^^^^   ^ type mismatch (str and c32)
```
Thirumalai-Shaktivel, line 5: We already get this message in the `main`:
```console
$ lpython examples/expr2.py
 --> examples/expr2.py:5:9
  |
5 |     s = "abc" + c
  |         ^^^^^   ^ type mismatch (str and c64)


```
I think this is fixed.

PR #679, opened 2022-06-23
decimal module inititiated

Madhav2310:


src/runtime/decimal.py
```diff
@@ -0,0 +1,26 @@
+from ltypes import i8, i16, i32, f32, f64, ccall
+
+@overload
+def Decimal(x: i64) -> i64:
+    return x
```
czgdp1807, line 5:
```
>>> from decimal import *
>>> Decimal(1)
Decimal('1')
>>> type(Decimal(1))
<class 'decimal.Decimal'>
```
Let's start off with something like the following,

```python
@dataclass
class Decimal:
    value: str
```

czgdp1807: We can remove `@dataclass` from `Decimal` in future. For now using them should be okay.

czgdp1807: cc: @certik

certik: I wonder if we should just implement these the right way, that is, they have to be implemented as a class, that implements some methods that you can then call. So we have to first implement classes in LPython.

Madhav2310: Will it have overloading too? Something like:

```
@dataclass
class Decimal:

    @overload
    def Decimal(x: i64) -> i64:
        return x
    @overload
    def Decimal(x: i32) -> i32:
        return x

    @overload
    def Decimal(x: f32) -> f64:
        return float(x)
    @overload
    def Decimal(x: f64) -> f64:
        return float(x)
    
```

Madhav2310: Also please explain why I am facing this particular error, what should be necessarily done differently?

integration_tests/test_decimal.py
```diff
@@ -0,0 +1,19 @@
...
+    assert i == 4
+    j: i64
+    j = Decimal(2)
+    k: f64
+    k = Decimal(0.5)
+    assert abs(float(j ** k) - 1.414213562373095048801688724) < eps
+
+def check():
+    test_Decimal_1()
+check()
```
czgdp1807, line 19: Add final new line.

src/runtime/decimal.py
certik: I think we can support overloading of methods, but as a second iteration, not first (it might be a bit involved, but doable).

Why don't you open up an issue for what is needed. I think what is needed is support for this:
```
class Decimal:
    def f(x: i64) -> i64:
        return x

def g() -> Decimal:
    return Decimal()

x: Decimal = g()
print(x.f())
```

And probably more, but the above is a start. Can you please open up an issue for this? Then let's implement it in LPython. Then you'll abe able to use it, and we can then open up issues for any missing features, etc.

Madhav2310:
@czgdp1807 please see if this is good for merging, and how can we improve it

integration_tests/test_decimal.py
```diff
@@ -0,0 +1,25 @@
+from decimal import (Decimal)
+from ltypes import i64, f32, f64, i64
+
+eps: f64
+eps = 1e-12
+
+def test_Decimal_1():
+    i: i64
```
czgdp1807, line 8:
```suggestion
    i: Decimal
```

integration_tests/test_decimal.py
```diff
@@ -0,0 +1,25 @@
...
+    assert i == 4
+    b: i64
+    b = 2
+    j: i64
+    j = Decimal(b)
+    k: f64
+    c:f64
+    c = 0.5
+    k = Decimal(c)
+    assert abs(float(j ** k) - 1.414213562373095048801688724) < eps
```
czgdp1807, line 21: Also check for,

`assert j ** k == 8`
`assert j ** k == Decimal(8)`

czgdp1807 (changes requested):

certik:
@czgdp1807 can you please review this PR? It looks like @Madhav2310 asked you to review it, but somehow we missed it.

czgdp1807:
This requires full class support (especially method calling, object reaction). Data classes alone won't be sufficient to implement this. I discussed this with @Madhav2310 in one of the GSoC weekly meetings and we arrived at the same conclusion.

Issue #683, opened 2022-06-23
Instead of offsetting indices, set dimensions correctly in variable type


czgdp1807:
https://github.com/lcompilers/lpython/blob/4591db8b087577046f548d21097d8c8c23fcb2f7/src/lpython/semantics/python_ast_to_asr.cpp#L887-L896

https://github.com/lcompilers/lpython/blob/4591db8b087577046f548d21097d8c8c23fcb2f7/src/lpython/semantics/python_ast_to_asr.cpp#L866-L870

The above code is hackish and it becomes to keep track whether to offset the index by 1 in an optimisation pass or not. The reason of confusion is,

1. We are setting lower bound as `1` and upper bound as `size` in both Python and Fortran.
2. We offsetting indices by `1` only in Python.
3. We only have ASR while performing an optimisation pass. In Python we set lower and upper bounds same as Fortran but we offset index by 1 only in Python. Hence no clear differences between the two.

I suggest doing the following,

1. Set lower bound as `0` and upper bound as `size - 1` in Python. In Fortran its provided by the user and default setting in LFortran is correct.
2. Don't offset indices by `1` in any LCompiler.
3. In an optimisation pass set the bounds of the loop (which we are creating manually in an optimisation pass) always to `0 + lower_bound(dimension)` and `size - 1 + lower_bound(dimension)`. In Python `lower_bound(dimension)` is always `0`, so no worries. In Fortran `lower_bound(dimension)` is by default `1` so loop bounds will transform to `1` and `size` automatically.

I discovered this problem while working on #675 and I can fix it in #675 itself.

@certik Please let me know what you think.

certik:
See also: https://gitlab.com/lfortran/lfortran/-/issues/666

As discussed there, I think we should move to `(size, start_index)` representation of arrays. I don't have an opinion whether we should get rid of `start_index`, which would mean adjusting/offsetting all array indexing (there are pros/cons of doing that). So first let's keep the `start_index`. We should not offset any indices (that's the whole point of storing `start_index` in the first place).

czgdp1807:
Well, my main concern is offsetting indices. Either you always do it in an optimisation pass or you never do it because once an ASR is formed you never know whether the source was Python or Fortran. So offsetting indices should never be done in any frontend irrespective of which array representation we are using. 

Anyways would it be okay if I remove offsetting indices by 1 in LPython? Let me know.

certik:
[...]

PR #685, opened 2022-06-24
Removed index_add_one and set dimensions correctly

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/683

An error needing a fix similar to https://gitlab.com/lfortran/lfortran/-/merge_requests/1740 is coming up. So I will fix it here itself.


certik (approved):
I think this is fine. For strings, we should just start from 1 (for now), so for LPython one must offset by 1.

We could start strings at 0, if it is easier in the LLVM backend, but I think we can change this later if needed.

czgdp1807:
> For strings, we should just start from 1 (for now), so for LPython one must offset by 1.

Fixed.

PR #686, opened 2022-06-24
[Parser] Parse list comprehension

### LINKED ISSUE #508: Parse Single line if and for statements
Example 1:
```python
b = 6 if a == 2 else 8
```
Error:
```console
syntax error: Token 'if' is unexpected here
 --> examples/expr2.py:1:7
  |
1 | b = 6 if a == 2 else 8
  |       ^^ 


```
Example 2:
```python
i = [x.strip() for x in args.bench_compare.split(',')]
```
Error:
```console
syntax error: Token 'for' is unexpected here
 --> examples/expr2.py:1:16
  |
1 | i = [x.strip() for x in args.bench_compare.split(',')]
  |                ^^^ 


```

akshanshbhatt:
Closes #508
Reference: https://docs.python.org/3/reference/expressions.html#displays-for-lists-sets-and-dictionaries


certik (approved):
I think that this is great. Can you please rebase on top of the latest `main`?

For small PRs like this one, let's not merge into the PR, but just rebase. It makes the git history a lot cleaner.

certik:
Awesome, thanks!

PR #687, opened 2022-06-25
Add support for `sep` keyword in print

### LINKED ISSUE #665: Implement the `sep` keyword argument in print()
Example:
```
print("a","b","c",sep=",")
```
This I think might be easy: simply extract/check it in ast->asr, and insert `,` instead of ` ` between the other arguments.

Smit-create:
Fixes #665


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -4139,10 +4139,17 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
     void handle_print(const T &x) {
         std::vector<llvm::Value *> args;
         std::vector<std::string> fmt;
-        for (size_t i=0; i<x.n_values; i++) {
+        // separator
+        this->visit_expr_wrapper(x.m_values[x.n_values - 1], true);
+        llvm::Value* separator = tmp;
+        for (size_t i=0; i<x.n_values - 1; i++) {
```
certik, line 4145: This change effectively encodes the separator as the first argument to Print.

Let's add it as a new argument to `Print`, called `string separator`. That will make it clean. Also force the change in LFortran. This change here would break LFortran, since it is not using this new convention. As an argument, it would not break it.

We ultimately need a language independent representation of string formatting, but I think this little change (adding `separator` in ASR's Print) goes in the right direction and is clean enough.

Smit-create: Sure!

src/libasr/ASR.asdl
```diff
@@ -190,7 +190,7 @@ stmt
               expr? read, expr? write, expr? readwrite, expr? delim,
               expr? pad, expr? flen, expr? blocksize, expr? convert,
               expr? carriagecontrol, expr? iolength)
-    | FileWrite(int label, expr? unit, expr? fmt, expr? iomsg, expr? iostat, expr? id, expr* values)
+    | FileWrite(int label, expr? unit, expr? fmt, expr? iomsg, expr? iostat, expr? id, expr* values, expr? separator)
```
certik, line 193: Since we are not hooking this up in this PR, I would not add it to `FileWrite` yet. I am hoping to figure out a general solution for file IO that works well for both LPython and LFortran.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3303,10 +3303,22 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
                 Vec<ASR::expr_t*> args_expr = ASRUtils::call_arg2expr(al, args);
+                ASR::expr_t *separator = nullptr;
+                if (c->n_keywords > 0) {
+                    std::string arg_name;
+                    for (size_t i = 0; i < c->n_keywords; i++) {
+                        arg_name = c->m_keywords[i].m_arg;
+                        if (arg_name == "sep") {
+                            visit_expr(*c->m_keywords[i].m_value);
+                            separator = ASRUtils::EXPR(tmp);
+                            break;
```
certik, line 3316: Let's check the type of `tmp`, make sure it's a string expression, otherwise give an error message, and let's write a test for the error message.

integration_tests/print_01.py
```diff
@@ -1 +1,8 @@
-print("Hello World!")
+def f():
+    print("Hello World!")
+    x: str
+    x = ","
+    print("a", "b", sep=x)
+    print("a", "b", sep=":")
```
certik, line 6: Let's also add tests for `sep` with more than 1 character.

certik (commented):
I think this looks very good. I left some minor comments, it would be nice if we can fix them while we are at this.

src/libasr/ASR.asdl
Smit-create: I need to make that change because `handle_print` is also used by `FileWrite`.

certik: I see, keep it there then.

certik (approved):
I think this looks good to go in.

certik:
Thanks @Smit-create !

PR #688, opened 2022-06-25
Add a test for type mismatch in pointers

Smit-create:
cc @czgdp1807

### DIFF
--- a/tests/reference/llvm-test_c_interop1-9657eef.stderr
+++ b/tests/reference/llvm-test_c_interop1-9657eef.stderr
@@ -1,0 +1,4 @@
+code generation error: asr_to_llvm: module failed verification. Error:
+Stored value type does not match pointer operand type!
+  store %array* %y, %array.0** %yptr1, align 8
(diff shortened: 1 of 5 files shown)


tests/reference/llvm-test_c_interop1-9657eef.stderr
```diff
@@ -0,0 +1,4 @@
+code generation error: asr_to_llvm: module failed verification. Error:
+Stored value type does not match pointer operand type!
+  store %array* %y, %array.0** %yptr1, align 8
+ %array.0*
```
certik, line 4: I would not be adding LLVM test failures here, because they might depend on LLVM version and ultimately our LLVM backend should never fail like this. It should fail by giving LFortran codegen error (which we can and should test) and ultimately this should be a semantic error, already in AST->ASR.

Smit-create: Okay, I'll open an issue for the same.

PR #697, opened 2022-06-26
Don't overload the `bool()` intrinsic function

### LINKED ISSUE #410: Don't overload the `bool` intrinsic function

namannimmo10:
Fixes #410


certik:
Something like that.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3452,6 +3452,72 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                al, loc, t, ASR::cast_kindType::IntegerToLogical, to_type, value));
+
+        } else if (ASRUtils::is_character(*type)) {
+            if (ASRUtils::expr_value(arg) != nullptr) {
+                char *c = ASR::down_cast<ASR::StringConstant_t>(
+                                        ASRUtils::expr_value(arg))->m_s;
+                value = ASR::down_cast<ASR::expr_t>(make_LogicalConstant_t(al,
+                                loc, std::string(c) != "", to_type));
+            }
+            // TODO: make bool() work for non-constant strings
```
namannimmo10, line 3500: How do you wanna do this? Call a runtime function or have a node for this, too?

certik: Can this be done with the existing `Cast` ASR node? Such as casting Integer to Bool.

czgdp1807: I would suggest using `Cast` node. What do you all think?

namannimmo10: How about we add `CharacterToLogical` and `ComplexToLogical` under `cast_kind`? 

> Can this be done with the existing `Cast` ASR node?

I don't see any existing node that does the job here.

czgdp1807: Well here’s what I think.
1. First check whether the call is being made to `bool`. In an abstract sense, I mean verify that the semantics of the code is to call `bool` over a single input.
2. Now, see if the input is of intrinsic type (i.e., integer, real/float, complex). If so then things are pretty much clear, source type is the type of input argument and destination type is logical. If the input type is say a class then I think Python provides some magic method (I think it is `__bool__`) for conversion. We will have to call `__bool__`.
3. On the basis of step 2 either create a `Cast` node (if the type is node) or a `FunctionCall` (if the type is class).

I think I have covered all the cases for `bool(x)` expression. It is fairly easy to implement the above steps in ASR (provided I am not missing out on anything).

czgdp1807: > How about we add CharacterToLogical and ComplexToLogical under cast_kind?

I think yes we should add and use them in step 3 of my comment above.

czgdp1807: For non-constant strings in the LLVM backend we might not need a new C runtime function. We can call already existing function which computes the length of the string. The integer received in return will easily determine the result. If 0 then false and true in all other cases.

namannimmo10:
@czgdp1807, @certik -- this is ready.

certik (approved):
I think that this is fine.

Issue #610, opened 2022-06-15
To reverse strings using slicing 


faze-geek:
Is str[ : : int ] supported right now ?
```
def main1():
    s1: str
    s1 = "abc"
    print(s1[::-1])

main1()

(lp) anurag-3:lpython user$ python try.py
cba
(lp) anurag-3:lpython user$ ./src/bin/lpython try.py && ./a.out 
abc    # Should be cba
```

certik:
Good point, we should give an error message for now ("not implemented"), and ultimately we should implement this.

faze-geek:
While playing with this I realised that the string slicing with step function is buggy .
```
def main():
    s: str
    s="abcdefg"
    print(s[0:7:2])

main()

(lp) anurag-3:lpython user$ python try.py
aceg
(lp) anurag-3:lpython user$ ./src/bin/lpython try.py && ./a.out
abcdefg
```

These lines need to be fixed I think , I don't get why `index_add_one` is needed here ?
https://github.com/lcompilers/lpython/blob/49a65ead06e011b34821c1add854a73894d142f7/src/lpython/semantics/python_ast_to_asr.cpp#L1793-L1796

@namannimmo10 Could you help for fixing this .Thanks !

namannimmo10:
This code produces a `StringSection` node in the ASR and that is handled by the LLVM visitor,

https://github.com/lcompilers/lpython/blob/0a7865ea64d54728858d89c095536ae9d99cdbd2/src/libasr/codegen/asr_to_llvm.cpp#L3188-L3200

`lfortran_str_copy` calls into this C function:

https://github.com/lcompilers/lpython/blob/0a7865ea64d54728858d89c095536ae9d99cdbd2/src/runtime/impure/lfortran_intrinsics.c#L638-L659 

So this should also take care of the `step` index. Would you like to try this?

PR #699, opened 2022-06-26
Added support for step in string slicing 

faze-geek:
Related to https://github.com/lcompilers/lpython/issues/610#issuecomment-1157516635

cc @namannimmo10


integration_tests/test_str_01.py
```diff
@@ -49,6 +49,12 @@ def test_constant_str_subscript():
     assert "abc"[2] == "c"
     assert "abc"[:2] == "ab"
 
+def test_str_slice_step():
+    s: str
+    s = "abcdefghijk"
+    assert s[3:12:3] == "dgj"
+    assert s[12:3:-3] == "mjg"
+
```
certik, line 57: You have to call this new function from below.

integration_tests/test_str_01.py
```diff
@@ -49,6 +49,12 @@ def test_constant_str_subscript():
     assert "abc"[2] == "c"
     assert "abc"[:2] == "ab"
 
+def test_str_slice_step():
+    s: str
+    s = "abcdefghijk"
+    assert s[3:12:3] == "dgj"
+    assert s[12:3:-3] == "mjg"
```
namannimmo10, line 56: Can you please add more tests? Like these:

```python
>>> s = "abcdefg"
>>> s[::]
'abcdefg'
>>> s[1:4:]
'bcd'
>>> s[1::-2]
'b'
>>> s[:4:5]
'a'
>>>
```

czgdp1807: In addition to the above also try passing non-constant expressions to the inputs. Say like this,

```python
start: i32
end: i32
step: i32
s = "abcdefg"
start = 1
end = 4
step = 1
assert s[start:end:step] == "bcd"
assert s[start:end - 1:step] == "bc"
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -788,19 +788,20 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         return builder->CreateCall(fn, {str});
     }
 
-    llvm::Value* lfortran_str_copy(llvm::Value* str, llvm::Value* idx1, llvm::Value* idx2)
+    llvm::Value* lfortran_str_copy(llvm::Value* str, llvm::Value* idx1, llvm::Value* idx2, llvm::Value* idx3 = nullptr)
```
czgdp1807, line 791: I think instead `idx3` we should use `step` as the name.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1204,6 +1205,8 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                     llvm::Value *idx2 = tmp;
                     this->visit_expr_wrapper(x.m_args[0].m_left, true);
                     llvm::Value *idx1 = tmp;
+                    //this->visit_expr_wrapper(x.m_args[0].m_step, true);
+                    //llvm::Value *idx3 = tmp;
```
czgdp1807, line 1209: Why this is commented out? Aren't we supposed to visit the step expression in ASR and then use it in `llvm::Value *step`?

src/runtime/impure/lfortran_intrinsics.c
```diff
@@ -636,7 +636,7 @@ LFORTRAN_API void _lfortran_strrepeat(char** s, int32_t n, char** dest)
 }
 
 // idx1 and idx2 both start from 1
-LFORTRAN_API char* _lfortran_str_copy(char* s, int32_t idx1, int32_t idx2) {
+LFORTRAN_API char* _lfortran_str_copy(char* s, int32_t idx1, int32_t idx2, int32_t idx3) {
```
czgdp1807, line 639: Same `idx3` -> `step`.

src/runtime/impure/lfortran_intrinsics.c
```diff
@@ -649,12 +649,20 @@ LFORTRAN_API char* _lfortran_str_copy(char* s, int32_t idx1, int32_t idx2) {
     if(idx2 <= 0) {
         idx2 = s_len + idx2;
     }
-    char* dest_char = (char*)malloc(idx2-idx1+2);
-    for (int i=idx1; i <= idx2; i++)
-    {
-        dest_char[i-idx1] = s[i-1];
+    char* dest_char;
+    if (idx3 > 0 && (idx1 <= idx2)) {
+        dest_char = (char*)malloc(idx2-idx1+2);
```
czgdp1807, line 654: Well in case `idx3` is present the length will not be `idx2 - idx1 + 2` I think. The factor `idx3` will come in. I think the correct length would be, `(idx2 - idx1)/idx3 + 1 + 2`. Using `idx2 - idx1 + 2` is wastage of memory by huge margins if `idx3` is not `1`.

src/runtime/impure/lfortran_intrinsics.c
```diff
@@ -649,12 +649,20 @@ LFORTRAN_API char* _lfortran_str_copy(char* s, int32_t idx1, int32_t idx2) {
...
+        for (int i=idx1; i <= idx2; i+=idx3){
+            dest_char[((i - idx1)/(idx3))] = s[i-1];
+        }
+        dest_char[idx2-idx1+1] = '\0';
+    } else if (idx3 < 0 && (idx1 > idx2)) {
```
czgdp1807, line 659: I think we might not need to implement `idx3 < 0` separately. Basically its just an arithmetic progression with start = `idx1`, end < `idx2` and `idx3` as the common difference. So, `idx1 + (number_of_terms - 1) * idx3 < idx2`. The sign doesn't matter here (so it shouldn't in the implementation as well I think).

Anyways upto you.

czgdp1807: For checking if the arithmetic progression is finite you can simply check if the number of terms coming from the above in-equality is positive. Simple do, `number_of_terms = floor((idx2 - idx1)/idx3 + 1)`. If `number_of_terms` is positive then the result will be finite sized. Otherwise it won't be finite sized. Consider, `idx2 > idx1` and `idx3 < 0` then, `idx2 - idx1 > 0` and hence `number_of_terms <= 0`, so no need to run the loop.

czgdp1807: If you are doing this for caching efficiency then also it won't matter. Because you need to fill the destination string in the right order. Given that `idx3 < 0` and `idx1 > idx2` If you traverse input string from `idx1` to `idx2` then you will move in the backwards direction (lower cache efficiency for input string) but in forward direction in the output string (higher cache efficiency for output string). If you travel from `idx2` to `idx1` in the input string then you will move in forward direction (higher cache efficiency for input string) but you will be required to fill output string in backward direction (lower cache efficiency). Theoretically, its better to achieve higher cache efficiency in output string (because its smaller and we are taking fixed step size of 1 while filling it i.e., greater spatial locality). In input string it all depends on step size if it is higher then it doesn't matter whether you move in forward or backward directions (indices are already far away so lower spatial locality already).

integration_tests/test_str_01.py
faze-geek: I have added more tests now !

src/libasr/codegen/asr_to_llvm.cpp
faze-geek: `idx3` was used just for consistency with `idx1` for left and `idx2`for right. I've updated the name now.

src/runtime/impure/lfortran_intrinsics.c
faze-geek: Thanks @czgdp1807 for the insight and suggestions. 
I was aiming to get a structure for working step function in `[start:end:step]` format . A couple of edge cases need to be handled first then I'll address all the suggestions .

faze-geek:
> @faze-geek Would you like to continue working on this?

Sorry, I have not been able to work on it after university reopening. Feel free to takeover if you can complete it .Thanks !

redbopo:
Hi @faze-geek  and @czgdp1807 , I'd like to take over it a fix later.

Smit-create:
Done in #853

PR #701, opened 2022-06-27
C: Implement separator in the C backend

certik:


certik:
This builds on top of #687.

PR #705, opened 2022-06-28
Ignore inlining of functions with return statements

czgdp1807:
https://github.com/lcompilers/lpython/issues/519 - A temporary fix to make things work.


certik (approved):
That looks good for now.

Issue #700, opened 2022-06-27
Implement Structs inside Structs


Smit-create:
Python: 
```py
from ltypes import i32, f32, dataclass

@dataclass
class A:
    y: f32
    x: i32

@dataclass
class B:
    a: A
    z: i32

def f(b: B):
    print(b.z)
    print(b.a.x)
    print(b.a.y)

def g():
    b: B
    b.z = 1
    b.a.x = 2
    b.a.y = 3.0
    f(b)

g()
```

Error:
```
semantic error: Only Name is supported for now in Attribute
  --> expr.py:15:11
   |
15 |     print(b.a.x)
   |           ^^^^^ 
```

Issue #691, opened 2022-06-25
Add Structs Array support


Smit-create:
Python code:
```py
from ltypes import i32, f32, dataclass

@dataclass
class A:
    y: f32
    x: i32

def g():
    y: A[2]
    y[0].x = 1
    y[0].y = 1.1
    y[1].x = 2
    y[1].y = 2.2

g()
```
Error:
```
semantic error: Only Name is supported for now in Attribute
  --> expr.py:10:5
   |
10 |     y[0].x = 1
   |     ^^^^^^ 
```

redbopo:
Hi @Smit-create , I think I can fix this with a patch.

Smit-create:
Sure, that would be great. Thank you!

redbopo:
Just the same in previous

```
from ltypes import i32, f32, dataclass

@dataclass
class A:
    y: f32
    x: i32

def g():
    y: A[2]
    y[0].x = 1
    y[0].y = 1.1
    y[1].x = 2
    y[1].y = 2.2

g()
```

czgdp1807:
@redbopo In response to https://github.com/lcompilers/lpython/issues/691#issuecomment-1168675985,

https://github.com/lcompilers/lpython/pull/718/commits/772eda6df1fbf11df7e05eabefc705fcfc12584c

Since we are supporting arrays of structs for the first time, `ArrayRef` visitor is required to set `der_type_name`. In the above commit I have done the same.

PR #707, opened 2022-06-28
Adding support for Structs inside Structs

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/700


czgdp1807:
So the scene is a bit tricky here,

1. Fortran (i.e., concerned with LFortran) - By default, members which are of derived type are stored by value. Which means when you initialise say as `employee(jack, 2003)` where, `jack = person("Jack", "Smith")` then `jack` will be deep copied and hence, changes in `jack.fname` or any of its data members won't affect `employee.person.fname`.
2. Python (i.e., concerned with LPython) - By default, members which are of derived type are stored by reference. Taking the same example in point 1, now changes in `jack.fname` would be reflected in `employee.person.fname` and vice-versa.

Since, LLVM backend is common in both so either to store a derived type member of a struct by value or by reference should be somehow encoded in ASR while constructing it from input language's AST. For now, what I have done is, in LPython I wrap the derived type member in a Pointer type and rest of the code as is. So, in LLVM backend, the derived type members are pointers and hence effectively they get stored by reference and point 2 is satisfied.

cc: @certik

certik (approved):
I think that this is fine to go in.

Regarding value/reference, I would need to understand the details more. The options on the table are either

* keep ASR as is, and somehow support a subset of Python that maps to the current design
* modify ASR possibly as reference, and modify LFortran to use it, assuming it is a superset
* Add a flag to ASR Struct node, to choose which one it is

Ideally I am trying to keep the ASR simple and small, the smallest design that covers Fortran and the ahead of time compiled Python.

czgdp1807:
Its the difference in semantics of Python and Fortran regarding derived types a.k.a classes which is the concern here. You can play around with the test in this PR and write something similar for Fortran, the differences will be clear.

czgdp1807:
> Without knowing the details, it seems by value and no flag is the simplest/fastest, but if it can't be made to work with Python, then we need to improve the design.

As far as I can understand, do you mean that we should always deep copy. I think this won't give the expected behaviour in Python.

certik:
It needs to behave like CPython. So either we give a compiler (or runtime) error, or it must behave like CPython.

I see, CPython is by reference. A simple solution is to never access the original variable (checked at compile time). Then it can be compiled by value.

czgdp1807:
Well depends on how we fix, #691 . I will fix that first and then add this restriction.

PR #709, opened 2022-06-28
[Parser] Parsing support for `in` (Membership Operator)

### LINKED ISSUE #511: [New] Parse IN and IS in the expression

akshanshbhatt:
Closes #511

- Solved the shift/reduce conflicts related to expressions containing `in`. The grammar was clashing with the syntax of for-loops (`for...in`).
- Refactored the tokens as asked in https://github.com/lcompilers/lpython/pull/578#discussion_r895094909.


certik (approved):
I think that's fine. Why is the `FOR` precedence needed?

akshanshbhatt:
> I think that's fine. Why is the `FOR` precedence needed?

There were shift/reduce conflicts between the `for-loop` and `in` expressions. The grammar pattern for a `for-loop` is as follows -

```bison
KW_FOR tuple_item KW_IN expr ":" sep statements
```
`tuple_item` node has rules that consider `expr` to be a valid grammar when reduced. Therefore,

```bison
KW_FOR expr KW_IN expr ":" sep statements
```
is also a valid pattern for `for-loop`. This can lead to grammar ambiguity, which was the reason behind the shift/reduce conflict. 

```py
for i in a in list1:
    pass
```
In this example, `i in a` and `a in list1` both are valid `expr`.
Therefore, I decided to give precedence to the `for-loop` instead of the expression so that the first `in` could be associated with the `for-loop`.

An alternative approach can be replacing `tuple_item` from the `for-loop` with `id_item` as we did in the case of `list_comprehension` -

https://github.com/lcompilers/lpython/blob/bdf09ac9d4b766fcb4421a8190cbb7259114e9e5/src/lpython/parser/parser.yy#L679-L684

General `tuples` can have expressions as items, but `id_tuple` will ensure that only `identifiers` are items. Using this, we will not require the `%precedence` token. I was unsure of what to do, so I decided to go with the option requiring the least number of changes to the existing grammar.

certik:
You did the right choice. The simplest way to get things working that is not hackish, you did it. And later if we figure out a better/cleaner solution we can iterate on it.

Thirumalai-Shaktivel:
Excellent, Thanks for this PR!

PR #710, opened 2022-06-29
Implement the end keyword argument for print()

### LINKED ISSUE #704: Implement the end keyword argument in print()
The motivation behind this issue is https://github.com/lcompilers/lpython/issues/665 and https://github.com/lcompilers/lpython/pull/687 .
Print in python is `print(value(s), sep= ‘ ‘, end = ‘\n’)` where `end` is keyword argument set to `"\n" ` by default to take the cursor to the next line. That user should be able to control that .

I have added it as a pull request here https://github.com/faze-geek/lpython/pull/2 . @certik do tell me if this is a possible addition then I'll make a pull request with tests and updated reference files here .

Use case and test on branch -
```
def main():
    print("a","b","c",sep = "=", end ="+")
    print("d","e","f",end ="!!" , sep = "*\n")

main()

a=b=c+d*
e*
f!!(lp) anurag-3:lpython user$ ./src/bin/lpython try.py && ./a.out 

a=b=c+d*
e*
f!!(lp) anurag-3:lpython user$

```

faze-geek:
Fix #704 

Once this goes in, I could add`end` in c/cpp backends as a separate pull request.


faze-geek:
The failed tests don't fail locally on my machine. Would appreicate if someone could figure out if the failing tests are relevant to this pr .

tests/reference/asr-test_print2-64acb15.stderr
```diff
@@ -0,0 +1,29 @@
+Traceback (most recent call last):
```
Smit-create, line 1: The expected output here should show the ASR semantic error. Here, you have a segmentation fault which means something is wrong with the code.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3284,12 +3285,27 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                            visit_expr(*c->m_keywords[i].m_value);
+                            end = ASRUtils::EXPR(tmp);
+                            ASR::ttype_t *type = ASRUtils::expr_type(end);
+                            if (!ASRUtils::is_character(*type)) {
+                                std::string found = ASRUtils::type_to_str(type);
+                                diag.add(diag::Diagnostic(
+                                    "End is expected to be of string type",
+                                    diag::Level::Error, diag::Stage::Semantic, {
+                                        diag::Label("Expected string, found: " + found,
+                                                {separator->base.loc})
```
Smit-create, line 3299:
```suggestion
                                                {end->base.loc})
```

This will fix the issue.

faze-geek: Thanks @Smit-create . I think the tests should pass now.

certik (approved):
I think this is good to merge. Thanks!

Issue #694, opened 2022-06-25
Show error for Pointer type mismatch at ASR level


Smit-create:
Python code:
```py
from ltypes import pointer, i16, Pointer, i32

def f():
    yptr1: Pointer[i16[:]]
    y: i32[2]
    y[0] = 1
    y[1] = 2
    yptr1 = pointer(y)

f()
```

This should raise an error at ASR level. https://github.com/lcompilers/lpython/pull/688#discussion_r906671577
cc @czgdp1807

certik:
Yes. All errors should be raised in AST->ASR phase. The only errors in ASR->LLVM should be "not implemented", nothing else.

PR #711, opened 2022-06-29
Fix `check_equal_type` and raise error while assigning pointers of different types

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/694

I noticed that in `main` branch `check_equal_type` is implemented incorrectly. It doesn't consider their kind (a.k.a bid-width) while checking equality of types. So, things like the following were getting through ASR unchecked,

1. Returning 32 bit variable/value from a function which is supposed to return 64 bit variable/value. ASR should always raise an error if such things are encountered. When I added check for equality of kind in `check_equal_type` several errors popped up while compiling `lpython_builtin.py`. I think many tests are not yet tested with LLVM? I don't know. I might be wrong.
2. Division of `f32`/`f32` resulting in `f64` - I think `f32`/`f32` should always return `f32` (returning `f64` is more like consuming 32 extra bit but with no extra accuracy).
3. Storing 64 bit constants in a tuple, accepting only 32 bit constants - Should never happen IMHO. Probably tuples are not yet supported in LLVM?

Anyways, I might be wrong but checking for equality of kind is very important while checking for equality of types. At least LLVM treats them as different and doesn't auto-casts type if they only have different bit-widths. C can auto cast so nothing to worry about in C backend.

If you feel that this PR is not necessary then feel free to close it.


certik (approved):
I think this is fine.

PR #716, opened 2022-06-30
Make `not` unary work for complex type

namannimmo10:


tests/tests.toml
```diff
@@ -223,6 +223,10 @@ llvm = true
 filename = "../integration_tests/test_builtin.py"
 asr = true
 
+[[test]]
+filename = "../integration_tests/test_complex.py"
+asr = true
+
```
czgdp1807, line 229: Is LLVM not supported for `test_complex.py`?

namannimmo10: It is supported.  We test this file with `cpython` and `llvm`.

PR #717, opened 2022-06-30
CPP/C: Consider `IntegerToLogical` cast kind in the backends

namannimmo10:
#410


tests/expr17.py
```diff
@@ -0,0 +1,8 @@
+def f():
+    i: i32
+    i = 3
+    b: bool
+    b = bool(i)
+    print(b)
+
+f()
```
czgdp1807, line 8: Can we shift this to `integration_tests`? And add it in `integration_tests/CMakeLists.txt` with `c`, `cpp`, `cpython` and `llvm` backends?

tests/reference/c-expr17-a800575.stdout
```diff
@@ -0,0 +1,54 @@
...
+{
+    f();
+}
+
+void f()
+{
+    bool b;
+    int32_t i;
+    i = 3;
+    b = (bool)(i);
```
czgdp1807, line 46: I don't think `C` has `bool` type. Probably, doing https://github.com/lcompilers/lpython/pull/717/files#r911632396 will bring clarity since the generated code will be compiled by `GCC` and `Clang` on CI.

namannimmo10: > I don't think `C` has `bool` type

If you see the C code that is generated, you'd see `stdbool.h` as a header file, so it should compile.  But I can shift this test under `integration_tests`,  just in case...

czgdp1807: Well checking if C code compiles is the best approach I feel. We have the infrastructure for that so why not use it.

czgdp1807: Try running `./run_tests.sh` under `integration_tests` directory.

tests/tests.toml
```diff
@@ -152,6 +152,11 @@ asr = true
 filename = "../integration_tests/expr_10.py"
 asr = true
 
+[[test]]
+filename = "../integration_tests/expr_11.py"
+c = true
+cpp = true
+
```
czgdp1807, line 159: Fine to have this but not necessary. We are already testing the generated code via `integration_tests/CMakeLists.txt`.

PR #726, opened 2022-07-02
Use `RealToLogical` cast kind directly

namannimmo10:


czgdp1807:
I am neutral on this change. We are saving one node per bool(real_input) expressions if we have it. If we don’t have it then also fine. Fine either way.

namannimmo10:
>  If we don’t have it then also fine. Fine either way.

Not really. See the test I have added. If we don't add this node, then the `RealToInteger` cast kind loses some information. `0.00000xxx` gets converted to `0`, then `bool(0)` will give us `False`, but it should actually turn out to be `True`.

czgdp1807:
Ah! I see. Makes sense.

czgdp1807 (approved):

PR #728, opened 2022-07-03
Fix bug in `int()` and `float()` intrinsic functions

namannimmo10:


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3376,12 +3376,15 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             return (ASR::asr_t *)ASR::down_cast<ASR::expr_t>(ASR::make_Cast_t(
             al, loc, arg, ASR::cast_kindType::LogicalToInteger,
             to_type, value));
-        } else if (!ASRUtils::is_integer(*type)) {
+        } else if (ASRUtils::is_integer(*type)) {
+            return (ASR::asr_t *)arg;
```
czgdp1807, line 3380: So the kind will be retained here, right? If input is 32 bit integer then answer will also be 32 bit integer and if the input is 64 bit integer then answer will also be 64 bit integer.

czgdp1807:
@namannimmo10 Do you want to continue working on this?

namannimmo10:
Yes, I will complete this today!

src/lpython/semantics/python_ast_to_asr.cpp
namannimmo10: The kind would be retained for `int()`, whereas `float()` now gives only 64-bit numbers.

czgdp1807: Why the difference in the behaviour of `int()` and `float()`? Is it not good to keep the behaviour for `int()` and `float()` similar?

namannimmo10:
@czgdp1807 -- please feel free to merge if this looks good to you.

PR #731, opened 2022-07-04
Multiple datatypes for pow and fabs

Madhav2310:


integration_tests/test_math.py
```diff
@@ -54,16 +54,18 @@ def test_radians():
 
 
 def test_exp():
-    i: f64
+    i: f64pow(2.4, 4.3)
```
czgdp1807, line 57: I think this is a syntactical mistake?

Madhav2310:
> If its taking too long then feel free to go ahead and copy changes though manually (so that you know what's being done) from https://github.com/czgdp1807/lpython/tree/madhav2 and push here by committing yourself. Let's complete it as early as possible.

I did run that too.. its giving the same error I'm receiving with my branch "Segfault: Signal SIGSEGV (segmentation fault) received".

integration_tests/test_math.py
```diff
@@ -60,10 +60,9 @@ def test_exp():
 
 
 def test_pow():
-    i: f64
-    i = pow(2.4, 4.3)
-    assert abs(i - 43.14280115650323) < eps
 
+    assert abs(pow(2.4, 4.3) - 43.14280115650323) < eps
+    assert abs(pow(2, 4) - 16) < eps
```
czgdp1807, line 65: What about using variables as input, expressions with multiply, divide, etc, recursive calls? Tests for `fabs` are better in this regard. Please do something similar for `pow` as well.

czgdp1807:
@Madhav2310 The tests pass. A slight fix was needed. To figure it out I executed `lpython src/runtime/math.py` and then it gave a type mismatch error along with the location of the line in the file. And pretty much it. Address https://github.com/lcompilers/lpython/pull/731#discussion_r913428791 and then we can merge it.

integration_tests/test_math.py
Madhav2310: Yes I will. I was unable to narrow down to the error as it kept giving ```Segfault: Signal SIGSEGV (segmentation fault) received``` error. on using the --show-asr command to debug as you had instructed

Madhav2310: I'll update this with more robust tests and once this pr is merged, I'll make another pr with error files for pow.

namannimmo10 (approved):

integration_tests/test_math.py
Madhav2310: This is done

czgdp1807: Well the tests are still not robust enough. But anyways, I have approved it. Waiting for @certik to review.

czgdp1807:
@certik Is this good to go?

certik:
Thanks @Madhav2310 for the PR and thanks @czgdp1807 for merging.

In the future, let's not merge PRs that have a history that is not polished. In this case, a lot of tests were added and removed a few times, which just increases the git repository size for no reason. It's also hard to review both in the PR and in the git history.

Once a PR is done and reviewed, then we need to polish the git history before merging. For simple PRs like this one, just a few simple commits will do.

If we are in a rush and we absolutely need this in, then squashing the commits from the github webinterface is also a better option in my opinion, than merging as is.

certik:
Ah, perfect. No problem then! Thank you.

PR #741, opened 2022-07-06
llvm refactor: IntegerToInteger cast case in Cast visitor 

namannimmo10:


czgdp1807:
Can you make the same PR/MR on LFortran repository? The reason being casting is more well tested there. So it would be good to run those CI tests with the changes here.

namannimmo10:
Corresponding MR: https://gitlab.com/lfortran/lfortran/-/merge_requests/1820

PR #742, opened 2022-07-06
Split ArrayRef into ArrayItem and ArraySection

czgdp1807:
Corresponding LFortran MR - https://gitlab.com/lfortran/lfortran/-/merge_requests/1819


src/libasr/ASR.asdl
```diff
@@ -263,7 +263,8 @@ expr
     | DictConstant(expr* keys, expr* values, ttype type)
     | DictLen(expr arg, ttype type, expr? value)
     | Var(symbol v)
-    | ArrayRef(symbol v, array_index* args, ttype type, expr? value)
+    | ArrayItem(symbol v, array_index* args, ttype type, expr? value)
+    | ArraySection(symbol v, array_index* args, ttype type, expr? value)
```
namannimmo10, line 267:
```suggestion
    | ArraySection(expr v, array_index* args, ttype type, expr? value)
```

czgdp1807: I wouldn't prefer to do this change in this PR. Fixing `ArrayItem` is the next stop in my priority list. So, I would focus on it once this gets in.

certik (approved):

PR #744, opened 2022-07-07
Implement string comparison

### LINKED ISSUE #197: String comparison
Currently string comparison such as `assert x = "abc"` doesn't work properly yet, I think we should implement the operation in LPython in the runtime library and the frontend should insert the call to it.

Abdelrahman-Kh-Fouad:
Fixes #197 

with basic testing `a == "sd"` it give me `Stored value type does not match pointer operand type!`, how to fix return type to be bool? or Is the problem something else?
```bash
  store i8* %4, i1* %s, align 8
 i1
```
@namannimmo10

### DIFF
--- a/integration_tests/test_str_comparison.py
+++ b/integration_tests/test_str_comparison.py
@@ -1,0 +1,36 @@
+def f():
+    s1: str = "abcd"
+    s2: str = "abcd"
+    assert s1 == s2
+    assert s1 <= s2
+    assert s1 >= s2
+    s1 = "abcde"
+    assert s1 >= s2 
+    assert s1 > s2
+    s1 = "abc"
+    assert s1 < s2 
+    assert s1 <= s2
+    s1 = "Abcd"
+    s2 = "abcd"
+    assert s1 < s2 
+    s1 = "orange"
+    s2 = "apple"
+    assert s1 >= s2 
+    assert s1 > s2 
+    s1 = "albatross"
+    s2 = "albany"
+    assert s1 >= s2
+    assert s1 > s2
+    assert s1 != s2
+    s1 = "maple"
+    s2 = "morning"
+    assert s1 <= s2
+    assert s1 < s2
+    assert s1 != s2
+    s1 = "Zebra"
+    s2 = "ant"	
+    assert s1 <= s2
+    assert s1 < s2
+    assert s1 != s2
+
+f()
--- a/src/runtime/impure/lfortran_intrinsics.c
+++ b/src/runtime/impure/lfortran_intrinsics.c
@@ -614,6 +614,53 @@ LFORTRAN_API void _lfortran_strcat(char** s1, char** s2, char** dest)
     *dest = &(dest_char[0]);
 }
 
+#define MIN(x, y) ((x < y) ? x : y)
+
+int str_compare(char **s1, char **s2)
+{
+    int s1_len = strlen(*s1);
+    int s2_len = strlen(*s2);
+    int lim = MIN(s1_len, s2_len);
+    int res = 0;
+    int i ;
+    for (i = 0; i < lim; i++) {
+        if ((*s1)[i] != (*s2)[i]) {
+            res = (*s1)[i] - (*s2)[i];
+            break;
+        }
+    }
+    res = (i == lim)? s1_len - s2_len : res;
+    return res;
+}
+LFORTRAN_API bool _lpython_str_compare_eq(char **s1, char **s2)
+{
+    return str_compare(s1, s2) == 0;
+}
+
+LFORTRAN_API bool _lpython_str_compare_noteq(char **s1, char **s2)
+{
+    return str_compare(s1, s2) != 0;
+}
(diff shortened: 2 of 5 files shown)


Abdelrahman-Kh-Fouad:
> You are using `lfortran_strop` to call into C, but it assumes that the third argument is of type `char**`. See `lfortran_strrepeat` for example, it assumes the second argument is of type `int32_t`. So I think you need a different function (maybe `lfortran_strcmp`) to handle string comparison.

i made `lfortran_str_cmp` to accept `int8` as a result and it didn't work, i tried to make functions return value like `str_len` (last commit) and it didn't work either and, they gave me same message 
```bash
  store i8 %3, i1* %s, align 1
 i1
```
is `i1` refere to void?

namannimmo10:
`i1` is a boolean value (`True` or `False`), a 1-bit integer.

namannimmo10:
This looks better now. Please add the tests covering all corner cases and test them with `cpython` and `llvm`.

namannimmo10:
You also need to add this file in https://github.com/lcompilers/lpython/blob/main/integration_tests/CMakeLists.txt with labels, `cpython` and `llvm`.

namannimmo10:
Please do a `git rebase` with the `main` branch rather than merge it with `main`. It leads to unnecessary commits, making a mess in the history. 
```console
$ git checkout main
$ git pull upstream main
$ git checkout stringsCompare
$ git rebase main   # if ever this leads to conflicts - resolve manually
```

Abdelrahman-Kh-Fouad:
> Please do a `git rebase` with the `main` branch rather than merge it with `main`. It leads to unnecessary commits, making a mess in the history.
> 
> ```
> $ git checkout main
> $ git pull upstream main
> $ git checkout stringsCompare
> $ git rebase main   # if ever this leads to conflicts - resolve manually
> ```

i rebased the branch as you said but, it didn't squash the branch.

namannimmo10:
You don't have to squash all the commits into one. Do `git rebase -i main` and it will show you the commits in an interactive mode. Now you can change `pick` to `s` to squash the commit, `d` to drop, and so on... For instance, you have three commits with the message, "Solve conflicts" that are not needed here, so change `pick` to `s`, so that we only have a relevant set of commits. When you're done, close the editor, and push it with `--force`.

tests/tests.toml
```diff
@@ -346,6 +346,12 @@ filename = "../integration_tests/print_01.py"
 c = true
 cpp = true
 
+[[test]]
+filename = "../integration_tests/test_str_comparison.py"
+llvm = true
+c = true
+cpp = true 
+
```
namannimmo10, line 354:
```suggestion
```
These are not needed. Just testing with `cpython` and `llvm` should be enough, I think.

Abdelrahman-Kh-Fouad: I deleted it.

namannimmo10:
You also need to delete those 6 generated files.

integration_tests/test_str_comparison.py
```diff
@@ -0,0 +1,33 @@
+def f():
+    s1: str = "abcd"
+    s2: str = "abcd"
+    assert s1 == s2
+    assert s1 <= s2
+    assert s1 >= s2
+    s1 = "abcde"
+    assert s1 >= s2
```
namannimmo10, line 8: A test for the `!=` operator would be nice too!

src/runtime/impure/lfortran_intrinsics.c
```diff
@@ -614,6 +614,53 @@ LFORTRAN_API void _lfortran_strcat(char** s1, char** s2, char** dest)
     *dest = &(dest_char[0]);
 }
 
+#define MIN(x, y) (((x) < (y)) ? (x) : (y))
```
namannimmo10, line 617:
```suggestion
#define MIN(x, y) ((x < y) ? x : y)
```

namannimmo10:
Please make these changes and it should be ready afterward.

integration_tests/out
```diff
@@ -0,0 +1,696 @@
+-- The C compiler identification is GNU 11.2.0
```
certik, line 1: I think this file should not be committed.

certik:
After everything is ready, let's create a simple commit or commits and rebased on top of the latest master.

Abdelrahman-Kh-Fouad:
> After everything is ready, let's create a simple commit or commits and rebased on top of the latest master.

it's ready now

certik (approved):
That looks good to me, I think we can merge it.

Would you mind creating a nice set of patches for this? Or do you want me to just squash it using github?

Abdelrahman-Kh-Fouad:
> Would you mind creating a nice set of patches for

i don't mind but, how to create a set of patches?

namannimmo10:
I squashed all the commits into one for now. @Abdelrahman-Kh-Fouad -- thanks a lot for working on this! If you still want to know how to squash commits, we can show you in our upcoming meeting.

PR #748, opened 2022-07-08
Raise error for dict section

namannimmo10:


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1868,6 +1868,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                 tmp = ASR::make_ListSection_t(al, x.base.base.loc, value, ai,
                         type, nullptr);
                 return;
+
```
Smit-create, line 1871: I don't think this is required here. It's readable without a new line.

certik (approved):

certik:
Thanks!

Issue #740, opened 2022-07-06
Raise an error for unsupported type annotation


namannimmo10:
```console
(lp) namannimmo@srf lpython % cat t.py
def f():
    i: i128
    i = 4
    print(i)

f()
(lp) namannimmo@srf lpython % lpython t.py
Traceback (most recent call last):
  File "/Users/namannimmo/Applications/oss/lpython/src/bin/lpython.cpp", line 864
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir, lpython_pass_manager, compiler_options, time_report);
  File "/Users/namannimmo/Applications/oss/lpython/src/bin/lpython.cpp", line 345
    r1 = LFortran::LPython::python_ast_to_asr(al, *ast, diagnostics, true,
  File "/Users/namannimmo/Applications/oss/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 3843
    auto res2 = body_visitor(al, *ast_m, diagnostics, unit, main_module,
  File "/Users/namannimmo/Applications/oss/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 3770
    b.visit_Module(ast);
  File "/Users/namannimmo/Applications/oss/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2335
    visit_stmt(*x.m_body[i]);
  File "/Users/namannimmo/Applications/oss/lpython/src/lpython/python_ast.h", line 1882
    void visit_stmt(const stmt_t &b) { visit_stmt_t(b, self()); }
  File "/Users/namannimmo/Applications/oss/lpython/src/lpython/python_ast.h", line 1750
    case stmtType::FunctionDef: { v.visit_FunctionDef((const FunctionDef_t &)x); return; }
  File "/Users/namannimmo/Applications/oss/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2362
    handle_fn(x, *ASR::down_cast<ASR::Subroutine_t>(t));
  File "/Users/namannimmo/Applications/oss/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2353
    transform_stmts(body, x.n_body, x.m_body);
  File "/Users/namannimmo/Applications/oss/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2310
    this->visit_stmt(*m_body[i]);
  File "/Users/namannimmo/Applications/oss/lpython/src/lpython/python_ast.h", line 1882
[...]

PR #749, opened 2022-07-09
fix issue 740.

redbopo:
fix [740](https://github.com/lcompilers/lpython/issues/740)

### DIFF
--- a/tests/reference/asr-test_unsupported_type-0d813dd.stderr
+++ b/tests/reference/asr-test_unsupported_type-0d813dd.stderr
@@ -1,0 +1,5 @@
+semantic error: Unsupported type annotation: i128
+ --> tests/errors/test_unsupported_type.py:2:5
+  |
+2 |     i: i128
+  |     ^^^^^^^ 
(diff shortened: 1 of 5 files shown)


tests/reference/asr-test_unsupported_type-0d813dd.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Unsupported type annotation: i128
+ --> tests/errors/test_unsupported_type.py:2:5
+  |
+2 |     i: i128
+  |     ^^^^^^^
```
namannimmo10, line 5: Can you try to improve the message like this?
```suggestion
2 |     i: i128
  |        ^^^^ 
```

certik: I think that can be done in a subsequent PR. This PR fixes the main problem to raise an error, which now works.

certik (approved):
I think this is fine. It changes the `symbol_get_past_external` function to pass through a null pointer. Alternatively we would add an assert here and fix it in the caller. I think this is fine for now.

redbopo:
Thanks @namannimmo10  and @certik , I will try a  better error message later.

Issue #745, opened 2022-07-07
Bug in builtin `str` function


namannimmo10:
```console
(lp) namannimmo@srf lpython % cat t.py
def f():
    print(str(45))



f()
(lp) namannimmo@srf lpython % lpython t.py
code generation error: List type not supported yet in ListAppend
 --> t.py:7:2706
  |
7 | 
  |                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).
```

Smit-create:
The error is raised from this line:
https://github.com/lcompilers/lpython/blob/0e39d6c1064e13415e4d433b3cb9efbf004dc478/src/runtime/lpython_builtin.py#L140

Because in llvm codegen we only support accessing elements from `list[i32]`. Since the implementation of `str(x: i32)` uses `list[str]`, we get the codegen error when we hit this line:
https://github.com/lcompilers/lpython/blob/0e39d6c1064e13415e4d433b3cb9efbf004dc478/src/libasr/codegen/asr_to_llvm.cpp#L1137-L1148

PR #750, opened 2022-07-09
LLVM: Fix error message in ListItem

Smit-create:
https://github.com/lcompilers/lpython/issues/745#issuecomment-1179528580


namannimmo10 (approved):
Thanks!

PR #754, opened 2022-07-10
Make `int()` work for non-constant strings

namannimmo10:


namannimmo10:
If you know a more robust way to do this (maybe in LLVM itself), please let me know.
CC: @czgdp1807 @Smit-create @certik

certik:
It has to be a runtime conversion for sure. See the discussion here:

* https://gitlab.com/lfortran/lfortran/-/merge_requests/1752

See also: https://github.com/lcompilers/lpython/pull/669#issuecomment-1161979090


Based on my [comment](https://gitlab.com/lfortran/lfortran/-/merge_requests/1752#note_961839380) it's a bit confusing where this should go. So far we have Cast (which changes the bits), BitCast (which does not change bit), we need StringOrd, StringChr, and then we also have this string integer to integer conversion. They probably can go into Cast, since they change bits.

integration_tests/test_str_to_int.py
```diff
@@ -6,5 +6,18 @@ def f():
...
+    assert i == 123
+    s = "-123"
+    i = int(s)
+    assert i == -123
+    s = "    1234"
+    i = int(s)
+    assert i == 1234
+    s = "    -1234   "
+    i = int(s)
+    assert i == -1234
```
czgdp1807, line 21: Let's check for something like, `int("2030.300")`. Supposed to give an error as follows,

```bash
Python 3.8.9 (default, Apr 13 2022, 08:48:06) 
[Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
>>> int("2030.300")
ValueError: invalid literal for int() with base 10: '2030.300'
```

namannimmo10: Unfortunately, this is very tricky. There are so many cases to consider. See how CPython does this:

https://github.com/python/cpython/blob/4b4439daed3992a5c5a83b86596d6e00ac3c1203/Objects/longobject.c#L2270-L2271

I suggest that we deal with this in a separate PR. Also, the changes in https://github.com/lcompilers/lpython/pull/591/files were incomplete.

namannimmo10:
Let me know if this looks good for now. I will also raise a corresponding MR to LFortran.

namannimmo10:
Corresponding MR: https://gitlab.com/lfortran/lfortran/-/merge_requests/1835.

integration_tests/test_str_to_int.py
faze-geek: The above input gives this error as expected, right ? It is a good test to have. I'll look into CPython examples for missed cases.

semantic error: invalid literal for int() with base 10: '2030.300'
 --> try.py:5:13
   |
 |     i = int("2030.300")

PR #759, opened 2022-07-13
CI: Make build/test faster

Smit-create:


certik (approved):
I think that's fine. Thanks!

Issue #762, opened 2022-07-13
Added support for step in string slicing


redbopo:
This issue related to [Draft_699](https://github.com/lcompilers/lpython/pull/699)
Make a new one here to record some ideas.

Current indexing like `s[start:end:step]` will get into intrinsics in `_lfortran_str_copy()`.

https://github.com/lcompilers/lpython/blob/2043379f67b6bbe2397500f6f881b9edb38684d2/src/runtime/impure/lfortran_intrinsics.c#L639-L644

It's fine to work with 

```
s = "abcdefgh"
print(s[8])
# To Raise Error "String index out of Bounds"
```

But deal with
```
s = "abcdefgh"
print(s[8::])
```
python will return str "",  lpython will raise same error mentioned.

I'd like to make a new intrinsics function like `_lfortran_str_step_copy()`  deal with step specific, which called by `visit_StringSection` .

PR #766, opened 2022-07-14
Support for step in string slicing

redbopo:
The draft originly in [699](https://github.com/lcompilers/lpython/pull/699) , I take over it here.
Some thought are list in [issue-762](https://github.com/lcompilers/lpython/issues/762)


certik:
Thanks for working on this!

redbopo:
cc @certik  and  @czgdp1807 .
And I'd like to  make another fix in lfortran_str_copy() function after this is merged,
since I move the part of slice copy with str_slice_copy() function,
and str_copy() contains redundant  variable to handle StringItem.

certik (approved):
Looks good to me. @czgdp1807 if this looks good to you, then go ahead and merge.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1911,25 +1913,31 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
                     ai.m_right = ASR::down_cast<ASR::expr_t>(
-                    ASR::make_IntegerConstant_t(al, loc, 0, int_type));
-                }
-                // If step is not present, assign it to 1 (step should be always present)
-                if (ai.m_step == nullptr) {
-                    ai.m_step = ASR::down_cast<ASR::expr_t>(
-                    ASR::make_IntegerConstant_t(al, loc, 1, int_type));
-                } else {
-                    ai.m_step = index_add_one(loc, ai.m_step);
+                    ASR::make_IntegerConstant_t(al, loc, step_forward ? MAX_INTEGER : MIN_INTEGER, int_type));
```
certik, line 1938: What is the logic here with regards the Min/Max integer? 

It doesn't feel right (it's using some arbitrary integer as a flag; what is the user actually puts such an integer in the Python code?). I think we should modify our ASR to accommodate the use case you are trying to fix.

redbopo: It deal with the start or end  not set like ` s[start:end:step]  s[::], s[:0:], s[::-1], s[:0:-1]` .
I move the m_step setting first to calculate, so we can also find it move forward or backward.
The previous method will set start to 0 ,  end to 1 if not set.  It's obvious wrong with str slice copy.
And in fact not like the str_copy in StrItem,  the start, end can be any number even if they are out of the boundary.
For example:
```
>>> s = "123456789"
>>> s[10::1]
''
>>> s[10:12:1]
''
>>> s[13:12:1]
''
>>> s[13:2:-1]
'987654'
>>> s[13:-102:-1]
'987654321'
>>> s[13:0:-1]
'98765432'
>>> 

```
So I think it can be a method set the max integer or min integer in INT32 to be its boundary.

Originnally I think we may left both of them with null pointer, but it will cause an building error. 
Probably it can be better if pass with nullptr success , but I don't know what to do with it now.
Seems that we should create a value with m_left , m_right.

certik: Thanks for the explanation. I need to think about this some more. Sorry for the delay, I'll get back to this soon.

czgdp1807: > Originnally I think we may left both of them with null pointer, but it will cause an building error.

In my opinion, a building error/segmentation fault due to `nullptr` is safer than incorrect outputs due to out of bounds. Since you have already set `start`/`end` to max-min INT32 adding or subtracting small numbers from them would lead to incorrect bounds which may lead to incorrect sliced strings. That would be very opaque and user may never know (if its a part of a very large codebase). `nullptr` instead will give segmentation fault always (so no chance of incorrect outputs) and we will be forced to handle this case.

redbopo: Thanks for review.
Yes, It causes an unexpected mistake due to the boundary problem.

```
s = "abcdefghijk"
print("s = ")
print(s)
print("s[2147483647+1::] = ")
print(s[2147483647+1::]) # expected to ""
print("s[2147483647::] = ")
print(s[2147483647::])

# ----
s = 
abcdefghijk
s[2147483647+1::] = 
abcdefghijk
s[2147483647::] = 
```

And  it's because the int32 range from -2147483648 to 2147483647.
If we edit "2147483647+1" as the start, it will be reset as "-2147483648" .
Since we're using `int_type =   ASR::make_Integer_t(al, loc, 4, nullptr, 0)`, this can not be avoided.
It will be a similar problem in INT64 which has a range boudary of integer.

But python suit this fine 
```
>>> s = "12345"
>>> s[2**62::]
''
>>> s[2**63::]
''
>>> s[-2**63::]
'12345'
>>> s[2**63+2::]
''

```

I don't have some ideas to handle this now,  please let me know if you have any  suggestions.

Smit-create:
Done in #853

PR #767, opened 2022-07-14
Make `str()` work for non-constant values

Abdelrahman-Kh-Fouad:
Related to #761 .
I have missed `ord()` on converting line.
there was a bug in reversing loop.


Abdelrahman-Kh-Fouad:
@certik @namannimmo10 
ready for reviewing.

src/runtime/lpython_builtin.py
```diff
@@ -135,11 +135,11 @@ def str(x: i32) -> str:
     rev_result_len: i32
     rev_result_len = 0
     while x > 0:
-        rev_result += chr(x - _lpython_floordiv(x, 10)*10)
+        rev_result += chr(ord('0') + (x - _lpython_floordiv(x, 10) * 10))
```
namannimmo10, line 138:
```suggestion
        rev_result += chr(48 + (x - _lpython_floordiv(x, 10) * 10))
```

namannimmo10:
To make sure that this works with LLVM, please rebase with `main`.

namannimmo10:
Thanks. Can you also make `str()` work for floats and other inputs?

PR #775, opened 2022-07-18
Supporting For in looping over strings

Shaikh-Ubaid:
towards #206 , #732 

This implements the approach suggested in this [comment](https://github.com/lcompilers/lpython/issues/732#issuecomment-1174866795)

Issue #732, opened 2022-07-04
Add support for loop iteration


Madhav2310:
Python Code: 
```
s: str
s = "samplestring"
char : str

for char in s:
    if ord(char) >= 97 and ord(char) <=122 :
        result += chr(ord(char) - 32)
    else :
        result += char
```

Error:
```
semantic error: Only function call `range(..)` supported as for loop iteration for now
   --> /home/madhav/lpython/src/runtime/lpython_builtin.py:156:5 - 160:26
    |
156 |        for char in s:
    |        ^^^^^^^^^^^^^^...


```

Madhav2310:
It is supported for lists i think, not totally sure. https://github.com/lcompilers/lpython/blob/4591db8b087577046f548d21097d8c8c23fcb2f7/src/runtime/ltypes/ltypes.py#L76

Madhav2310:
@certik @czgdp1807 @namannimmo10

namannimmo10:
Similar to #206.

Madhav2310:
> Similar to #206.

Right, yes. Who can I discuss with on how to implement this?

czgdp1807:
What I would do here is, convert the following,

```python
for char in s:
    if ord(char) >= 97 and ord(char) <=122 :
        result += chr(ord(char) - 32)
    else :
        result += char
```

to,

```python
for i in range(len(s)):
    char = s[i]
    if ord(char) >= 97 and ord(char) <=122 :
        result += chr(ord(char) - 32)
    else :
        result += char
```

We can do it simply in the AST to ASR transition itself (in the node below). However we won't be able to get back the original code after this (AST to Source code transition).

https://github.com/lcompilers/lpython/blob/989b0a2c5442e4b0452ff0cdaf3b87ed872058bf/src/lpython/semantics/python_ast_to_asr.cpp#L2605-L2687

@certik @Madhav2310 What do you think?

Madhav2310:
I don't understand the bit about AST to ASR transition, I'll look into it! But yes I'll go ahead and implement the python code that way to implement iterate over the strings. Thank you!

Issue #206, opened 2022-03-08
Implement looping through a string


namannimmo10:
Reference code:
```python
def main0():
    s: str
    s = 'aabbcc'
    c: str
    for c in s:
        print(c)

main0()
```
ASR:
```
semantic error: Only function call `range(..)` supported as for loop iteration for now
 --> c.py:5:5 - 6:16
  |
5 |        for c in s:
  |        ^^^^^^^^^^^...
...
  |
6 |            print(c)
  | ...^^^^^^^^^^^^^^^^
```

Smit-create:
This could be generalised for all `iterable` objects like tuples, list, etc.

Smit-create:
This works now:
```
% cat a.py
def main0():
    s: str
    s = 'aabbcc'
    c: str
    for c in s:
        print(c)

main0()
lpython % lpython a.py
a
a
b
b
c
c

```

PR #781, opened 2022-07-18
C, CPP: Strings

Shaikh-Ubaid:
towards #206 , #732 

This is a continuation of #775 . Only the recent `5` commits are part of this `PR`, the rest of the commits are part of #775 . I will rebase this after #775 is merged.


Shaikh-Ubaid:
The tests, currently do/might not pass. It seems there is a following minor issue/challenge. 

For the following code (`integration_tests/loop_01.py`)
```py
def main0():
    s: str
    s = 'aabbcc'
    c: str
    for c in s:
        print(c)

main0()
```
the `--show-c` output is as follows (displaying only relevant part for simplicity):
```c
void main0()
{
    int32_t __explicit_iterator;
    char * c;
    char * s;
    s = "aabbcc";
    for (__explicit_iterator=0; __explicit_iterator<=strlen(s) - 1; __explicit_iterator++) {
        c = s[__explicit_iterator + 1 - 1]; // line causing failure
        printf("%s\n", c);
    }
}
```
In the above `c` code, we see that on line `c = s[__explicit_iterator + 1 - 1];`, the type of the target is `char *` where as the type of value is `char`, which I guess is causing the failure. In `Python`, `char` and `str` are similar/same but in `c` both are treated differently. Please, could someone possibly share how we should go about `str` and `char` assignments for the `C` Backend?

Shaikh-Ubaid:
In the `C` Backend, if we are indexing a string
**for example:** `element = a[i]`, where `a` is of type `char *` and `i` is of type `int`

**doubt:** what should be the type of the `element`? Should it be `char` or `char *`? 

Essentially, the doubt is, in the `C` Backend, are we considering `characters` as `char` or as strings of length `1`? 

Please possibly clarify.

certik:
ASR currently has a notion of a String, but not a Character. That is, character is a string of length 1. Just like in Python and Fortran.

How the backend implements it is solely left to the backend. The easiest I think is to treat characters as `char *`, strings of length 1. Then it is consistent and everything should just work. An optional optimization could be to use just `char` if you know at compile time (in the backend) that the length is 1, but then you have to adjust all accesses to such a character. So it's more complicated, but it can definitely be done. I personally would not bother with such optimizations, at least not yet. I would focus on the big fish, that is, ASR->ASR optimizations. And leave these "micro-optimizations" to LLVM (for now), or the C compiler. That will give us the best bang for our buck. Later, we can implement these micro-optimizations too.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -549,7 +549,11 @@ R"(#include <stdio.h>
         if (ASR::is_a<ASR::Var_t>(*x.m_target)) {
             visit_Var(*ASR::down_cast<ASR::Var_t>(x.m_target));
             target = src;
-            if( ASRUtils::is_array(ASRUtils::expr_type(x.m_target)) && !is_c ) {
+            // auto v = ASRUtils::EXPR2VAR(x.m_target);
+            // if (is_c && ASRUtils::is_character(*v->m_type)) {
+            //     target = "free(" + target + "); " + target;
+            // }
```
Shaikh-Ubaid, line 555: Using `free` here fails few tests. 
```bash
90% tests passed, 3 tests failed out of 29

Label Time Summary:
c          =   0.71 sec*proc (29 tests)
cpython    =   0.70 sec*proc (21 tests)
llvm       =   0.71 sec*proc (29 tests)

Total Test time (real) =   0.73 sec

The following tests FAILED:
         10 - print_01 (Subprocess aborted)
         18 - loop_01 (Subprocess aborted)
         21 - test_c_interop_02 (Subprocess aborted)
```

These tests have string allocated in `stack memory` and I guess calling `free` on `stack memory` is causing the failure.

src/libasr/codegen/asr_to_c.cpp
```diff
@@ -634,6 +635,20 @@ R"(
         src = out;
     }
 
+    void visit_StringItem(const ASR::StringItem_t& x) {
+        this->visit_expr(*x.m_idx);
+        std::string idx = std::move(src);
+        this->visit_expr(*x.m_arg);
+        std::string str = std::move(src);
+        src = "(char *)memcpy((char *)calloc(2U, sizeof(char)), "
+                + str + " + " + idx + " - 1, 1U)";
```
Shaikh-Ubaid, line 644: For the `C` Backend, currently, I am allocating memory into heap using `calloc`. Ideally, I guess, the same memory should be freed. I tried freeing it [here](https://github.com/lcompilers/lpython/pull/781#discussion_r928645818).

certik:
The deallocation I think has to be done via the ImplicitDeallocate ASR node, that LFortran correctly inserts for allocatable arrays. So we have to insert these nodes correctly, and then call "free" in the backend when we visit this node.

certik:
What is the status of this --- ready for review?

Shaikh-Ubaid:
Inserting the `ImplicitDeallocate` ASR node is still left. I am sorry, I couldn't look into. I will work on it soon.

certik:
The deallocation is a bigger task, which I would recommend to tackle in a separate PR. Just polish this PR, resolve conflicts, and we can merge it after review.

Shaikh-Ubaid:
Strings currently leak memory in the `C` Backend. Currently, in `visit_StringItem`, a space for strings of length `2` is allocated on heap which is never deallocated.

Shaikh-Ubaid:

certik (approved):
Thanks! Go ahead and merge it.

PR #798, opened 2022-07-20
Adding support for `np.reshape`

czgdp1807:


certik (approved):
I think that this looks good!

I don't see any issues. Go ahead and merge. Thanks!

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -64,4 +64,59 @@ namespace LFortran {
         return LLVM::CreateInBoundsGEP(*builder, ptr, idx_vec);
     }
 
+    llvm::Type* LLVMUtils::getIntType(int a_kind, bool get_pointer) {
```
namannimmo10, line 67: We have this in `asr_to_llvm`: 
https://github.com/lcompilers/lpython/blob/0a9ee91213f7bfb9a0798ef3530980a59813f0ed/src/libasr/codegen/asr_to_llvm.cpp#L411
So it would be nice to drop that and do `llvm_utils->getIntType(kind)` everywhere in that file.

namannimmo10: Perhaps we can do that in a subsequent PR.

certik: Yes, I think there is a large opportunity to clean up our "ASR API" in `libasr`. We'll just do it iteratively.

czgdp1807: We can do the replacements. Possibly all the `getXYZType` can go into `LLVMUtils` class and then called from there.

PR #799, opened 2022-07-20
[New] Parse function call in the Subscription

### LINKED ISSUE #792: [New] Subscript for Function Call is not recognized
```py
$ lpython --show-ast examples/expr2.py --tree --new-parser
syntax error: Token '[' is unexpected here
 --> examples/expr2.py:1:22
  |
1 | return get_versions()["version"]
  |                      ^ 
```

Thirumalai-Shaktivel:
Closes #792


akshanshbhatt:
```py
func()[1][1]
```

We should also add support for statements like this.

Thirumalai-Shaktivel:
Thanks for the test. I parsed the Nested Subscript as well.

certik:
The conflicts must be resolved.

certik (approved):
I think this is fine.

PR #806, opened 2022-07-22
Supporting `np.vectorize` a.k.a `elemental`

czgdp1807:


czgdp1807:

src/libasr/ASR.asdl
```diff
@@ -88,7 +88,7 @@ symbol
         abi abi, access access, deftype deftype, string? bindc_name, bool pure,
         bool module)
     | Function(symbol_table symtab, identifier name, expr* args, stmt* body,
-        expr return_var, abi abi, access access, deftype deftype,
+        expr return_var, abi abi, access access, deftype deftype, bool elemental,
```
certik, line 91: Note: a subroutine can also be elemental, but let's do that later.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -34,9 +34,8 @@ namespace LFortran::LPython {
 // * First the current directory (this is incorrect, we need to do it relative to the current file)
 // * Then the LPython runtime directory
 LFortran::Result<std::string> get_full_path(const std::string &filename,
-        const std::string &runtime_library_dir, bool &ltypes, bool &numpy) {
+        const std::string &runtime_library_dir, bool &ltypes) {
```
certik, line 37: Why was the `numpy` removed? I think we'll need it. But we can put it back later.

certik (approved):
Awesome! Thanks for doing this. I think this can be merged as is.

src/libasr/ASR.asdl
czgdp1807: Yes. Planning to do that later. Didn't want to make changes which were out of scope.

src/lpython/semantics/python_ast_to_asr.cpp
czgdp1807: `numpy` (and similarly `ltypes`, see the check below) was being used to completely ignore any import being made from these modules.

https://github.com/lcompilers/lpython/blob/04bf5b7cba1003ea165062aa3886169c77b1f04a/src/lpython/semantics/python_ast_to_asr.cpp#L111-L112

https://github.com/lcompilers/lpython/blob/04bf5b7cba1003ea165062aa3886169c77b1f04a/src/lpython/semantics/python_ast_to_asr.cpp#L2296-L2302

That's  why we were able to do `from numpy import empty, reshape`, even though `lpython_intrinsic_numpy.py` is empty. However since now we have added `sin`, `cos` in `lpython_intrinsic_numpy.py` and we actually need to import these functions from `lpython_intrinsic_numpy.py`. So, I removed `numpy` from here (because we cannot ignore `sin`, `cos` and other elemental functions imports). However you can still ignore the imports from `numpy` by specifying the functions/symbols, which you want to ignore, in `ProceduresDatabase`. In future if you figure out some python implementation for any NumPy function, just remove it from `ProceduresDatabase.to_be_ignored` map and LPython will start importing from `lpython_intrinsic_numpy.py`.

certik: Ok!

Issue #785, opened 2022-07-19
[New] Yield is not recognized in the expression


Thirumalai-Shaktivel:
Example:
```py
def test():
    if hasattr(i, "__parameters__"):
        yield from i.__parameters__
    elif isinstance(i, TypeVar):
        yield i
```
Error:
```console
$ lpython --show-ast examples/expr2.py --tree --new-parser
syntax error: Token 'yield' is unexpected here
 --> examples/expr2.py:3:9
  |
3 |         yield from i.__parameters__
  |         ^^^^^ 

```

krshrimali:
Hi, @Thirumalai-Shaktivel - If you haven't started (or plan to) working on it already, is it okay if I pick this up and try resolving it?

Thirumalai-Shaktivel:
Sure! Go ahead, work on it and submit a PR.

PR #816, opened 2022-07-24
[New Parser] Add support for `yield`

krshrimali:
Fixes https://github.com/lcompilers/lpython/issues/785

This PR:

* Adds support for `yield` for the new parser.
* Adds relevant tests + registers the same in `tests.toml`.

This adds support for the following variants of using `yield`:

```python
yield () # Empty tuple
yield (x)
yield # Empty yield
yield x
```

**Sample Code for testing:**

```python
def get():
    i: list[i32]
    i = [1, 2, 3]
    x: i32
    for x in i:
        yield x
```

**AST Generated:**

```python
(Module [(FunctionDef get ([] [] [] [] [] [] []) [(AnnAssign (Name i Store) (Subscript (Name list Load) (Name i32 Load) Load) () 1) (Assign [(Name i Store)] (List [(ConstantInt 1 ()) (ConstantInt 2 ()) (ConstantInt 3 ())] Load) ()) (AnnAssign (Name x Store) (Name i32 Load) () 1) (For (Name x Store) (Name i Load) [(Expr (Yield (Name x Load)))] [] ())] [] () ())] [])
```

cc: @Thirumalai-Shaktivel @certik


czgdp1807:
Supporting this in backends would be an interesting problem. Reason being neither LLVM nor C/C++ support yield at their syntax levels (AFAIK). We will be required to store the state of the function in a temporary struct and then exit the function when yield is executed. When the function is called again then continue from the previous state. If the previous state is NULL then start the function body execution from scratch. A combination of goto statements might be helpful but I think storing the state of the function would be inevitable. Doing this in a general way would be a bit tricky (AFAICT) as it depends on the semantics of the function body (in other words we will be required to know what the function actually wants to do in order to figure out what to store in its state). Not necessary may be, we can figure out a generic strategy to find the items to be stored in the state of the function. After all Python implements yield in a generic fashion so we should also be able to do the same just would be a little bit tricky.

P.S. I encountered a similar problem of implementing yield functionality in cython in https://github.com/scipy/scipy/pull/14540/files#diff-1de80bb633c0dfde7628093eace96ef527a450a2d005a8d97c548374b21ef2e0R958 and I used a similar strategy (which I described above) there to achieve the same.

certik (approved):

certik:
Thanks @krshrimali !

@czgdp1807 I opened up #819 for how to implement it in the backend.

PR #818, opened 2022-07-24
Fix lpython_floordiv(x: i32,y: i32) for 9 digit inputs

### LINKED ISSUE #768: Bug in `_lpython_floordiv(x: i32, y: i32)`
with `i32` numbers with 9 digits it return wrong answers.
```bash
$ cat a.py
def ff():
    x: i32
    y :i32
    y = 10
    x = 123456789
    print(_lpython_floordiv(x, y))

ff()
$ cat b.py
def ff():
    x: i64
    y :i64
    y = 10
    x = 123456789
    print(_lpython_floordiv(x, y))
ff()
$ ./a
12345679
$ ./b
12345678

```

faze-geek:
Fixes #768 
```
def main():
    x: i32
    y :i32
    y = 10
    x = 123456789
    print(_lpython_floordiv(x, y))

main()
12345678     #  gave 12345679 earlier which is incorrect
```


integration_tests/test_lpython_floordiv
```diff
@@ -0,0 +1,10 @@
+from ltypes import i32
+
+def main():
+    x : i32
+    y : i32
+    y = 10
+    x = 123456789
+    assert(_lpython_floordiv(x, y)) == 12345678
```
namannimmo10, line 8:
```suggestion
    assert x//y == 12345678
```

integration_tests/test_lpython_floordiv
```diff
@@ -0,0 +1,10 @@
...
+
+main()
```
namannimmo10, line 10:
```suggestion
main()

```

integration_tests/test_lpython_floordiv
faze-geek: Thanks @namannimmo10. I have made the changes !

namannimmo10:
You need to register this test in `integration_tests/CMakeLists.txt` with the labels, `cpython` and `llvm`.

faze-geek:
I can see that the indentation is wrong, is there a certain count of whitespaces that should be added between file name and `Labels` ?

namannimmo10:
No, you forgot to add the `.py` extension :)

src/runtime/lpython_builtin.py
```diff
@@ -481,7 +481,7 @@ def _lpython_floordiv(a: f32, b: f32) -> f32:
 
 @overload
 def _lpython_floordiv(a: i32, b: i32) -> i32:
-    r: f32
+    r: f64
     r = a/b
     result: i32
     result = int(r)
```
czgdp1807, line 487: Why don't we use `i64` as the type of `r`?

faze-geek: It won't change the result in this case, although the `r ` variables in all overloaded functions are of floating point `f32/f64` type.

faze-geek: This lead to another error being spotted - 

```
def main():
    x : f32
    y : f32
    y = 10.0
    x = 123456789.0
    print(x//y)

12345679.000000
You have new mail in /var/mail/user
12345678.0
```

czgdp1807: I don't think type of `r` matters here. `a` is `i32` and `b` is `i32` so `a/b` will anyways be `i32`. In the low level code it would just get casted to the type of `r` before storing. The final answer is already computed when `a/b` executes. Then it doesn't matter where you store it, in fact you can directly return `a/b` (conceptually, if lpython gives error then it should be fixed).

Try seeing the LLVM code being generated for the new error you found. The bug might be resolved if we carefully track what's happening at each step of code generation (AST, ASR, LLVM).

faze-geek: >  The final answer is already computed when `a/b` executes. Then it doesn't matter where you store it, in fact you can directly return `a/b` (conceptually, if lpython gives error then it should be fixed).

If `r` type is irrelevent then the current implementation has loopholes because as of now you can expect different answers as per the type of r. Also directly returning `a/b` as you mentioned gives error in lpython.
> Try seeing the LLVM code being generated for the new error you found. 

Okay thanks.

czgdp1807:
Any progress on this?

czgdp1807:
Closing this in favour of https://github.com/lcompilers/lpython/pull/1067.

src/runtime/lpython_builtin.py
czgdp1807: > If r type is irrelevent then the current implementation has loopholes because as of now you can expect different answers as per the type of r.

Yes, you are right, `r` type is relevant. It should be kept to the highest possible floating point kind. That way `float(a)/float(b)` won't result in rounding up and getting stored in `r`. By using anything lower than `f64` causes rounding up which leads to errors in final integer result. Suggested the same in https://github.com/lcompilers/lpython/pull/1067#discussion_r959185801

PR #771, opened 2022-07-15
Do_loop increment variable generalized

Madhav2310:
Support for #770

PR #822, opened 2022-07-26
Fix do loop variable type

### LINKED ISSUE #770: DoLoops not working for i64 increment variable
For the following code: 
```
    i: i64
    for i in range(len(s)):
        char = s[i]
        if ord(char) >= 65 and ord(char) <=90 :
            result += chr(ord(char) + 32)
        else :
            result += char
```
This works for `i: i32` but not for `i: i64`

Smit-create:
Fixes #770 
Closes #771

### DIFF
--- a/integration_tests/loop_01.py
+++ b/integration_tests/loop_01.py
@@ -5,4 +5,13 @@ def main0():
     for c in s:
         print(c)
 
+def test_issue_770():
+    i: i64
+    res: i64
+    res = 0
+    for i in range(10):
+        res += i
+    assert res == 45
+
 main0()
+test_issue_770()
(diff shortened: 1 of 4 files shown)


czgdp1807:
There are a couple of places (AFAICT) where the `ASR::DoLoop` is generated programatically (i.e., by calling `ASR::make_DoLoop_t`). There the kind of increment variable is 32 bit but it should be taken from the kind of iterator variable. Can you just check for those and add a fix if necessary? Like search for `make_DoLoop_t` and read the code to see if things are correct.

namannimmo10 (approved):
LGTM. Thank you 👍

czgdp1807 (approved):

PR #826, opened 2022-07-27
Recognize TypeIgnore and store in the AST

### LINKED ISSUE #788: [Old] Type comment Issue
```py
def _256Bit(NBitBase):  # type: ignore[misc]
    pass
```
Error:
```py
$ lpython --show-ast examples/expr2.py --tree
  File "/home/thirumalai/Open_Source/lpython/src/bin/../runtime/lpython_parser.py", line 110, in <module>
    a2 = v.visit(a)
  File "/home/thirumalai/conda_root/envs/lf/lib/python3.10/ast.py", line 410, in visit
    return visitor(node)
  File "/home/thirumalai/Open_Source/lpython/src/bin/../runtime/lpython_parser.py", line 77, in generic_visit
    new_list.append(self.visit(item))
  File "/home/thirumalai/conda_root/envs/lf/lib/python3.10/ast.py", line 410, in visit
    return visitor(node)
  File "/home/thirumalai/Open_Source/lpython/src/bin/../runtime/lpython_parser.py", line 100, in generic_visit
    new_node.first = linecol_to_pos(node.lineno, node.col_offset+1, newlines)
AttributeError: 'TypeIgnore' object has no attribute 'col_offset'
The command 'python /home/thirumalai/Open_Source/lpython/src/bin/../runtime/lpython_parser.py examples/expr2.py examples/expr2.py62D3531' failed.
  Binary file "/home/thirumalai/Open_Source/lpython/src/bin/lpython", in _start()
  File "./csu/../csu/libc-start.c", line 392, in __libc_start_main_impl()
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main()
  File "/home/thirumalai/Open_Source/lpython/src/bin/lpython.cpp", line 827, in ??
    return emit_ast(arg_file, runtime_library_dir, compiler_options);
AssertFailed: diagnostics.has_error()
```

Thirumalai-Shaktivel:
Closes #788


src/lpython/parser/semantics.h
```diff
@@ -258,8 +258,29 @@ static inline char *extract_type_comment(Allocator &al, LFortran::Str &s) {
...
+            break;
+        }
+    }
+
+    if (flag) {
+        pos = 6;
+        str = str.substr(pos, str.size());
+        s.from_str_view(str);
+        p.type_ignore.push_back(p.m_a, down_cast<type_ignore_t>(
+            make_TypeIgnore_t(p.m_a, loc, 0, s.c_str(p.m_a))));
```
Thirumalai-Shaktivel, line 278: cc @certik.
Do we store the Line number information in the `tokenizer`? As I can see only the start and end of the token information in the `Location`.
Here we have to pass Line number instead of `0`

src/runtime/lpython_parser.py
```diff
@@ -96,7 +96,7 @@ def generic_visit(self, node):
                 raise Exception("Unsupported value type")
         new_ast = getattr(python_ast, class_name)
         new_node = new_ast(**d)
-        if hasattr(node, "lineno"):
+        if hasattr(node, "col_offset"):
```
Thirumalai-Shaktivel, line 99: This change was made to fix the Old Parser issue: https://github.com/lcompilers/lpython/issues/788#issue-1309054967

src/lpython/parser/semantics.h
certik: We can compute the corresponding line using the LocationManager class from the `loc` variable. For now just set 0, we can figure out later how to do this -- either after AST is constructed, or (probably better), we pass in the location manager into the `Parser` class and then access it using `p.m_location_manager` above, and convert the linear location `start` into line+column.

certik: We do not store the `lineno` members, we use our location information instead. I think this `lineno` is there specified manually (instead of using the `attributes` part in ASDL). I think we can just remove it from `Python.asdl`, and then we don't have to set it to anything. The full location is stored there using the `loc` approach.

certik (approved):

src/lpython/parser/semantics.h
Thirumalai-Shaktivel: I left `TypeIgnore` unchanged in the `Python.asdl`, because it causes the Old Parser to fail.

Thirumalai-Shaktivel:
Thanks for the review!

src/lpython/parser/semantics.h
certik: Yes, the old parser would also need to be modified.

PR #837, opened 2022-07-28
Testing and fixing array operation expressions

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/868

### DIFF
--- a/integration_tests/array_expr_01.py
+++ b/integration_tests/array_expr_01.py
@@ -1,0 +1,41 @@
+from ltypes import i32, f32, f64
+from numpy import empty, reshape, int32, float64
+
+def array_expr_01():
+    dim1: i32
+    dim2: i32
+    dim3: i32
+    dim1d: i32
+    i: i32
+    shape1d: i32[1] = empty(1, dtype=int32)
+    shape3d: i32[3] = empty(3, dtype=int32)
+    eps: f64
+    eps = 1e-12
+
+    dim1 = 10
+    dim2 = 10
+    dim3 = 5
+    dim1d = dim1 * dim2 * dim3
+
+    e: f64[10, 10, 5] = empty((dim1, dim2, dim3), dtype=float64)
+    f: f64[10, 10, 5] = empty((dim1, dim2, dim3), dtype=float64)
+    g: f64[500] = empty(dim1d, dtype=float64)
+    e1d: f64[500] = empty((dim1d), dtype=float64)
+    f1d: f64[500] = empty((dim1d), dtype=float64)
+
+    for i in range(dim1d):
+        e1d[i] = float(i + 1)
+        f1d[i] = float(i + 1)
+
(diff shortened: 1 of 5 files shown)


integration_tests/array_expr_01.py
```diff
@@ -0,0 +1,42 @@
...
+    eps: f64
+    eps = 1e-12
+
+    dim1 = 10
+    dim2 = 10
+    dim3 = 5
+    dim1d = dim1 * dim2 * dim3
+
+    # TODO: Replace with dim1, dim2, dim3 in array type annotations
+    e: f64[10, 10, 5] = empty((dim1, dim2, dim3))
```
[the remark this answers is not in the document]
certik: Arguements are allowed. Variables cannot be allowed, because, as you said, they are not initialized.

Constant variables / parameters can be allowed. They are allowed in LFortran. LPython currently does not support constant variables, but eventually it will, and then we can use those as well.

integration_tests/array_expr_01.py
```diff
@@ -0,0 +1,42 @@
...
+from numpy import empty, reshape
+
+def array_expr_01():
+    dim1: i32
+    dim2: i32
+    dim3: i32
+    dim1d: i32
+    i: i32
+    shape1d: i32[1] = empty(1, dtype=int)
+    shape3d: i32[3] = empty(3, dtype=int)
```
certik, line 11: I think we have to set these types to int32, where `int32` is imported from NumPy.
```suggestion
    shape1d: i32[1] = empty(1, dtype=int32)
    shape3d: i32[3] = empty(3, dtype=int32)
```

integration_tests/array_expr_01.py
certik, line 21: This might be the default for `empty`, so not strictly needed, but should not hurt:
```suggestion
    e: f64[10, 10, 5] = empty((dim1, dim2, dim3), dtype=float64)
```

integration_tests/array_expr_02.py
```diff
@@ -0,0 +1,37 @@
...
+        assert abs(array_a[i]**2 + sqrt(array_b[i]) - result[i]) <= eps
+
+def f():
+    i: i32
+    j: i32
+    k: i32
+
+    array_a: f32[256] = empty(256)
+    array_b: f32[256] = empty(256)
+    array_c: f32[256] = empty(256)
```
certik, line 25: Here we need this:
```suggestion
    array_a: f32[256] = empty(256, dtype=float32)
    array_b: f32[256] = empty(256, dtype=float32)
    array_c: f32[256] = empty(256, dtype=float32)
```
Otherwise CPython would use a different type than LPython.

czgdp1807:
@certik I have addressed your comments. Is there anything else to be updated here?

certik (approved):
Looks great!

PR #839, opened 2022-07-28
Remove redundant `CharacterToCharacter` cast kind

namannimmo10:


namannimmo10:
@Abdelrahman-Kh-Fouad -- I don't see any error here. Can you review this?

Abdelrahman-Kh-Fouad:
> @Abdelrahman-Kh-Fouad -- I don't see any error here. Can you review this?

it works fine as you said, but where is the main implementation for `CharacterToCharacter`?

Abdelrahman-Kh-Fouad:
> @Abdelrahman-Kh-Fouad -- I don't see any error here. Can you review this?

yes it works fine as you said.

PR #844, opened 2022-07-29
Parse type ignore in class definition

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
This PR is good to go in, so I'm merging this.

PR #853, opened 2022-07-30
Support string slicing

### LINKED ISSUE #610: To reverse strings using slicing

### LINKED ISSUE #762: Added support for step in string slicing

Smit-create:
Fixes #762 
Fixes #610

Closes #766, #699


certik (approved):
Overall I think this is great.

One note: the `start`, `end` and `step` is now optional in `StringSection`. That means we need to do all kinds of logic in all backends to determine the actual range. Since we know it is always a section and not an item (represented by `StringItem`), the prior design was that `start`, `end` and `step` are always mandatory, and always properly filled in (with `1`, `0`, `StringLen(x)-1` or whatever) and in the backend you don't have to do any kind of logic, you just implement `start:stop:step` explicitly.

We can try this change in a subsequent PR, so that we can see exactly the impact, but it seems to me it would be a better design. The only possible downside is that in some corner case, the `StringLen(x)` might be evaluated twice (not actually sure if that is the case or not), but in that case, our optimizer (to be done!) should be able to recognize this at the ASR level and optimize it out. Also if we choose to represent strings by `(length, pointer)` tuple, it might not even be a problem.

certik:
So we can merge now, and improve upon it later.

PR #854, opened 2022-07-31
Simplify the grammar rules of the parameter_list in Function_def

Thirumalai-Shaktivel:
Reference: https://docs.python.org/3/reference/compound_stmts.html#function-definitions


certik (approved):
Looks good!

Thirumalai-Shaktivel:
Thanks for the review!

PR #797, opened 2022-07-19
Add `str()` to casting

Abdelrahman-Kh-Fouad:

PR #856, opened 2022-07-31
BUG: Fix bool to str

Smit-create:
cc @Abdelrahman-Kh-Fouad @certik


czgdp1807 (approved):

czgdp1807:
Yeah. This is the way for now. Will abstract this stuff out later.

namannimmo10:
I don't see any bug in `main`:

```console
def f():
    var: bool
    var = True
    assert str(var) == "True"
    var = False
    assert str(var) == "False"
    assert str(True) == "True"


f()
```

namannimmo10:
I merged Abdel's PR and he made sure this test worked:
https://github.com/lcompilers/lpython/pull/797/files

certik:
We should add the test. If master already passes with this test, then I guess we don't need to add the change in `array_op`.

czgdp1807:
This test is not the same as what @Abdelrahman-Kh-Fouad posted on Zulip. Can you try the exact thing that he posted so that we are sure before concluding this PR?

czgdp1807:
@Smit-create I think your test is also not the same as one which @Abdelrahman-Kh-Fouad reported the bug with? Can you add his test as well? Can you confirm that error is thrown in main branch with which test (the one in your PR right now or the one reported on Zulip)?

Abdelrahman-Kh-Fouad:
```bash 
$cat a.py
def f():
    f: bool
    f = True
    assert str(f) == "True"
    assert str(True) == "True"

f()
```
 with above example everything goes well.
but when we remove any statement that contains `if\else` some functions like `print` and located before `str(True)`, the error appear(the same one i posted on zulip):
```bash
$ cat a.py
def f():
    f: bool
    f = True
    assert str(True) == "True"
f()
```
or 

```bash
$ cat a.py
def f():
    f: bool
    f = True
    assert str(True) == "True"
    assert str(f) == "True"
f()
```

Abdelrahman-Kh-Fouad:
i thing everything in stmt in asr, if i put any of them (and `assert` one of them)before casting it will run well. 
```asdl
stmt
    = Allocate(alloc_arg* args, expr? stat, expr? errmsg, expr? source)
    | Assign(int label, identifier variable)
    | Assignment(expr target, expr value, stmt? overloaded)
    | Associate(expr target, expr value)
    | Cycle()
    -- deallocates if allocated otherwise throws a runtime error
    | ExplicitDeallocate(symbol* vars)
    -- deallocates if allocated otherwise does nothing
    | ImplicitDeallocate(symbol* vars)
    | DoConcurrentLoop(do_loop_head head, stmt* body)
    | DoLoop(do_loop_head head, stmt* body)
    | ErrorStop(expr? code)
    | Exit()
    | ForAllSingle(do_loop_head head, stmt assign_stmt)
        -- GoTo points to a GoToTarget with the corresponding target_id within
        -- the same procedure. We currently use `int` IDs to link GoTo with
        -- GoToTarget to avoid issues with serialization.
    | GoTo(int target_id)
        -- An empty statement, a target of zero or more GoTo statements
        -- the `id` is only unique within a procedure
    | GoToTarget(int id)
    | If(expr test, stmt* body, stmt* orelse)
    | IfArithmetic(expr test, int lt_label, int eq_label, int gt_label)
    | Print(expr? fmt, expr* values, expr? separator, expr? end)
...
```
i'm not sure, but this what i got from testing, we can open issue for that.

Smit-create:
This is the failing test on main:
```py
def ff():
    print(str(True))

ff()
```

This is fixed in this PR.

czgdp1807:
Open an issue with https://github.com/lcompilers/lpython/pull/856#issuecomment-1200535626 and ping me. I will look into it. @Abdelrahman-Kh-Fouad For now let's merge this anyways.

czgdp1807 (approved):

src/libasr/pass/array_op.cpp
```diff
@@ -353,6 +353,10 @@ class ArrayOpVisitor : public PassUtils::PassVisitor<ArrayOpVisitor>
         tmp_val = const_cast<ASR::expr_t*>(&(x.base));
     }
 
+    void visit_LogicalConstant(const ASR::LogicalConstant_t& x) {
+        tmp_val = const_cast<ASR::expr_t*>(&(x.base));
```
certik, line 357: Don't cast like this, rather, do:
```c++
tmp_val = ASR::down_cast<ASR::expr_t*>(&x);
```
Would you please mind sending a PR fixing this?

Smit-create: Sure, will do!

PR #858, opened 2022-08-01
Implement array function for numpy module

Thirumalai-Shaktivel:


integration_tests/test_numpy_04.py
```diff
@@ -0,0 +1,14 @@
+from ltypes import f64
+from numpy import array
+
+def test_array():
+    x: list[f64] = array([1.0, 2.0, 3.0])
```
certik, line 5:
```suggestion
    x: f64[3] = array([1.0, 2.0, 3.0])
```

Thirumalai-Shaktivel: If we change, I get:
```console
$ lpython integration_tests/test_numpy_04.py
semantic error: Type mismatch in annotation-assignment, the types must be compatible
 --> integration_tests/test_numpy_04.py:5:5
  |
5 |     x: f64[3] = array([1.0, 2.0, 3.0])
  |     ^           ^^^^^^^^^^^^^^^^^^^^^^ type mismatch ('f64[4]' and 'list[f64]')
```
how do I differentiate between `list` and `array`

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4068,6 +4068,19 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                    throw SemanticError("array accepts only 1 argument for now, got " +
+                                        std::to_string(args.size()) + " arguments instead.",
+                                        x.base.base.loc);
+                }
+                ASR::ListConstant_t* list = ASR::down_cast<ASR::ListConstant_t>(
+                                                args[0].m_value);
+                ASR::ttype_t *type = ASRUtils::expr_type(args[0].m_value);
+                ASR::expr_t **m_args = list->m_args;
+                size_t n_args = list->n_args;
+                tmp = ASR::make_ListConstant_t(al, x.base.base.loc, m_args, n_args, type);
```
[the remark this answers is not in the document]
Thirumalai-Shaktivel: At first, I had the thought to use Array, later I changed to List :)

integration_tests/test_numpy_04.py
certik: A list is a type `List_t`, while an array is a type (like an `Integer_t` or `Real_t`) with dimensions set.

Thirumalai-Shaktivel: Got it, Thank you!

certik (approved):
I think this looks good, great job!

The old `ListConstant` is not present in ASR anymore, you replaced it with an `ArrayConstant`. So we "stole" the `ListConstant`'s arguments which is fine.

If both were used, we would have to copy everything, so that ASR->ASR passes can modify the ASR in place (and it would not accidentally modify some unrelated part that happened to reuse the ASR nodes).

Thirumalai-Shaktivel:
Thanks for the review!

src/lpython/semantics/python_ast_to_asr.cpp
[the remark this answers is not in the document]
Thirumalai-Shaktivel: How to implement for `(runtime) List type`?
For which, I checked weather `list` is a `Var_t`, after that how do I extract the contents of `list`, as we do above for ConstantList?
How to access the `value` of the variable in `python_ast_to_asr`?
If value is not yet set, then what should I return for the `(runtime) List type`?
cc @certik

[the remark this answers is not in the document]
Thirumalai-Shaktivel: If we return `empty array` with `list_type`, Where should I later handle it?

PR #878, opened 2022-08-03
Raise error while changing immutable objects

### LINKED ISSUE #820: Raise error while changing immutable objects
```console
def f():
    s: str
    s = "abcd"
    s[0] = 'e'
    print(s)

f()
  File "/Users/namannimmo/Applications/oss/lpython/t.py", line 10, in <module>
    f()
  File "/Users/namannimmo/Applications/oss/lpython/t.py", line 4, in f
    s[0] = 'e'
TypeError: 'str' object does not support item assignment
(lp) namannimmo@srf lpython % asr t.py # this works but should raise an error.
```
```console
def f():
    t: tuple[i32, i32]
    t = (1, 2)
    t[0] = 3


f()
  File "/Users/namannimmo/Applications/oss/lpython/t.py", line 7, in <module>
    f()
  File "/Users/namannimmo/Applications/oss/lpython/t.py", line 4, in f
    t[0] = 3
TypeError: 'tuple' object does not support item assignment
(lp) namannimmo@srf lpython % asr t.py # works. should raise an error instead.
```

namannimmo10:
Fixes #820


certik (approved):
Very good!

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2588,6 +2588,10 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                                 ASR::make_Var_t(al, x.base.base.loc, s));
                         tmp = make_DictInsert_t(al, x.base.base.loc, se, key, val);
                         return;
+                    } else if (ASR::is_a<ASR::Character_t>(*type)) {
+                        throw SemanticError("'str' object does not support item assignment", x.base.base.loc);
+                    } else if (ASR::is_a<ASR::Tuple_t>(*type)) {
+                        throw SemanticError("'tuple' object does not support item assignment", x.base.base.loc);
```
Smit-create, line 2594: How about instead of chaining this with if-else, we create a new function say `is_immutable` to check whether the type is immutable or not?

certik: What other types are immutable?

Smit-create: One of the examples is here: https://github.com/lcompilers/lpython/issues/820#issuecomment-1193970996. Separating them to a new function might be easy to maintain.

Smit-create: One of the examples is here: https://github.com/lcompilers/lpython/issues/820#issuecomment-1193970996. Separating them to a new function might be easy to maintain.

namannimmo10: All python numeric types (numbers) are designed to be immutable. So I don't think having a separate function makes a lot of sense. Anyway, I still added it, so that it is readable for us.

czgdp1807: Well immutability is a very subjective (with respect to the language under consideration and the type under consideration) concept. In CPython, AFAIK, all objects are passed by reference or while assigning, a name (variable) points to an object. As you said all CPython numeric types are immutable. But that doesn't mean you can't change the value of a floating point name or an integer name (variable). You can definitely do that. However, when you assign one integer variable to another integer variable then two names (variables) start pointing to a common integer object. Changing one integer variable to another integer object doesn't make any modification to the original object (see example below).

```python
>>> a = 12
>>> b = a
>>> id(a), id(b)
(4314587736, 4314587736) # 12 is an integer object with a, b names pointing to it
>>> b = 13
>>> id(a), id(b)
(4314587736, 4314587768) # a still points to 12, b points to 13
```

So immutability with respect to numeric types means you cannot modify `12` or `13` (both are integer objects in CPython, `integer != integer objects` (see https://docs.python.org/3/c-api/long.html). And anyways you cannot modify, `12`, `13` or any constant number in LPython as well. You can modify an integer variable (name) though which is inline with what CPython does.

Now coming on to data types like tuple. They are containers and you can index them in an attempt to modify any single value present inside these containers. CPython doesn't allow that. But indexing and modifying a list object is allowed.

Bottomline - Immutability of constant numeric types is anyways retained implicitly. For enforcing immutability of container data types like tuple, string, frozenset, we need to en-force via checks as the one you are adding in this PR. In fact a user can make their class type sort of immutable by raising an error whenever `__setitem__` is called. So having `is_immutable` will help in encapsulating tricky and complicated checks if needed in future.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2538,6 +2538,14 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                 targets.size());
     }
 
+    bool is_immutable(const ASR::ttype_t *type) {
+        if (ASR::is_a<ASR::Character_t>(*type) || ASR::is_a<ASR::Tuple_t>(*type)
+            || ASR::is_a<ASR::Complex_t>(*type)) {
+            return true;
+        }
+        return false;
```
czgdp1807, line 2546:
```suggestion
        return ((ASR::is_a<ASR::Character_t>(*type) || ASR::is_a<ASR::Tuple_t>(*type) || ASR::is_a<ASR::Complex_t>(*type)));
```

czgdp1807: Let's put it in `asr_utils.h`. We might need to re-use it in some ASR->ASR passes.

src/lpython/semantics/python_ast_to_asr.cpp
certik: (The added complication is that LPython restricts Python to such a subset so that we can modify the semantics to actually always copy things over for `a = b`, no references or reference counted pointers.)

czgdp1807: Yes. In LPython, deepcopy happens whenever assigning one variable to another. I think that's the best way to not restrict performance. Like as [we discussed with lists](https://github.com/lcompilers/lpython/pull/835#discussion_r932485046), we do deepcopy always when `a = b` (`a` and `b` both are lists).

certik: I documented this at https://github.com/lcompilers/lpython/wiki/LPython-Semantics

src/libasr/asr_utils.h
```diff
@@ -809,6 +809,11 @@ static inline std::string binop_to_str_python(const ASR::binopType t) {
     }
 }
 
+static inline bool is_immutable(const ASR::ttype_t *type) {
+    return ((ASR::is_a<ASR::Character_t>(*type) || ASR::is_a<ASR::Tuple_t>(*type)
+        || ASR::is_a<ASR::Complex_t>(*type)));
+}
```
Smit-create, line 815: @namannimmo10 This will fail for:
```py
def ff():
    a: c32[2]
    a[0] = complex(1, 2)
    print(a)
    a[1] = complex(3, 4)
    print(a)

ff()
```

czgdp1807: I think the error should only be raised for,

`a.real = 1.0` or `a.img = 2.0` (i.e., only for updating attributes of complex type). We can always assign a complex variable name to another value. The original value shouldn't have any change. 

If there's an error can you fix this @Smit-create ?

Issue #859, opened 2022-08-01
segfault in visit_AnnAssignUtil for `a: list[i32] = []`


certik:
I want to test another benchmark (https://lemire.me/blog/2012/06/20/do-not-waste-time-with-stl-vectors/):
```python
from ltypes import i32

def test_list(n: i32) -> i32:
    a: list[i32] = []
    i: i32
    for i in range(n):
        a.append(i)
    sum: i32 = 0
    for i in range(n):
        sum += a[i]
    return sum

print(test_list(101))
```
Currently it fails with:
```console
$ lpython --new-parser b.py                                     
Traceback (most recent call last):
  File "/Users/ondrej/repos/lpython/src/bin/lpython.cpp", line 883
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir, lpython_pass_manager, compiler_options, time_report);
  File "/Users/ondrej/repos/lpython/src/bin/lpython.cpp", line 349
    r1 = LFortran::LPython::python_ast_to_asr(al, *ast, diagnostics, true,
  File "/Users/ondrej/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 4169
    auto res2 = body_visitor(al, *ast_m, diagnostics, unit, main_module,
  File "/Users/ondrej/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 4096
    b.visit_Module(ast);
  File "/Users/ondrej/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2432
    visit_stmt(*x.m_body[i]);
  File "/Users/ondrej/repos/lpython/src/lpython/python_ast.h", line 1882
    void visit_stmt(const stmt_t &b) { visit_stmt_t(b, self()); }
  File "/Users/ondrej/repos/lpython/src/lpython/python_ast.h", line 1750
    case stmtType::FunctionDef: { v.visit_FunctionDef((const FunctionDef_t &)x); return; }
  File "/Users/ondrej/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2461
    ASR::Function_t *f = ASR::down_cast<ASR::Function_t>(t);
  File "/Users/ondrej/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2450
    transform_stmts(body, x.n_body, x.m_body);
  File "/Users/ondrej/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2407
[...]

PR #879, opened 2022-08-03
Supporting initialisation with empty list in ASR

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/859


czgdp1807:
Well its segfaulting on Ubuntu. Will try to reproduce this on my linux machine tomorrow.

czgdp1807:
I think this is happening because initial capacity of the `list` is `0` and hence doubling always produces `0`. On my mac this works fine (probably mac safe gaurds `malloc` implementation against this corner case of `0` input but linux doesn't). I will confirm tomorrow.

certik:
That's a fine solution I think. The PR looks ok.

certik:
Thanks!

PR #884, opened 2022-08-04
Resolve "Case sensitive keywords and not in issue" in the tokenizer

Thirumalai-Shaktivel:
Resolve following issues:
```py
$ lpython --show-ast examples/expr2.py --indent --new-parser
syntax error: Token 'RAISE' (of type 'raise') is unexpected here
 --> examples/expr2.py:1:21
  |
1 | assert self.conv(np.RAISE) == 'NPY_RAISE'
  |                     ^^^^^ 


```
```py
$ lpython --show-ast examples/expr2.py --indent --new-parser
syntax error: Token 'not in' is unexpected here
 --> examples/expr2.py:7:4
  |
7 | if not info or name in info[-1]: # next processor
  |    ^^^^^^ 


```


build0.sh
```diff
@@ -16,7 +16,7 @@ python grammar/asdl_cpp.py src/libasr/ASR.asdl src/libasr/asr.h
 python src/libasr/wasm_instructions_visitor.py
 
 # Generate the tokenizer and parser
-(cd src/lpython/parser && re2c -W -b tokenizer.re -o tokenizer.cpp)
+(cd src/lpython/parser && re2c -W -b tokenizer.re -o tokenizer.cpp --case-inverted)
```
certik, line 19: Let's not use the `--case-inverted`, because I think it will become confusing with LFortran.

I think all we need is to change `'raise'` to `"raise"` and the same for all the other keywords.

Thirumalai-Shaktivel: Yup, its works. Thank you!

certik (approved):
That looks good!

Thirumalai-Shaktivel:
Thanks for the review!

PR #899, opened 2022-08-04
[Parser] Parse multiple statements in single line as `body` of `multiline statements`

### LINKED ISSUE #886: Singleline multi-statements are not recognized
```py
def dtype(self) -> _DType_co: ...

class sub(np.ndarray): pass

If True: x
else: x

try: shutil.rmtree(os.path.join(asv_build_cache, c))
except OSError: pass
```

akshanshbhatt:
Closes #886


akshanshbhatt:
The names of these grammar rules are a mess :)

certik (approved):
I think that looks good. Thanks for getting this done!

Thirumalai-Shaktivel:
LGTM as well! Thanks for fixing this.

PR #900, opened 2022-08-05
Initial implementation of generics

certik:


src/libasr/asdl_cpp.py
```diff
@@ -868,7 +866,7 @@ def visitConstructor(self, cons, _):
 
     def make_visitor(self, name, fields):
         self.emit("")
-        self.emit("ASR::asr_t* duplicate_%s(%s_t* x) {" % (name, name), 1)
+        self.emit("virtual ASR::asr_t* duplicate_%s(%s_t* x) {" % (name, name), 1)
```
certik, line 869: This will get fixed in a subsequent PR, @ansharlubis has a work in progress fix already.

certik: Fixed in https://github.com/lcompilers/lpython/pull/903.

PR #907, opened 2022-08-05
Adding generic array support

ansharlubis:
Allowed frontend generic array support. Does not work with CPython right now.


integration_tests/CMakeLists.txt
```diff
@@ -213,6 +213,7 @@ RUN(NAME vec_01              LABELS cpython llvm)
 RUN(NAME test_str_comparison LABELS cpython llvm)
 
 RUN(NAME generics_01         LABELS cpython llvm)
+RUN(NAME generics_array_01   LABELS llvm)
```
certik, line 216: Add a CPython test too:
```suggestion
RUN(NAME generics_array_01   LABELS cpython llvm)
```

certik:
Excellent, great idea to get this working.

integration_tests/CMakeLists.txt
ansharlubis: The problem is that a TypeVar object, in this case T, is not subscriptable in CPython.

certik: I see, that is indeed a problem.

I think we should probably design our own "TypeVar" that can handle subscripting, we'll add it to `ltypes` and we have to make sure exactly the same syntax works both in LPython and CPython.

integration_tests/generics_array_01.py
```diff
@@ -0,0 +1,15 @@
+from ltypes import TypeVar
+
+T = TypeVar('T')
+
+def f(lst: T[:], i: T) -> T:
```
czgdp1807, line 5: A more interesting case would be to support `def f(lst: T[...], i: T) -> T`. We can do it later but we should keep in mind to support it.

certik:
Thanks! I squashed and merged. I think this is fine to get started. We have to rework the syntax to get this working with CPython, and so on.

integration_tests/generics_array_01.py
ansharlubis: I haven't tried that out with CPython, but I bet the same issue as I mentioned above has to be solved first as well.

ansharlubis:
Got it! I also think that defining our own TypeVar would be a good idea.

PR #909, opened 2022-08-05
Update compiler_tester from LFortran

certik:


run_tests.py
```diff
@@ -28,91 +110,50 @@ def main():
...
-                    filename, update_reference, extra_args)
-
-        if tokens:
-            run_test("tokens", "lpython --no-color --show-tokens {infile} -o {outfile}",
-                    filename, update_reference, extra_args)
-
-        print()
-
+        + os.pathsep + os.environ["PATH"]
+    test_data = toml.load(open("tests/tests.toml"))
```
Smit-create, line 114: A minor suggestion: we can use the absolute path to avoid path errors if we are running that from some other directory. But that's fine, I can fix as a follow up.

Smit-create (approved):

run_tests.py
certik: I merged all PRs. This code now lives in `src/libasr/compiler_tester` and it is shared with LFortran. So you can go ahead and modify /  polish it more.

Thirumalai-Shaktivel:
The `./run_tester.py` is really fast now. Thank you for working on this!

Thirumalai-Shaktivel:
The `ast_new` is not included in the `run_tests.py`, so #913 adds it.

PR #914, opened 2022-08-05
Implement for loop with slice/subscript objects

### LINKED ISSUE #883: Implement looping through a reversed string
```console
def f():
    s: str
    s = "abcde"
    c: str
    res: str
    for c in s[::-1]:
        print(c)


f()
e
d
c
b
a
semantic error: Only function call `range(..)` supported as for loop iteration for now
 --> t.py:6:5 - 7:16
  |
6 |        for c in s[::-1]:
  |        ^^^^^^^^^^^^^^^^^...
...
  |
7 |            print(c)
  | ...^^^^^^^^^^^^^^^^ 


(lp) namannimmo@srf lpython %
```

Smit-create:
Fixes #883


integration_tests/test_builtin_str.py
```diff
@@ -63,6 +63,15 @@ def test_str_slice_step():
     assert s[::-1] == "kjihgfedcba"
 
 
+def test_issue_883():
+    s: str
+    s = "abcde"
+    c: str
+    for c in s[::-1]:
+        print(c)
```
czgdp1807, line 71: This looks okay. Do we support `reversed` intrinsic as well?

czgdp1807, line 71: This looks okay. Do we support `reversed` intrinsic as well?

Smit-create: Not yet.

czgdp1807: Okay. Makes sense. Let's support that in a subsequent PR.

czgdp1807 (approved):
@certik Is this good to merge?

certik (approved):
Looks good to me, thanks!

Issue #882, opened 2022-08-04
Add printing support for list, tuple, set, and dict 


namannimmo10:
```console
(lp) namannimmo@srf lpython % cat t.py
def f():
    l: list[str]
    l = ['a', 'b', 'c']
    print(l)


f()
(lp) namannimmo@srf lpython % lpython t.py && ./a.out 
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/namannimmo/Applications/oss/lpython/src/bin/lpython.cpp", line 1081
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir, lpython_pass_manager, compiler_options, time_report);
  File "/Users/namannimmo/Applications/oss/lpython/src/bin/lpython.cpp", line 559
    res = fe.get_llvm3(*asr, pass_manager, diagnostics);
  File "/Users/namannimmo/Applications/oss/lpython/src/lpython/python_evaluator.cpp", line 57
    e->get_context(), al, lpm, compiler_options.platform,
  File "/Users/namannimmo/Applications/oss/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 5464
    v.nested_func_types = pass_find_nested_vars(asr, context,
  File "../libasr/asr.h", line 3691
  File "../libasr/asr.h", line 3669
  File "../libasr/asr.h", line 3692
  File "../libasr/asr.h", line 3470
  File "/Users/namannimmo/Applications/oss/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 1068
    visit_symbol(*item.second);
  File "../libasr/asr.h", line 3694
  File "../libasr/asr.h", line 3480
  File "/Users/namannimmo/Applications/oss/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 2414
    visit_procedures(x);
  File "/Users/namannimmo/Applications/oss/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 2488
    this->visit_stmt(*x.m_body[i]);
  File "../libasr/asr.h", line 3708
  File "../libasr/asr.h", line 3514
  File "/Users/namannimmo/Applications/oss/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 4569
    handle_print(x);
  File "/Users/namannimmo/Applications/oss/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 4735
    ASRUtils::type_to_str(t) + " type.");
[...]

PR #927, opened 2022-08-06
add print for ListConstant.

redbopo:
@namannimmo10 
Just a start fix for print with list type in [882](https://github.com/lcompilers/lpython/issues/882).

This patch make it possible for ListConstant type in follow as:

```
print(["ab", "abc", "abcd"])
print([1, 2, 3, 4])
print([1.23 , 324.3 , 56.431, 90.5, 34.1])
print(['a', 'b', 'c', 'd' , 'e', 'f'])
```
Mainly in asr_to_llvm step.

I don't know whether it's right to change the test_output not related in removing  `align`.
Just to run the command `./run_tests.py -u` 

---------------

Something more, 

And now 
`print([ ])`
entering an empty List will raise an error in ast_to_asr step .
Need to get a fix in future.

<details>

```
  Binary file "/home/r/Desktop/workSpace/lpython/src/bin/lpython", in _start()
  File "/home/r/Desktop/workSpace/lpython/src/bin/lpython.cpp", line 1081, in ??
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir, lpython_pass_manager, compiler_options, time_report);
  File "/home/r/Desktop/workSpace/lpython/src/bin/lpython.cpp", line 541, in ??
    compiler_options.disable_main, compiler_options.symtab_only, infile);
  File "/home/r/Desktop/workSpace/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 4409, in LFortran::LPython::python_ast_to_asr(Allocator&, LFortran::LPython::AST::ast_t&, LFortran::diag::Diagnostics&, bool, bool, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)
    ast_overload);
  File "/home/r/Desktop/workSpace/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 4335, in LFortran::LPython::body_visitor(Allocator&, LFortran::LPython::AST::Module_t const&, LFortran::diag::Diagnostics&, LFortran::ASR::asr_t*, bool, std::map<int, LFortran::ASR::symbol_t*, std::less<int>, std::allocator<std::pair<int const, LFortran::ASR::symbol_t*> > >&)
    b.visit_Module(ast);
  File "/home/r/Desktop/workSpace/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2611, in LFortran::LPython::BodyVisitor::visit_Module(LFortran::LPython::AST::Module_t const&)
    visit_stmt(*x.m_body[i]);
  File "/home/r/Desktop/workSpace/lpython/src/lpython/python_ast.h", line 1882, in LFortran::LPython::AST::BaseVisitor<LFortran::LPython::BodyVisitor>::visit_stmt(LFortran::LPython::AST::stmt_t const&)
  File "/home/r/Desktop/workSpace/lpython/src/lpython/python_ast.h", line 1750, in ??
    case stmtType::FunctionDef: { v.visit_FunctionDef((const FunctionDef_t &)x); return; }
  File "/home/r/Desktop/workSpace/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2638, in LFortran::LPython::BodyVisitor::visit_FunctionDef(LFortran::LPython::AST::FunctionDef_t const&)
    handle_fn(x, *f);
  File "/home/r/Desktop/workSpace/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2628, in LFortran::LPython::BodyVisitor::handle_fn(LFortran::LPython::AST::FunctionDef_t const&, LFortran::ASR::Function_t&)
    transform_stmts(body, x.n_body, x.m_body);
  File "/home/r/Desktop/workSpace/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 2586, in LFortran::LPython::BodyVisitor::transform_stmts(LFortran::Vec<LFortran::ASR::stmt_t*>&, unsigned long, LFortran::LPython::AST::stmt_t**)
    this->visit_stmt(*m_body[i]);
  File "/home/r/Desktop/workSpace/lpython/src/lpython/python_ast.h", line 1882, in LFortran::LPython::AST::BaseVisitor<LFortran::LPython::BodyVisitor>::visit_stmt(LFortran::LPython::AST::stmt_t const&)
  File "/home/r/Desktop/workSpace/lpython/src/lpython/python_ast.h", line 1772, in ??
    case stmtType::Expr: { v.visit_Expr((const Expr_t &)x); return; }
  File "/home/r/Desktop/workSpace/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 3616, in LFortran::LPython::BodyVisitor::visit_Expr(LFortran::LPython::AST::Expr_t const&)
    visit_expr_list(c->m_args, c->n_args, args);
  File "/home/r/Desktop/workSpace/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 467, in LFortran::LPython::CommonVisitor<LFortran::LPython::BodyVisitor>::visit_expr_list(LFortran::LPython::AST::expr_t**, unsigned long, LFortran::Vec<LFortran::ASR::call_arg_t>&)
    ASR::expr_t* expr = ASRUtils::EXPR(tmp);
  File "/home/r/Desktop/workSpace/lpython/src/libasr/asr_utils.h", line 23, in ??
    return ASR::down_cast<ASR::expr_t>(f);
AssertFailed: f != nullptr
```

</details>


Thirumalai-Shaktivel:
@redbopo, Thanks for your effort in fixing the `list` print.
Regarding the references update, please make a clean build and update the tests. The changes in the references (align) should be removed.

integration_tests/print_02.py
```diff
@@ -0,0 +1,9 @@
+
+
+def f():
+    print(["ab", "abc", "abcd"])
+    print([1, 2, 3, 4])
+    print([1.23 , 324.3 , 56.431, 90.5, 34.1])
+    print(['a', 'b', 'c', 'd' , 'e', 'f'])
+
+f()
```
Thirumalai-Shaktivel, line 9:
```suggestion
# Test: Printing ListConstant
def f():
    print(["ab", "abc", "abcd"])
    print([1, 2, 3, 4])
    print([1.23 , 324.3 , 56.431, 90.5, 34.1])
    print(['a', 'b', 'c', 'd' , 'e', 'f'])

f()

```

Thirumalai-Shaktivel:
Any issue with the references?

redbopo:
Thanks for reviews @Thirumalai-Shaktivel .
As for the change in removing align,  I think it's strange.
I think the right thing is we run `./run_tests.py -u` , it will get its match output.
And it still remove the align.
Although I can remove it myself, but have a conflict with the command.

redbopo:
and I don't think I get the build system changed, since it's unrelated reference output.
but in the picture follow, the output is no align.
Screenshot from 2022-08-06 20-05-59

Thirumalai-Shaktivel:
I tested in my local machine, I did not face this issue. Shall I update the references and push it here?

redbopo:
Yes, thanks. @Thirumalai-Shaktivel .  
I will to figure out the problems first.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -4599,6 +4599,52 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                        }
+                    }
+                    fmt.push_back("%s");
+                    llvm::Value *close_bracket = builder->CreateGlobalStringPtr("]");
+                    args.push_back(close_bracket);
+                }
+                else {
+                    throw LCompilersException("Printing List is only support for "
+                        "ASR type ListConstant now.");
+                }
```
czgdp1807, line 4647: I think we should do generic printing of list variables with the help of an ASR->ASR pass, so that all backend have the printing support at once. How I see printing it is as follows,

```python
print(l) # l is a list
```

should be converted to,

```python
print("[", end="")
for i in range(len(l)):
    if i < len(l) - 1:
        print(l[i], end=",")
    else:
        print(l[i], end="")
print("]")
```

czgdp1807: @certik What do you say?

redbopo: I thing it will be cool. A higher abstraction can save a lot of code, although I have few experience on asr to other backend now.

certik:
These things can effectively be implemented in ASR, as @czgdp1807 said, or we need to call into a C runtime library, as we currently do. One could implement it even in the frontend language (Python or Fortran), although the variable number of arguments is an issue that gets solved by an ASR->ASR pass, so we can just implement the actual printing of each value.

redbopo:
Update with ASR pass, but not Finish yet. Mainly with the  `AssertFailed: x.base.type == asrType::expr` error now. Will be back later.

redbopo:
Works now. 
The float format is decided by backend part, there will be a small mount of error and I don't like it . 
```
./src/bin/lpython --pass=print_list_constant   ./build/t.py  && ./a.out 

['ab', 'abc', 'abcd']
[1, 2, 3, 4]
[    1.22999999999999998,   324.30000000000001137,    56.43099999999999739,    90.50000000000000000,    34.10000000000000142]
['a', 'b', 'c', 'd', 'e', 'f']

```

certik (approved):
Looks good to me. If @czgdp1807 is fine with it, we can merge.

czgdp1807:
Will review it tomorrow.

certik:
Regarding printing of floating point numbers, that's a separate issue: https://github.com/lcompilers/lpython/issues/949, we just need to implement a better algorithm.

czgdp1807 (changes requested):
The doc string is not in sync with the code. The doc string says that the print(list) is converted to a loop but in the implementation of the visitor I wasn’t able to find any loop creation code. Am I right @redbopo? If so then can you update the doc string.

czgdp1807 (commented):
So, since we are printing constant lists for which we have its size available at compile time. Now calling print statement each time to print a single value is not something optimal provided we can easily achieve that. Like, if constant list contains 5 elements (and we know it via n_args attribute of ListConstant node), then they can be printed via a single call to Print. Similarly if the constant list contains 100 elements then calling print 100 times would add 100 new statements to the ASR which otherwise could be done by just 10 calls to Print each printing 10 elements at a time with right formatting.

Bottomline - I think the current logic can be easily improved by just adding some extra logic to print multiple elements of a constant list in a single ASR Print statement so that the size of the ASR and generated code are optimal. If you want to do it here then it would be great. If you have already done that then I will approve this PR, just let me know.

src/libasr/pass/print_list_constant.cpp
```diff
@@ -0,0 +1,132 @@
...
+                                    &listC->m_args[j], 1, nullptr, empty_str)));
+          pass_result.push_back(al, print_single_quote);
+        } else if (ASRUtils::is_real(*arg_type) ||
+                   ASRUtils::is_integer(*arg_type)) {
+          pass_result.push_back(al,
+                                LFortran::ASRUtils::STMT(ASR::make_Print_t(
+                                    al, x.base.base.loc, nullptr,
+                                    &listC->m_args[j], 1, nullptr, empty_str)));
+        } else {
+          throw LCompilersException("Printing ConstantList is only support for"
```
czgdp1807, line 114: I am a bit confused here. Why the type of the list element is a constraint for printing? LLVM backend supports everything for printing and if not then also at least ASR shouldn’t be limited because of LLVM. Is there any blocker at ASR level that compelled you to add this check?

redbopo: Yes, In fact the day before I commit in on LLVM backend.
At that time I notice print statement not support with list_constant .
And the logic is almost the same as now, but in ASR pass now.
I just not know how to initilize a list_constant with complex item type and so on, and not include it in the test file the same as the implementation.

czgdp1807: What happens if you remove this check? Do the current tests pass? Which ones fail and what's the error message you receive?

redbopo: https://github.com/lcompilers/lpython/blob/d720d6abcf918106e2c9bd7a5f50bcd5db51c027/src/libasr/codegen/asr_to_llvm.cpp#L4559-L4693

The print ListConstant type is not supported in LLVM backend.
And I think you want to support an ASR pass is to transform the 
print list_constant ---to--->  print item ( n times)
which will fix other backend too, if we only support print listConstant in llvm backend.

redbopo: Now like the code, it will triger an error 

```
def f():
    print([complex(6, 8)])

f()
```

redbopo: I will need to add more on other type.

czgdp1807: Sure. Take your time.

redbopo:
Hi @czgdp1807 ,  I think it will arise some problems if we'd like to 
convert the  `PRINT STMT` ,here is print list_constant, to `DOLOOP STMT`.

If I don't misunderstand, to achieve the pass
we will need to add another `ASSIGN STMT` like `a: list[i32] = [1, 2, 3, 4] ` in previous.
So `print([1,2,3,4])` will be converted to sth. like `a: list[i32] = [1,2,3,4]; for i in range(4): print(a[i])`
It also means we also need to add `a, i` in the function symbol_table .

In this way, will it be more proper to only implement a llvm backend support now?

czgdp1807:
I think you don't need a list variable to print `ListConstant`. Anyways, thanks for letting me know, I will try to look into your PR and complete it. If you are interested you can work on https://github.com/lcompilers/lpython/issues/960. Its a good issue and we can use your help in let. Let me know if you want to work on it. I will un-assign myself from there in that case.

redbopo:
Thanks for quick reply. I will look in #960 later.

Shaikh-Ubaid:
Please, could someone possibly share if we are trying to modify the `print(l)` where `l` is of type `list` to the following:
```python
print("[", end="")
for i in range(len(l)):
    if i < len(l) - 1:
        print(l[i], end=", ")
    else:
        print(l[i], end="")
print("]")
```

redbopo:
Of course @Shaikh-Ubaid

czgdp1807:
Closing as @Shaikh-Ubaid added a pass for printing lists. Relevant issue open - #1129.

PR #932, opened 2022-08-06
Implement `ListInsert` in the LLVM Backend

Thirumalai-Shaktivel:


integration_tests/test_list_03.py
```diff
@@ -10,7 +10,14 @@ def test_list(n: i32) -> i32:
         sum += a[i]
     return sum
 
+def test_list_02(pos: i32, value: i32) -> i32:
+    x: list[i32]
+    x = [1, 3, 4]
+    x.insert(pos, value)
```
czgdp1807, line 16: Let's test multiple inserts on the same list variable.

Thirumalai-Shaktivel: Now, I understood what the issue is! Instead of inserting at the `pos`, I replace it.
Thank you

czgdp1807: Yes. Writing moderate to strong tests is important to implement the feature robustly. The above was very simple. Leaving out corner cases is okay but leaving out a large class of use cases is not recommended.

Thirumalai-Shaktivel:
@czgdp1807
I'm planning to implement in this way:
```c++
int main() {
  int a[] = {1, 3, 4, 0};
  for (int i = 3; i > 1; i--) {
    a[i] = a[i-1];
  }
  a[1] = 2;
  for (int i = 0; i <=3; i ++) {
    std::cout << a[i] << " ";
  }
  return 0;
}
```

Thirumalai-Shaktivel:
Can you please some explanation:
- What does `get_pointer_to_current_end_point`? (How does it calculate the value?)
- How to get the List size?
```c++
llvm::Value* arg_size = builder->CreateMul(llvm::ConstantInt::get(context,
                                                   llvm::APInt(32, type_size)), src_capacity);
```
Does this work as intended?
- The following retrieves the value at the specific `pos` in the list, right?
`llvm_utils->create_gep(list, pos)`

czgdp1807:
Well, it needs the following steps IMO,

1. First resize if needed. Since, `n` is always pointing to the index of the next incoming element you can simply call `LLVMList::resize_if_needed`.
2. Now start shifting element to the right by one step (you need to generate loop in LLVM for that) starting from the `n - 1` position till `pos`.
3. Now call `LLVMList::write_item` with `list`, `pos` and `value`.

That should be it.

Thirumalai-Shaktivel:
Yup, I had the same thought!

czgdp1807:
> What does get_pointer_to_current_end_point do? (How does it actually calculate the value?)

It points to the index of the next incoming element. Say you have 4 elements in the list then its value will be 4 (0, 1, 2, 3 will be the indices of the already existing elements and 4 will be the index of the next incoming list).

> How to get the List size?

If you are planning to use `size` for resizing for inserting the new element then you don't need it. `resize_if_needed` already does that. `capacity` is the maximum element of elements a list can accommodate, and `n` (current end point) is the number of elements currently in the list.

Thirumalai-Shaktivel:
Thank you, I got it!

czgdp1807:
Well doing the second step in https://github.com/lcompilers/lpython/pull/932#issuecomment-1207232359 from `n - 1` to `pos` (in reverse direction) might not be cache efficient. Starting from `pos` should be better. You just need to store `pos + 1`'s value in a temporary variable. Something like the following,

```python
tmp1 = l[pos]
for i in range(pos, n):
    tmp2 = l[i + 1]
    l[i + 1] = tmp1
    tmp1 = tmp2
```

I tested the above approach but not so sure. You can see the code here and try it out for yourself before implementing the same logic in LPython (https://ideone.com/JSDhuh).

Thirumalai-Shaktivel:
I have implemented as mentioned, but I got struck in handling the loop, Can you please review and help debug the issue here. The while loop goes to an infinite loop!

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -295,6 +295,47 @@ namespace LFortran {
...
+        builder->CreateCondBr(cond, loopbody, loopend);
+
+        // body
+        llvm_utils->start_new_block(loopbody);
+        next_element_ptr = builder->CreateAdd(copy_pos, next_element_ptr);
+        tmp2 = LLVM::CreateLoad(*builder, read_item(list, next_element_ptr, true));
+        write_item(list, next_element_ptr, tmp1);
+        tmp1 = tmp2;
+        pos = builder->CreateAdd(pos, llvm::ConstantInt::get(context, llvm::APInt(32, 1)));
+        builder->CreateBr(loophead);
```
Thirumalai-Shaktivel, line 331: I tried to debug here, 
Initially `pos` is `1` as per the example.
`pos` incremented by `1` resulting in `2` in the body, but when it returns to the loop head, `pos` has the same value i.e., `1`

Thirumalai-Shaktivel:
@czgdp1807, this PR is ready, Can you please review it?

czgdp1807 (approved):

Thirumalai-Shaktivel:
Thank you very much for your guidance and review.

certik (approved):
Fine to merge.

We can create some nicer API to create loops in LLVM in some easier way.

Thirumalai-Shaktivel:
Thank you!

Thirumalai-Shaktivel:
> Fine to merge.
> 
> We can create some nicer API to create loops in LLVM in some easier way.

How do I implement API for loops, or if possible, `if` statements?

certik:
You implement some nice C++ utility classes or functions, that make using loops/if statements in LLVM easier.

Perhaps something with C++ lambda functions: `if(<lambda condition>, <lamba then>, optional <lambda else>)`, and you just fill in the condition and then / else, and the `if` function generates the correct boilerplate.

PR #935, opened 2022-08-08
Make `len()` work with tuples

namannimmo10:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1267,6 +1267,16 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
     }
 
+    void visit_TupleLen(const ASR::TupleLen_t& x) {
+        if (x.m_value) {
+            this->visit_expr(*x.m_value);
+        } else {
+            ASR::Tuple_t* asr_tuple = ASR::down_cast<ASR::Tuple_t>(
+                                            ASRUtils::expr_type(x.m_arg));
+            tmp = llvm::ConstantInt::get(context, llvm::APInt(32, asr_tuple->n_type));
+        }
```
certik, line 1277: I think the implementation should be just this:
```suggestion
        LCOMPILERS_ASSERT(x.m_value);
        this->visit_expr(*x.m_value);
```
Then in the frontend, we have to make sure we always set the `value` attribute when generating the `TupleLen` node, since the length is a compile time feature.

I think we still want to keep `TupleLen` in ASR as a node, in order to nicely represent this operation (so that once we can print ASR as Python code that it prints naturally), but since it is a fully compile time feature, it would always have the `value` set.

namannimmo10: We can simply make this change, no?
```diff
diff --git a/src/libasr/ASR.asdl b/src/libasr/ASR.asdl
index 35d9e6045..f91c8cd0c 100644
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -247,7 +247,7 @@ expr
     | SetLen(expr arg, ttype type, expr? value)
 
     | TupleConstant(expr* elements, ttype type)
-    | TupleLen(expr arg, ttype type, expr? value)
+    | TupleLen(expr arg, ttype type, expr value)
 
     | StringConstant(string s, ttype type)
     | StringConcat(expr left, expr right, ttype type, expr? value)
```

certik (approved):
Looks good to merge I think.

src/libasr/codegen/asr_to_llvm.cpp
certik: Yes, I think that should work.

PR #936, opened 2022-08-08
Implement `ListRemove` in the LLVM Backend

### LINKED ISSUE #934: List Remove is not implemented in the llvm backend
```py
def test():
    x: list[i32] = [1, 2, 3]
    x.remove(5)
    i: i32
    for i in range(len(x)):
        print(x[i])

test()
```
```py
case stmtType::ListRemove: { v.visit_ListRemove((const ListRemove_t &)x); return; }
LCompilersException: visit_ListRemove() not implemented
```

Thirumalai-Shaktivel:
Closes #934


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -373,6 +373,129 @@ namespace LFortran {
...
+        builder->CreateCondBr(cond, thenBB, elseBB);
+
+        builder->SetInsertPoint(thenBB);
+            // "Element: `item` not found in the list"
+            // replace this with Diagnostic Error message.
+            // Print the `item` in the error message.
+        builder->CreateBr(mergeBB);
+
+        llvm_utils->start_new_block(elseBB);
+        llvm_utils->start_new_block(mergeBB);
```
Thirumalai-Shaktivel, line 438:
```c++
         if (i == end_point) {
            std::cout << "Element: `item` not found in the list";
         }
```
Is this how we implement the above If statement?

I need help moving `print_util` to use here as well.
Also, How do I use `Diagnostic message` here?
@certik @czgdp1807

certik: This happens at runtime in the generated code, while the `Diagnostic` class is in LFortran at compile time, so the two live in different worlds at different times.

In here if the item cannot be found we simply do whatever the semantics of the operation is. That is, we have to choose what happens if `ListRemove` can't find the item. I would do what we do for if `ArrayItem` is out of bounds --- we should check it (optionally) and stop the program (with a nice stacktrace eventually).

certik:
Excellent, thanks for working on this.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -373,6 +373,129 @@ namespace LFortran {
...
+        llvm::Function *fn = builder->GetInsertBlock()->getParent();
+        llvm::BasicBlock *thenBB = llvm::BasicBlock::Create(context, "then", fn);
+        llvm::BasicBlock *elseBB = llvm::BasicBlock::Create(context, "else");
+        llvm::BasicBlock *mergeBB = llvm::BasicBlock::Create(context, "ifcont");
+        builder->CreateCondBr(cond, thenBB, elseBB);
+
+        builder->SetInsertPoint(thenBB);
+            // "Element: `item` not found in the list"
+            // replace this with Diagnostic Error message.
+            // Print the `item` in the error message.
```
Thirumalai-Shaktivel, line 434: I need to replace this with an Exception saying "Element not found in the list", right?
For that I added `throw LCompilersException()`, I don't know what the problem is, but even when the `cond` is false the stacktrace appears.

Thirumalai-Shaktivel: You mean, The non-existing element has to be handled before it comes to the Back-end? (Like in the Python_ast_to_asr or something similar)

czgdp1807: How can you handle any input element at compile time when its value will be known only at run-time? I can call, `list.remove(x)` where is `x` is available only at runtime (say loaded from a file, comes from C runtime or whatever).

Thirumalai-Shaktivel: Sorry my bad, I got it!
How do I handle the above issue?

czgdp1807: Well I don’t know for sure. So, LPython raises AssertionError if assert check fails. We can do something similar for ValueError here. It won’t be be as good as CPython but anyways it would be something working and familiar to the users. @certik thoughts?

Thirumalai-Shaktivel: > In here if the item cannot be found we simply do whatever the semantics of the operation is. That is, we have to choose what happens if ListRemove can't find the item. I would do what we do for if ArrayItem is out of bounds --- we should check it (optionally) and stop the program (with a nice stacktrace eventually).

@certik commented this in the previous discussion.

certik: For now just stop the program at runtime with an error. Later we'll add some more unified and better runtime error handling.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -373,6 +373,129 @@ namespace LFortran {
...
+        llvm::BasicBlock *mergeBB = llvm::BasicBlock::Create(context, "ifcont");
+
+        llvm::Value* cond = builder->CreateICmpEQ(
+                              LLVM::CreateLoad(*builder, i), current_end_point);
+        builder->CreateCondBr(cond, thenBB, elseBB);
+        builder->SetInsertPoint(thenBB);
+        {
+            LCompilersException("The list does not contain the element");
+            // Not working!
+        }
```
Thirumalai-Shaktivel, line 437: I assume LCompilersException is used to stop the program. 
Here it `doesn't trigger` if the condition is `true`. 
So, I used `throw LCompilersException()` to resolve it, but this leads to `triggering` even when the condition is `false`.

czgdp1807: Well I feel `throw LCompilersException` will be executed always during compiling the code with LPython.

Reason - `LCompilersException("The list does not contain the element");` is a part of LPython's code so it will execute when LPython is called. Its not a part of generated LLVM code. What you want is generating LLVM code to throw exceptions.

Anyways, I will fix this in your branch and push the changes here.

Thirumalai-Shaktivel: I see! Yup, thank you!

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -373,6 +373,129 @@ namespace LFortran {
...
+        builder->CreateBr(mergeBB);
+
+        llvm_utils->start_new_block(elseBB);
+        llvm_utils->start_new_block(mergeBB);
+
+        return LLVM::CreateLoad(*builder, i);
+    }
+
+    void LLVMList::remove(llvm::Value* list, llvm::Value* item) {
+        llvm::Type* pos_type = std::get<2>(typecode2listtype["164"]);
```
Thirumalai-Shaktivel, line 447: This line is causing `SEGFAULT`
So, I did this to fix:
```diff
diff --git a/src/libasr/codegen/llvm_utils.cpp b/src/libasr/codegen/llvm_utils.cpp
index b9fb5ebdd..1f43aa37e 100644
--- a/src/libasr/codegen/llvm_utils.cpp
+++ b/src/libasr/codegen/llvm_utils.cpp
@@ -375,7 +375,7 @@ namespace LFortran {
 
     llvm::Value* LLVMList::find_item_position(llvm::Value* list,
                                     llvm::Value* item) {
-        llvm::Type* pos_type = std::get<2>(typecode2listtype["i64"]);
+        llvm::Type* pos_type = llvm::Type::getInt32Ty(context);
         llvm::Value* current_end_point = LLVM::CreateLoad(*builder,
                                         get_pointer_to_current_end_point(list));
         llvm::AllocaInst *i = builder->CreateAlloca(pos_type, nullptr);
@@ -444,7 +444,7 @@ namespace LFortran {
     }
 
     void LLVMList::remove(llvm::Value* list, llvm::Value* item) {
-        llvm::Type* pos_type = std::get<2>(typecode2listtype["164"]);
+        llvm::Type* pos_type = llvm::Type::getInt32Ty(context);
         llvm::Value* current_end_point = LLVM::CreateLoad(*builder,
                                         get_pointer_to_current_end_point(list));
         llvm::AllocaInst *item_pos = builder->CreateAlloca(pos_type, nullptr);
``

Comment (czgdp1807) [reply src/libasr/codegen/llvm_utils.cpp:447]: Yes. We already know the type of position so no need to access it from `typecode2listtype`.

File: src/libasr/codegen/llvm_utils.cpp [hunk 5]
```diff
@@ -373,6 +397,146 @@ namespace LFortran {
...
+        llvm::BasicBlock *thenBB = llvm::BasicBlock::Create(context, "then", fn);
+        llvm::BasicBlock *elseBB = llvm::BasicBlock::Create(context, "else");
+        llvm::BasicBlock *mergeBB = llvm::BasicBlock::Create(context, "ifcont");
+
+        llvm::Value* cond = builder->CreateICmpEQ(
+                              LLVM::CreateLoad(*builder, i), current_end_point);
+        builder->CreateCondBr(cond, thenBB, elseBB);
+        builder->SetInsertPoint(thenBB);
+        {
+            // TODO: Allow runtime information like the exact element which is
```
czgdp1807, line 467: This is simple to do but I would do it in a subsequent PR.

czgdp1807:

certik (approved):
Looks good to me. This PR could use the result of https://github.com/lcompilers/lpython/issues/938 (to be done).

src/libasr/codegen/llvm_utils.cpp
Thirumalai-Shaktivel: Thank you for fixing the Exception part. Yup, we can do in the subsequent PR.

Thirumalai-Shaktivel:
Thank you for fixing the string part, I was little confused while handling the string(I don't know where the issue was!).  Now the issue is fixed!

Thirumalai-Shaktivel:
Thank you @certik for the review!

Thirumalai-Shaktivel:
If the PR is good to go in, @czgdp1807 please merge it.

czgdp1807:
Let's clean up the git history first? You already know the steps to do that. 🙂.

redbopo:
I notice that every time to merge will get every commits in the PR get merged. If just want single commit with single PR, maybe we can try git squash merge .

czgdp1807:
I think this PR is big enough and shouldn't be squash merged. See https://github.com/lcompilers/lpython/blob/main/readme_utils/rebasing.md#merging. It has steps to clean up git history of a PR.

czgdp1807:
Yes. That merge commit should be removed.

Thirumalai-Shaktivel:
This PR is ready, any other changes @czgdp1807 ?

czgdp1807 (approved):
@certik Should we merge this?

czgdp1807:
I noticed you approved above so merged it.

Issue #873, opened 2022-08-02
Error:  `x: list[i64] = [1, 2, 3]`


Smit-create:
```
def main0():
    x: list[i64] = [1, 2, 3]

main0()

```

Error:
```
semantic error: Type mismatch in annotation-assignment, the types must be compatible
 --> expr.py:2:5
  |
2 |     x: list[i64] = [1, 2, 3]
  |     ^              ^^^^^^^^^ type mismatch ('list[i64]' and 'list[i32]')

```

certik:
Just like for `x: i64 = 5` we cast the default integer(4) to integer(8), we should cast the integer(4) [1, 2, 3] to integer(8) [1, 2, 3]. So indeed the above is a bug that we need to fix.

Smit-create:
Yeah, I agree too for separating the casting logic in a new file.

Thirumalai-Shaktivel:
Other issues
```py
$ lpython examples/expr2.py --new-parser --show-asr
semantic error: Type mismatch in annotation-assignment, the types must be compatible
 --> examples/expr2.py:4:5
  |
4 |     x: list[i64] = [1, 2]
  |     ^              ^^^^^^ type mismatch ('list[i64]' and 'list[i32]')


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).
```
```py
$lpython examples/expr2.py --new-parser --show-asr
semantic error: Type mismatch in annotation-assignment, the types must be compatible
 --> examples/expr2.py:4:5
  |
4 |     x: tuple[i32, f32] = (1, 1.0)
  |     ^                    ^^^^^^^^ type mismatch ('tuple[i32, f32]' and 'tuple[i32, f64]')


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).
```
```py
$ lpython examples/expr2.py --new-parser --show-asr
semantic error: Type mismatch in annotation-assignment, the types must be compatible
 --> examples/expr2.py:4:5
  |
4 |     x: tuple[i64, f64] = (1, 1.0)
  |     ^                    ^^^^^^^^ type mismatch ('tuple[i64, f64]' and 'tuple[i32, f64]')


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).
```

czgdp1807:
So here are my thoughts on your examples,

[...]

PR #940, opened 2022-08-08
Revamping casting mechanism

czgdp1807:
Addresses - https://github.com/lcompilers/lpython/issues/873#issuecomment-1202577019


tests/reference/cpp-expr3-9c516d4.stdout
```diff
@@ -34,11 +34,11 @@ void test_cast()
...
     a = 2;
     b =   4.20000000000000018e+00;
-    a = (float)(a)*b;
+    a = (int)((float)(a)*b);
     b = b + (float)(1);
     a = 5;
-    a = (float)(a) -   3.89999999999999991e+00;
-    a = (float)(a)/b;
+    a = (int)((float)(a) -   3.89999999999999991e+00);
+    a = (int)((float)(a)/b);
```
certik, line 41: This is not correct. It needs to return a compiler error. Consider this code:
```python
def test_cast():
    a = 2
    b = 4.2
    a *= b
    b += 1
    a = 5
    print(a, type(a))
    a -= 3.9
    print(a, type(a))
    a /= b
    b = 3/4
    if a < b:
        print("a < b")

test_cast()
```
Which prints in CPython:
```console
$ python tests/expr3.py
(5, <type 'int'>)
(1.1, <type 'float'>)
```
Essentially, the `a -= 3.9` operation changes `a` from an int to a float.

In LPython variables cannot change their types, so we need to give a compiler error for this and not allow it.

The user must explicitly type this.

certik:

tests/reference/cpp-expr3-9c516d4.stdout
czgdp1807: Okay. Well the above code compiles in `main` branch as well. So, you mean any binary expression like `int - float` or `float + complex` shouldn't be allowed right?

czgdp1807: Or you mean assigning integer to float and float to integer shouldn't be allowed?

czgdp1807: > In LPython variables cannot change their types, so we need to give a compiler error for this and not allow it.

Well then we will have to remove the casting mechanism all together from LPython, because many test expressions work in `main` branch because casting is performed on `left`, `right` parts of `BinOp`? Should I remove casting and fix the tests and make all the expressions strictly typed?

Smit-create (approved):
Thanks!!!

tests/reference/cpp-expr3-9c516d4.stdout
czgdp1807: Or specifically the problem is with shorthand operations, `a -= 3.9`? If so then this can be fixed by making `AugAssign` visitor strictly typed.

czgdp1807: So I assumed that https://github.com/lcompilers/lpython/pull/940#discussion_r942294345 is the case. Since, the tested you posted in https://github.com/lcompilers/lpython/pull/940#discussion_r942279864 is same as the one present in https://github.com/lcompilers/lpython/blob/main/tests/expr3.py. So if we want to raise a compiler error for `a -= 3.9` then we will have to remove the following file https://github.com/lcompilers/lpython/blob/main/tests/reference/asr-expr3-10eb0a4.stdout and replace with `stderr` version.

certik: Yes

certik:
It's better. But the file `tests/expr3.py` cannot compile to ASR, because it changes the type of the variable from int to float, which is not allowed in LPython, and instead we fix it up by casting (or be letting it use the original type), both of which is incorrect and will give different results than CPython. We have to modify it and change it to be conforming.

tests/reference/cpp-expr3-9c516d4.stdout
czgdp1807: Done. Needed a little bit of fixing at some places. Did that.

czgdp1807:
@certik I have made the necessary changes. Is there anything else to be done? If not then I will clean up the git history and merge.

integration_tests/expr_05.py
```diff
@@ -14,7 +14,7 @@ def main0():
     assert test_multiply(a, b) == -50
     i: i64
     i = 1
-    i += 1
+    i += int(1)
```
certik, line 17: We'll have to figure out how to create i64 vs i32 in some systematic way, but for now this is fine.

certik (approved):
I think this looks good now. Good to merge after git history cleanup.

PR #948, opened 2022-08-09
[Parser] Fix bug with non-ASCII characters in the new parser

akshanshbhatt:
Partially fixes #916 (New parser only)

CPython supports Unicode characters for identifier names, but the old definition only restricted it to the ASCII character set.
https://github.com/lcompilers/lpython/blob/d40c2c1076df2381437aebf3052377d4ef594a8a/src/lpython/parser/tokenizer.re#L271-L272

However, we cannot simply fix this by adding all the non-ASCII characters with the old char definition, like this -

```re
char = [^\x00-\x7F]|[a-zA-Z_];
```

This is because CPython restricts the first letter of an identifier to be a numeral, which holds for any language. Here is an example -

```py
❯ bat examples/expr2.py
───────┬─────────────────────────────────────────────────────────────
       │ File: examples/expr2.py
───────┼─────────────────────────────────────────────────────────────
   1 ~ │ def क१():
   2 ~ │     print("Hello")
   3   │
   4 ~ │ क१()
───────┴─────────────────────────────────────────────────────────────
❯ python3.11 examples/expr2.py
Hello
```
```py
❯ bat examples/expr2.py
───────┬─────────────────────────────────────────────────────────────
       │ File: examples/expr2.py
───────┼─────────────────────────────────────────────────────────────
   1 ~ │ def १क():
   2 ~ │     print("Hello")
   3   │
   4 ~ │ १क()
───────┴─────────────────────────────────────────────────────────────
❯ python3.11 examples/expr2.py
  File "/Users/akshansh/Documents/GitHub/lpython/examples/expr2.py", line 1
    def १क():
        ^
SyntaxError: invalid character '१' (U+0967)
```

`१` is a Hindi numeral, so it throws an error if used as the first character of the identifier.
I came across this article https://www.regular-expressions.info/unicode.html which suggested there are well-defined Unicode categories that can be used in a regular expression. We require all the characters in the `letter` category for `char`. In re2c, we can import these categories by including a header comment like this -

```c
/*!include:re2c "re2c-2.2/include/unicode_categories.re" */
```

I tried this, but it did not recognize the file's path for some reason. So instead, I just copied the regex for the letter category from the file mentioned in https://github.com/skvadrik/re2c/issues/235#issuecomment-580955677.

PR #951, opened 2022-08-09
[Parser] Add support for unicode identifiers

akshanshbhatt:
Continuation of #948


certik:
This change looks much better. How much bigger is the generated tokenizer compared to master?

certik:
On my computer:
```console
$ time lpython --show-ast --new-parser a.py --no-color > a.txt
lpython --show-ast --new-parser a.py --no-color > a.txt  0.13s user 0.03s system 96% cpu 0.157 total
```

certik (approved):
Let's merge this change, since it's a lot simpler.

The only possible downside of this is that it does not reject technically invalid names that start with a Hindi numeral, but if users were requesting this (I don't think they would, or not soon), we could add an optional ASR pass that would check this and report it as an error and users who care about this can run that pass.

Thirumalai-Shaktivel:
Thank you!

Issue #857, opened 2022-07-31
List append benchmark


certik:
Here is a simple benchmark for appending to a list in Python:
```python
from ltypes import i32

def test_list():
    a: list[i32] = [0, 1, 2, 3, 4]
    n: i32 = 100000000
    i: i32
    for i in range(n):
        a.append(i + 5)
    print(a[n])

test_list()
```
and C++:
```c++
#include <vector>
#include <iostream>

int main() {
    std::vector<int32_t> a = {0, 1, 2, 3, 4};
    int32_t n = 100000000;
    for (int32_t i = 0; i < n; i++) {
        a.push_back(i + 5);
    }
    std::cout << a[n] << std::endl;
    return 0;
}
```

Results on Apple M1 Max (I ran each benchmark many times, took the lowest numbers):
```console
$ clang++ -std=c++17 -O3 a.cpp
$ time ./a.out
100000000
./a.out  0.07s user 0.09s system 94% cpu 0.170 total
$ lpython --fast a.py
$ time ./a.out
100000000
./a.out  0.05s user 0.03s system 93% cpu 0.089 total
$ time PYTHONPATH=../src/runtime/ltypes python a.py                 
100000000
PYTHONPATH=../src/runtime/ltypes python a.py  4.21s user 0.51s system 99% cpu 4.730 total
```

| Compiler | Time | Relative |
| --------- | ----- | -------- |
| LPython | 0.089 | 1.0 |
| Clang++ | 0.170 | 1.91 |
| Python | 4.730 | 53.15 |

Versions:
```console
$ clang++ --version
Apple clang version 13.0.0 (clang-1300.0.29.30)
Target: arm64-apple-darwin21.3.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
$ lpython --version
LPython version: 0.3.0-350-g6d29003ea
Platform: macOS ARM
Default target: arm64-apple-darwin21.3.0
$ python --version
Python 3.10.2
```

Thanks @czgdp1807 for implementing lists in our LLVM backend (https://github.com/lcompilers/lpython/pull/835)! This is just a first implementation, but I already like the results. :)

certik:
I think we currently do not deallocate at the end, so one can also benchmark just the append cycle:
```c++
#include <vector>
#include <iostream>
#include <chrono>

int main() {
    std::vector<int32_t> a = {0, 1, 2, 3, 4};
    int32_t n = 100000000;
[...]

PR #952, opened 2022-08-10
Supporting list of lists, list of tuples and tuples with lists

czgdp1807:


certik:
Excellent!

namannimmo10:
Thanks @czgdp1807!

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -587,4 +594,43 @@ namespace LFortran {
         }
     }
 
+    llvm::Value* LLVMTuple::check_tuple_equality(llvm::Value* t1, llvm::Value* t2,
```
czgdp1807, line 597: This needs to go into a more generic function I believe.

czgdp1807: I will do that in the next commit, so that list of lists and tuple with list work well.

integration_tests/test_list_06.py
```diff
@@ -0,0 +1,48 @@
...
+    for k in range(rows):
+        tensor.append(mat)
+        for i in range(rows):
+            for j in range(cols):
+                mat[i][j] += float(1)
+
+    for k in range(rows):
+        for i in range(rows):
+            for j in range(cols):
+                assert mat[i][j] - tensor[k][i][j] == rows - k
```
czgdp1807, line 46: @certik This check would fail with CPython because in CPython lists are shallow copied but in LPython we deepcopy lists (see https://github.com/lcompilers/lpython/pull/835#discussion_r932470488). So I haven't enabled `cpython` for now in the tests. We can make it work with CPython by calling `copy.deepcopy` and ignoring the same in LPython as we anyways do deepcopy by default. What do you say?

certik: Yes, using `deepcopy` should work.

czgdp1807:

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -250,15 +304,73 @@ namespace LFortran {
...
+                llvm::Value* tmp = builder->CreateAdd(
+                            pos,
+                            llvm::ConstantInt::get(context, llvm::APInt(32, 1)));
+                LLVM::CreateStore(*builder, tmp, pos_ptr);
+            }
+
+            builder->CreateBr(loophead);
+
+            // end
+            llvm_utils->start_new_block(loopend);
```
certik, line 353: Does this slow anything down in the already existing list (or any other) functionality in master? For example in our current benchmarks for lists? Or is this only used for new features implemented in this PR, but does not affect any feature already in master.

certik (approved):
Amazing. That's super exciting that all these features now work.

Everything looks good. I left one question below. If this does not slow down anything that is already in master, then it can be merged.

src/libasr/codegen/llvm_utils.cpp
czgdp1807: This code path will only be taken for nesting of types inside lists. For `list[f64]` (i.e., a list with intrinsic element type, the code path in `main` will be taken). So, theoretically no change in benchmarks should happen.

certik: When you say "code path", you mean code path at compile time? If so, then this is good.

If you mean "runtime", then that must mean there might be some "if statement" being executed at runtime, and we don't want that.

czgdp1807: > When you say "code path", you mean code path at compile time? 

Yes code path at compile time.

czgdp1807: Updated the benchmarks for my linux system in https://github.com/lcompilers/lpython/issues/857#issuecomment-1200960150. I don't see any change with `list05` (the branch of this PR).

czgdp1807: Updated https://github.com/lcompilers/lpython/issues/857#issuecomment-1200960150 with results from my macbook, this and main branch give no difference for me.

czgdp1807: Bottomline - This and main branch give same results on both of my systems. So, theoretically and practically from my perspective this PR wouldn't slow down anything for features already existing in `main`. @certik Do you want to merge it?

certik:
Yes, for sure. Thank you so much for this. Very exciting.

PR #954, opened 2022-08-11
strings: capialize, lower

Abdelrahman-Kh-Fouad:


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -737,6 +737,19 @@ LFORTRAN_API int32_t _lpython_bit_length8(int64_t num)
...
+    char *res = (char*)malloc(len+1);
+    strcpy(res, str);
+    if (len != 0) {
+        if(res[0] >= 'a' && res[0] <= 'z') {
+            res[0] -= 32;
+        }
+    }
+    res[len] = '\0';
+    return res;
+}
```
certik, line 751: This function I think we can implement in LPython, so we should, and remove it from ASR.

We should only put things into ASR that cannot be implemented in LPython.

Abdelrahman-Kh-Fouad: i've implemented  like `_lpython_floordiv` is that the correct way?

certik:
@Abdelrahman-Kh-Fouad is this ready for review?

Abdelrahman-Kh-Fouad:
> @Abdelrahman-Kh-Fouad is this ready for review?

yes it's

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -650,6 +653,43 @@ struct PythonIntrinsicProcedures {
...
+                std::to_string(args.size()) + " given)", loc);
+        }
+        ASR::expr_t *arg1 = args[0];
+        ASR::ttype_t *arg1_type = ASRUtils::expr_type(arg1);
+        if (ASRUtils::is_character(*arg1_type)) {
+            std::string val = ASR::down_cast<ASR::StringConstant_t>(arg1)->m_s;
+            ASR::ttype_t *type = ASRUtils::TYPE(ASR::make_Character_t(al, loc,
+                                        1, 1, nullptr, nullptr, 0));
+            ASR::ttype_t *res_type = ASRUtils::TYPE(ASR::make_StringConstant_t(al, loc, s2c(al, ""), type));
+            return ASR::down_cast<ASR::expr_t>(ASR::make_StringConstant_t(al, loc, s2c(al, val),  res_type));
```
certik, line 669: Shouldn't the code here implement the capitalization at compile time? It seems it just repackages the first argument as `StringConstant`, but does not capitalize.

certik:
I think this is generally good, thanks for implementing it!

src/lpython/semantics/python_comptime_eval.h
certik: No, just keep adding commits until it is completely reviewed and ready. Only as the very last thing before merge you can rebase /  polish the history.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4276,6 +4307,39 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
                 }
+            } else if (AST::is_a<AST::ConstantStr_t>(*at->m_value)) {
+                if (std::string(at->m_attr) == std::string("capitalize")) {
+                    if(args.size() != 0) {
+                        throw SemanticError("str.capitalize() takes no arguments",
+                            x.base.base.loc);
+                    }
+                    AST::ConstantStr_t *n = AST::down_cast<AST::ConstantStr_t>(at->m_value);
+                    std::string res = n->m_value;
+                    res[0] = toupper(res[0]);
```
certik, line 4318: What if it is an empty string?

integration_tests/test_str_attributes.py
```diff
@@ -0,0 +1,19 @@
...
+    s = " tom and jerry"
+    assert s.capitalize() == s
+    assert "empty string" .capitalize() == "Empty string"
+
+
+def lower():
+    s: str
+    s = "AaaaAABBbbbbBB!@12223BN"
+    assert s.lower() == "aaaaaabbbbbbbb!@12223bn"
+    assert "DDd12Vv" .lower() == "ddd12vv"
```
certik, line 16: I am confused here --- I would expect this line to fail until you actually implemented the compile time version in 6c7edc877fe1dc202f61731bf66cce70fbec6651, but it seemed it was passing --- is there a bug somewhere? Is this not exercising the compile time version?

integration_tests/test_str_attributes.py
```diff
@@ -0,0 +1,19 @@
+def capitalize():
+    s: str
+    s = "tom and jerry"
+    assert s.capitalize() == "Tom and jerry"
+    s = "12wddd"
+    assert s.capitalize() == s
+    s = " tom and jerry"
+    assert s.capitalize() == s
+    assert "empty string" .capitalize() == "Empty string"
```
certik, line 9: Also add a test for `"".capitalize()`.

src/lpython/semantics/python_ast_to_asr.cpp
Abdelrahman-Kh-Fouad: I added it.

integration_tests/test_str_attributes.py
Abdelrahman-Kh-Fouad: i spend alot of time for tracing, i found that in evaluation condition `arg_values.size() == args.size()` doesn't doesn't happen, so `eval__lpython_str_capitalize` function doesn't called.

certik (approved):
I think this is fine.

PR #877, opened 2022-08-03
Fmean, mean overload functions implemented

Madhav2310:

PR #958, opened 2022-08-12
Initial implementation of statistics.mode()

Madhav2310:
We can improve it by having an efficient internal sort function(which I was having a little trouble with https://github.com/lcompilers/lpython/pull/877#discussion_r938295994 ) and also by using the collections.count() method. We can look into implementing these soon

### DIFF
--- a/integration_tests/test_statistics.py
+++ b/integration_tests/test_statistics.py
@@ -1,6 +1,8 @@
 from statistics import (mean, fmean, geometric_mean, harmonic_mean, variance,
-                        stdev, pvariance, pstdev, correlation, covariance, linear_regression)
+                        stdev, pvariance, pstdev, correlation, covariance,
+                        linear_regression, mode)
 from ltypes import i32, f64, i64, f32
+
 
 eps: f64
 eps = 1e-12
@@ -187,6 +189,31 @@ def test_linear_regression():
     assert abs(slope + 0.18514007308160782) < eps
     assert abs(intercept - 25.750304506699152) < eps
 
+def test_mode():
+    a: list[i32]
+    a = [3, 1, 12, 4, 0]
+    i: i32
+    i = mode(a)
+    assert i == 3
+
+    b: list[i32]
+    b = [4, 2, 4, 4, 2, 3, 5]
+    j: i32
+    j = mode(b)
+    assert j == 4
+
+    c: list[i32]
+    c = [2, 3, 4, 1, 2, 4, 5]
+    k: i32
+    k = mode(c)
+    assert k == 2
+
+    d: list[i32]
+    d = [-1, 2, -3, -5, -3, -1, 4, -2, 4, -5, -3, 4, -3]
+    k = mode(d)
+    assert k == -3
+
+
 def check():
     test_mean()
     test_geometric_mean()
@@ -199,5 +226,6 @@ def check():
     test_linear_regression()
     test_correlation()
     test_covariance()
+    test_mode()
 
 check()
--- a/src/runtime/statistics.py
+++ b/src/runtime/statistics.py
@@ -220,6 +220,28 @@ def harmonic_mean(x: list[f64]) -> f64:
 
     return k / sum
 
+
+def mode(x: list[i32]) -> i32:
+    k: i32 = len(x)
+    c: i32
+    count: dict[i32, i32] = {0: 0}
+
+    # insert keys in the dictionary
+    for c in range(k):
+        count[x[c]] = 0
+
+    # update the frequencies
+    for c in range(k):
+        count[x[c]] = count[x[c]] + 1
+
+    max_count: i32 = 0
+    ans: i32
+    for c in range(k):
+        if max_count < count[x[c]]:
+            max_count = count[x[c]]
+            ans = x[c]
+    return ans
+
 @overload
 def variance(x: list[f64]) -> f64:
     """
@@ -270,6 +292,7 @@ def stdev(x: list[i32]) -> f64:
     Returns the standard deviation of a data sequence of numbers
     """
     return variance(x)**0.5
+
 
 @overload
 def pvariance(x: list[f64]) -> f64:
@@ -497,4 +520,3 @@ def linear_regression(x: list[f64], y: list[f64]) -> tuple[f64, f64]:
     LinReg: tuple[f64, f64] = (slope, intercept)
 
     return LinReg
-    
--- a/src/runtime/statistics.py
+++ b/src/runtime/statistics.py
@@ -221,6 +221,7 @@ def harmonic_mean(x: list[f64]) -> f64:
     return k / sum
 
 
+# TODO: Use generics to support other types.
 def mode(x: list[i32]) -> i32:
     k: i32 = len(x)
     c: i32


src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+    Returns the median of a data sequence of numbers
+    """
+    k: i32 = len(x)
+    if k == 0:
+        raise Exception("No mode for empty data")
+    i: i32
+    j: i32
+    for i in range(k):
+        for j in range(i+1,k):
+            if x[i]>x[j]:
```
czgdp1807, line 153:
```suggestion
            if x[i] > x[j]:
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+    """
+    Returns the median of a data sequence of numbers
+    """
+    k: i32 = len(x)
+    if k == 0:
+        raise Exception("No mode for empty data")
+    i: i32
+    j: i32
+    for i in range(k):
+        for j in range(i+1,k):
```
czgdp1807, line 152:
```suggestion
        for j in range(i + 1, k):
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+            if x[i]>x[j]:
+                x[i],x[j]=x[j],x[i]
```
czgdp1807, line 154:
```suggestion
                x[i], x[j] = x[j] ,x[i]
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+    ans: i32 =x[0]
```
czgdp1807, line 155:
```suggestion
    ans: i32 = x[0]
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+    init: i32 = x[0]
+    count: i32 = 0
+    max_count: i32 = 0
+    c: i32
+    for c in range(k):
+        if(x[c] == init):
```
czgdp1807, line 161:
```suggestion
        if x[c] == init:
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+            count = count +1
+            if max_count>count:
```
czgdp1807, line 163:
```suggestion
            if max_count > count:
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+                x[i],x[j]=x[j],x[i]
+    ans: f64 =x[0]
+    init: f64 = x[0]
+    count: i32 = 0
+    max_count: i32 = 0
+    c: i32
+    for c in range(k):
+        if(x[c] == init):
+            count = count +1
+            if max_count>count:
```
czgdp1807, line 192:
```suggestion
            if max_count > count:
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+    Returns the median of a data sequence of numbers
+    """
+    k: i32 = len(x)
+    if k == 0:
+        raise Exception("No mode for empty data")
+    i: i32
+    j: i32
+    for i in range(k):
+        for j in range(i+1,k):
+            if x[i]>x[j]:
```
czgdp1807, line 182:
```suggestion
            if x[i] > x[j]:
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+                x[i],x[j]=x[j],x[i]
```
czgdp1807, line 183:
```suggestion
                x[i], x[j] = x[j], x[i]
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+    ans: f64 =x[0]
+    init: f64 = x[0]
+    count: i32 = 0
+    max_count: i32 = 0
+    c: i32
+    for c in range(k):
+        if(x[c] == init):
```
czgdp1807, line 190:
```suggestion
        if x[c] == init:
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+            count = count +1
```
czgdp1807, line 191:
```suggestion
            count = count + 1
```

src/runtime/statistics.py
czgdp1807, line 192:
```suggestion
            if max_count > count:
```

czgdp1807 (changes requested):
Please clean up the code.

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
 
     return k/sum
 
+@overload
+def mode(x: list[i32]) -> i32:
+    """
+    Returns the median of a data sequence of numbers
+    """
```
Smit-create, line 145:
```suggestion
    """
    Returns the mode of a data sequence of numbers
    """
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,64 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+                ans = x[c]
+        else:
+            init = x[c]
+            count =1
+    return ans
+
+@overload
+def mode(x: list[f64]) -> f64:
+    """
+    Returns the median of a data sequence of numbers
```
Smit-create, line 173:
```suggestion
    Returns the mode of a data sequence of numbers
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,66 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+    Returns the mode of a data sequence of numbers
+    """
+    k: i32 = len(x)
+    if k == 0:
+        raise Exception("No mode for empty data")
+    i: i32
+    j: i32
+    ans: i32 = x[0]
+    for i in range(k):
+        for j in range(i+1,k):
```
czgdp1807, line 153:
```suggestion
        for j in range(i + 1, k):
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,66 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+            if x[i] > x[j]:
+                x[i], x[j] = x[j], x[i]
+
+    init: i32 = x[0]
+    count: i32 = 0
+    max_count: i32 = 0
+    c: i32
+    for c in range(k):
+        if x[c] == init:
+            count = count +1
```
czgdp1807, line 163:
```suggestion
            count = count + 1
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,66 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+    Returns the mode of a data sequence of numbers
+    """
+    k: i32 = len(x)
+    if k == 0:
+        raise Exception("No mode for empty data")
+    i: i32
+    j: i32
+    ans: f64 = x[0]
+    for i in range(k):
+        for j in range(i+1,k):
```
czgdp1807, line 183:
```suggestion
        for j in range(i + 1, k):
```

src/runtime/statistics.py
```diff
@@ -138,6 +138,66 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+            if x[i] > x[j]:
+                x[i], x[j] = x[j], x[i]
+
+    init: f64 = x[0]
+    count: i32 = 0
+    max_count: i32 = 0
+    c: i32
+    for c in range(k):
+        if(x[c] == init):
+            count = count +1
```
czgdp1807, line 193:
```suggestion
            count = count + 1
```

czgdp1807 (changes requested):

integration_tests/test_statistics.py
```diff
@@ -120,6 +79,36 @@ def test_stdev():
...
+    d = [13.2, 24.0, 3.1, 41.9, 5.0]
+    l: f64
+    l = mode(d)
+    assert l == 13.2
+
+    e: list[f64]
+    e = [1.7, 2.1, 2.1, 30.0, 49.7, 49.7, 49.7, 5.0]
+    m: f64
+    m = mode(e)
+    assert m == 49.7
```
czgdp1807, line 111:
```suggestion
    assert abs(m - 49.7) < 1e-12
```

@Smit-create Would you mind changing these equality checks over floating points in a similar manner in this whole diff? It would be really helpful.

czgdp1807 (changes requested):

src/runtime/statistics.py
```diff
@@ -138,6 +137,66 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+    if k == 0:
+        raise Exception("No mode for empty data")
+    i: i32
+    j: i32
+    ans: i32 = x[0]
+    for i in range(k):
+        for j in range(i + 1, k):
+            if x[i] > x[j]:
+                x[i], x[j] = x[j], x[i]
+
```
Smit-create, line 155: I am not sure that this works in lpython currently, and that is why the CI fails.

czgdp1807: Well making it work is easy.

1. First check whether the assignment is `TupleConstant` to `TupleConstant`. Assume that both `TupleConstant`s have same type in LLVM backend. Reason - If they weren't having same types then AST to ASR transition should have raised a semantic error.
2. Now, take `deepcopy` of each of the elements present in the value `TupleConstant` and store them in a vector of type, `Vec<llvm::Value*> src_values`.  Now do,

```cpp
ptr_loads_copy = ptr_loads;
ptr_loads = 0;
for( size_t i = 0; i < target_tuple->n_elements; i++ ) {
     visit_expr_wrapper(*target_tuple->m_elements[i], false);
     LLVM::CreateStore(*builder, src_values[i], tmp);
}
ptr_loads = ptr_loads_copy;
```

Do you want to try out working on this @Smit-create?

src/runtime/statistics.py
```diff
@@ -138,6 +137,66 @@ def harmonic_mean(x: list[i32]) -> f64:
...
+    count: i32 = 0
+    max_count: i32 = 0
+    c: i32
+    for c in range(k):
+        if x[c] == init:
+            count = count + 1
+            if max_count > count:
+                ans = x[c]
+        else:
+            init = x[c]
```
Smit-create, line 166: The logic of this algorithm is wrong. This is an incorrect way of computing mode.

Smit-create:
@czgdp1807 This looks good to go.

src/runtime/statistics.py
```diff
@@ -220,6 +220,64 @@ def harmonic_mean(x: list[f64]) -> f64:
...
+# @overload
+# def mode(x: list[f64]) -> f64:
+#     """
+#     Returns the mode of a data sequence of numbers
+#     """
+#     k: i32 = len(x)
+#     if k == 0:
+#         raise Exception("No mode for empty data")
+#     ans: f64
+#     count: dict[f64, i32] = {0.0: 0}
```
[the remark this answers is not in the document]
czgdp1807: https://stackoverflow.com/a/690791. I think we shouldn't use dicts here and allow for an error tolerance as input. Anyways removing `f64` overloads is better for now, instead of waiting for hash function for `f64` to appear. `49.7999999999` and `49.8` are same or different. See the ambiguity below,

```python
>>> import numpy as np
>>> from scipy import stats
>>> d = np.array([1.7, 2.1, 2.1, 30.0, 49.7999999999, 49.7999999999, 49.7999999999, 49.8, 5.0])
>>> m = stats.mode(d)
>>> m
ModeResult(mode=array([49.8]), count=array([3])) # 49.8 is the mode but then the frequency is wrong.
```

czgdp1807 (changes requested):
Overall remove `f64` overloads for `mode` for now, even the commented ones.

src/runtime/statistics.py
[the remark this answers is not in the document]
certik: I would not use floats as keys, only as values in dicts.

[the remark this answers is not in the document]
certik: We should disallow it in AST->ASR, with a nice error message and explanation.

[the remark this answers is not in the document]
certik: If users really want it, they can cast the f64 to i64 and use that as keys.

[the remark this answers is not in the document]
czgdp1807: Yes. If users want to use `f64` keys in LPython they should provide their own hash function and their own equality check to compare floats in case of collisions. Until we support these advanced features an error message at ASR level should be fine.

integration_tests/test_statistics.py
```diff
@@ -164,6 +165,26 @@ def test_correlation():
...
+    assert i == 3
+
+    b: list[i32]
+    b = [4, 2, 4, 4, 2, 3, 5]
+    j: i32
+    j = mode(b)
+    assert j == 4
+
+    c: list[i32]
+    c = [2, 3, 4, 1, 2, 4, 5]
```
czgdp1807, line 182: Try for negative numbers as well?

Smit-create: I see that negative keys produce some incorrect result.

integration_tests/test_statistics.py
```diff
@@ -187,6 +189,33 @@ def test_linear_regression():
...
+    k: i32
+    k = mode(c)
+    assert k == 2
+
+    # TODO: dictionary giving incorrect results with negative numbers
+    # d: list[i32]
+    # d = [-1, 2, -3, -5, -3, -1, 4, -2, 4, -5, -3, 4, -3]
+    # k = mode(d)
+    # print(k)
+    # assert k == -3
```
czgdp1807, line 216: I will look into this.

czgdp1807: @Smit-create Try using https://github.com/lcompilers/lpython/pull/1090/commits/a8be9663b7cb5d87f23ff33f8a02dc78c179cb3f commit.

Smit-create: Thanks, fixed.

certik:
Looks good to me. @czgdp1807 can you please review and merge?

Does LPython support deepcopy? I generally don't like using it, but if you have to make a copy, that would be the way to do it.

czgdp1807:
> Does LPython support deepcopy?

Yes in LLVM backend I added a generalised deepcopy generator for any ASR type. See below. We need to update it to cover more ASR types. Better to do it iteratively as the need arises. Though whenever adding a new feature we should make sure we are not making a shallow copy of stuff. Its very important to maintain this behaviour consistently.

https://github.com/lcompilers/lpython/blob/91cb772ef86a58d8b5020a8dde864754bc06d06a/src/libasr/codegen/llvm_utils.cpp#L300-L326



> I generally don't like using it, but if you have to make a copy, that would be the way to do it.

Correct. For ensuring `deepcopy` in CPython you have to call deepcopy function, otherwise it will always to shallow copy by default for mutable data structures.

src/runtime/statistics.py
```diff
@@ -220,6 +220,39 @@ def harmonic_mean(x: list[f64]) -> f64:
 
     return k / sum
 
+
+def mode(x: list[i32]) -> i32:
```
czgdp1807, line 224: We don't have support for `i64` right? Probably generics will be helpful here. Not sure what's the right time to start using them but this is a good use case of using generics along with restrictions.

src/runtime/statistics.py
```diff
@@ -220,6 +220,39 @@ def harmonic_mean(x: list[f64]) -> f64:
...
+    k: i32 = len(x)
+    y: list[i32]
+    y = deepcopy(x)
+    c: i32; i: i32; j: i32;
+    for i in range(k):
+        for j in range(0, k-1):
+            if y[j] > y[j+1]:
+                c = y[j]
+                y[j] = y[j+1]
+                y[j+1] = c
```
czgdp1807, line 234: This is a sort operation right?

czgdp1807: Why are we doing this if we are going to use `dict` anyways for storing value to frequency?

czgdp1807: Is this to preserve consistency of outputs because `dict` is a unordered?

czgdp1807:
> can you please review and merge?

Left a bunch of questions. Waiting for the answer.

src/runtime/statistics.py
certik: I would start using generics in more tests first, not in the runtime library. I think there will be a lot of things to fix up and improve for generics, especially syntax wise. So we just have to start slowly playing with it and improving it.

czgdp1807: I see. Let's add a comment here then `TODO: Use generics to support other types`.

Smit-create: Done

src/runtime/statistics.py
Smit-create: Oh, I changed the algorithm. It was to preserve the order but that can be done other way as in the updated diff.

czgdp1807 (approved):

src/runtime/statistics.py
czgdp1807: Yes. Its O(N) in both space and time which is the best we can go? Probably if we want to preserve space then we can sort the list in O(NlogN) by using O(logN) space (max size of recursion stack I think) but then it would be too complicated and slower.

PR #965, opened 2022-08-13
Declare variables on the same line

certik:
This is the most natural Python syntax to declare variables on the same
line.

### DIFF
--- a/integration_tests/loop_01.py
+++ b/integration_tests/loop_01.py
@@ -1,14 +1,10 @@
 def main0():
-    s: str
-    s = 'aabbcc'
-    c: str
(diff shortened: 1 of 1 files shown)


czgdp1807 (approved):
This is cool.

czgdp1807:
I think we can merge this?

certik:
Sure. I think this is good enough. It's slightly annoying to do `x: i32; y: i32; z: i32` instead of `x, y, z: i32`, but the Python syntax does not allow it. And doing  something like `i32("x y z")` seems not worth it. I think the above is good enough for now.

czgdp1807:
Yeah at least every declaration stays in the same line.

Issue #964, opened 2022-08-13
Supportding `dict` in LLVM backend


czgdp1807:
I gave a thought on how should we go about implementing this. Before writing my conclusion following points should be noted,

1. Implementing a Hash function is tricky for now in LLVM backend. In addition, using a pre-existing one written in C language sounds like a good choice for me.
2. Scaling for various combination of types is extremely tricky and unsafe in C. Our `asr_to_llvm.cpp` allows to scale very easily. We did this for lists and tuples in LLVM backend by generating only the types required by the user code.
3. In future I believe, we will be able to implement complicated hash functions in Python itself and then compile them via LPython.

So now my conclusion is,

1. Implement everything other than hash function in LLVM itself, like collision handling, accessing key-value pair, etc.
2. Implement hash function in C for now and call it from LLVM. In future once we will be able to write and compile hash function in Python itself then we will directly call it from LLVM and stop using the one defined in C runtime.
3. We should implement everything in dict so that drop in replacement of hash functions is possible. 

Rest of the details, blockers I will try to figure out while actually implementing `dict`.

czgdp1807:
@certik What are your thoughts on this?

certik:
Yes, I think that seems like a good plan forward. Let's start in LLVM/C, and then later as LPython becomes stronger, we can move some parts of it to LPython and compile to ASR and then call it from the LLVM backend.

PR #975, opened 2022-08-15
Initial support for `dict` data structure

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/964

Following is the plan,

1. Represent `dict` with the help of a struct in LLVM. Maintain original key-value pairs for the purpose of collision handling and in case a collision happens then distinguishing between update and insert operation. We need original key-value pairs anyways for all kinds of collision handling strategy.
2. If collision happens while insertion then do [linear probing](https://web.stanford.edu/class/archive/cs/cs166/cs166.1166/lectures/12/Small12.pdf) (simple as well as cache efficient).
3. Resizing strategy - [All-at-once resizing strategy](https://en.wikipedia.org/wiki/Hash_table#Resizing_by_moving_all_entries). Will happen if load factor touches a threshold value. Simpler than lazy resizing in which don't copy all the older key-value pairs into new array but instead only when they are accessed. Its a bit complicated to implement so I will do it in future.

TODOs after this PR - https://github.com/lcompilers/lpython/issues/983#issuecomment-1218136744


czgdp1807:
@certik This is ready for review. I have mentioned my TODOs after this PR in https://github.com/lcompilers/lpython/issues/983#issuecomment-1218136744. You can review each commit individually if you want. Let me know if anything needs a change here.

certik (approved):
I reviewed every commit, everything looks great.

My only possible question is if the CPython version of the test takes a long time, if so, we should reduce the number of loop iterations in the test. 

That can be adjusted later.

So +1 to merge.

Great job!

certik:
@czgdp1807 is this ordered or unordered dictionary?

czgdp1807:
I think this is unordered dictionary (because hash maps are unordered in nature). For ordered dictionary we will require an implementation similar to std::map.

certik:
Here is another way to implement OrderedDict: https://github.com/Aqcurate/OrderedDict#ordered-dictionary-implementation

PR #982, opened 2022-08-17
LSP: Add support for goto def in compiler for VSCode extension

ankitaS11:
This PR adds support for `goto-def` option in the compiler, which will be used by VSCode extension. Usage is:

```
lpython --goto-def <index> <file_name>
```

PR for the extension depends on this: https://github.com/lcompilers/lpython-vscode-extension/pull/3

cc: @certik

### DIFF
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -449,6 +449,151 @@ int get_errors (const std::string &infile,
     return 0;
 }
 
+bool input_in_scope_identifier(int input_location,
+                               std::pair<int, int> token_location) {
+  return token_location.first <= input_location &&
+         input_location <= token_location.second;
+}
+
+bool find_tok_in_scopes(const std::string &infile,
+                        const std::string &runtime_library_dir,
+                        CompilerOptions &compiler_options,
+                        const std::string tok_name) {
+  Allocator al(4 * 1024);
+  LFortran::diag::Diagnostics diagnostics;
+  LFortran::LocationManager lm;
+  lm.in_filename = infile;
+  std::string input = LFortran::read_file(infile);
+  lm.init_simple(input);
+  LFortran::Result<LFortran::LPython::AST::ast_t *> r1 =
+      LFortran::parse_python_file(al, runtime_library_dir, infile, diagnostics,
+                                  compiler_options.new_parser);
+  if (r1.ok) {
+    LFortran::LPython::AST::ast_t *ast = r1.result;
+    LFortran::Result<LFortran::ASR::TranslationUnit_t *> x =
+        LFortran::LPython::python_ast_to_asr(
+            al, *ast, diagnostics, true, compiler_options.disable_main,
+            compiler_options.symtab_only, infile);
+    if (!x.ok) {
+      std::cout << "{}\n";
+      return false;
+    }
+    std::vector<LFortran::LPython::document_symbols> symbol_lists;
+    LFortran::LPython::document_symbols loc;
+    for (auto &a : x.result->m_global_scope->get_scope()) {
+      std::string symbol_name = a.first;
+      uint32_t first_line;
+      uint32_t last_line;
+      uint32_t first_column;
+      uint32_t last_column;
+      lm.pos_to_linecol(a.second->base.loc.first, first_line, first_column);
+      lm.pos_to_linecol(a.second->base.loc.last, last_line, last_column);
+      loc.first_column = first_column;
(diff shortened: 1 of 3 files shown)


src/lpython/parser/tokenizer.re
```diff
@@ -696,5 +696,24 @@ std::string pickle_token(int token, const LFortran::YYSTYPE &yystype)
...
+    token_name += std::to_string(yystype.f);
+  } else if (token == yytokentype::TK_IMAG_NUM) {
+    token_name += std::to_string(yystype.f) + "j";
+  } else if (token == yytokentype::TK_STRING) {
+    token_name = "\"" + yystype.string.str() + "\"";
+  } else if (token == yytokentype::TK_TYPE_COMMENT) {
+    token_name = "\"" + yystype.string.str() + "\"";
+  }
+  return token_name;
+}
```
certik, line 715: Should we call this function in `pickle_token`, so that we do not repeat the code?

src/bin/lpython.cpp
```diff
@@ -449,6 +449,151 @@ int get_errors (const std::string &infile,
...
+    toks = res.result;
+  } else {
+    std::cout << "{}\n";
+    return {};
+  }
+
+  for (size_t i = 0; i < toks.size(); ++i) {
+    if (input_in_scope_identifier(
+            index,
+            std::make_pair<int, int>(locations[i].first, locations[i].last))) {
```
certik, line 585: I would just do:
```suggestion
    if (locations[i].first <= index && index <= locations[i].last) {
}
```
And remove `input_in_scope_identifier`.

src/bin/lpython.cpp
```diff
@@ -449,6 +449,151 @@ int get_errors (const std::string &infile,
...
+        LFortran::LPython::python_ast_to_asr(
+            al, *ast, diagnostics, true, compiler_options.disable_main,
+            compiler_options.symtab_only, infile);
+    if (!x.ok) {
+      std::cout << "{}\n";
+      return false;
+    }
+    std::vector<LFortran::LPython::document_symbols> symbol_lists;
+    LFortran::LPython::document_symbols loc;
+    for (auto &a : x.result->m_global_scope->get_scope()) {
```
certik, line 483: This will only be able to lookup symbols at global scope. Rather, we need to use `resolve_symbol` to lookup the symbol in the scoped symbol table. For that, we need to know which ASR node we are in.

To do that, we need to implement "index" -> ASR node calculation.

This will have to be one by the "overlapping intervals" problem, as we discussed some time ago.

src/bin/lpython.cpp
```diff
@@ -859,6 +1005,7 @@ int main(int argc, char *argv[])
         // LSP specific options
         app.add_flag("--show-errors", show_errors, "Show errors when LSP is running in the background");
         app.add_flag("--show-document-symbols", show_document_symbols, "Show symbols in lpython file");
+        app.add_option("--goto-def", goto_def_index, "Pass index for GoTo Definition");
```
certik, line 1008: VSCode calls it `Go to Definition`, so maybe we can call this option `--go-to-definition`?

certik: Also, it seems the input almost always is a line and column number? If so, that should be the input here as well. So the description would be "Return a definition for the symbol at the given line and column", and if this is too long, we can shorten it.

gnikit:
@ankitaS11 Are you launching an `lpython` process every time you issue a server request? If so, do you use any persistent caching or do you have to reparse the source files to AST and ASR with every request?

certik:
Currently we don't, but we will probably have to add it.

gnikit:
I would probably guess that you need to leave the process running, since more interactive requests like hover will potentially overwhelm the users computer by "spamming" hover requests.

czgdp1807:
Any news on this?

gnikit:
> Any news on this?

I would recommend you hold the merge on this (possibly turn it into a draft) until there is an agreement on the server design. If merged it would be harder to undo in the future.

The server should really be a running process with an open communications channel otherwise you are parsing the sources + serving LSP every time you make a request, you should only be doing the latter. For small projects that is probably fine but if this is to be a functional language server it needs to be able to store and query the ASR objects.

certik:
More discussion is needed on this, so I converted into a Draft.

I didn't have time to push this yet, but I think the design has to be to have a long running LPython/LFortran server process. We had this design, so it is in the history. I think we should revert to it and start improving it. I'll try to get to this soon.

Issue #983, opened 2022-08-17
Dict implementation design


certik:
Here are the freedoms that we have in implementation dictionaries:

* Hash function
* load factor / initial size of an array
* array resizing strategy (rehashing or not)
* key collision strategy
* memory allocation

Here is how we can do this:

* [ ] We start with some defaults (just one choice for all of the above)
* [ ] We implement more (eventually all) choices, and allow the user to select using a type annotation, such as `x: Dict[hash3, loadfactor(5), resizing("x"), collision_strategy("list")]`. 
* [ ] We can then benchmark various combinations and choose some good solid defaults
* [ ] We can later add a user defined hash function that gets compiled and inlined in LLVM, so optimal performance.
* [ ] Ideally the hash function should be defined in the frontend language too

czgdp1807:
Adding TODOs here after #975 is merged (in decreasing priority order),

- [ ] Create iterators using `builder->CreateAlloca` for `LLVMList` and `LLVMDict` outside user loops. In case we enter a loop in LLVM backend then let the `list_api`, `dict_api` so that they can set their iterators before hand.
- [ ] Benchmarks dict right after the above issue is resolved.
- [ ] Implement [separate chaining](https://en.wikipedia.org/wiki/Hash_table#Separate_chaining) key collision resolution using arrays of arrays (basically wrapped under lists so that we can re-use a lot of code already written for lists).
- [ ] Implement, delete a.k.a erase key feature in dict for all the collision resolution strategies.
- [ ] Try different hash functions for different key types.
- [ ] Implement lazy re-hashing. Very tricky to implement but still worth trying out.
- [ ] Compare lazy re-hashing and re-hashing all at once.

czgdp1807:
Question - Should we allow floats/reals as keys?

The reason why I am asking this question is because floats are always approximately equal. So, converting them to int of same size by memcpy won't work well. Plus, its not a good practice to use floats as keys. 123.000 and 123.000001 are approximately same, now user may expect them to work as a single key but due to slight difference both will give totally different hash value. So disallowing these as keys will prevent a lot of bad practices.

Also, in case of collisions we will go for exact equality checks in keys but as I said above floats are approximately equal mostly, so users may get very random results due to this property of floats.

Even if we want to allow floats as keys, we should ask the user to provide an equality function in the dict type which we should use to compare floats in case of collisions.

certik:
[...]

PR #986, opened 2022-08-18
Creating dict iterators outside the user loops

czgdp1807:

Benchmarks with this branch - https://gist.github.com/czgdp1807/e7f7b6ca52c57b16b27ec8d0259c6d4a

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1756,6 +1756,8 @@ public:
     }
 
     void visit_Program(const ASR::Program_t &x) {
+        bool is_dict_present_copy = dict_api->is_dict_present;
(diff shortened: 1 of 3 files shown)


certik (approved):
Looks good. Are the `dict_api->set_iterators();` a no-op if no dictionaries are used? Or is the some performance overhead for each while loop, whether or not it uses a dictionary?

czgdp1807:
> Are the `dict_api->set_iterators();` a no-op if no dictionaries are used?

Yes. I am yet to optimize it slightly like setting `dict_api->is_dict_present=false` at the entry point of `visit_Program`, `visit_Function` so that we don't create iterators if no dict is used in a given scope.

czgdp1807:
@certik This is ready. Is there anything to be changed here?

certik (approved):
No, go ahead and merge.

PR #957, opened 2022-08-12
Adding type parameters to functions without return values

ansharlubis:

PR #989, opened 2022-08-19
Adding type parameters to functions without return values (cleaned history)

ansharlubis:
Cleaned up PR #957


certik (approved):
Very cool, thanks for the PR, I reviewed all commits, I think this looks good.

Issue #967, opened 2022-08-15
In the dictionaries, values without keys is not recognized


Thirumalai-Shaktivel:
Error:
```py
syntax error: Token '**' is unexpected here
 --> examples/expr2.py:1:2
  |
1 | {**self._derived_units, **derived_units}
  |  ^^ 
```

PR #998, opened 2022-08-20
Recognize dictionary unpacking operation

Thirumalai-Shaktivel:
Closes: #967 
ping @akshanshbhatt


tests/parser/dictionary1.py
```diff
@@ -24,4 +24,6 @@ def test():
           """
         }
 
+    {**self._derived_units, **derived_units}
```
Thirumalai-Shaktivel, line 27: I have a doubt @certik,
For the above example, we store `nullptr` for keys
```py
└-Module
  |-body=↧
  | └-Expr
  |   └-value=Dict
  |     |-keys=↧
  |     └-values=↧
  |       |-Attribute
  |       | |-value=Name
  |       | | |-id=self
  |       | | └-expr_contextType=Load
  |       | |-attr=_derived_units
  |       | └-expr_contextType=Load
  |       └-Name
  |         |-id=derived_units
  |         └-expr_contextType=Load
  └-type_ignores=↧
```
But CPython stores `None`
```py
Module(
   body=[
      Expr(
         value=Dict(
            keys=[
               None,
               None],
            values=[
               Attribute(
                  value=Name(id='self', ctx=Load()),
                  attr='_derived_units',
                  ctx=Load()),
               Name(id='derived_units', ctx=Load())]))],
   type_ignores=[])
```
Should we store None as well? `or` the above LPython Output looks good?

certik: How do we handle None in other contexts?

Thirumalai-Shaktivel: We only store Constant None in the AST
Example:
```py
x = {None: "none"}
```
CPython
```py
Module(
   body=[
      Assign(
         targets=[
            Name(id='x', ctx=Store())],
         value=Dict(
            keys=[
               Constant(value=None)],
            values=[
               Constant(value='none')]))],
   type_ignores=[])
```

certik: I see. It seems to me that `keys` and `values` should always have the same length, like in CPython, so storing ConstantNone seems like a better solution for LPython? Let me know what you think.

Thirumalai-Shaktivel: Now, I get some clarity!
So, here we cannnot have `nullptr` when the key is not available.
Consider an example:
```py
{
    **a,
    x: "Text",
    **y,
    z: "Text"
}
```
```py
└-Module
  |-body=↧
  | └-Expr
  |   └-value=Dict
  |     |-keys=↧
  |     | |-Name
  |     | | |-id=x
  |     | | └-expr_contextType=Load
  |     | └-Name
  |     |   |-id=z
  |     |   └-expr_contextType=Load
  |     └-values=↧
  |       |-Name
  |       | |-id=a
  |       | └-expr_contextType=Load
  |       |-ConstantStr
  |       | |-value="Text"
  |       | └-kind=()
  |       |-Name
  |       | |-id=y
  |       | └-expr_contextType=Load
  |       └-ConstantStr
  |         |-value="Text"
  |         └-kind=()
  └-type_ignores=↧
```
which is wrong, as the index gets wrongly matched.
Now, what should we do with this?
Add `ConstantNone`? I think that would cause the conflict!
Consider another example:
```py
{
    **a,
    None: "Text",
    **y,
    z: "Text"
}
```
```py
Module(
   body=[
      Expr(
         value=Dict(
            keys=[
               None,
               Constant(value=None),
               None,
               Name(id='z', ctx=Load())],
            values=[
               Name(id='a', ctx=Load()),
               Constant(value='Text'),
               Name(id='y', ctx=Load()),
               Constant(value='Text')]))],
   type_ignores=[])
```
This is the simple example in which the key is not available
```py
a = {
    "a": "Text 1"
}
b = {
    "b" : "Text 2"
}

x = {
    **a,
    None: "Text 3",
    **b,
    "c": "Text 4"
}

print(x["a"])
```
I think we can add a new node for this?

certik: I would just do whatever CPython does. Do we have means to store None in the dict? If so, then I would store it just like CPython does, because then we don't have to worry about the design. We just do whatever they do, and we can check it against them.

If there is something that our AST cannot fundamentally represent, then let's discuss that and find a solution. Otherwise I would just follow CPython's lead on this.

Thirumalai-Shaktivel: Currently, we only have `ConstatntNone`. So, I'm planning to create a new node `None` and pass this for the dict.

Thirumalai-Shaktivel:
Something like this @certik, Do you have any other suggestions?

Thirumalai-Shaktivel:
Example:
```py
x = {
    **a,
    "c": "Text 4"
}
```
LPython AST:
```py
└-Module
  |-body=↧
  | └-Assign
  |   |-targets=↧
  |   | └-Name
  |   |   |-id=x
  |   |   └-expr_contextType=Store
  |   |-value=Dict
  |   | |-keys=↧
  |   | | |-NONE
  |   | | └-ConstantStr
  |   | |   |-value="c"
  |   | |   └-kind=()
  |   | └-values=↧
  |   |   |-Name
  |   |   | |-id=a
  |   |   | └-expr_contextType=Load
  |   |   └-ConstantStr
  |   |     |-value="Text 4"
  |   |     └-kind=()
  |   └-type_comment=()
  └-type_ignores=↧
```
CPython AST:
```py
Module(
   body=[
      Assign(
         targets=[
            Name(id='x', ctx=Store())],
         value=Dict(
            keys=[
               None,
               Constant(value='c')],
            values=[
               Name(id='a', ctx=Load()),
               Constant(value='Text 4')]))],
   type_ignores=[])
```

grammar/Python.asdl
```diff
@@ -86,6 +86,7 @@ module LPython
          | ConstantEllipsis(string? kind)
          | ConstantNone(string? kind)
          | ConstantBytes(string value, string? kind)
+         | NONE()
```
certik, line 89: Why cannot we use ConstantNone?

Thirumalai-Shaktivel: Consider this example:
```py
x = {
    **a,
    None: "Text 3",
}
```
When we use the Constant we get:
```py
└-Module
  |-body=↧
  | └-Assign
  |   |-targets=↧
  |   | └-Name
  |   |   |-id=x
  |   |   └-expr_contextType=Store
  |   |-value=Dict
  |   | |-keys=↧
  |   | | |-ConstantNone
  |   | | | └-kind=()
  |   | | └-ConstantNone
  |   | |   └-kind=()
  |   | └-values=↧
  |   |   |-Name
  |   |   | |-id=a
  |   |   | └-expr_contextType=Load
  |   |   └-ConstantStr
  |   |     |-value="Text 3"
  |   |     └-kind=()
  |   └-type_comment=()
  └-type_ignores=↧
```
is this fine?

certik:
How does CPython represent None? I don't see any None here: https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Parser/Python.asdl

akshanshbhatt:
I just want to confirm that `Constant(value=None)` in CPython AST is equivalent to `ConstantNone` in LPython, right?

Thirumalai-Shaktivel:
Yes, it is.

akshanshbhatt:
> Yes, it is.

Then I guess this is what is causing the issue here. Till now, we didn't encounter any case where `None` was present isolated (i.e. `None` and not `Constant(value=None)`). Here, we must find a way to pass `None` independently.

Thirumalai-Shaktivel:
https://github.com/python/cpython/blob/755be9b1505af591b9f2ee424a6525b6c2b65ce9/Grammar/python.gram#L898
Here they pass `NULL` when they encounter `'**' expr` in the parser and
https://github.com/python/cpython/blob/755be9b1505af591b9f2ee424a6525b6c2b65ce9/Parser/action_helpers.c#L389-L398
they store `NULL` in the key.
I guess in the `ast` module, if the key is `NULL` they print `None`.

certik:
So CPython stores `nullptr` in the AST? We can also do that, we do that by using "optional" in AST.asdl with the `?`.

However, in the Dict's case: `Dict(expr* keys, expr* values)`, we would have to modify it to something like:
```
Dict(dict_key* keys, expr* values)
...
dict_key = (expr? key)
```
That's one way to do it. I think that would work, would not it?

The other way is to use `ConstantNone` as the "key". Do you see any technical issue with this?

Any other options on the table?

Thirumalai-Shaktivel:
Perfect, I had the same thought (1st approach)!
But, I think for now,  let's use `ConstantNone`. 
If we face any issue in the future, later we can modify for our needs.

Thirumalai-Shaktivel:
This PR is ready, @certik! Any other changes to be made?

certik (approved):
Looks good now!

Thirumalai-Shaktivel:
Thanks for the review!

Issue #992, opened 2022-08-19
Compiling Python packages with LPython


czgdp1807:
Following is the plan to achieve the goal mentioned in issue title,

1. Write down small type annotated test packages in Python syntax. The nesting can vary from just 1 folder/package having an `__init__.py` file to 3-4 folders/subpackages one inside the other each level having an `__init__.py`. We can write packages for different sorting algorithms, graph algorithms and some practical stuff like that. We should try not to use workarounds but write code in a way that feels natural to us.
2. Try to compile those test packages starting from easy ones (with just 1 folder) to difficult ones. Do sprints to quickly make them work then send PRs each having a single fix.
3. Then proceed with small packages from PyPI (there must be many). Try to compile and fix LPython as described in step 2. 
4. Once all of the above is done, move on to advanced Python packages.

Alongside we should also try to interface with CPython code (#703).

@certik What do you say?

Shaikh-Ubaid:
Interested!

Shaikh-Ubaid:
Interested

certik:
Yes, we should get LPython working with actual Python packages, that we write ourselves to ensure we stick to the subset that LPython supports. And the goal here is to fix up the inevitable bugs that we either know about or that we will hit. After most things work, it will allow people to upload packages to PyPi and to have dependencies, etc. We can use regular Python tooling to install dependencies and download/upload from PyPi. We have to improve LPython to be able to compile such a package with dependencies.

We probably also need to add a mode to compile the library and all dependencies to some kind of a "mod" file that stores the ASR in it, for quick compilation, so that when you modify a script, only that script has to be parsed and compiled to ASR, all other modules can be just deserialized. LFortran works this way, and it works great.

czgdp1807:
I think we can re-use the serialization-deserialisation mechanism of LFortran and use it here to compile Python modules till ASR level. In fact, modules written by us right now compile (parsed + AST->ASR) only when we import them. However LFortran pre-compiles them till ASR and loads them if we import in our program. LPython should behave in a similar manner I think. Also, the question is how to integrate LPython with a build system (writing build files in simple words) so that LPython compiles all the intrinsic modules defined in it when its being built from source.

certik:
Yes. I think CPython compiles every module into `.pyc`, so LPython would compile every module to `.lpy` (or even to `.pyc` to reuse the extension), and the compiled module would be just serialized ASR. Just like LFortran works with the `.mod` files.

[...]

PR #999, opened 2022-08-20
Saving LPython's intrinsic modules as ``pyc`` files

czgdp1807:
https://github.com/lcompilers/lpython/issues/992#issuecomment-1221244033

For now I just save the ASR of a Python file compiled with `-c` option in `.pyc` files. Next steps,

1. Add the possibility first try loading the module from `.pyc` file generated earlier. If not found then compile and re-save in `.pyc` file.
2. Update CMake build files to compile intrinsic modules and save their ASR during LPython's build time. 
3. Test the update end to end.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,9 @@ set(WITH_LFORTRAN_BINARY_MODFILES YES
 set(WITH_RUNTIME_LIBRARY YES
     CACHE BOOL "Compile and install the runtime library")
 
+set(WITH_INTRINSIC_MODULES no
+    CACHE BOOL "Compile intrinsic modules to .pyc (ASR) at build time")
+
 # Find ZLIB with our custom finder before including LLVM since the finder for LLVM
 # might search for ZLIB again and find the shared libraries instead of the static ones
 find_package(StaticZLIB REQUIRED)
--- a/src/libasr/modfile.cpp
+++ b/src/libasr/modfile.cpp
@@ -57,6 +57,34 @@ std::string save_modfile(const ASR::TranslationUnit_t &m) {
     return b.get_str();
 }
 
+std::string save_pycfile(const ASR::TranslationUnit_t &m) {
+#ifdef WITH_LFORTRAN_BINARY_MODFILES
+    BinaryWriter b;
+#else
+    TextWriter b;
+#endif
+    // Header
+    b.write_string(lfortran_modfile_type_string);
+    b.write_string(LFORTRAN_VERSION);
+
+    // AST section: Original module source code:
+    // Currently empty.
+    // Note: in the future we can save here:
+    // * A path to the original source code
+    // * Hash of the orig source code
+    // * AST binary export of it (this AST only changes if the hash changes)
+
+    // ASR section:
+
+    // Export ASR:
+    // Currently empty.
+
+    // Full ASR:
+    b.write_string(serialize(m));
+
+    return b.get_str();
+}
+
 ASR::TranslationUnit_t* load_modfile(Allocator &al, const std::string &s,
         bool load_symtab_id, SymbolTable &symtab) {
 #ifdef WITH_LFORTRAN_BINARY_MODFILES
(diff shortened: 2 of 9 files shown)


czgdp1807:
@certik Is this PR a good approach for implementing https://github.com/lcompilers/lpython/issues/992#issuecomment-1221244033?

src/libasr/modfile.cpp
```diff
@@ -57,6 +57,34 @@ std::string save_modfile(const ASR::TranslationUnit_t &m) {
...
+    // ASR section:
+
+    // Export ASR:
+    // Currently empty.
+
+    // Full ASR:
+    b.write_string(serialize(m));
+
+    return b.get_str();
+}
```
certik, line 86: I would merge this with LFortran --- I think all LCompilers can then reuse it, just change the extension. I think there is nothing special about LPython's `.pyc` compared to LFortran's `.mod`, I think they can be exactly identical. It's just ASR that is saved, it's independent of the frontend.

czgdp1807: There is one thing different though. Fortran files have specific module symbols. Only those are saved in the `.mod` files. However in the case Python, whole file can be a module. We just have to import something from it and it becomes a module. But if you call it using `python` command then it acts as an "executable". So what I have done is if any file is compiled with `-c` option of LPython then its ASR (i.e., the full `ASR::TranslationUnit_t`) gets saved in a `.pyc` file. Does that make sense?

certik: We can discuss it more. What I had in mind is that a single Python file is exactly 100% equivalent to a single Fortran module file, at the ASR level.

certik: (Well, there is a difference that Python modules can be nested, but this is something that people have requested for Fortran also to do, so we should allow that at the ASR level in some clean way.)

czgdp1807: Cool. Let me try out implementing the rest of the steps I have written in https://github.com/lcompilers/lpython/pull/999#issue-1345166526. Let's see if it works till the end.

czgdp1807:
@certik Could you please review this? So far I have added support for saving LPython's generated ASR into `.pyc` files when `-c` and `--disable-options` are enabled. If these changes look good to you then I will proceed forward with compiling intrinsic modules at build time of LPython.

src/libasr/modfile.cpp
```diff
@@ -79,4 +107,27 @@ ASR::TranslationUnit_t* load_modfile(Allocator &al, const std::string &s,
...
+    std::string version = b.read_string();
+    if (version != LFORTRAN_VERSION) {
+        throw LCompilersException("Incompatible format: LFortran Modfile was generated using version '" + version + "', but current LFortran version is '" + LFORTRAN_VERSION + "'");
+    }
+    std::string asr_binary = b.read_string();
+    ASR::asr_t *asr = deserialize_asr(al, asr_binary, load_symtab_id);
+
+    ASR::TranslationUnit_t *tu = ASR::down_cast2<ASR::TranslationUnit_t>(asr);
+    return tu;
+}
```
certik, line 131: Why not merging this with LFortran's save/load? It seems it is exactly the same code.

czgdp1807: Yeah. Well I would prefer to do in a separate PR. Reason - Don' want to break something in LFortran which already works. Also, for now this is exactly similar but we don't know if it will diverge in future. Other than similarity is there any other thing that is of concern?

certik (approved):
I think it's fine.

My only concern is with the duplication, since we rather want to extend / amend ASR and whatever is needed, and then exactly reuse the saving / loading functionality. It will benefit both LPython and LFortran. In fact, I think we could even mix and match ASR modules, that is, take an LPython `.pyc` module and "use" it from LFortran and vice versa. It's a little hard to wrap the head around it, but in `libasr` it is "just" ASR, it is *not* Python and it is *not* Fortran; or equivalently, it is both Python and Fortran.  So rather than saving Python/Fortran, I want to think about it as saving ASR as a module. That's it. So it should be exactly the same code, and if it is not, that would be a concern.

czgdp1807:
Yes. I agree. I will try to combine the functions (or as much as is common between them) here. Rest I will try to leave the scope of handling differences if any.

certik:
Regarding the behavior --- should LPython always save `.pyc` files (as CPython I think does)? Or should we make it configurable? Since LFortran always saves `.mod` files and *if* CPython also always saves `.pyc` files then perhaps we should also always save it. Does CPython decide based on the timestamp of the `.py` file if the `.pyc` file needs to be recompiled? LFortran never checks any timestamps, it just expects a `.mod` file and if it is there it will use it. It is the job of the build system to ensure the `.mod` file is up-to-date. CPython, on the other hand, does not require a build system. 

It seems we almost need two modes:
* The CPython like mode which does not require a build system and two submodes:
  * Use `.pyc` and check timestamps
  * No `.pyc` and it will always recompile everything (that is the current mode in master) 
* A build system mode, where we do not check timestamps and just require `.pyc` to be present if we import a module (we do not care about the `.py` files at all of imported modules) --- that is effectively the LFortran mode.

czgdp1807:
By timestamp you mean two things right, the time at which `pyc` file was generated and the time at which the `py` was last modified. If the second happened after the first then recompile, right? Makes. Sense. Well if LPython is called by the build system then LPython's timestamp verification would be redundant. So, we need a way to distinguish between whether LPython was called by the user or the build system.

certik:
Yes regarding timestamp.

Yes, we need to distinguish the two (or three) modes above.

For example `lpython a.py` is the CPython mode (either `.pyc` or not). And `lpython -c a.py -o a.o` is the build system mode.

certik:
Regarding mix and matching LPython / LFortran, we will (later) version `libasr` itself, and just use the `libasr` version in the `.pyc` / `.mod` file. If the version exactly matches, then we can mix and match.

CMakeLists.txt
```diff
@@ -80,6 +80,9 @@ set(WITH_LFORTRAN_BINARY_MODFILES YES
 set(WITH_RUNTIME_LIBRARY YES
     CACHE BOOL "Compile and install the runtime library")
 
+set(WITH_INTRINSIC_MODULES no
+    CACHE BOOL "Compile intrinsic modules till ASR level at build time")
```
certik, line 84: I would maybe write it as:
```suggestion
    CACHE BOOL "Compile intrinsic modules to .pyc (ASR) at build time")
```
But that can be done later.

certik (approved):
I think that's fine. Further improvements can be done once this is merged.

PR #1002, opened 2022-08-20
Sinh & cosh numpy_intrinsic functions implemented

Madhav2310:


Madhav2310:
@czgdp1807 could you please help me with this error?

integration_tests/elemental_05.py
```diff
@@ -0,0 +1,143 @@
...
+    k: i32
+    l: i32
+    eps: f64
+    eps = 1e-12
+
+    for i in range(size1):
+        for j in range(size2):
+            for k in range(size3):
+                for l in range(size4):
+                    assert abs( ((sinh(array[i, j, k, l])%3) +2)/2 - result[i, j, k, l]) <= eps
```
czgdp1807, line 24: @certik Builtin `%` in LPython doesn't work on arrays by default because it is implemented as a function and not a `binop`. So the question is which approach to follow,
1. To make `_mod` function defined in `lpython_builtin.py` vectorised;
2. Define `Mod` as `binop` and then use `builder->CreateSRem` in LLVM backend. Note that not all backends might support modulus operation on types other than integers.
3. Define [`mod`](https://numpy.org/doc/stable/reference/generated/numpy.mod.html) and then make it vectorisable.

I think 3rd approach is a good one and complies with Python and NumPy standards. What do you say @certik?

czgdp1807 (approved):
@certik This is ready to merge.

certik (approved):
I think this is good, thanks!

Issue #984, opened 2022-08-17
Draft: Restrictions for generics


ansharlubis:
Here's an idea of how to have user-defined restrictions.
```` python
from ltypes import TypeVar
from ltypes import f64, i32

T = TypeVar('T')

# Interfaces for checking the type
def plus(x: T, y: T) -> T:
    pass

def zero(x: T) -> T:
    pass

def div(x: T, y: i32) -> f64:
    pass

# To allow str to have a zero value
def empty_string(x: str) -> str:
    return ""

# The generic function
def mean(x: list[T]) -> f64:
    k: i32 = len(x)
    if k == 0:
        return zero(x[0])
    sum: T
    sum = zero(x[0])
    i: i32
    for i in range(k):
        sum = plus(sum, x[i])
    return div(sum, k)

# print(mean<plus->int.__add__>([1,2,3]))
print(mean([1,2,3]))
# print(mean<plus->float.__add__>([1,2,3]))
print(mean([1.0,2.0,3.0]))
# print(mean<plus->str.__add__,zero->empty_string>([1,2,3]))
print(mean(["a","b"]))
````

ansharlubis:
Before going to the above example, let's consider a simpler example:
````
from ltypes import TypeVar

T = TypeVar('T')

@restriction(T)
def plus(x: T, y: T) -> T:
    pass

def f(x: T, y: T) -> T:
    return plus(x,y)

print(f(1,2, plus=int.__add__))
print(f("a","b", plus=str.__add__))
print(f("c","d", plus=std.__add__))
````
The syntax `f<plus->int.__add__>(1,2)` isn't supported by Python, so I figure it can be treated as an extra argument in `f(1,2, plus=int.__add__)` because if I'm not mistaken Python ignores the extra arguments.

PR #1007, opened 2022-08-22
Adding hard-coded restrictions to generics (clean history)

ansharlubis:


src/libasr/pass/instantiate_template.cpp
```diff
@@ -191,28 +227,26 @@ class FunctionInstantiator : public ASR::BaseExprStmtDuplicator<FunctionInstanti
                     ASR::Real_t* tnew = ASR::down_cast<ASR::Real_t>(t);
                     return ASRUtils::TYPE(ASR::make_Real_t(al, t->base.loc,
                             tnew->m_kind, param->m_dims, param->n_dims));
-                }  
+                }
```
certik, line 230: You should configure your editor to remove white space, then there is no issue. I am going to merge it, and you can fix it later.

certik (approved):
This looks good. I am going to merge it, and we can iterate upon it with subsequent PRs.

Thank you so much @ansharlubis for getting this done. This is very very good foundation.

certik:
The next step is to implement something like this:

* https://github.com/lcompilers/lpython/issues/984

and also do all the restrictions checking in `libasr`, that way we can share it with LFortran.

ansharlubis:
Got it!

Issue #892, opened 2022-08-04
Redeclaration within different scopes


Abdelrahman-Kh-Fouad:
```bash
$ cat a.py
from math import log2, floor
def ff():
    i: i32
    for i in range(1, 1000):
        x: i32 
        x = i
        y: f32
        y = float(x)
        assert x.bit_length() == floor(log2(y)) + 1

    for i in range((1 << 30), (1 << 30) + 100 , 2):
        x: i32
        x = i
        y: f32
        y = float(x)
        assert x.bit_length() == floor(log2(y)) + 1
ff()
```
result:
```bash
$ ./src/bin/lpython -o out a.py
semantic error: Symbol is already declared in the same scope
 --> a.py:5:9
  |
5 |         x: i32 
  |         ~~~~~~ original declaration
  |
12 |         x: i32
  |         ^^^^^^ redeclaration


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).
```

certik:
This is an interesting one. In ASR we currently do not create a new scope in a for loop. There are two ways out:

* Either always create a new scope in a for loop
* Or the LPython frontend has to insert a "Block" ASR node (which has a scope) if there are any local variables in the loop, and declare them there

Probably the second option seems the way to start.

Smit-create:
CPython does something like the following, it allows to declare variables inside for loop but they are also accessible outside the loop:
```py
def f():
    i: int
    for i in range(10):
        x: int
        x = i
    print(x)
    for i in range(15):
        x: int
        x = i
    print(x)
f()
```

The output is:
```
9
14
```

czgdp1807:
To enable [the above](https://github.com/lcompilers/lpython/issues/892#issuecomment-1222330814), one might need to move the variables declared inside the loop at a line just before the starting of the loop. Seems like CPython has a runtime scope and loops don't have any scope of their own. So, CPython adds `x` into the current runtime scope and when the loop ends `x` is still present inside the scope and it is accessible.

PR #1008, opened 2022-08-22
Use block call to allow variables declaration in for loop

Smit-create:
#892


Smit-create:
The PR is ready for review. I have fixed most of the errors with only one remaining in: `integration_tests/generics_list_01.py`. I am not much familiar with the recent updates on generics.

czgdp1807 (changes requested):
Some changes are needed. I will shepherd this PR later this week.

czgdp1807:
@certik This is ready for review. Let me know if there is anything to be updated here. Otherwise I will clean up the git history.

certik (approved):
I think this is fine. Why did the ASR tests change?

czgdp1807:
> Why did the ASR tests change?

Its because to store variables declared inside loop body, we need to create a new scope. So, the change is the Symtab ID of some symbol tables increased by some amount (1 or 2 mostly). Nothing to worry about though as Symtab IDs for internal usage of the compiler only and doesn't affect code generation in the backends or the passes.

certik:
I see, that's fine.

czgdp1807 (approved):

PR #1011, opened 2022-08-22
Implementing ``dict.pop`` for LLVM backend

czgdp1807:

Benchmarks - https://gist.github.com/czgdp1807/e7f7b6ca52c57b16b27ec8d0259c6d4a#file-dict_bench_3-md

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1294,6 +1294,23 @@ public:
                                   LLVM::is_llvm_struct(dict_type->m_value_type));
     }
 
+    void visit_DictPop(const ASR::DictPop_t& x) {
+        ASR::Dict_t* dict_type = ASR::down_cast<ASR::Dict_t>(
+                                    ASRUtils::expr_type(x.m_a));
+        uint64_t ptr_loads_copy = ptr_loads;
+        ptr_loads = 0;
+        this->visit_expr(*x.m_a);
+        llvm::Value* pdict = tmp;
+
+        ptr_loads = !LLVM::is_llvm_struct(dict_type->m_key_type);
+        this->visit_expr_wrapper(x.m_key, true);
+        ptr_loads = ptr_loads_copy;
+        llvm::Value *key = tmp;
+
+        tmp = dict_api->pop_item(pdict, key, *module, dict_type,
+                                 LLVM::is_llvm_struct(dict_type->m_value_type));
+    }
+
     void visit_ListLen(const ASR::ListLen_t& x) {
         if (x.m_value) {
             this->visit_expr(*x.m_value);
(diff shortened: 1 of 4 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1294,6 +1294,23 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                                   LLVM::is_llvm_struct(dict_type->m_value_type));
     }
 
+    void visit_DictPop(const ASR::DictPop_t& x) {
+        ASR::Dict_t* dict_type = ASR::down_cast<ASR::Dict_t>(
+                                    ASRUtils::expr_type(x.m_a));
+        uint64_t ptr_loads_copy = ptr_loads;
```
Thirumalai-Shaktivel, line 1300: I have a doubt, what is `ptr_loads`? (Like what is it used for?)

czgdp1807: Posting the discussion I had about this ~10ish days ago,

1. Var can store both pointer and non-pointers.
2. And pointer can be used as a non-pointer variable (in Fortran, `libasr` is shared).
3. So, sometimes its not clear how many times you have load to the llvm::Value associated with a Var.
4. Like say,
```fortran
integer, pointer :: pi
integer :: i
integer :: a(4)
pi => i
a(pi) = 0
```
5. Now the ASR will have `Assignment(ArrayItem(Var(a), Var(pi)), 0)`.
6. Now in LLVM a is a `struct*` (pointer to struct), `pi` is `i32**`.
7. Well when you visit `Var(a)`, you can't load `llvm::Value* tmp`. Because you need a pointer.
8. But when you visit `Var(pi)` you need to load twice becuase you need `i32` and not `i32*` or `i32**`.
9. For now I am using `ptr_loads` which specifies how many times to load a pointer when you `visit_Var`.
10. Keep using `ptr_loads` as the general solution for now as it works and is simple. Easy to override and restore between multiple calls.

Thirumalai-Shaktivel: I got it, Thank you

czgdp1807: This is one of the ambiguities which should be handled in ASR ideally but doing too much abstraction in ASR right now is not a good idea. Instead we should learn, experiment and then after some time when backends stabilise and we implement all the crucial features, we can make ASR handle these kinds of ambiguities.

czgdp1807:
@certik This is ready for review. Benchmarks in https://gist.github.com/czgdp1807/e7f7b6ca52c57b16b27ec8d0259c6d4a#file-dict_bench_3-md.

certik (approved):
I think that's fine.

src/libasr/codegen/asr_to_llvm.cpp
certik: Yes, these are things where we can probably improve the ASR, but we need more experience.

Writing a C/C++ frontend might also help with this regard. For now we are focusing more on getting it done, we just have to be a little careful about these things.

PR #1037, opened 2022-08-24
Add support for strings as keys in ``dict``

czgdp1807:

Benchmarks -

* String with keys - https://gist.github.com/czgdp1807/e7f7b6ca52c57b16b27ec8d0259c6d4a#file-dict_bench_str-md
* Same benchmark but 64 bit integers as keys - https://gist.github.com/czgdp1807/e7f7b6ca52c57b16b27ec8d0259c6d4a#file-dict_bench_int64-md

So, why LPython lags behind clang, gcc with strings as keys but beats every other compiler comfortably when the same keys are presented as integers? I have listed the reasons below which I could figure out after analysing the situation,

1. **No benefit of Spatial Locality with Strings** - So, the way `str` works in LPython is that first the call is redirected to C runtime library and then `malloc` is called there. This returns the pointer to the requested memory available in the heap. We use this pointer as strings in LLVM backend (by storing it in a pointer of type `i8*`). So it can be seen that each string can be anywhere in the heap memory. We store the pointers in an array so these pointers are spatially local but not the contents of these pointers (which are the actual strings). This leads to lower cache efficiency each time we access a key-value pair from the dict.
2. **Too many malloc calls** - As I said in the above point, we make a malloc call each time we create a string. So for every insertion, reading operation we call malloc to obtain the string. For example if I remove all the dict operations from my benchmark and only keep two loops each doing `key = str(c + i)`, the resulting program takes ~1.3s to execute. Its a significant percentage of the total time with `dict08` branch.
3. **Resizing technique** - Since each time we resize/rehash the dict, we need to insert elements into the new hash table. I noticed that when strings as keys are used, tripling (instead of doubling) gives way better results (see my benchmarks above with `dict08` branch).

What are the **next steps** to resolve this problem,

1. **Implement separate chaining for strings** - We use linear probing right now and it beats every other compiler (which use separate chaining) very comfortably. The main advantage of this approach is spatial locality (technically higher cache efficiency). However, for strings as I noted above, spatial locality cannot be leveraged because each string gets its memory from different `malloc` calls. Using separate chaining instead will help in resizing on the basis of average number of key-value pairs in all the buckets of the dict. So, we will avoid resizing as long as possible for strings.
2. **Make possible to choose resizing factor via dict syntax** - Since tripling the dict for resizing produced much better results in both macOS and Linux, we should note that this should be exposed and the user should be able to decide whether to double, triple or quadruple or whatever the dict when resizing happens.


certik:
Perfect. Yes, we need to fix it. I think we should implement deallocation for strings, and then we need a custom allocator for strings

czgdp1807:
Yes. Anyways, in future we need to expose all the collision resolution strategies we have to the user via syntax. Set the defaults but allow for modifications via syntax. That we won’t be rigid as opposed to unordered_map and dict.

czgdp1807:
Anyways, should I merge this. This is ready. Next step would be to implement separate chaining and benchmark again.

certik (approved):
I think that this is fine.

PR #1038, opened 2022-08-24
Bug fix and register ``test_dict_02``

czgdp1807:


czgdp1807:
@certik This is a fix that creeped in due to `test_dict_02` not being registered in `CMakeLists.txt` and `current_scope->get_scope().find(...)` usage instead of `current_scope->resolve_symbol` in `visit_Call`. I am fixing this here and merging it in `main` once the tests pass. Any other issue (there shouldn't be any) with `dict`s will be handled in https://github.com/lcompilers/lpython/pull/1037.

Issue #865, opened 2022-08-01
Add "inline" feature for functions


certik:
When a function is annotated with `inline`, it will get inlined.

For gcc/clang, the `inline` keyword is [effectively ignored](https://stackoverflow.com/questions/3212571/usefulness-of-the-inline-feature), but you can do `__attribute__(always_inline)` which should force it to inline it. I think our "inline" feature should "force" it as well. It can possibly only do it when optimizations are enabled, so in AST->ASR, it would still represent it as a function, but it would set an "inline" attribute.

We have to choose Python syntax how to annotate a function to be inlined, such as an `@inline` decorator.

rebcabin:
in some of our ARC code, the attribute is denoted `__atribute__((always_inline))`, with double round brackets. I do not know whether the doubling of the round brackets is required or optional.

certik:
Here is relevant documentation how C and other languages handle this:

* https://stackoverflow.com/questions/22767523/what-inline-attribute-always-inline-means-in-the-function
* https://gcc.gnu.org/onlinedocs/gcc/Inline.html
* https://en.cppreference.com/w/cpp/language/inline
* https://awesomekling.github.io/Smarter-C++-inlining-with-attribute-flatten/
* https://nnethercote.github.io/perf-book/inlining.html
* https://swiftrocks.com/the-forbidden-inline-attribute-in-swift

Smit-create:
Closing this via #1041

PR #1041, opened 2022-08-25
Add inline decorator for functions

Smit-create:
#865


certik (approved):
Awesome, thanks for this! I am glad we only have to do this once now for Function, no Subroutine. :)

czgdp1807:
I will implement this in the backend. Thanks for adding to ASR.

PR #1048, opened 2022-08-26
User-defined restrictions

ansharlubis:
This follows the idea in #984


certik (approved):
I think this is good to go in.

In subsequent PRs we should add more tests for error messages.

certik:
In subsequent PRs we can also then provide some explicit functions (add/zero/etc) for integers, and use the general mechanism even for integers, and then remove the special casing for integers from the instantiation code.

certik:
Closed in favor of #1125.

PR #1055, opened 2022-08-28
WASM: Define EMSCRIPTEN_KEEPALIVE functions to be used by Web-Frontend

Shaikh-Ubaid:
This `PR` adds the `EMSCRIPTEN_KEEPALIVE` functions to be used by Web-Frontend. There is a Web Frontend deployed at https://www.ubaidshaikh.me/lcompilers_web_frontend/. The default example currently used is `examples/expr2.f90`.

```python
def main0():
    x: i32
    x = (2+3)*5
    print(x)

main0()
```

### DIFF
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -33,6 +33,10 @@
 #include <cpp-terminal/terminal.h>
 #include <cpp-terminal/prompt0.h>
 
+#ifdef HAVE_BUILD_TO_WASM
+    #include <emscripten/emscripten.h>
+#endif
+
 #ifdef HAVE_LFORTRAN_RAPIDJSON
     #include <rapidjson/document.h>
     #include <rapidjson/stringbuffer.h>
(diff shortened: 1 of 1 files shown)


Shaikh-Ubaid:
Also, another example that I tested on the Web-Frontend:
```python
def main0():
    a: i32
    b: i32
    c: i32
    i: i32
    no_of_fib_terms: i32
    no_of_fib_terms = 11
    
    a = 0
    b = 1
    
    for i in range(no_of_fib_terms):
        print(i, "th fibonacci number =", a)
        c = a + b
        a = b
        b = c

main0()
```

Shaikh-Ubaid:
`numpy` arrays also seem to work. It seems they are currently following the `1-based` indexing. I think/guess, we need to fix it.

Shaikh-Ubaid:
The repository for the web-frontend https://www.ubaidshaikh.me/lcompilers_web_frontend/ is currently at https://github.com/Shaikh-Ubaid/lcompilers_web_frontend.

src/bin/lpython.cpp
```diff
@@ -804,6 +808,158 @@ int link_executable(const std::vector<std::string> &infiles,
...
+    INITIALIZE_VARS;
+    lm.init_simple(input);
+    LFortran::Result<LFortran::LPython::AST::Module_t*> ast = LFortran::parse(al, input, diagnostics);
+    out = diagnostics.render(input, lm, compiler_options);
+    if (ast.ok) {
+        auto casted_ast = (LFortran::LPython::AST::ast_t*)ast.result;
+        LFortran::Result<LFortran::ASR::TranslationUnit_t*>
+        asr = LFortran::LPython::python_ast_to_asr(al, *casted_ast, diagnostics, true,
+            compiler_options.disable_main, compiler_options.symtab_only, "input");
+        out += diagnostics.render(input, lm, compiler_options);
```
Shaikh-Ubaid, line 852: **doubt:**
- Please, could someone possibly share if on Line `852` (and other similar lines), we shall just do `out = diagnostics.render(input, lm, compiler_options);` instead of `out += diagnostics.render(input, lm, compiler_options);`? 
    - Since, we did not execute `diagnostics.diagnostics.clear()` so all the previous diagnostics messages might still be present in the current diagnostics and therefore we can just assign the output of `diagnostics.render()` to `out` instead of append to `out`.

Shaikh-Ubaid:

certik (approved):

src/bin/lpython.cpp
certik: I think the `diagnostics` variable itself contains the accumulation of all error messages, so I guess just `out = ...` is enough.

certik (approved):

certik (approved):

src/bin/lpython.cpp
Shaikh-Ubaid: Got it. I just pushed a commit with this change.

Shaikh-Ubaid:
I am adding this to auto-merge as it seems to be approved.

PR #1057, opened 2022-08-29
Fixing & for C backend

ronnuriel:


certik:
@ronnuriel I merged the first two commits, and also added a new commit (https://github.com/lcompilers/lpython/pull/1057/commits/693ed9ffeb3fa7b22455c66bb814110ee4313869) that updates the reference tests, I just did:
```
./run_tests.py -u
```
and committed.

integration_tests/expr_01.py
```diff
@@ -4,13 +4,23 @@
 def add(x: i32, y: i32) -> i32:
     return x + y
 
+@inline
+def and_op(x: i32, y: i32) -> i32:
```
czgdp1807, line 8: I see. I think this feature isn't available in LLVM backend yet (will confirm though). I will add the same (after confirming).

cc: @certik @ronnuriel @rebcabin

certik: Go for it. Tests pass, but I assume LLVM ignores it.

czgdp1807: Well I am wondering if we should use the function inline pass for this. A function will be inlined in LLVM backend if it's decorated with `@inline`. Make sense?

In `--fast` flag all functions (provided they satisfy some conditions) are anyways inlined automatically.

czgdp1807: Yes. LLVM ignores this decorator. Making it work via function inline pass.

PR #1061, opened 2022-08-29
Added binary op and set & op precedence back to 11 (same as the table)

ronnuriel:


certik (approved):
I think this looks good, great job!

Issue #960, opened 2022-08-12
Implementing inverse trigonometric and angle utility functions in NumPy


czgdp1807:
**Links to get started**

See https://github.com/lcompilers/lpython/issues/959#issue-1337177101

**What you have to do?**

- [x] `arcsin` - https://github.com/lcompilers/lpython/pull/1054
- [x] `arccos` - https://github.com/lcompilers/lpython/pull/1054
- [ ] `arctan` - https://github.com/lcompilers/lpython/pull/1062
- [ ] `degrees` - https://github.com/lcompilers/lpython/pull/1062
- [ ] `radians` - https://github.com/lcompilers/lpython/pull/1062

czgdp1807:
@redbopo Feel free to take it over whenever you have time.

PR #1062, opened 2022-08-29
Implement arctan, degrees, radians in Numpy.

redbopo:
Seems like degrees, radians not provided in `src/libasr/runtime/lfortran_intrinsics.c` 
I just add each with float and double input type function.

close [960](https://github.com/lcompilers/lpython/issues/960)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -591,6 +591,29 @@ LFORTRAN_API double_complex_t _lfortran_zatanh(double_complex_t x)
     return catanh(x);
 }
 
+// degrees -----------------------------------------------------------------------
+
+LFORTRAN_API float _lfortran_sdegrees(float x)
+{
+    return x*(180.0/M_PI);
+}
```
czgdp1807, line 599: I think we don't need this at all. This can be implemented directly in `lpython_intrinsic_numpy.py` as Python code.

src/runtime/lpython_intrinsic_numpy.py
```diff
@@ -165,3 +165,63 @@ def _lfortran_sacos(x: f32) -> f32:
...
+########## degrees ##########
+
+@ccall
+def _lfortran_ddegrees(x: f64) -> f64:
+    pass
+
+@overload
+@vectorize
+def degrees(x: f64) -> f64:
+    return _lfortran_ddegrees(x)
```
czgdp1807, line 198: Just implement the formula here. There isn't a need to call C-runtime just for multiplication-division operation.

redbopo:
Thanks for reviews. @czgdp1807 
Now remove the degrees, radians in c runtime.
Instead add pi_32, pi_64 for f32, f64 with them, like cmath.py.

PR #1065, opened 2022-08-30
Auto-saving and loading modules at compile time

czgdp1807:
https://github.com/lcompilers/lpython/issues/992#issuecomment-1225763992


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -153,9 +163,30 @@ namespace CastingUtil {
...
+    }
+
+    LFortran::Location loc;
+    LFortran::ASR::asr_t *asr = LFortran::ASR::make_TranslationUnit_t(al, loc,
+        symtab, nullptr, 0);
+    LFortran::ASR::TranslationUnit_t *tu =
+        LFortran::ASR::down_cast2<LFortran::ASR::TranslationUnit_t>(asr);
+    LFORTRAN_ASSERT(LFortran::asr_verify(*tu));
+
+    std::string modfile_binary = LFortran::save_pycfile(*tu);
```
czgdp1807, line 189: This is the same approach as LFortran i.e., https://github.com/lfortran/lfortran/blob/f07bc66f885d4bc365be140b23a2b62f7e205d85/src/bin/lfortran.cpp#L609-L649 but still [this issue](https://github.com/lcompilers/lpython/issues/992#issuecomment-1229976152) persists. The reason I think is because we keep using the same process to compile the modules as well as the main program. However LFortran launches a different process to compile modules and a different one to compile the main program. If we want to follow the auto-compile, load and save approach in LPython then probably we shouldn't compare Symtab ID for exact matches while comparing the reference tests and the ASR output. @certik Thoughts?

czgdp1807: Or you can add an internal option to not use pre-compiled modules (i.e., ignore pyc files) but compile every-time while generating reference tests.

czgdp1807: > The reason I think is because we keep using the same process to compile the modules as well as the main program.

I made some changes (see in https://github.com/lcompilers/lpython/pull/1065/commits/8317729c7698aa7a50433af585ec3ba7dfbebae5) where I launched a separate lpython process to compile modules and save them in pyc files. And it worked on my mac. Between consecutive clean builds reference tests matched. So I think the issue is not the one in https://github.com/lcompilers/lpython/issues/992#issuecomment-1229976152. The real issue is modules getting compiled with other programs because of which the module's symbol tables are affected. If modules are compiled by an isolated lpython process then I don't see any such thing happening.

czgdp1807: The tests pass with https://github.com/lcompilers/lpython/commit/8317729c7698aa7a50433af585ec3ba7dfbebae5 (an **experimental commit**). So clearly the issue is what I explained in my above comment. Now `system` is not a safe way to compile modules in an isolated `lpython` process. Some better ways include,

* `popen` - See the comparison between `system` and `popen` in https://stackoverflow.com/a/8538345.
* `spawn` - https://stackoverflow.com/a/55490499

I would say encapsulate the logic to safely compile a module in a API inside LPython and then call it everywhere. Current approach in `main` is very non-deterministic so we need to change it anyways.

certik: We should not be using any kind of isolated process. We simply can always compile using a new class for compilation. We just have to handle the symbol table ID correctly, for example as a local variable, etc.

Furthermore, we should use exactly the same code with LFortran, not maintain two separate codes to load/save ASR to mod files.

czgdp1807: > Furthermore, we should use exactly the same code with LFortran, not maintain two separate codes to load/save ASR to mod files.

The code and logic is exactly the same. The difference between LFortran and LPython is explained below,

> The reason I think is because we keep using the same process to compile the modules as well as the main program. However LFortran launches a different process to compile modules and a different one to compile the main program.

Basically LFortran compiles modules according to the instructions in the build system. However LPython compiles modules on the fly. So a module compiled with one program will give different ASR when compiled along with another program. This is not the issue in LFortran because there we compile the modules, save them in `.mod` files before compiling the main program.

czgdp1807: So let's think what happens when we compile modules on the fly. Consider three files `a.py, `b.py` and `module.py`. `a.py` and `b.py` import `module.py`.

**Case 1**

We execute, `lpython a.py` first and then `lpython b.py`. Since we compile modules on the fly during execution of `lpython a.py`, when we will reach the following call to compile `module.py`, 

https://github.com/lcompilers/lpython/blob/d4e2f74eaa657ae36078e735cf3c4fe4c532781d/src/lpython/semantics/python_ast_to_asr.cpp#L243-L244

some symbol tables would have been created already by then. Say `n` symbol tables were created. So the ID of first symbol table while compiling `module.py` will be `n` (symbol table counter starts from `0` and its global for a single `lpython` process). The same thing will be stored in `module.pyc`. So minimum symbol table ID in `module.pyc` will be `n`.

**Case 2**
After a cleaning our repository, we execute `lpython b.py` first and then `lpython a.py`. Now again we will reach the following call,

https://github.com/lcompilers/lpython/blob/d4e2f74eaa657ae36078e735cf3c4fe4c532781d/src/lpython/semantics/python_ast_to_asr.cpp#L243-L244

However this time (since we are compiling b.py) say `m` symbol tables were created till the above call. So now minimum symbol table ID in `module.pyc` will be `m`. 

So you can see how compiling on the fly produces different symbol table IDs if we change the order of programs. If we compile `module.py` before then minimum symbol table ID will always be 0 for any program and hence every symbol table ID will be fixed irrespective of the order in which we compile `a.py` and `b.py`.

Thirumalai-Shaktivel:
What's need to be done to complete this PR?

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -30,6 +32,16 @@
...
 
+#include <sys/types.h>
+#include <sys/stat.h>
+#ifndef _WIN32
+#include <unistd.h>
+#endif
+
+#ifdef _WIN32
+#define stat _stat
+#endif
```
certik, line 43: This should be moved to a separate file, the `python_ast_to_asr` should be platform independent.

src/lpython/semantics/python_ast_to_asr.cpp
certik: Yes, the symbol table ID will differ based on who compiled it. This is not a problem, as the deserialization renumbers the Symbol IDs already. LFortran also has to (and does) it.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -305,7 +368,14 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
     if (ltypes) return nullptr;
 
     if( compile_module ) {
-        mod1 = compile_module_till_asr(al, rl_path, infile, loc, err);
+        // mod1 = compile_module_till_asr(al, rl_path, infile, loc, err);
+        std::string cmd = "lpython -c --disable-main " + infile;
+        system(cmd.c_str());
```
certik, line 373: For now this is fine, but it is fragile (it will fail if `lpython` is not in path, or a different `lpython` gets picked up). We will later rework it to simply call the appropriate C++ functions directly to do the compilation.

Shaikh-Ubaid:
It seems the tests pass locally, but fail at the `CI`.

PR #1066, opened 2022-08-30
Implement separate chaining collision resolution for ``dict``

czgdp1807:
Background and reference comments,

https://github.com/lcompilers/lpython/pull/1037#issue-1349578265

Benchmarks - https://gist.github.com/czgdp1807/e7f7b6ca52c57b16b27ec8d0259c6d4a (see `SeparateChaining` suffixed rows).

Conclusions

1. [As expected](https://github.com/lcompilers/lpython/pull/1037#issue-1349578265) `SeparateChaining` beats all the other approaches for strings as keys. Fulfill the purpose of creating this PR. Since separate chaining avoids re-hashing unless extremely necessary, in case of strings as keys this seems to help as we are able to avoid expensive `malloc` calls.
2. Again as expected `SeparateChaining` is faster than C++, Python's hash maps but slower than linear probing. This is because linear probing takes advantage of spatial locality in case of integers as keys.

Overall, I would say both linear probing and separate chaining in LPython perform better than other approaches with integers as keys. For strings as keys separate chaining performs better than all the other approaches. So, now we have the best collision resolution strategies implemented in LPython for both integers and strings. Floats are anyways disallowed (see #1080).

**Next steps**

1. Expose linear probing and separate chaining via LPython's syntax so that the users can choose their own collision resolution strategy according to their own use cases.
2. By default Linear probing will be used as collision resolution strategy.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -221,7 +221,8 @@ public:
     std::unique_ptr<LLVMUtils> llvm_utils;
     std::unique_ptr<LLVMList> list_api;
     std::unique_ptr<LLVMTuple> tuple_api;
-    std::unique_ptr<LLVMDict> dict_api;
+    std::unique_ptr<LLVMDictInterface> dict_api_lp;
+    std::unique_ptr<LLVMDictInterface> dict_api_sc;
     std::unique_ptr<LLVMArrUtils::Descriptor> arr_descr;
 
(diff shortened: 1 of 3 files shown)


czgdp1807:
@certik This PR is ready. Please review. The description has been updated as well.

certik (approved):
I think this is good to merge.

certik:
I would call this automatically for `dict[str,?]` for now. Later we can expose this explicitly.

PR #1067, opened 2022-08-30
Floor division 9 digits test case

Shaikh-Ubaid:
towards https://github.com/lcompilers/lpython/issues/768

### DIFF
--- a/src/libasr/codegen/asr_to_c_cpp.h
+++ b/src/libasr/codegen/asr_to_c_cpp.h
@@ -582,7 +582,7 @@ R"(#include <stdio.h>
         self().visit_expr(*x.m_arg);
         switch (x.m_kind) {
             case (ASR::cast_kindType::IntegerToReal) : {
-                src = "(float)(" + src + ")";
+                src = "(double)(" + src + ")";
                 break;
             }
             case (ASR::cast_kindType::RealToInteger) : {
--- a/src/libasr/codegen/asr_to_c_cpp.h
+++ b/src/libasr/codegen/asr_to_c_cpp.h
@@ -578,19 +578,11 @@ R"(#include <stdio.h>
         }
     }
 
-    void extract_kinds(const ASR::Cast_t &x, int &arg_kind, int &dest_kind) {
-        dest_kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);
-        ASR::ttype_t *curr_type = ASRUtils::expr_type(x.m_arg);
-        LFORTRAN_ASSERT(curr_type != nullptr)
-        arg_kind = ASRUtils::extract_kind_from_ttype_t(curr_type);
-    }
-
     void visit_Cast(const ASR::Cast_t &x) {
         self().visit_expr(*x.m_arg);
         switch (x.m_kind) {
             case (ASR::cast_kindType::IntegerToReal) : {
-                int arg_kind = -1, dest_kind = -1; // arg_kind is not used currently
-                extract_kinds(x, arg_kind, dest_kind);
+                int dest_kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);
                 switch (dest_kind) {
                     case 4: src = "(float)(" + src + ")"; break;
                     case 8: src = "(double)(" + src + ")"; break;
@@ -599,8 +591,7 @@ R"(#include <stdio.h>
                 break;
             }
             case (ASR::cast_kindType::RealToInteger) : {
-                int arg_kind = -1, dest_kind = -1; // arg_kind is not used currently
-                extract_kinds(x, arg_kind, dest_kind);
+                int dest_kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);
                 switch (dest_kind) {
                     case 4: src = "(int32_t)(" + src + ")"; break;
                     case 8: src = "(int64_t)(" + src + ")"; break;
--- a/src/libasr/codegen/asr_to_c_cpp.h
+++ b/src/libasr/codegen/asr_to_c_cpp.h
@@ -592,11 +592,7 @@ R"(#include <stdio.h>
             }
             case (ASR::cast_kindType::RealToInteger) : {
                 int dest_kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);
-                switch (dest_kind) {
-                    case 4: src = "(int32_t)(" + src + ")"; break;
-                    case 8: src = "(int64_t)(" + src + ")"; break;
-                    default: throw CodeGenError("Cast RealToInteger: Unsupported Kind " + std::to_string(dest_kind));
-                }
+                src = "(int" + std::to_string(dest_kind * 8) + "_t)(" + src + ")";
                 break;
             }
             case (ASR::cast_kindType::RealToReal) : {
--- a/src/libasr/codegen/asr_to_c_cpp.h
+++ b/src/libasr/codegen/asr_to_c_cpp.h
@@ -589,11 +589,23 @@ R"(#include <stdio.h>
         self().visit_expr(*x.m_arg);
         switch (x.m_kind) {
             case (ASR::cast_kindType::IntegerToReal) : {
-                src = "(double)(" + src + ")";
+                int arg_kind = -1, dest_kind = -1; // arg_kind is not used currently
+                extract_kinds(x, arg_kind, dest_kind);
+                switch (dest_kind) {
+                    case 4: src = "(float)(" + src + ")"; break;
+                    case 8: src = "(double)(" + src + ")"; break;
+                    default: throw CodeGenError("Cast IntegerToReal: Unsupported Kind");
+                }
                 break;
             }
             case (ASR::cast_kindType::RealToInteger) : {
-                src = "(int)(" + src + ")";
+                int arg_kind = -1, dest_kind = -1; // arg_kind is not used currently
+                extract_kinds(x, arg_kind, dest_kind);
+                switch (dest_kind) {
+                    case 4: src = "(int32_t)(" + src + ")"; break;
+                    case 8: src = "(int64_t)(" + src + ")"; break;
+                    default: throw CodeGenError("Cast RealToInteger: Unsupported Kind");
+                }
                 break;
             }
             case (ASR::cast_kindType::RealToReal) : {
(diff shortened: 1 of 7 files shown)


Shaikh-Ubaid:

Smit-create (approved):
LGTM, thanks!!

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -582,7 +582,7 @@ R"(#include <stdio.h>
         self().visit_expr(*x.m_arg);
         switch (x.m_kind) {
             case (ASR::cast_kindType::IntegerToReal) : {
-                src = "(float)(" + src + ")";
+                src = "(double)(" + src + ")";
```
czgdp1807, line 585: It should depend on the destination kind ideally. If the destination is 64-bit real then we should do `double` otherwise `float`. Doing `double` all the time might not be a good idea.

czgdp1807: Instead of this update the definition of `_lpython_floordiv` to use `f64` for doing and storing division results,

https://github.com/lcompilers/lpython/blob/0fc1847c56b470521a2afe1bef983facd6f6c220/src/runtime/lpython_builtin.py#L457

can be changed to `r = float(a)/float(b)` because `f32` rounds things up.

Shaikh-Ubaid: Please, could you possibly share how `float()` could cast to a `double`? Currently, it seems `float(x)` casts to `f32` instead of the desired `f64`. 

I tried the suggested change. Please, find the relevant function from the output of `--show-c` below.  It casts the `a` and `b` to `f32`

```c
int32_t __lpython_overloaded_2___lpython_floordiv(int32_t a, int32_t b)
{
    double r;
    int32_t result;
    r = (float)(a)/(float)(b);
    result = (int)(r);
    if (r >=   0.00000000000000000e+00 || (float)(result) == r) {
    }
    _lpython_return_variable = result - 1;
}
```

It seem we need the line ` r = (float)(a)/(float)(b);` to be ` r = (double)(a)/(double)(b);`.

Shaikh-Ubaid:
>It should depend on the destination kind ideally. If the destination is 64-bit real then we should do double otherwise float. Doing double all the time might not be a good idea.

I updated it to (hopefully) use the destination kind. Please, possibly review.

src/libasr/codegen/asr_to_c_cpp.h
czgdp1807: > Please, could you possibly share how `float()` could cast to a `double`? Currently, it seems `float(x)` casts to `f32` instead of the desired `f64`.

I am talking about `float` intrinsic in Python. It always casts (or is supposed to cast to `f64`). Its different C/C++ `float`. The change I recommended in https://github.com/lcompilers/lpython/pull/1067#discussion_r959185801 should be done in the following functions,

https://github.com/lcompilers/lpython/blob/a227df8692cef722eb2630b64c0aa88b133512ed/src/runtime/lpython_builtin.py#L441-L472

C/C++ backend shouldn't be changed incorrectly to make `_lpython_floordiv` work.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -578,15 +578,34 @@ R"(#include <stdio.h>
...
         self().visit_expr(*x.m_arg);
         switch (x.m_kind) {
             case (ASR::cast_kindType::IntegerToReal) : {
-                src = "(float)(" + src + ")";
+                int arg_kind = -1, dest_kind = -1; // arg_kind is not used currently
+                extract_kinds(x, arg_kind, dest_kind);
+                switch (dest_kind) {
+                    case 4: src = "(float)(" + src + ")"; break;
+                    case 8: src = "(double)(" + src + ")"; break;
+                    default: throw CodeGenError("Cast IntegerToReal: Unsupported Kind");
```
czgdp1807, line 597:
```suggestion
                    default: throw CodeGenError("Cast IntegerToReal: Unsupported Kind " + std::to_string(dest_kind));
```

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -578,15 +578,34 @@ R"(#include <stdio.h>
...
+        LFORTRAN_ASSERT(curr_type != nullptr)
+        arg_kind = ASRUtils::extract_kind_from_ttype_t(curr_type);
+    }
+
     void visit_Cast(const ASR::Cast_t &x) {
         self().visit_expr(*x.m_arg);
         switch (x.m_kind) {
             case (ASR::cast_kindType::IntegerToReal) : {
-                src = "(float)(" + src + ")";
+                int arg_kind = -1, dest_kind = -1; // arg_kind is not used currently
```
czgdp1807, line 592: Why is `arg_kind` needed? Isn't only the `dest_kind` important in this case?

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -578,15 +578,34 @@ R"(#include <stdio.h>
...
                 break;
             }
             case (ASR::cast_kindType::RealToInteger) : {
-                src = "(int)(" + src + ")";
+                int arg_kind = -1, dest_kind = -1; // arg_kind is not used currently
+                extract_kinds(x, arg_kind, dest_kind);
+                switch (dest_kind) {
+                    case 4: src = "(int32_t)(" + src + ")"; break;
+                    case 8: src = "(int64_t)(" + src + ")"; break;
+                    default: throw CodeGenError("Cast RealToInteger: Unsupported Kind");
```
czgdp1807, line 607:
```suggestion
                    default: throw CodeGenError("Cast RealToInteger: Unsupported Kind " + std::to_string(dest_kind));
```

czgdp1807 (changes requested):

src/libasr/codegen/asr_to_c_cpp.h
Shaikh-Ubaid: Please, possibly share if it is to be removed. Along with it, it seems we could also remove the function `extract_kinds()`.

Shaikh-Ubaid: I just noticed it. It seems it is not unused. We are passing it as a parameter to `extract_kinds()`.

Shaikh-Ubaid: Please, possibly share if it is to be removed. Along with it, it seems we could also remove the function `extract_kinds()`. We can get the `dest_kind` directly using `dest_kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);`.

src/libasr/codegen/asr_to_c_cpp.h
Shaikh-Ubaid: https://github.com/lcompilers/lpython/blob/a227df8692cef722eb2630b64c0aa88b133512ed/src/lpython/semantics/python_ast_to_asr.cpp#L1302-L1304

It seems that for division, the kind is already taken to be `8`. Please, possibly share if we still need to cast integers to floats using `float()` in the above functions that you shared.

Currently, it seems the `C` and `CPP` backends did not yet have support for casting `integers` to `double`. I attempted to add this support in https://github.com/lcompilers/lpython/pull/1067/commits/0613bd2f46c40eb6e647d121120376c024ec6d89, https://github.com/lcompilers/lpython/pull/1067/commits/ccc29fd3b1de0a9b40a0161059ec7456fe660ff3. Please possibly share if it is fine.

Shaikh-Ubaid:
This seems ready. Please possibly review and please share feedback. I shared the current concerns in comments https://github.com/lcompilers/lpython/pull/1067#discussion_r959614630 and https://github.com/lcompilers/lpython/pull/1067#discussion_r959588608.

src/libasr/codegen/asr_to_c_cpp.h
czgdp1807: > It seems that for division, the kind is already taken to be 8.

That's incorrect IMO. If both operands are 32-bit floating points then the result should also be 32-bit floating point. 

> Please, possibly share if we still need to cast integers to floats using float() in the above functions that you shared.

We should still do `float()` in Python code. Python code should be correct, its the compiler that needs to be fixed.

src/libasr/codegen/asr_to_c_cpp.h
czgdp1807: > It is mostly present because it seems that there could be a possibility that we might need the `arg_kind` later.

Well, then we will make the change later. :-). Let's use  only`dest_kind` for now.

> Please, possibly share if it is to be removed. Along with it, it seems we could also remove the function `extract_kinds()`. We can get the `dest_kind` directly using `dest_kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);`. (where `x` is of type `const ASR::Cast_t`).

Yes, exactly.

Shaikh-Ubaid: Got it.

src/libasr/codegen/asr_to_c_cpp.h
Shaikh-Ubaid: Got it.

src/libasr/codegen/asr_to_c_cpp.h
Shaikh-Ubaid: I just added a commit with this suggestion.

src/libasr/codegen/asr_to_c_cpp.h
Shaikh-Ubaid: I just added a commit to use `float()` to convert `i32` to `f64`. Gangandeep Sir, please, could you possibly share if we also need to cast `f32` to `f64` in `def _lpython_floordiv(a: f32, b: f32) -> f32`?

Shaikh-Ubaid:
I added some commits. I shared a concern here https://github.com/lcompilers/lpython/pull/1067#discussion_r960916229. This is ready. Please possibly review.

src/libasr/codegen/asr_to_c_cpp.h
czgdp1807: Yes. Otherwise it will round things up. You can play with to find such case yourself. I figured one case while working on other PR.

Shaikh-Ubaid: Got it.

Shaikh-Ubaid:
This seems ready. Please possibly review and please share feedback.

czgdp1807 (approved):
LGTM. Will merge this by today EOD. cc: @certik

Shaikh-Ubaid:

PR #950, opened 2022-08-09
Support for print ASR after pass.

redbopo:
Add command `--print-after-asr-pass`  on [issue-943](https://github.com/lcompilers/lpython/issues/943)
The extra space end of line is removed by editor plugIns .

But  there will be problems if we enter the python file a little large, like `./integration_tests/func_inline_01.py`

Issue #943, opened 2022-08-09
Improving ASR passes infrastructure


certik:
Currently one can print all passes using `lfortran --pass xx examples/expr2.f90`.

High priority:

* easier registration of new passes
* printing ASR output after every pass

Low priority:

* figure out how to handle parameters for each pass. Some kind of a map of argument name, type and value; dynamically constructed from the ASR passes cpp files; keep the door open for creating plugins for ASR passes, that are registered at runtime (using a shared library)
* Expose arguments in the driver
* Add `lfortran --print-passes` which would print all passes, one on each line, with a "name" and a "one line description"
* Add `lfortran --print-pass=fma` which will print "name", "one line description", "full multi-paragraph description", "all parameters: names, types, default values"
* Add `lfortran --print-pass-pipeline --show-llvm examples/expr2.f90` that will print half a page of all the passes in some format such as `arr_slice,do_loops,fma[ax=5,b=7],...` including all parameters that the LLVM backend (let's say) sets for the given pass (and the given compiler options such as `--fast` and architecture, for loop unrolling for example).
* Add `lfortran --pass="arr_slice,fma[ax=5,b=7]" --show-llvm examples/expr2.f90` that would reproduce the default LLVM pass pipeline; then one can modify it and the command line (add/remove passes, change pass parameters)

czgdp1807:
> easier registration of new passes

Added to my list. Will work on it.

redbopo:
I'd like to add on `printing ASR output after every pass` .

PR #1069, opened 2022-08-31
Add an option to skip pass in default pipeline

redbopo:
Add an option to skip/ignore a pass in the default pipeline in [issue_943](https://github.com/lcompilers/lpython/issues/943)

Although there will be conflict with print asr pass [PR](https://github.com/lcompilers/lpython/pull/950) 
I think it's okay to just start.

### DIFF
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -1297,6 +1297,7 @@ int main(int argc, char *argv[])
         bool show_errors = false;
         bool with_intrinsic_modules = false;
         std::string arg_pass;
+        std::string skip_pass;
         bool arg_no_color = false;
         bool show_llvm = false;
         bool show_asm = false;
@@ -1363,6 +1364,7 @@ int main(int argc, char *argv[])
         app.add_flag("--indent", compiler_options.indent, "Indented print ASR/AST");
         app.add_flag("--tree", compiler_options.tree, "Tree structure print ASR/AST");
         app.add_option("--pass", arg_pass, "Apply the ASR pass and show ASR (implies --show-asr)");
+        app.add_option("--skip_pass", skip_pass, "Skip an ASR pass in default pipeline");
         app.add_flag("--disable-main", compiler_options.disable_main, "Do not generate any code for the `main` function");
         app.add_flag("--symtab-only", compiler_options.symtab_only, "Only create symbol tables in ASR (skip executable stmt)");
         app.add_flag("--time-report", time_report, "Show compilation time report");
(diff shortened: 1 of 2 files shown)


src/bin/lpython.cpp
```diff
@@ -902,6 +903,7 @@ int main(int argc, char *argv[])
         app.add_flag("--indent", compiler_options.indent, "Indented print ASR/AST");
         app.add_flag("--tree", compiler_options.tree, "Tree structure print ASR/AST");
         app.add_option("--pass", arg_pass, "Apply the ASR pass and show ASR (implies --show-asr)");
+        app.add_option("--skip_pass", skip_pass, "Skip an ASR pass in default pipeline");
```
czgdp1807, line 906: What if I want to skip multiple passes? Like loop unrolling and function inlining? I think we should treat `skip_pass` and `pass` command line arguments in the same fashion while parsing them but just ignore the pass which is present `skip_pass` vector.

czgdp1807:
Rest this is a good PR. @redbopo Would you like to complete it so that we can merge?

src/bin/lpython.cpp
redbopo: Yes, I consider it will be like the way in parse_pass_arg().
although in the issue mentioned just said ` Add an option to skip/ignore a pass in the default pipeline`
So I left it here.

redbopo:
Thanks for reviews, @czgdp1807 , I'll add it soon.

src/libasr/pass/pass_manager.h
```diff
@@ -129,34 +158,15 @@ namespace LCompilers {
...
-                            std::cerr << it.first << std::endl;
-                        }
-                        exit(1);
-                    }
-                    _user_defined_passes.push_back(current_pass);
-                    current_pass.clear();
-                }
-            }
+            _parse_pass_arg(arg_pass, _user_defined_passes);
+            _parse_pass_arg(skip_pass, _skip_passes);
```
czgdp1807, line 169: 👍🏻.

redbopo:
Thanks for reviews @czgdp1807 , is it okay to merge it now?

czgdp1807 (approved):
Fine by me. We just need to update it with `main` and then get it in. @certik Any ideas for testing this other than manual verification?

certik (approved):
I think this is fine for now. I think we have to ultimately rework how passes are managed and make this more general. Until then I think this works.

czgdp1807 (approved):
LGTM. But merge after #1411 is complete.

czgdp1807:
@Smit-create Please rebase on top of latest `main` and get this in. Thanks.

Issue #1035, opened 2022-08-24
``mod`` for arrays


czgdp1807:
@certik Builtin `%` in LPython doesn't work on arrays by default because it is implemented as a function and not a `binop`. So the question is which approach to follow,
1. To make `_mod` function defined in `lpython_builtin.py` vectorised;
2. Define `Mod` as `binop` and then use `builder->CreateSRem` in LLVM backend. Note that not all backends might support modulus operation on types other than integers.
3. Define [`mod`](https://numpy.org/doc/stable/reference/generated/numpy.mod.html) and then make it vectorisable.

I think 3rd approach is a good one and complies with Python and NumPy standards. What do you say @certik?

_Originally posted by @czgdp1807 in https://github.com/lcompilers/lpython/pull/1002#discussion_r951450426_

certik:
If the third approach works, we can go with that I think. Otherwise we can do 1., or 2.

PR #1076, opened 2022-09-02
Implementing ``numpy.mod`` for array inputs

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1035


rebcabin:
Is numpy.mod defined for negative arguments? If so, are they tested here?

czgdp1807:
From what I am able to figure out, `numpy.mod` doesn’t specify anything on sign of integers. Though its `numpy.fmod` which is supposed to work with negative integers. Plus `numpy.mod` is re-using Python’s `%` operator (which `numpy.mod` is supposed to do), so the behaviour will be equivalent and depend on how we define `%`. 
In any case, I will add a test for negative arrays and arrays with positives, negatives and 0s on Monday.

czgdp1807:
> Is numpy.mod defined for negative arguments? If so, are they tested here?

Done.

certik (approved):
This looks good.

Issue #733, opened 2022-07-04
bug in list implementation


Madhav2310:
Code:
```
s: str
s= "sample text"
l: list[str]
l[:0] = s
print(l)
```
Expected output:
`['s', 'a', 'm', 'p', 'l', 'e', ' ', 't', 'e', 'x', 't'] `

Error:
```
semantic error: Type mismatch in assignment, the types must be compatible
   --> /home/madhav/lpython/src/runtime/lpython_builtin.py:156:5
    |
156 |     l[:0] = s
    |     ^^^^^   ^ type mismatch ('list[str]' and 'str')

```

ArpanShah2k:
I am interested in fixing this issue . It will be a great opportunity if you allot me to solve this issue . 
@czgdp1807 @Madhav2310 @certik @rebcabin @HaoZeke

PR #1079, opened 2022-09-04
ASR: String to List Conversion

Smit-create:
#733


Thirumalai-Shaktivel:
LGTM, Thank you!
To be implemented after this PR is merged:
```py
x = list()
x = list([])
x = list("lpython")
```

czgdp1807:
Will this work for runtime string variables?

Smit-create:
I will add the LLVM backend as a follow-up for runtime values.

tests/cast.py
```diff
@@ -0,0 +1,10 @@
+def f():
+    s: str
+    s = "lpython"
+    x: list[str]
+    x = list(s)
+    y: list[str]
+    y = ['a', 'b', 'c']
+    x = list(y)
```
czgdp1807, line 8: The verification is not happening whether the generated list is correct or now? Can you add some asserts for constant values here? Something like,

```
assert list("lpython") == ['l', 'p', 'y', 't', 'h', 'o', 'n']
```

czgdp1807 (changes requested):

tests/cast.py
czgdp1807: Covering empty strings, large strings, single character strings would be good.

integration_tests/str_to_list_cast.py
```diff
@@ -0,0 +1,14 @@
...
+    s = "lpython"
+    i: i32
+    for i in range(len(s)):
+        assert x[i] == s[i]
+    x = list("")
+    assert len(x) == 0
+    x = list("L")
+    assert len(x) == 1 and x[0] == 'L'
+
+f()
```
czgdp1807, line 14: This is okay. Really surprising though. None of the variables (`s`, `x`) are marked as constant but still they work with only the compile time logic you have added.

czgdp1807 (approved):

integration_tests/str_to_list_cast.py
Smit-create: That is because if we have a string value at compile time then we can directly make a list constant out of it. So no cast, in that case, is required.

PR #1080, opened 2022-09-04
Disallow ``f64``, ``f32``, ``c64``, ``c32`` as keys in dict

### LINKED ISSUE #1074: Disallow ``f64``, ``f32``, ``c64``, ``c32`` as keys in ``dict``
For background please read the following comments,

> Question - Should we allow floats/reals as keys?
> The reason why I am asking this question is because floats are always approximately equal. So, converting them to int of same size by memcpy won't work well. Plus, its not a good practice to use floats as keys. 123.000 and 123.000001 are approximately same, now user may expect them to work as a single key but due to slight difference both will give totally different hash value. So disallowing these as keys will prevent a lot of bad practices.
> Also, in case of collisions we will go for exact equality checks in keys but as I said above floats are approximately equal mostly, so users may get very random results due to this property of floats.
> Even if we want to allow floats as keys, we should ask the user to provide an equality function in the dict type which we should use to compare floats in case of collisions.
> See https://softwareengineering.stackexchange.com/a/391107 and https://diego.assencio.com/?index=67e5393c40a627818513f9bcacd6a70d for details.

What is to be done?

> We should disallow it in AST->ASR, with a nice error message and explanation.

How to be done?

We need to check for `ASR::is_a<ASR::Real_t>(*key_type) || ASR::is_a<ASR::Complex>(*key_type)` in the following places and throw a `SemanticError` with a nice explanation (take ideas from above and your own knowledge),

1. https://github.com/lcompilers/lpython/blob/4ea45c5338fbbf58b35b237134ca3da78d5c3c1b/src/lpython/semantics/python_ast_to_asr.cpp#L1117-L1119
2. https://github.com/lcompilers/lpython/blob/4ea45c5338fbbf58b35b237134ca3da78d5c3c1b/src/lpython/semantics/python_ast_to_asr.cpp#L3588

ronnuriel:

Closes #1074

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -1147,6 +1147,15 @@ public:
         return false;
     }
 
+    void raise_error_when_dict_key_is_float_or_complex(ASR::ttype_t* key_type, const Location& loc) {
+        if(ASR::is_a<ASR::Real_t>(*key_type) || ASR::is_a<ASR::Complex_t>(*key_type)) {
+            throw SemanticError("'dict' key type cannot be float/complex because resolving collisions "
+                                "by exact comparison of float/complex values will result in unexpected "
+                                "behaviours. In addition fuzzy equality checks with a certain tolerance "
+                                "does not follow transitivity with float/complex values.", loc);
+        }
+    }
+
     // Convert Python AST type annotation to an ASR type
     // Examples:
     // i32, i64, f32, f64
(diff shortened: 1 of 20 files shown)


czgdp1807 (commented):
Thanks. I will shepherd this tomorrow.

czgdp1807 (approved):
@certik This is ready. Let me know if this looks good to you.

certik (approved):
This looks great, thanks for implementing this!

PR #1083, opened 2022-09-05
Implement numpy.floor and numpy.ceil functions

Madhav2310:


Madhav2310:
```
@overload
@vectorize
def floor(x: f32) -> f32:
    result: i32
    result = int(x)
    if x >= 0 or x == result:
        return float(result)
    return float(result - 1)
```
Since float returns f64, I tried other ways to return f32, but none seems to work. Please suggest how I can resolve it. @czgdp1807 @Smit-create

Thirumalai-Shaktivel:
Try assigning the value to the variable of type `f32` and return the variable, Maybe that would work.

Madhav2310:
> Try assigning the value to the variable of type `f32` and return the variable, Maybe that would work.

Thank you, that worked!

Madhav2310:
@czgdp1807 @Smit-create Please review

Smit-create:
> @Smit-create Would you mind completing this and get it merged?

Will do

czgdp1807 (approved):

PR #1085, opened 2022-09-05
PASS: print list

Shaikh-Ubaid:
towards https://github.com/lcompilers/lpython/issues/882

This `PR` adds an `ASR` pass to support printing lists.

**Notes:**
- It seems parsing of empty `list` constant in print statement is not supported at the `AST` level. For example: `print([])` at the `ASR` level is received as `print()` that is as an empty print statement.
    - Therefore, printing empty `list` constant is currently not supported.


Shaikh-Ubaid:

Shaikh-Ubaid:
Please, could someone possibly share if we shall add the functionality to print tuples as well in the same pass? It seems, for printing `tuple`, we just need to print `(` inplace of `[` and `)` inplace of `]`.

We could also implement another ASR pass to support printing tuples, but I guess incorporating both the support for printing `list` and `tuple` in the same pass saves an extra pass and thus would (hopefully) be more efficient.

redbopo:
I think the idea place of issue 882 is to provide with executable a.out file, which means asr itself also works to lower to llvm next.
And I just try the PR, it will failed to llvm.

integration_tests/print_02.py
```diff
@@ -0,0 +1,19 @@
+from ltypes import i32, f64
+
+# Test: Printing ListConstant
+def f():
+    a: list[str] = ["ab", "abc", "abcd"]
```
czgdp1807, line 5: What if the list is nested? Something like, `anested: list[list[i32]]`

Can you add a test for the same?

Shaikh-Ubaid: Thank you for sharing. I think this pass does not support it currently.

Shaikh-Ubaid: I added the support for nested lists and also tests for the same.

Shaikh-Ubaid:

czgdp1807:
Will review tomorrow.

integration_tests/print_02.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # print list constant
+    print([-3, 2, 1, 0])
+    print(['a', 'b', 'c', 'd' , 'e', 'f'])
+
+def test_nested_lists():
+    w: list[list[list[list[list[f64]]]]] = [[[[[2.13, -98.17]]], [[[1.11]]]]]
+    x: list[list[list[i32]]] = [[[3, -1], [-2, 5], [5]], [[-3, 1], [2, -5], [-5]]]
+    y: list[list[f64]] = [[3.14, -1.0012], [-2.38, 5.51]]
+    z: list[list[str]] = [["bat", "ball"], ["cat", "dog"], ["c", "c++", "java", "python"]]
```
czgdp1807, line 23: These are small lists. Test them on scale as well (like lists of size 50 to 80).

czgdp1807 (changes requested):

integration_tests/print_02.py
Shaikh-Ubaid: Got it. I just added a stronger test case. Please possibly review and please share feedback.

certik (approved):
Looks good to me. Let's wait also for @czgdp1807 to review.

czgdp1807 (approved):
Seems fine to me. Please add support for printing tuples in the next PR so that we can test your pass with mixed types like, `list[tuple[i32, f64, str]]`. I read the code and it should work with mixed types but still testing is better.

Issue #772, opened 2022-07-15
Create runtime stacktraces


certik:
The generated LLVM code needs the ability to create Python stacktraces. So that when bounds check error, or string->int conversion error happens, we print a runtime stacktrace. So we need to get the stack addresses, convert them to LLVM functions, then use LLVM debugging information to find ASR functions that they correspond to, and finally ASR location information (that we need to store in the binary) to find the original Python file and line/column number where the error comes from. Then we can print Python style stacktraces at runtime.

Thirumalai-Shaktivel:
> Later we need some solid way to handle runtime errors, see https://github.com/lcompilers/lpython/issues/772. Maybe you can create a function runtime_error that will print the error message and exits, and we'll call it from the LLVM backend when we need to give a runtime error message. Later we will extend this to give a nice stacktrace.

..

certik:
One (orthogonal) issue to fix is:

* [ ] https://github.com/lcompilers/lpython/issues/1218

certik:
To create runtime stacktraces, I would do the following to get started:

* [ ] On all platforms we would follow the macOS convention that creates a `.dSYM` with `lines.txt` and `lines.dat`.
* [ ] We port the relevant parts from src/libasr/stacktrace.cpp into the runtime library (we have to port from C++ to C)
* [ ] We have to figure out how to create `lines.txt` and `lines.dat` in LPython/LFortran at compilation time

certik:
One can test this by:
```c
int main() {
    int i;
    i = 1;
    i = i + 2;
    i = i + 3;
    i = i + 4;
    return i;
}
```
Compile using:
```
clang -emit-llvm -S -g a.c
```
and LLVM then stores debugging information for each instruction (and function), such as:
```llvm
define i32 @main() #0 !dbg !13 {
...
  store i32 1, i32* %2, align 4, !dbg !19
  %3 = load i32, i32* %2, align 4, !dbg !20
  %4 = add nsw i32 %3, 2, !dbg !21
  store i32 %4, i32* %2, align 4, !dbg !22
  %5 = load i32, i32* %2, align 4, !dbg !23
  %6 = add nsw i32 %5, 3, !dbg !24
  store i32 %6, i32* %2, align 4, !dbg !25
  %7 = load i32, i32* %2, align 4, !dbg !26
  %8 = add nsw i32 %7, 4, !dbg !27
  store i32 %8, i32* %2, align 4, !dbg !28
[...]

PR #1093, opened 2022-09-06
Add Index bounds checking option to check the `out of range` in lists

Thirumalai-Shaktivel:


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -589,8 +589,37 @@ namespace LFortran {
...
+        llvm::BasicBlock *mergeBB = llvm::BasicBlock::Create(context, "ifcont");
+
+        llvm::Value* cond = builder->CreateOr(
+                                builder->CreateICmpSGT(pos, end_point),
+                                builder->CreateICmpSGT(zero, pos));
+        builder->CreateCondBr(cond, thenBB, elseBB);
+        builder->SetInsertPoint(thenBB);
+        {
+            std::string message = "list index out of range";
+            llvm::Value *fmt_ptr = builder->CreateGlobalStringPtr("IndexError: %s\n");
```
czgdp1807, line 610: I think its time to work out a method to provide runtime information (like in this the index and the bounds of the list). Its an easy-medium thing. Would you like to do that?

czgdp1807: May be do it in a follow up PR.

czgdp1807: Another pain point is to test whether this error is raised or not for out of range indices. In  CPython we use `pytest.raises` (or something like that, don't remember the exact details). I am not sure what to use here?

Thirumalai-Shaktivel:
> Would you like to do that?
Yup, definitely!
Can you please share, how one can implement it?

czgdp1807:
Tests are failing.

Thirumalai-Shaktivel:
@czgdp1807 I think this PR is ready, any other changes to be made!

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -589,25 +589,56 @@ namespace LFortran {
...
+            int exit_code_int = 1;
+            llvm::Value *exit_code = llvm::ConstantInt::get(context,
+                    llvm::APInt(32, exit_code_int));
+            exit(context, module, *builder, exit_code);
+        }
+        builder->CreateBr(mergeBB);
+
+        llvm_utils->start_new_block(elseBB);
+        llvm_utils->start_new_block(mergeBB);
+        return;
```
czgdp1807, line 623:
```suggestion
```

certik:
TODO:
* [ ] only call this in debug mode, for now let's only do it when an option is enabled, such as `--enable-bounds-checking`. Later we'll figure out what options to do by default in Debug mode.


Later we need some solid way to handle runtime errors, see #772. Maybe you can create a function `runtime_error` that will print the error message and exits, and we'll call it from the LLVM backend when we need to give a runtime error message. Later we will extend this to give a nice stacktrace.

czgdp1807 (approved):
The changes here are anyways unused in CI so I am neutral on merging this into main.

Thirumalai-Shaktivel:
Thanks for the review!

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -160,6 +160,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
     std::unique_ptr<llvm::Module> module;
     std::unique_ptr<llvm::IRBuilder<>> builder;
     Platform platform;
+    bool check_index_bound;
```
certik, line 163:
```suggestion
    bool bounds_checking;
```

src/libasr/utils.h
```diff
@@ -32,6 +32,7 @@ struct CompilerOptions {
     bool generate_object_code = false;
     bool no_warnings = false;
     bool no_error_banner = false;
+    bool check_index_bound = false;
```
certik, line 35:
```suggestion
    bool enable_bounds_checking = false;
```

certik (approved):
I think this looks great.

We should add tests for going out of bounds, test the error messages, etc. But we need to figure out how to handle it, so we can do that in subsequent PRs.

certik:
This might be a way to test this: https://github.com/lcompilers/lpython/issues/1128

Thirumalai-Shaktivel:
This PR is ready. if everything looks good, please merge it.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3315,7 +3318,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 llvm::Value* list = tmp;
                 this->visit_expr_wrapper(asr_target0->m_pos, true);
                 llvm::Value* pos = tmp;
-                target = list_api->read_item(list, pos, true);
+                target = list_api->read_item(list, pos, *module, true, enable_bounds_checking);
```
czgdp1807, line 3321: I wouldn't pass it to each function. Instead make `enable_bounds_checking` a static data member of `LLVMList` because it is a global flag for the whole compiler so should be shared across all the `LLVMList` objects.

czgdp1807 (changes requested):

src/libasr/codegen/asr_to_llvm.cpp
Thirumalai-Shaktivel: Done, any other changes?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1283,7 +1285,8 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         ptr_loads = ptr_loads_copy;
         llvm::Value *pos = tmp;
 
-        tmp = list_api->read_item(plist, pos,
+        list_api->enable_bounds_checking = enable_bounds_checking;
```
czgdp1807, line 1288: `enable_bounds_checking` should be a static data member of LLVMList. Right now its localised to only one object. So passing it as an argument or doing `list_api->enable_bounds_checking = enable_bounds_checking` is the same thing.

Thirumalai-Shaktivel: In that case, how do I assign the `enable_bounds_checking` value?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -229,11 +230,12 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
     bool is_assignment_target;
 
     ASRToLLVMVisitor(Allocator &al, llvm::LLVMContext &context, Platform platform,
-        diag::Diagnostics &diagnostics) :
+        bool enable_bounds_checking, diag::Diagnostics &diagnostics) :
     diag{diagnostics},
     context(context),
     builder(std::make_unique<llvm::IRBuilder<>>(context)),
     platform{platform},
+    enable_bounds_checking{enable_bounds_checking},
```
czgdp1807, line 238: Just do the following inside the body of `ASRToLLVMVisitor` constructor,

```cpp
LLVMList::enable_bounds_checking = enable_bounds_checking
```

src/libasr/codegen/asr_to_llvm.cpp
czgdp1807: Remove these assignments from everywhere in this file.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -251,6 +251,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         llvm_utils->tuple_api = tuple_api.get();
         llvm_utils->list_api = list_api.get();
         llvm_utils->dict_api = nullptr;
+        LLVMList::enable_bounds_checking = enable_bounds_checking;
```
Thirumalai-Shaktivel, line 254: causes:
```
[ 98%] Built target test_lpython
/usr/bin/ld: ../libasr/libasr.a(asr_to_llvm.cpp.o): in function `LFortran::ASRToLLVMVisitor::ASRToLLVMVisitor(Allocator&, llvm::LLVMContext&, LFortran::Platform, bool, LFortran::diag::Diagnostics&)':
/home/thirumalai/Open_Source/lpython/src/libasr/codegen/asr_to_llvm.cpp:254: undefined reference to `LFortran::LLVMList::enable_bounds_checking'
/usr/bin/ld: ../libasr/libasr.a(llvm_utils.cpp.o): in function `LFortran::LLVMList::write_item(llvm::Value*, llvm::Value*, llvm::Value*, LFortran::ASR::ttype_t*, llvm::Module&)':
/home/thirumalai/Open_Source/lpython/src/libasr/codegen/llvm_utils.cpp:990: undefined reference to `LFortran::LLVMList::enable_bounds_checking'
/usr/bin/ld: ../libasr/libasr.a(llvm_utils.cpp.o): in function `LFortran::LLVMList::write_item(llvm::Value*, llvm::Value*, llvm::Value*, llvm::Module&)':
/home/thirumalai/Open_Source/lpython/src/libasr/codegen/llvm_utils.cpp:1000: undefined reference to `LFortran::LLVMList::enable_bounds_checking'
/usr/bin/ld: ../libasr/libasr.a(llvm_utils.cpp.o): in function `LFortran::LLVMList::read_item(llvm::Value*, llvm::Value*, llvm::Module&, bool)':
/home/thirumalai/Open_Source/lpython/src/libasr/codegen/llvm_utils.cpp:2174: undefined reference to `LFortran::LLVMList::enable_bounds_checking'
collect2: error: ld returned 1 exit status
make[2]: *** [src/bin/CMakeFiles/lpython.dir/build.make:150: src/bin/lpython] Error 1
make[1]: *** [CMakeFiles/Makefile2:425: src/bin/CMakeFiles/lpython.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

Thirumalai-Shaktivel: Was this because of `static` charcteristic of the data menmeber?

czgdp1807:
```suggestion
        bool LLVMList::enable_bounds_checking = enable_bounds_checking;
```

Thirumalai-Shaktivel:
```
```

czgdp1807:
To be noted that this change is not tested so we don't know if it works or now. I think following https://github.com/lcompilers/lpython/issues/1128 to test this PR is important before we merge this.

src/libasr/codegen/llvm_utils.h
```diff
@@ -128,6 +128,10 @@ namespace LFortran {
 
         public:
 
+            static bool enable_bounds_checking;
```
czgdp1807, line 131: Making this private would be better now.

Thirumalai-Shaktivel: Done!

Thirumalai-Shaktivel:
That works, thank your!

src/libasr/codegen/llvm_utils.cpp
certik: Yes, we need that, so Thirumalai and I created an issue for this: https://github.com/lcompilers/lpython/issues/1128

src/libasr/codegen/asr_to_llvm.cpp
certik: Hm I don't think that's a good design: that's like having a global variable. What if we are doing multiple compilation (even in parallel) with multiple options? Then this would fail.

I think we should always be passing all input data as arguments whenever possible. We should not use global or static variables.

certik:
Everything looks good. I would not use static functions, nor global functions. There are a very few places in LPython/LFortran that use global variables, and we should eventually remove all of them, and keep everything stateless.

A compromise is the `compiler_options` struct, that is typically passed in as arguments, that way it's easy to add more options.

src/libasr/codegen/asr_to_llvm.cpp
czgdp1807: > Hm I don't think that's a good design: that's like having a global variable.

Well, `lpython --enable_bounds_checking abc.py` means that enable bound checking for lists while compiling `abc.py`. Now `static bool enable_bounds_checking` is not a global variable for the compiler, its a shared variable for all the `LLVMList`'s class objects under `lpython` process which is compiling `abc.py`. So its not the same as having a global variable IMO. The point here is that when we repeatedly pass `enable_bounds_checking` as an argument to all the methods of `LLVMList` class then its fairly easy to make mistakes by passing an wrong value to some methods. This would lead to some selective bound checking on some lists and not some other lists, or bound checking enable for read and not for write (just because we passed false by mistake and provided we don't test this right now).

> What if we are doing multiple compilation (even in parallel) with multiple options? Then this would fail.

By in parallel if you mean `lpython --enable_bounds_checking abc.py` and `lpython xyz.py` launched simultaneously in the background, then it wouldn't fail because the two processes are isolated from each other I think. As I said the static member which is added here is shared across `LLVMList`'s objects and not more than that.

If we don't want to use `static bool LLVMList::enable_bounds_checking` then at least we should properly test this PR and shouldn't merge otherwise it will pile up mistakes and it will be hard to debug in the future once we have the ability to test this. So yes I am not completely against passing `enable_bounds_checking` as an argument but then we should make sure that we are not making any mistakes and doing bounds check selectively for some lists even though user wants it to be enabled for all of their lists and other container data structures.

certik: We will be compiling a large project in parallel, in the same process. We parse everything to AST (fully parallel), then based on the file dependencies, we compile everything to ASR. There can in principle be different options for different files. My understanding of the static variable is that it will not work, as in one file we set it to true, for another file we set it to false, bit it overwrites the previous true to false, so the first file will now have false, and things will break. Is my understanding correct? If so, we can't use this approach.

czgdp1807: Its okay not to use static variables but then we shouldn’t merge this PR without being tested for all the list operations. My main concern is piling up of mistakes if we merge this without testing.

Thirumalai-Shaktivel: Yup, we will wait until https://github.com/lcompilers/lpython/issues/1128 is implemented!

certik:
Let's do this the right way:

* Not use static variables. If the code is big, let's think how to make it less verbose, but keep things local as much as possible, do not use global variables. It might be that we need to instantiate the LLVMUtils as a regular instance inside the LLVM backend class. The mistake seems to be coming from using LLVMUtils as a singleton, which now need state. We should not add a (global) state. But making LLVMUtils local instance, we can then store a state.

* Add tests that check that all the out of bounds code is being correctly executed in each case.

src/libasr/codegen/asr_to_llvm.cpp
czgdp1807: What’s preventing it from being implemented? Any blockers? If not then may be let’s go ahead and implementing it.

src/bin/lpython.cpp
```diff
@@ -1129,6 +1129,7 @@ int main(int argc, char *argv[])
         app.add_flag("--static", static_link, "Create a static executable");
         app.add_flag("--no-warnings", compiler_options.no_warnings, "Turn off all warnings");
         app.add_flag("--no-error-banner", compiler_options.no_error_banner, "Turn off error banner");
+        app.add_flag("--enable-bounds-checking", compiler_options.enable_bounds_checking, "Turn on index bounds checking");
```
czgdp1807, line 1132: I think its default value should be `true` according to LPython contract defined in https://github.com/lcompilers/lpython/issues/1139 (see quotation below). However if I enable `--fast` option then I think this should be `false` by default. Note that I think at the end user's choices should be valued the most. If they enable `--fast --enable-bounds-checking` then `compiler_options.enable_bounds_checking` should be `true`. See the quotation below,

> The default option in LPython should be Debug mode. The Debug mode must have all runtime safety options on. In this Debug mode, you get very fast compilation, and safe runtime behavior. No silent surprises. Beautiful runtime error messages with stack traces. If you don't get any runtime error and things compiles (obviously), then it will behave just like in CPython.

src/libasr/codegen/asr_to_llvm.cpp
czgdp1807: > There can in principle be different options for different files. My understanding of the static variable is that it will not work, as in one file we set it to true, for another file we set it to false, bit it overwrites the previous true to false, so the first file will now have false, and things will break. Is my understanding correct?

If you are launching separate "lpython file1.py", "lpython file2.py" "lpython --enable-bounds-checking file3.py" commands then no, nothing would break because these are separate processes. One process cannot see and overwrite the static variables of a class in another process. So the options you set for one `lpython` command to compile one file won't have any affect on other command with different set of options. In fact, **the change I proposed is limited to the private scope of `LLVMList` and not even classes outside `LLVMList` will be able to touch any of its static variables**. So nothing to worry about. But again I provided two options,

1. `static bool LLVMList::enable_bounds_checking` - This helps us in applying the bounds check by setting this variable one line, this is a safe choice as anyways we launch multiple "lpython" commands to compile multiple files and each command is isolated from the other.
2. Pass it as argument - Well, here you will have to make sure that we test each operation properly and there are very good chances of making a mistake, so very strong testing is needed with this approach.

Its okay to follow 2, I don't have any issues.

czgdp1807:
> It might be that we need to instantiate the LLVMUtils as a regular instance inside the LLVM backend class. 

`LLVMList`, `LLVMUtils`, `LLVMDict` all are already regular instances inside LLVM backend class. See the code snippet below,

https://github.com/lcompilers/lpython/blob/91cb772ef86a58d8b5020a8dde864754bc06d06a/src/libasr/codegen/asr_to_llvm.cpp#L240-L244

> The mistake seems to be coming from using LLVMUtils as a singleton, which now needs state. We should not add a (global) state. But making LLVMUtils local instance, we can then store a state.

`LLVMList`, `LLVMUtils`, `LLVMDict` are not singleton classes. Their constructors are public and you can create multiple objects for these classes. See code snippets below (all the constructors are public and they don't follow singleton pattern anywhere),

https://github.com/lcompilers/lpython/blob/91cb772ef86a58d8b5020a8dde864754bc06d06a/src/libasr/codegen/llvm_utils.h#L129-L133

https://github.com/lcompilers/lpython/blob/91cb772ef86a58d8b5020a8dde864754bc06d06a/src/libasr/codegen/llvm_utils.h#L201-L206

https://github.com/lcompilers/lpython/blob/91cb772ef86a58d8b5020a8dde864754bc06d06a/src/libasr/codegen/llvm_utils.h#L247-L254


> * We should not add a (global) state. But making LLVMUtils local instance, we can then store a state.

Correct, that's what I proposed in Case 2 of [this comment](https://github.com/lcompilers/lpython/pull/1093#discussion_r979144269). In fact static variable is good for multiprocessing based parallel compilation as you apply the same restriction uniformly to all the LLVMList objects (you can create multiple objects for any class defined in `llvm_utils.h`). For multithreading based parallel compilation you can make the same variable a data member of the object which will localise things to an object and since an object is available to its own thread no other thread will be to overwrite it.

src/libasr/codegen/asr_to_llvm.cpp
certik: We want to make sure we do not close a door to single process multiple threads. It seems static variables would close the door, so let's not do that?

certik:
That's good that LLVMUtils is a local object. In which case you can store state in it if needed, and not use static variables. We do not want to close the door to multiple threads, where each wants to apply different setting, and it seems static variables would close the door.

czgdp1807:
Correct. Using a data member is fine because there too within a thread you have to enable the option only once and then call the list API functions as it is without passing any option. Lesser mistakes and more uniformity will enhance user and developer experience of LPython.

certik:
static members are shared across threads (https://en.cppreference.com/w/cpp/language/static). Even within one thread, if you instantiate LLVMUtils twice (not realizing it has a global state via the static members), it will break. It's better not to use static variables, nor any kind of global state.

czgdp1807:
Yes I agree. I proposed static variable approach because I thought we will be following multi-processing based approach for parallel compilation. But in multithreading setting using objects local to a function call (and each function call is made in a separate thread) is the best approach. We will avoid locks and all thread safety issues that way.

src/libasr/codegen/asr_to_llvm.cpp
czgdp1807: > It seems static variables would close the door, so let's not do that?

I agree with you. I am opening a new issue for proposing these guidelines and we should include it in README.md so that new developers know that we are open to multithreading based parallel compilation. I feel we will be able to achieve sandboxing and safety in this setting as well.

czgdp1807: In fact multi-threading based approach will avoid all kinds of dependencies on how fast the build system is because we will parallelise the build on our own. So that's good thing I feel.

certik (approved):
I think the PR looks good to me.

We just need to test this, so let's create a testing infrastructure for runtime tests in a separate PR, get it in, and then we can use it here.

* [x] https://github.com/lcompilers/lpython/issues/1128

Thirumalai-Shaktivel:
> I think its default value should be `true` according to LPython contract defined in #1139 (see quotation below). However if I enable `--fast` option then I think this should be `false` by default. Note that I think at the end user's choices should be valued the most. If they enable `--fast --enable-bounds-checking` then `compiler_options.enable_bounds_checking` should be `true`. See the quotation below,
> 
> > The default option in LPython should be Debug mode. The Debug mode must have all runtime safety options on. In this Debug mode, you get very fast compilation, and safe runtime behavior. No silent surprises. Beautiful runtime error messages with stack traces. If you don't get any runtime error and things compiles (obviously), then it will behave just like in CPython.

I'm applying this approach for `enable-bounds-checking`.

tests/reference/runtime-test_list_01-3ee9b3e.stderr
```diff
@@ -1 +1 @@
-ValueError: The list does not contain the element
+sh: 1: ./a.out: Text file busy
```
Thirumalai-Shaktivel, line 1: This happens when I run `./run_tests,py -u`.
So this is happening because of: https://stackoverflow.com/questions/16764946/what-generates-the-text-file-busy-message-in-unix
So, I think for every file compilation we should generate a unique executable (but not `a.out`). And after printing the output, I think we should delete the executable?
cc @certik What are your thoughts?

Thirumalai-Shaktivel:
There is some problem with the `run` option in run_tests.py
So, here is the output generated by hand:
```console
$ lpython tests/runtime_errors/test_list_01.py
ValueError: The list does not contain the element
$ lpython tests/runtime_errors/test_list_02.py
IndexError: list index out of range
```
But, When I update the tests using `run_tests.py -u`. Each time it gets different output in *.stderr.
I thought this was caused by writing both the output to "a.out". So, I used function to create a unique filename for each executable. But, still the problem persists.

certik:
We have to debug it.

Thirumalai-Shaktivel:
As can be seen from the updated tests, the same output (`ERROR STOP`) gets updated for almost all the files in the runtime_errors.

Thirumalai-Shaktivel:
This PR is ready to go in!
@certik @czgdp1807 kindly review this and provide feedback.

src/libasr/codegen/llvm_utils.h
```diff
@@ -170,28 +170,32 @@ namespace LFortran {
...
             llvm::Value* len(llvm::Value* list);
 
-            void check_index_within_bounds(llvm::Value* list, llvm::Value* pos);
+            void check_index_within_bounds(llvm::Value* list, llvm::Value* pos,
+                                           llvm::Module& module);
 
             void write_item(llvm::Value* list, llvm::Value* pos,
                             llvm::Value* item, ASR::ttype_t* asr_type,
-                            llvm::Module& module, bool check_index_bound=true);
+                            bool enable_bounds_checking, llvm::Module& module);
```
Thirumalai-Shaktivel, line 183: I think, there will be no need for bounds_checking in write_item.
@czgdp1807 What do you think?

certik (approved):
I have not reviewed very carefully, but at first glance it looks ok.

@czgdp1807 can you please review carefully and merge if it is ok?

Thirumalai-Shaktivel:
ping @certik @czgdp1807

tests/reference/runtime-test_list_02-5f7db5f.stderr
```diff
@@ -0,0 +1 @@
+IndexError: list index out of range
```
czgdp1807, line 1: Which index is out of range and what's the range? Useful information to add. I think its possible, we just need to modify the arguments to print call, correct?

czgdp1807 (changes requested):

tests/reference/runtime-test_list_02-5f7db5f.stderr
Thirumalai-Shaktivel: Yup, that's possible! I will give it a try in this PR, if possible
Otherwise, I will to submit a PR for that after completing this issue: #772.

tests/reference/runtime-test_list_02-5f7db5f.stderr
```diff
@@ -0,0 +1,2 @@
+IndexError: List index is out of range.
+Allocated index: 0:2, but given: 3
```
Thirumalai-Shaktivel, line 2: Does the error message look good or understandable?
Do you have other suggestions?
@certik @czgdp1807

tests/reference/runtime-test_list_01-3ee9b3e.stderr
```diff
@@ -1 +1 @@
-ValueError: The list does not contain the element
+ValueError: The list does not contain the element: 4
```
Thirumalai-Shaktivel, line 1: ..

tests/reference/runtime-test_list_02-5f7db5f.stderr
Thirumalai-Shaktivel: Done!

certik:
We should use the ASR node itself to pass information about the name of the array or list that is out of bounds and also store the line/column number with the error message.

certik (approved):
I think this looks good.

tests/reference/runtime-test_list_02-5f7db5f.stderr
czgdp1807: For me following is better,

`IndexError: List index is out of range. Index range is (0, 2) but given index is 3.`

Thirumalai-Shaktivel: Done!

czgdp1807:
Please make sure that this is only available in ReleaseSafe mode. Performance modes of LPython should have this check disabled.

czgdp1807 (approved):

src/bin/lpython.cpp
```diff
@@ -1328,6 +1325,16 @@ int main(int argc, char *argv[])
...
 
+        if( compiler_options.fast && compiler_options.enable_bounds_checking ) {
+        // ReleaseSafe Mode
+        } else if ( compiler_options.fast ) {
+        // Release Mode
+            lpython_pass_manager.use_optimization_passes();
+        } else {
+        // Debug Mode
+            compiler_options.enable_bounds_checking = true;
+        }
```
Thirumalai-Shaktivel, line 1336: Yup, I keep track of it using this check.

Thirumalai-Shaktivel: >  Performance modes of LPython should have this check disabled.

By `Performance modes`, you meant Release mode, right?

Thirumalai-Shaktivel:
Thanks for the approval, Let's merge this PR.

tests/reference/runtime-test_list_01-3ee9b3e.stderr
Thirumalai-Shaktivel: This issue was fixed by writing the executable into a new unique file, say `filename.out`.

Issue #1077, opened 2022-09-02
``list`` in C++/C backend


Smit-create:
I was working on #210 and saw some test failures in CPP backend because of the List being not supported in CPP codegen. I can try adding their support and wanted to discuss one of the possibilities. 
1. We can directly use CPP's `std::vector`, which provides almost all the operations as CPython's list like insert, remove, accessing element at any position, etc.

Any other suggestions for the same?
@certik @czgdp1807

czgdp1807:
`std::vector` is slow. I would say going along the same path of LLVM would be better.

czgdp1807:
I am self-assigning this. Here's how I will do it,

1. Generate C/C++ code for lists only if lists are actually used (i.e., present in the ASR).
2. There are two ways to generate C/C++ code for lists,

i. Follow the LLVM logic which keeps everything inlined. Like when you call append in LLVM then the code is generated at that spot and no function call is made to fulfil the purpose. The benefit of this approach is that we don't depend on C/C++ compiler to inline code for us, we are already doing it. The downside is that folks who want to re-use our generated C/C++ code won't be able to edit that easily, because everything is inlined.
ii. If a list is present then generate a struct and all the utility functions for lists associated with it. And just call those functions where-ever `ListAppend`, `ListRemove`, etc nodes are present. The benefit is that users will be able to edit our generated C/C++ code because we provide re-usable implementation of list. The downside is that we depend on C/C++ compiler to inline our code for lists.

IMO, approach ii should be good enough as it provides the option of re-usability. If the user want higher performance then they can go for LLVM backend directly. In addition in future if user provides `--fast` flag with `--show-c` flag then we can implement approach i as well. What do you say @certik ?

ArpanShah2k:
I am interested in fixing this issue . It will be a great opportunity if you allot me to solve this issue . @czgdp1807 @Smit-create @certik @rebcabin @HaoZeke

czgdp1807:
@ArpanShah2k I am already working on this in #1095. Thanks for your interest though. Please let us know if you are interested in any other issue.

certik:
@ArpanShah2k you don't need to ask for permissions, simply pick an issue and implement it, open a PR and go from there. It's very unlikely that somebody works on exactly the same thing at the same time (hour or minute), and in the very unlikely even that somebody does, it's easy to simply join forces and merge the PRs.

PR #1095, opened 2022-09-07
List in C/C++ backend

czgdp1807:
Implementing 2.ii in https://github.com/lcompilers/lpython/issues/1077#issuecomment-1236134901.

Closes https://github.com/lcompilers/lpython/issues/1077

Benchmark - https://github.com/lcompilers/lpython/issues/857#issuecomment-1200960150 (see `list04` row). Our generated C code (with no optimisation flags enabled for C compiler) is faster than un-optimised (no -O3 flag) executable generated `clang++`.


czgdp1807:
@certik This is ready to review. Feel free to review the diff in total and let me know if any change is to be made. Otherwise I will clean up the git history and merge.

certik (approved):
Thanks for implementing it!

I think this is fine to merge.

As an improvement in subsequent PRs, the C++ version should just use `std::vector` for lists, as that is the natural data structure that people would use in C++, pretty much equivalent to our list in ASR.

czgdp1807:
> As an improvement in subsequent PRs, the C++ version should just use `std::vector` for lists, as that is the natural data structure that people would use in C++, pretty much equivalent to our list in ASR.

I think we should expose this with something like `--use-stl` for C++ backend. The scene is that our implementation is very competitive and light weight (no templates because we already know the types so we directly create the correctly typed lists and hence we don't depend on templating support of C++). So, it should be left to the user to decide whether they want to use our implementations for container data structures or STL implementations.

certik:
Yes, making it configurable is even better.

certik:
How do the list benchmarks behave? Are we now getting the same performance in C/C++ as in LLVM, and if not, why not?

czgdp1807:
I don’t know how to enable -O3 optimisation for C compiler in integration tests. Please let me know and I will share the benchmark data with you.

certik:
I would execute the benchmarks by hand. Just compile it by hand and enable all optimization options.

PR #1096, opened 2022-09-07
Ignore forward declarations of ``ccall`` functions in C backend

### LINKED ISSUE #1094: Declare `@ccall` only when asked by the user
So as of now any function declared with `@ccall` is declared always in the generated code. However if someone has already declared their function in some other `.h` file then function decorated `@ccall` shouldn't be re-declared in the generated code. So we have to make the declaration of `@ccall` functions optional. Some options,

1. `ccall(static=true)`.
2. `ccall(declare=false)`.

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1094

### DIFF
--- a/integration_tests/bindc_03.py
+++ b/integration_tests/bindc_03.py
@@ -5,11 +5,11 @@ def g(a: CPtr, value: i32) -> None:
     pass
 
 @ccall
-def get_array(size: i32) -> CPtr:
+def get_array(size: i32) -> CPtr: # this function can be called from Python but is defined in C code
     pass
 
 @ccallable
-def f(q_void: CPtr) -> None:
+def f(q_void: CPtr) -> None: # this function can be called from C code
     i: i32
     el: i32
     q: Pointer[i32[:]]
--- a/src/libasr/codegen/asr_to_c_cpp.h
+++ b/src/libasr/codegen/asr_to_c_cpp.h
@@ -336,10 +336,29 @@ R"(#include <stdio.h>
 
     std::string declare_all_functions(const SymbolTable &scope) {
         std::string code;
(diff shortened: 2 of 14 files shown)


certik:
It's fine as a quick fix.

What is the logic / use case how this will be used?

It seems this should be made configurable using some compiler option. Even better: why do we need them in the first place? Well, they are needed because gcc will not compile it if we call a function that is declared after it is called. It seems that is the way in C, you have to forward declare them. However, it also seems some C compilers do not support forward declarations, in which case, what is the proper way to support those C compilers?

czgdp1807:
@certik So should we merge this?

czgdp1807:
@certik Can we merge this?

certik:
I would like to know the answers to my question above (https://github.com/lcompilers/lpython/pull/1096#issuecomment-1239530894).

One way to give an answer is to test this by actually compiling some code with some special (?) gcc options that require it.

czgdp1807:
> Even better: why do we need them in the first place? Well, they are needed because gcc will not compile it if we call a function that is declared after it is called.

AFAICT, it is needed because **some C compilers** raise error when we forward declare a function which is already present in an included header.

> One way to give an answer is to test this by actually compiling some code with some special (?) gcc options that require it.

GCC by default doesn't raise any error on re-declaring the same function which is already present in an included header. It is that some specific compiler raise error when we re-declare a function. which is already present in an included header.

czgdp1807:
So I and @certik figured that instead of the current workaround in this PR we should go ahead with emitting the functions in the right order. See https://github.com/lfortran/lfortran/issues/802#issuecomment-1263994992 for more details.

PR #1097, opened 2022-09-07
Using all proccessors for building (in build1.sh script)

ronnuriel:

### DIFF
--- a/build1.sh
+++ b/build1.sh
@@ -13,4 +13,4 @@ cmake \
     -DCMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH_LPYTHON;$CONDA_PREFIX" \
     -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
     .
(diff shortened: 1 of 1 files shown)


czgdp1807 (approved):
I am +1 on this change.

certik (approved):

certik:
This change is fine with me, but note that this does *not* mean "all processors", but rather it means maximum threads allowed by the source file dependencies. So the source code allows building 200 files in parallel, it will launch 200 parallel jobs, which on a small laptop with just a few cores (say 4) might be an issue.

In practice, for me, `-j16` gives pretty much the most parallelism anyway, and it will not break machines with smaller number of cores.

czgdp1807:
This can be closed I think? Manually specifying `-j8` or `-j5` or `-j16` should be fine. Using `-j` has the potential to stall someone's system.

certik:
I think we can close it for now.

PR #1098, opened 2022-09-07
Add String to List cast in LLVM

Smit-create:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -4769,6 +4769,73 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                builder->CreateCondBr(cond, loopbody, loopend);
+
+                // body
+                start_new_block(loopbody);
+                llvm::Value* item = lfortran_str_copy(str, ptr, ptr);
+                list_api->append(const_list, item, list_type->m_type, *module);
+                ptr = builder->CreateAdd(ptr, i32_one);
+
+                builder->CreateBr(loophead);
+
```
Smit-create, line 4825: This does work on the following example:
```py
def f():
    x: list[str]
    s: str
    s = "lpython"
    x = list(s)
    print(x[0])

f()
```
It generates LLVM code but goes into an infinite loop. Any suggestions for LLVM bug that may be visible here?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -4769,6 +4769,73 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                // head
+                start_new_block(loophead);
+                llvm::Value *cond = builder->CreateICmpSLT(ptr, str_len);
+                builder->CreateCondBr(cond, loopbody, loopend);
+
+                // body
+                start_new_block(loopbody);
+                llvm::Value* item = lfortran_str_copy(str, ptr, ptr);
+                list_api->append(const_list, item, list_type->m_type, *module);
+                ptr = builder->CreateAdd(ptr, i32_one);
```
Smit-create, line 4822: I tried printing the value of `ptr`, and it seems `ptr` is `0` and doesn't get updated here. Any suggestion to fix this?

src/libasr/codegen/asr_to_llvm.cpp
Thirumalai-Shaktivel: I faced the same issue, while implementing the `List remove`.

Thirumalai-Shaktivel: I think you have to Store and Load the value from and to ptr.

Thirumalai-Shaktivel: Do this, Create a new variable of type Alloca, and store the ptr value in that. Later use Store to assign value, Load to acces the value. Maybe that would help you get out of infinite loop.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -4769,6 +4769,73 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+
+                llvm::BasicBlock *loophead = llvm::BasicBlock::Create(context, "loop.head");
+                llvm::BasicBlock *loopbody = llvm::BasicBlock::Create(context, "loop.body");
+                llvm::BasicBlock *loopend = llvm::BasicBlock::Create(context, "loop.end");
+                this->current_loophead = loophead;
+                this->current_loopend = loopend;
+
+                // head
+                start_new_block(loophead);
+                llvm::Value *cond = builder->CreateICmpSLT(ptr, str_len);
```
Thirumalai-Shaktivel, line 4815: @Smit-create, I don't understand properly, Here the initial value of ptr is `"_character_to_list_itr"`?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -4769,6 +4769,73 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 tmp = builder->CreateSelect(cmp, zero_str, one_str);
                 break;
             }
+            case (ASR::cast_kindType::CharacterToList) : {
+                this->visit_expr_wrapper(x.m_arg, true);
```
Thirumalai-Shaktivel, line 4773: I think this line is not required. At the start of the switch case (`Line: 4520`), it's already been called, so you can simply use `tmp` to get the value.

Thirumalai-Shaktivel:
Applying this diff, made things to work, but the ouput
```diff
diff --git a/src/libasr/codegen/asr_to_llvm.cpp b/src/libasr/codegen/asr_to_llvm.cpp
index be298fb24..9a5ea3488 100644
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -4773,19 +4773,17 @@ public:
                 break;
             }
             case (ASR::cast_kindType::CharacterToList) : {
-                this->visit_expr_wrapper(x.m_arg, true);
                 llvm::Value *str = CreateLoad(tmp);
                 llvm::AllocaInst *parg = builder->CreateAlloca(character_type, nullptr);
-                builder->CreateStore(str, parg);
+                LLVM::CreateStore(*builder, str, parg);
                 llvm::Value *str_len = lfortran_str_len(parg);
+
                 llvm::Type *type = getIntType(4);
-                llvm::Value *ptr = module->getOrInsertGlobal("_character_to_list_itr",
-                    type);
-                ptr = builder->CreateLoad(ptr);
-                module->getNamedGlobal("_character_to_list_itr")->setInitializer(
-                        llvm::ConstantInt::get(context,
-                            llvm::APInt(32, 0)));
+                llvm::AllocaInst *ptr = builder->CreateAlloca(type, nullptr);
                 llvm::Value* i32_one = llvm::ConstantInt::get(context, llvm::APInt(32, 1));
+                llvm::Value* i32_zero = llvm::ConstantInt::get(context, llvm::APInt(32, 0));
+                LLVM::CreateStore(*builder, i32_zero, ptr);
 
                 ASR::List_t* list_type = ASR::down_cast<ASR::List_t>(x.m_type);
                 bool is_array_type_local = false, is_malloc_array_type_local = false;
@@ -4815,14 +4813,16 @@ public:
 
                 // head
                 start_new_block(loophead);
-                llvm::Value *cond = builder->CreateICmpSLT(ptr, str_len);
+                llvm::Value *cond = builder->CreateICmpSLT(LLVM::CreateLoad(*builder, ptr), str_len);
                 builder->CreateCondBr(cond, loopbody, loopend);
 
                 // body
                 start_new_block(loopbody);
-                llvm::Value* item = lfortran_str_copy(str, ptr, ptr);
+                llvm::Value* item = lfortran_str_copy(str, LLVM::CreateLoad(*builder, ptr), LLVM::CreateLoad(*builder, ptr));
                 list_api->append(const_list, item, list_type->m_type, *module);
-                ptr = builder->CreateAdd(ptr, i32_one);
+                LLVM::CreateStore(*builder, builder->CreateAdd(LLVM::CreateLoad(*builder, ptr), i32_one), ptr);
 
                 builder->CreateBr(loophead);
```
```py
def f():
    x: list[str]
    s: str
    s = "lpython"
    x = list(s)
    i:i32
    for i in range(len(x)):
        print(x[i], end=", ")
    print()

f()
```
```console
$ lpython examples/expr2.py && ./a.out
n, l, p, y, t, h, o,
```

Thirumalai-Shaktivel:
I figured it out, the string index starts from 1 to len(str). But I used 0 to len(str)
The following diff works, @Smit-create you can make the changes as you wished. Any doubts please ping me.
```diff
diff --git a/src/libasr/codegen/asr_to_llvm.cpp b/src/libasr/codegen/asr_to_llvm.cpp
index be298fb24..9989e4b07 100644
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -4773,20 +4773,18 @@ public:
                 break;
             }
             case (ASR::cast_kindType::CharacterToList) : {
-                this->visit_expr_wrapper(x.m_arg, true);
                 llvm::Value *str = CreateLoad(tmp);
+                // Store string into the register
                 llvm::AllocaInst *parg = builder->CreateAlloca(character_type, nullptr);
-                builder->CreateStore(str, parg);
+                LLVM::CreateStore(*builder, str, parg);
                 llvm::Value *str_len = lfortran_str_len(parg);
+
                 llvm::Type *type = getIntType(4);
-                llvm::Value *ptr = module->getOrInsertGlobal("_character_to_list_itr",
-                    type);
-                ptr = builder->CreateLoad(ptr);
-                module->getNamedGlobal("_character_to_list_itr")->setInitializer(
-                        llvm::ConstantInt::get(context,
-                            llvm::APInt(32, 0)));
+                llvm::AllocaInst *ptr = builder->CreateAlloca(type, nullptr);
                 llvm::Value* i32_one = llvm::ConstantInt::get(context, llvm::APInt(32, 1));
+                LLVM::CreateStore(*builder, i32_one, ptr);
 
+                // Empty List Initialization
                 ASR::List_t* list_type = ASR::down_cast<ASR::List_t>(x.m_type);
                 bool is_array_type_local = false, is_malloc_array_type_local = false;
                 bool is_list_local = false;
@@ -4805,8 +4803,8 @@ public:
                 uint64_t ptr_loads_copy = ptr_loads;
                 ptr_loads = 1;
 
+                // Looping statement
                 dict_api->set_iterators();
-
                 llvm::BasicBlock *loophead = llvm::BasicBlock::Create(context, "loop.head");
                 llvm::BasicBlock *loopbody = llvm::BasicBlock::Create(context, "loop.body");
                 llvm::BasicBlock *loopend = llvm::BasicBlock::Create(context, "loop.end");
@@ -4815,15 +4813,16 @@ public:
 
                 // head
                 start_new_block(loophead);
-                llvm::Value *cond = builder->CreateICmpSLT(ptr, str_len);
+                llvm::Value *cond = builder->CreateICmpSLE(LLVM::CreateLoad(*builder, ptr), str_len);
                 builder->CreateCondBr(cond, loopbody, loopend);
 
                 // body
                 start_new_block(loopbody);
-                llvm::Value* item = lfortran_str_copy(str, ptr, ptr);
-                list_api->append(const_list, item, list_type->m_type, *module);
-                ptr = builder->CreateAdd(ptr, i32_one);
-
+                {
+                    llvm::Value* item = lfortran_str_copy(str, LLVM::CreateLoad(*builder, ptr), LLVM::CreateLoad(*builder, ptr));
+                    list_api->append(const_list, item, list_type->m_type, *module);
+                    LLVM::CreateStore(*builder, builder->CreateAdd(LLVM::CreateLoad(*builder, ptr), i32_one), ptr);
+                }
                 builder->CreateBr(loophead);
 
                 // end
```

Smit-create:
Thanks, that worked.

Thirumalai-Shaktivel (approved):
LGTM!

integration_tests/str_to_list_cast.py
```diff
@@ -10,5 +10,14 @@ def f():
     assert len(x) == 0
     x = list("L")
     assert len(x) == 1 and x[0] == 'L'
+    x = list(s)
+    assert len(x) == len(s)
+    for i in range(len(x)):
+        assert x[i] == s[i]
+    s = "agowietg348203wk.smg.afejwp398273wd.a,23to0MEG.F,"
```
czgdp1807, line 17: Create a loop and do `s += str(i)` then do `x = list(s)` in the next line.

czgdp1807 (changes requested):

czgdp1807:
Correct me if I am wrong, so string to list cast can be implemented as follows at the Python's language level,

```python
l: list[str] = []
for char in string:
    l.append(char)
```

If so then I think this cast should be handled at ASR pass level so that it can be shared across all the backends at once.

Smit-create:
> Correct me if I am wrong, so string to list cast can be implemented as follows at the Python's language level,
> 
> ```python
> l: list[str] = []
> for char in string:
>     l.append(char)
> ```
> 
> If so then I think this cast should be handled at ASR pass level so that it can be shared across all the backends at once.

I tried to implement this, but I have a question now how we will make it to return `l`?

certik:
I think @czgdp1807 is probably right. So effectively this:
```python
l: list[str] = list(string)
```
is equivalent to this:
```python
l: list[str] = []
for char in string:
    l.append(char)
```
But in terms of performance, we might (later) do something like:

```python
l: list[str]
l.reserve(len(string))
for i in range(len(string)):
    l[i] = string[i]
```
If it is faster but we'll get to these optimizations later. Probably our (future) optimizer might be able to do such optimizations also.

How to return `l`?

Here is one approach: turn `l: list[str] = list(string)` into `l: list[str] = _lcompilers_string_to_list(string)` and implement a function (even by hand in our runtime library):
```python
def _lcompilers_string_to_list(string: str) -> list[str]:
    l: list[str] = []
    for char in string:
        l.append(char)
    return l
```
And make the frontend AST->ASR insert it. That's all that is needed I think.

Smit-create:
I have updated the string to list cast in the ASR frontend. Please review.

czgdp1807:
> I think @czgdp1807 is probably right. So effectively this:
> 
> ```python
> l: list[str] = list(string)
> ```
> 
> is equivalent to this:
> 
> ```python
> l: list[str] = []
> for char in string:
>     l.append(char)
> ```
> 
> But in terms of performance, we might (later) do something like:
> 
> ```python
> l: list[str]
> l.reserve(len(string))
> for i in range(len(string)):
>     l[i] = string[i]
> ```
> 
> If it is faster but we'll get to these optimizations later. Probably our (future) optimizer might be able to do such optimizations also.
> 
> How to return `l`?
> 
> Here is one approach: turn `l: list[str] = list(string)` into `l: list[str] = _lcompilers_string_to_list(string)` and implement a function (even by hand in our runtime library):
> 
> ```python
> def _lcompilers_string_to_list(string: str) -> list[str]:
>     l: list[str] = []
>     for char in string:
>         l.append(char)
>     return l
> ```
> 
> And make the frontend AST->ASR insert it. That's all that is needed I think. (So you don't even need to create a new pass.)

I would prefer this approach over others.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -849,8 +849,21 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
         }
+
+        if (call_name == "list" && (args.size() == 0 ||  args[0].m_value == nullptr)) {
+            if (ann_assign_target_type) {
+                ASR::ttype_t *type = ASRUtils::get_contained_type(ann_assign_target_type);
+                ASR::ttype_t* list_type = ASRUtils::TYPE(ASR::make_List_t(al, loc, type));
+                Vec<ASR::expr_t*> list;
+                list.reserve(al, 1);
+                return ASR::make_ListConstant_t(al, loc, list.p,
+                    list.size(), list_type);
```
certik, line 862: Is this creating an empty list constant?

src/lpython/semantics/python_comptime_eval.h
```diff
@@ -371,6 +372,47 @@ struct PythonIntrinsicProcedures {
...
+                }
+                return ASR::down_cast<ASR::expr_t>(ASR::make_ListConstant_t(al, loc, list.p,
+                    list.size(), list_type));
+            }
+        } else {
+            throw SemanticError("'" + ASRUtils::type_to_str_python(type) +
+                "' object conversion to List is not implemented ",
+                arg->base.loc);
+        }
+    }
```
certik, line 414: This is handling compile time `list("abcd")` stuff? If so, we need to add tests for that. The above tests seem to test runtime conversion, which is not implemented in this PR yet (I think).

src/runtime/lpython_builtin.py
```diff
@@ -732,3 +732,12 @@ def _lpython_str_startswith(s: str ,sub: str) -> bool:
...
     return res
+
+
+def list(s: str) -> list[str]:
+    l: list[str]
+    l = []
+    i: i32
+    for i in range(len(s)):
+        l.append(s[i])
+    return l
```
certik, line 743: Ah I see, it's doing the runtime as well. So we should test both compile time and runtime.

For runtime, should we rename this function to `_lcompilers_str_to_list`, just to make it clear that this is an internal function? 

CC @czgdp1807.

certik:
> I would prefer this approach over others.

I think it is doing that approach, but there are a few things to polish, see above.

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: Yes

src/lpython/semantics/python_comptime_eval.h
Smit-create: Runtime also works, those tests are present in the same file from line 1-10.

src/runtime/lpython_builtin.py
Smit-create: I think `list` is more approriate because it is a builtin function or intrinsic in python much like `ord`, `hex`, `bin` etc. and we can handle it the same way.

certik (approved):
I think this is good to go in. Everything is needed in this PR.

PR #1099, opened 2022-09-08
Implement negative indexing in Lists

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
I made these changes to implement negative index, but I get the following error:
```py
Instruction does not dominate all uses!
  %35 = add i32 %34, -1
  call void (i8*, ...) @_lfortran_printf(i8* getelementptr inbounds ([11 x i8], [11 x i8]* @3, i32 0, i32 0), i32 %35)
Instruction does not dominate all uses!
  %35 = add i32 %34, -1
  %38 = getelementptr inbounds i32, i32* %37, i32 %35
```
I don't know where I made wrong.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1245,11 +1245,37 @@ namespace LFortran {
...
+        llvm::Function *fn = builder->GetInsertBlock()->getParent();
+        llvm::BasicBlock *thenBB = llvm::BasicBlock::Create(context, "then", fn);
+        llvm::BasicBlock *elseBB = llvm::BasicBlock::Create(context, "else");
+        llvm::BasicBlock *mergeBB = llvm::BasicBlock::Create(context, "ifcont");
+
+        llvm::Value* cond = builder->CreateICmpSGT(zero, pos);
+        builder->CreateCondBr(cond, thenBB, elseBB);
+        builder->SetInsertPoint(thenBB);
+        {
+            pos = builder->CreateAdd(len, pos);
```
czgdp1807, line 1262: This happens because you are storing into a register conditionally. Instead create a temporary variable outside if-else block and store pos in it. Then load temporary variable inside if and else separately, update it and then store back into the temporary variable.

Thirumalai-Shaktivel:
This PR is ready to go in!

czgdp1807:
Try running the benchmark with #857.

Thirumalai-Shaktivel:
You mean, I should try the negative index, right?

Thirumalai-Shaktivel:
Yup, I got it!

Thirumalai-Shaktivel:
This PR has increased the time!!!!
Compiler | Time [s]
-- | -- | --
lpython --fast | 0.517
lpython | 0.838

Thirumalai-Shaktivel:
What to do now?

czgdp1807:
Yes. There can be two reasons. One, each index depends on `if-else` check. Branching reduces throughput of instructions executed by the processer. What you can do is `(index + len) % len` which translates to `builder->CreateSRem(builder->CreateAdd(index, len), len);`. Second, we are creating a temporary variable for every index so that causes `alloca` to be called for each instruction. I suspect first playes a major role but fixing second will also help in avoiding segmentation fault for very large number of list indexing.

Thirumalai-Shaktivel:
Compiler | Time [s]
-- | --
lpython --fast | 0.245
lpython | 0.400

czgdp1807:
Please update https://github.com/lcompilers/lpython/issues/857#issuecomment-1201061763 as well.

Thirumalai-Shaktivel:
Examples:
```py
from ltypes import i32

def test_insert():
    x: list[i32] = []
    x.insert(len(x), 2)
    i: i32
    for i in range(len(x)):
        print(x[i])

def test_remove():
    x: list[i32] = []
    x.remove(2)
    i: i32
    for i in range(len(x)):
        print(x[i])


test_insert()
test_remove()
```

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1245,11 +1245,26 @@ namespace LFortran {
...
+    llvm::Value* LLVMList::get_list_pos(llvm::Value* list, llvm::Value* pos) {
+        llvm::Value* list_len = LLVMList::len(list);
+
+        llvm::Type* type = llvm::Type::getInt32Ty(context);
+        llvm::AllocaInst *tmp_pos = builder->CreateAlloca(type, nullptr);
+        llvm::Value* zero = llvm::ConstantInt::get(llvm::Type::getInt32Ty(context),
+                                            llvm::APInt(32, 0));
+        LLVM::CreateStore(*builder,
+            builder->CreateSRem(builder->CreateAdd(pos, list_len), list_len),
+            tmp_pos);
```
certik, line 1257: In here we do `(pos+list_len) % list_len`. That is much slower than just `pos`. So I think we should not allow this in ASR, I think we should always assume in `ListItem` has index between `0` and `size-1`.

In Debug mode we'll add a runtime check for this.

For things like `x[-1]` or `x[-5]` the frontend will create ASR of the form `x[len(x)-1]` and `x[len(x)-5]`. So that would still work.

Finally, for `x[i]` where `i=-5` I think we should simply not allow such usage. I think I might have never used this in practice anyway, so our (future) runtime Debug check would simply give an error message, and the user would have to fix the code, in this case `x[i]` -> `x[len(x)+i]` if `i < 0`.

If the user absolutely requires this usage, he can do something like `x[(i+len(x)) % len(x)]`, that way the behavior is as intended and it is explicit what is going on, and (most importantly!) we are not losing any performance.

integration_tests/test_list_01.py
```diff
@@ -42,8 +42,22 @@ def test_list_02():
...
+    x: list[f64] = []
+    size: i32 = 0
+
+    i: i32
+    for i in range(100):
+        x.append((i * size)/2)
+        size = len(x)
+
+    for i in range(size):
+        assert x[i] == x[i-size]
```
certik, line 56:
```suggestion
        assert x[i] == x[((i-size) + len(x)) % len(x)]
```

src/libasr/codegen/llvm_utils.cpp
czgdp1807: Correct. Note that figuring out whether a variable is negative or not is a runtime operation. For some cases you can figure out in compile time (like for constant negative integers as cited above). Now the problem is allowing negative indexing as `x[-5]` but not via `x[i]` where `i = -5` is simply a source of confusion. For example I will be able to do the following,

```python
if flag:
    x[-5] = 0
else:
    x[-20] = 0
```

but not the following,

```python
if flag:
    i = -5
else:
    i = -20
x[i] = 0
```

Even though both codes are semantically mean the same. I would say add an option to allow negative indexing and generate code in backends accordingly. Explain to the users in the documentation that how negative indexing degrades performance, hence we don't support it by default.

Either allow negative indexing completely or don't allow it.

Thirumalai-Shaktivel:
ASR:
```py
(ListItem
    (Var 2 x)
    (IntegerBinOp
        (ListLen
            (Var 2 x)
            (Integer 4 [])
            ()
        )
        Sub
        (IntegerConstant 4 (Integer 4 []))
        (Integer 4 [])
        ()
    )
    (Real 8 [])
    ()
)
```

integration_tests/test_list_01.py
```diff
@@ -42,8 +42,28 @@ def test_list_02():
...
+    assert x[0] == x[-2]
+
+    size: i32 = 2
+    for i in range(100):
+        x.append((i * size)/2)
+        size = len(x)
+
+    for i in range(size):
+        assert x[i] == x[((i-len(x)) + size) % size]
+        assert x[i] == x[-(len(x)-i)]
```
certik, line 62:
```suggestion
```

integration_tests/test_list_01.py
```diff
@@ -42,8 +42,28 @@ def test_list_02():
...
+# Negative Indexing
+def test_list_03():
+    x: list[f64] = []
+
+    i: i32
+    for i in range(2):
+        x.append(float(i))
+
+    assert x[0] == x[-2]
+
```
certik, line 54:
```suggestion
    assert x[1] == x[-1]
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2349,6 +2349,16 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
 
             } else if (ASR::is_a<ASR::List_t>(*type)) {
                 index = ASRUtils::EXPR(tmp);
+                if(ASR::is_a<ASR::IntegerUnaryMinus_t>(*index)) {
```
certik, line 2352:
```suggestion
                if(expr_value(index) && ASR::is_a<ASR::IntegerConstant_t>(*expr_value(index))) {
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2349,6 +2349,16 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                    // Replace `x[-1]` to `x[len(x)-1]`
+                    ASR::ttype_t *int_type = ASRUtils::TYPE(ASR::make_Integer_t(
+                                                    al, loc, 4, nullptr, 0));
+                    ASR::expr_t *list_len = ASRUtils::EXPR(ASR::make_ListLen_t(
+                                al, loc, value, int_type, nullptr));
+                    ASR::expr_t *neg_idx = ASR::down_cast<ASR::IntegerUnaryMinus_t>(index)->m_arg;
```
certik, line 2358:
```suggestion
                    ASR::expr_t *neg_idx = expr_value(index);
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2349,6 +2349,16 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                    index = ASRUtils::EXPR(ASR::make_IntegerBinOp_t(al, loc,
+                        list_len, ASR::binopType::Sub, neg_idx, int_type, nullptr));
```
certik, line 2360:
```suggestion
                        list_len, ASR::binopType::Add, neg_idx, int_type, nullptr));
```

src/lpython/semantics/python_ast_to_asr.cpp
certik: We have to add another check underneath:
```c++
if (ASR::down_cast<ASR::IntegerConstant_t>(expr_value(index))->m_n > 0) {
```

src/libasr/codegen/llvm_utils.cpp
certik: > Either allow negative indexing completely or don't allow it.

That would be ideal, but I don't know how to enable it for general expressions like "i", since we only know at runtime if it is positive or negative. So for performance reasons I would only allow it for compile time constant integer expressions.

Thirumalai-Shaktivel:
This PR is ready. if everything looks good, please merge it.

integration_tests/test_list_01.py
```diff
@@ -42,8 +42,30 @@ def test_list_02():
...
+    x: list[f64] = []
+
+    i: i32
+    for i in range(2):
+        x.append(float(i))
+
+    assert x[1] == x[-1]
+    assert x[0] == x[-2]
+    assert x[-1] == 1.0
+    assert x[-2] == 0.0
```
czgdp1807, line 56: So only constant negative indices are allowed? If the same thing is done by a variable then it won't be allowed?

Thirumalai-Shaktivel: Yup, Only Constant negative index works.
If the user pass `variable` which contains the `neg value`, the output value will be `zero`.
But if the user `enables bound check`, then it will be caught as `out of range`.

certik: You can do it with a variable, but it has to have a constant value. Using `x[-1]` is very common in Python and I think it's harmless to allow it without losing any performance. Using `x[i]` for negative `i` is  much less common, and we can't easily support it without losing performance. So this is a good compromise I think.

certik (approved):
Looks great, thanks!

Thirumalai-Shaktivel:
I don't quite understand here, why does this index yields an output? (I thought they will print 0's)
```py
def t():
    a: list[i32] = [1, 2, 3, 4, 5, 6]
    i: i32
    for i in range(3, 9):
        print(a[-i])

t()
```
```console
$ lpython examples/expr2.py && ./a.out
6
5
4
3
2
1
```

Thirumalai-Shaktivel:
I reverted this PR changes, still it generates the same output!

czgdp1807:
> I don't quite understand here, why does this index yield an output? (I thought they will print 0's)

Printing 0s silently for negative indices is extremely dangerous in my opinion. User/developer will never know what's going on until they go deep into the documentation which someone rarely does.

I really think we should fail loudly by default and use flags to print 0s silently. Personally, I would any day prefer safety over performance. Or we if just want to allow only compile time negative indices and not runtime indices, at least we should give safety warnings by default. We can provide options to silence those but at least there must be a way to let the users know about this.

integration_tests/test_list_01.py
certik: It's a good design point. The contract is that if LPython compiles your code *and* there are no runtime errors, then it will behave exactly as in Python. You *have* to use bounds checking in Debug mode.

Whether or not we allow compile time negative values, your example `traverse_list` would still silently fail at runtime if bounds checking is not enabled. This is an unrelated problem. Nothing is stopping a user to call it with negative values.

The same issue is with any runtime indexing, especially arrays. It can be out of bounds and we can't be checking it all the time in Release mode.  The same with integer overflow. We have to be using smaller integers like i32 or i64, but they can overflow. There is nothing we can do about it, and we should not be checking it all the time, it will be slow. But in Debug mode I think we can check it reasonably fast.

We can have a ReleaseSafe mode which keeps all these runtime checks in place, but otherwise optimizes everything out as much as it can. In this mode, it would indeed not fail silently. But the ReleaseSafe mode will almost for sure be slower than just Release mode.

certik:
The default option in LPython should be Debug mode. The Debug mode must have all runtime safety options on. In this Debug mode, you get very fast compilation, and safe runtime behavior. No silent surprises. Beautiful runtime error messages with stack traces. If you don't get any runtime error and things compiles (obviously), then it will behave just like in CPython.

Then we have two more options: Release and ReleaseSafe. ReleaseSafe should never get any segfaults or undefined/random behavior, assuming there are no bugs in our compiler, but it might not be as fast as Release. The Release mode is the fastest, but you have to check each of your inputs in Debug mode (or ReleaseSafe mode). This mode is great for computational tasks where (or if) you don't mind it might segfault. As a backend in a webbrowser code you should use ReleaseSafe to prevent exploits.

integration_tests/test_list_01.py
czgdp1807: > We can have a ReleaseSafe mode which keeps all these runtime checks in place, but otherwise optimizes everything out as much as it can. In this mode, it would indeed not fail silently. But the ReleaseSafe mode will almost for sure be slower than just Release mode.

This is also okay for me. At least we should provide a release which is as safe as possible and gives runtime errors/warnings for issues we think are possible. We can specify at the top of our docs that we provide following releases (ReleaseSafe, Production, Debug) and then users will be responsible for their choice.

certik: We can call it Debug, ReleaseSafe, ReleaseFast. 

Both of the releases will be used in production. Webbrowser will use ReleaseSafe, an HPC computational code would use ReleaseFast.

Also, if webassembly is safe (?), then we could use ReleaseFast with a WASM output as well. It just depends on what you want to do.

czgdp1807:
Yes I commented above. Providing multiple releases is fine with me. That way users will be able to try out safer options before puting their code into production.

certik:
This is a common source of confusion, and indeed we should clarify this contract high up in the docs.

You first have to use Debug mode with any C, C++, Fortran and LPython codes. Always. Check your inputs, make sure no runtime bounds checks trigger. Then you can do ReleaseFast and get very fast code. If anything behaves weird, check in Debug mode. That is the workflow.

integration_tests/test_list_01.py
certik: https://github.com/lcompilers/lpython/issues/1139

Issue #942, opened 2022-08-09
Generics for arrays


certik:
As of https://github.com/lcompilers/lpython/pull/907, the following now works:
```python
def f(lst: T[:], i: T) -> T:
    lst[0] = i
    return lst[0]

array: i32[1]
x: i32 = 69
print(f(array, x))
```

Other things to get working:
```python
def f(n: i32, a: T[n], b: T[n]) -> T[n]:
    return a + b
```
To make progress, simply let's write it using an explicit type, such as:
```python
def f(n: i32, a: f64[n], b: f64[n]) -> f64[n]:
    return a + b
```
and make sure LPython can compile it. If not, then let's report a bug, and then modify the code to compile:
```python
def f(n: i32, a: f64[n], b: f64[n]) -> f64[n]:
    r: f64[n]
    for i in range(n):
        r[i] = a[i] + b[i]
    return r
```
And then write is using generics.

certik:
Another use case:
```
def f(a: f64[:], b: f64[:]) -> f64[size(a)]:
    r: f64[size(a)]
    for i in range(size(a)):
        r[i] = a[i] + b[i]
    return r
```

certik:
And then 2D and 3D:
```
def f(n: i32, m: i32, a: f64[n,m], b: f64[n,m]) -> f64[n,m]:
    r: f64[n, m]
    for i in range(n):
        for j in range(m):
            r[i,j] = a[i,j] + b[i,j]
    return r
```

ansharlubis:
````
def f(n: i32, a: f64[n], b: f64[n]) -> f64[n]:
    return a + b
````
Works on ASR, does not work on LLVM.

ansharlubis:
```` python
def f(n: i32, m: i32, a: f64[n,m], b: f64[n,m]) -> f64[n,m]:
    r: f64[n, m]
    for i in range(n):
        for j in range(m):
            r[i,j] = a[i,j] + b[i,j]
    return r
````
Similarly, this compiles into ASR, but does not work in LLVM.

ansharlubis:
I got all the examples rewritten into generic functions in #1102 and they can compile into ASR.

PR #1102, opened 2022-09-08
Added more generic array examples

ansharlubis:
The examples where size of the arrays are variables (T[n]) do not compile into LLVM.


ansharlubis:
@certik I added a couple more generic array examples based on #942

certik:
Let's enable CPython tests. Also, there are error messages from LPython, so it means we need to improve LPython to compile these?

ansharlubis:
There does still seem to be some issues with the generated functions. I'll try to look into it.

ansharlubis:
@certik I got the additional array examples to work in both CPython and LLVM. I removed the example that uses size(..) in the type annotation.

src/libasr/pass/instantiate_template.cpp
```diff
@@ -26,16 +26,16 @@ class FunctionInstantiator : public ASR::BaseExprStmtDuplicator<FunctionInstanti
         new_func_name{new_func_name}
         {}
 
-    ASR::asr_t* instantiate_Function(ASR::Function_t &x) {
+    ASR::asr_t* instantiate_Function(ASR::Function_t *x) {
```
certik, line 29: What is the motivation of this change? Can `x` be `nullptr`?

certik (approved):
I think this is good enough to merge, I left a comment above.

Thanks!

src/libasr/pass/instantiate_template.cpp
ansharlubis: I think I made some changes that requires `&x` but then I made more changes that actually work with `*x`. I don't really have any reason for it. Should I revert it back?

PR #1105, opened 2022-09-09
C/CPP: Visit value if it is not null

Smit-create:


src/libasr/codegen/asr_to_c.cpp
```diff
@@ -516,6 +516,14 @@ R"(
         src = out;
     }
 
+    void visit_ComplexConstant(const ASR::ComplexConstant_t &x) {
+        std::string re = std::to_string(x.m_re);
+        std::string im = std::to_string(x.m_im);
+        headers.insert("complex");
+        src = "CMPLX(" + re + ", " + im + ")";
+        last_expr_precedence = 2;
+    }
```
czgdp1807, line 525: I think this same feature is being implemented in #1108 ?

Smit-create: Yeah maybe because it was giving errors in multiple files while generating C code.

Smit-create:
@czgdp1807 This is ready for review.

certik:
Ok, so this is a tough one.

It depends on what our goal is with the C and C++ backends. We'll later also add a Python backend (printing ASR as Python code) and Fortran backend. People also asked for Julia backend.

So far the goal of the C++ backend was to print things in such a way that a developer would write, so that one can use it as a transpiler. Say from Python to C++, or from Fortran to C++, and get a high quality code as output. The C backend is new, but I think it should also follow this design: a high quality C code that you would write by hand.

Another use case is to treat the C backend like LLVM: noone cares how ugly it is, as long as it produces the fastest results.

I think to satisfy the two use cases, we always want to produce code that gives the highest performance. We do not want to produce C or C++ code that compromises runtime performance. However, within this constrain, I think we always want to print it in such a way that is as human readable as possible and as natural as we can make it.

For this reason, we did not use the `value` of expression nodes, and rather printed them in full. The only backends that use the value are WebAssembly and LLVM, because those are "lowering" backends, we are not transpiling, we are trying to get the fastest answer, and adding more operations increases a chance that things will not run as fast. With C, C++, Python and Fortran backends, I think they are "transpiling" backends, effectively transforming ASR back to the surface language (all of them), with as much semantics as possible. 

What is the issue that this PR fixes? There might be another way to fix it.

We can also make this optional, enabled with a compiler option.

Here is another idea: why not writing an ASR->ASR pass, that takes compile time values and substitutes them for the given expression? So `(2+3)*5` will become `25`. This effectively does the same as this PR, but for all backends. For performance reasons, we do not want to probably execute this in Debug build (when speed of compilation is essential), but in Release mode this can be the first pass to start implementing compile time code evaluation. The next pass would take the `x=25`, and see that we never modify it, and make it a constant `x`. Then the next pass propagates constant expressions. And so on.

Smit-create:
> What is the issue that this PR fixes? There might be another way to fix it.

I was trying to compile and run this file `test_builtin_pow.py` in C backend just as in the LLVM backend. The issue I faced while trying to generate C code was the difference in return types of `pow(int a, int b)`. When `b>=0`, the result is `int` but when `b<0` it is `float`. Debugging more into this that why that works in LLVM but not in C, I found that LLVM uses the `value` which C wasn't using. So this PR was trying to fix that issue.

czgdp1807:
> Debugging more into this that why that works in LLVM but not in C, I found that LLVM uses the value which C wasn't using. So this PR was trying to fix that issue.

Its more of a workaround than a fix. Compile time values are not always available. So its the runtime code which should work? Try not using value and instead fix the C backend or the ASR to make the test work.

certik:
I forgot the details, but I remember this was an issue, since a function in LPython/ASR cannot change its type based on the argument like this. We created some workaround.

Smit-create:
> I forgot the details, but I remember this was an issue, since a function in LPython/ASR cannot change its type based on the argument like this. We created some workaround.

I think that is still remaining to fix:
See:
```py
def f():
    a: i32; b: f64
    a = pow(2, 2)
    b = pow(2, -2)
    print(a, b)

f()
```

This code gives:
```
% lpython a.py && ./a.out 
 --> a.py:3:9
  |
3 |     a = pow(2, 2)
  |         ^^^^^^^^^ '**' could be used instead

semantic error: Assigning integer to float is not supported
 --> a.py:4:9
  |
4 |     b = pow(2, -2)
  |         ^^^^^^^^^^ 
```

Now, if we change the type of `b` to `i32` which is not expected (actually it should be `f64` as in the above code)
Code:
```py
def f():
    a: i32; b: i32
    a = pow(2, 2)
    b = pow(2, -2) # this is unexpected, but let's check
    print(a, b)

f()
```
This gives:
```
  store double 2.500000e-01, i32* %b, align 8
 i32
```

Smit-create:
> We can also make this optional, enabled with a compiler option.

What if we use the same `--fast` flag in C/CPP backend, where we would try to avoid some extra function calls if the value is not `nullptr`?

certik:
I started writing up the design here: https://github.com/lcompilers/lpython/wiki/ASR-Design, it touches the issue of not using values for the C backend (but it's ok for the LLVM backend).

I think you discovered a bug in verify, that should have caught the last error you got.

Do you have some plan forward here?

Smit-create:
> Do you have some plan forward here?

I don't think there is any possible way to overload functions based on the return time, so this is a complex bug that is still unsolved in the `main` branch. I think we had some discussions on the same earlier.

certik:
We can discuss out our next meeting.

czgdp1807:
What was the result of the discussion? I think the issue is not that compile time values are not getting used. In fact, even if we don't use compile time values in any backend correct code should still be generated.

Smit-create:
This is fixed now, so closing.

PR #1108, opened 2022-09-09
Enable "c" test for test_builtin_bool.py

### LINKED ISSUE #1104: Enable `c` test for `integration_tests/test_builtin_bool.py`
On executing, `lpython --show-c integration_tests/test_builtin_bool.py` I get the following error,

```bash
code generation error: Cast kind 9 not implemented
  --> integration_tests/test_builtin_bool.py:26:16
   |
26 |     assert not bool(f)
   |                ^^^^^^^ 


```

xmnlab:
In order to enable the "c" test for test_builtin_bool.py it was necessary to add some "cast" that was missing in [asr_to_c_cpp.h](https://github.com/lcompilers/lpython/compare/main...xmnlab:lpython:1104-enable-tests-for-test_builtin_bool.py?expand=1#diff-b57347717b40c348eefd0b276ac88c5aaecab2e92f3e17da4a5738158457c4b1)

Resolve #1104 

PS: @czgdp1807 I am not sure about this check: https://github.com/lcompilers/lpython/compare/main...xmnlab:lpython:1104-enable-tests-for-test_builtin_bool.py?expand=1#diff-71deb66202ad6a1e88305b3bee772e479643e037d278446a610f561216ee65c5R500 do you have any idea if it would be necessary and how to check the type there?


src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -643,6 +643,27 @@ R"(#include <stdio.h>
...
+                src = "(bool)(" + src + ")";
+                break;
+            }
+            case (ASR::cast_kindType::LogicalToCharacter) : {
+                src = "(char)(" + src + ")";
+                break;
+            }
+            case (ASR::cast_kindType::CharacterToLogical) : {
+
+                src = "(bool)(strlen(" + src + ") > 0 ? 1 : 0)";
```
Smit-create, line 660:
```suggestion
                src = "strlen(" + src + ") > 0 ? true : false)";
```

src/libasr/codegen/asr_to_c.cpp
```diff
@@ -490,6 +490,20 @@ R"(
...
     }
 
+    void visit_ComplexConstant(const ASR::ComplexConstant_t &x) {
+        headers.insert("complex");
+
+        std::string re = std::to_string(x.m_re);
+        std::string im = std::to_string(x.m_im);
+
+        std::string type_name = "float complex";
+        // if (x.m_type == 8) type_name = "double complex";
```
czgdp1807, line 500: Why this is commented out?

src/libasr/codegen/asr_to_c_cpp.h
xmnlab: And what about `"(bool)(strlen(" + src + ")>0)"`?

src/libasr/codegen/asr_to_c.cpp
xmnlab: I got the idea from here: https://github.com/lcompilers/lpython/blob/main/src/libasr/codegen/asr_to_c.cpp#L226

but m_type is a pointer: `ttype_t* m_type`

do you have any idea how to deal with that?

another question, I copied it from another function: `last_expr_precedence = 2;`
is that correct? what is it used for?

xmnlab: I tried to debug it here but I didn't find any useful information here:

```
$14 = (const LFortran::ASR::ComplexConstant_t &) @0x7ffff784c9a8: {static class_type = LFortran::ASR::ComplexConstant, base = {
    static class_type = LFortran::ASR::expr, base = {type = LFortran::ASR::expr, loc = {first = 6108, last = 6108}}, 
    type = LFortran::ASR::ComplexConstant}, m_re = 0, m_im = 0, m_type = 0x7ffff784c958}
```

any idea about how I should proceed here? thanks

xmnlab: @czgdp1807 any comment about that?

src/libasr/codegen/asr_to_c.cpp
```diff
@@ -490,6 +490,20 @@ R"(
...
+
+        src = "(" + type_name + ") (" + re + ", " + im + ")";
+
+        last_expr_precedence = 2;
```
czgdp1807, line 504: Its fine to have it as the same pattern is followed in other constant nodes (IntegerConstant, LogicalConstant, etc.).

src/libasr/codegen/asr_to_c.cpp
czgdp1807: I would uncomment the line as `m_kind` signifies the number of bytes a constant/type will occupy. So for `8` kind we should `double complex`.

xmnlab: yep, but as you can see here: https://github.com/lcompilers/lpython/pull/1108#discussion_r967696075

```
$14 = (const LFortran::ASR::ComplexConstant_t &) @0x7ffff784c9a8: {static class_type = LFortran::ASR::ComplexConstant, base = {
    static class_type = LFortran::ASR::expr, base = {type = LFortran::ASR::expr, loc = {first = 6108, last = 6108}}, 
    type = LFortran::ASR::ComplexConstant}, m_re = 0, m_im = 0, m_type = 0x7ffff784c958}
```
it doesn't have m_kind there. the question is, is there any way to access this information from `ASR::ComplexConstant_t`?

src/libasr/codegen/asr_to_c.cpp
xmnlab: nice! thanks!

src/libasr/codegen/asr_to_c.cpp
czgdp1807: m_kind is in the type of expression. Fetch the type, down cast it to ASR::Complex and then you can obtain the m_kind.

xmnlab:
this PR is ready again for  a new round of review. thanks!!
cc @czgdp1807

xmnlab:
@czgdp1807 let me know if there is anything else missing here :) thanks!!

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -647,6 +647,23 @@ R"(#include <stdio.h>
                 src = "(bool)(" + src + ")";
                 break;
             }
+            case (ASR::cast_kindType::LogicalToReal) : {
+                src = "(float)(" + src + ")";
```
czgdp1807, line 651: Depends on the kind of destination `Real_t` type. `float` should be used for kind `4` and `double` should be used for kind `8`.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -647,6 +647,23 @@ R"(#include <stdio.h>
...
+            }
+            case (ASR::cast_kindType::CharacterToLogical) : {
+
+                src = "(bool)(strlen(" + src + ") > 0)";
+                break;
+            }
+            case (ASR::cast_kindType::ComplexToLogical) : {
+                src = "(bool)(" + src + ")";
+                break;
+            }
```
czgdp1807, line 666: Is this case tested in this PR? If not then update `test_builtin_bool.py` to test this complex to logical case.

czgdp1807 (changes requested):

src/libasr/codegen/asr_to_c_cpp.h
xmnlab: It seems it is already tested by:

```
    c: c32
    c = complex(2, 3)
    assert bool(c)
    c = complex(0, 0)
    assert not bool(c)
    assert not bool(0 + 0j)
```

src/libasr/codegen/asr_to_c_cpp.h
xmnlab: done! thanks!

xmnlab:
@czgdp1807 suggestions applied .. thank you!

czgdp1807 (approved):

certik:
Thanks @xmnlab !

PR #1109, opened 2022-09-10
Add some missing Cast in C/CPP

### LINKED ISSUE #1152: Bug in C codegen
I was trying to compile and run the C code generated from `integration_tests/test_builtin_str.py` (#1109 ). I found the following issue with the C codegen:
Python:
```py
def f():
    s: str
    s = "abcde"
    c: str
    for c in s[::-1]:
        print(c)

f()
```

The generated C code on the main branch is:
```c
void f()
{
    int32_t __explicit_iterator;
    char * __tmp_assign_for_loop = (char*) malloc(40 * sizeof(char));
    char * c = (char*) malloc(40 * sizeof(char));
    char * s = (char*) malloc(40 * sizeof(char));
    strcpy(s, "abcde");
    for (__explicit_iterator=0; __explicit_iterator<=strlen(__tmp_assign_for_loop) - 1; __explicit_iterator++) {
        strcpy(c, (char *)memcpy((char *)calloc(2U, sizeof(char)), __tmp_assign_for_loop + __explicit_iterator + 1 - 1, 1U));
        printf("%s\n", c);
    }
}
```

The above code is incorrect because we expect `__tmp_assign_for_loop` to contain the value of `s[::-1]`. This is because of this line in the ASR frontend:
https://github.com/lcompilers/lpython/blob/91cb772ef86a58d8b5020a8dde864754bc06d06a/src/lpython/semantics/python_ast_to_asr.cpp#L3653

The variable value is ignored in C code generation as seen in the above code `__tmp_assign_for_loop` does not contain the correct value and so essentially will give segmentation fault due to memory errors.

I tried to fix that by using the following diff(changed made to wherever variables are declared):
```diff
diff --git a/src/libasr/codegen/asr_to_c_cpp.h b/src/libasr/codegen/asr_to_c_cpp.h
index b696d8d1e..bed985fc9 100644
--- a/src/libasr/codegen/asr_to_c_cpp.h
+++ b/src/libasr/codegen/asr_to_c_cpp.h
@@ -520,6 +520,17 @@ R"(#include <stdio.h>
             }
         }

+        for (auto &item : x.m_symtab->get_scope()) {
+            if (ASR::is_a<ASR::Variable_t>(*item.second)) {
+                ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(item.second);
+                if (v->m_value) {
+                    self().visit_expr(*v->m_value);
+                    std::string name(v->m_name);
+                    decl += name +  " = " + src + ";\n";
+                }
+            }
+        }
+
```
And got the following C code:
```c
void f()
{
    int32_t __explicit_iterator;
    char * __tmp_assign_for_loop = (char*) malloc(40 * sizeof(char));
    char * c = (char*) malloc(40 * sizeof(char));
    char * s = (char*) malloc(40 * sizeof(char));
    __tmp_assign_for_loop = _lfortran_str_slice(s, 0, 0, -1, false, false);
    strcpy(s, "abcde");
    for (__explicit_iterator=0; __explicit_iterator<=strlen(__tmp_assign_for_loop) - 1; __explicit_iterator++) {
        strcpy(c, (char *)memcpy((char *)calloc(2U, sizeof(char)), __tmp_assign_for_loop + __explicit_iterator + 1 - 1, 1U));
        printf("%s\n", c);
    }
}
```

That still won't work because `__tmp_assign_for_loop` is assigned before `s` is assigned. Any suggestions to fix it?

Smit-create:
Fixes #1152


czgdp1807:
Can you test this as well? Like try enabling integration tests which work after this change? If none works then add new tests.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -117,6 +117,30 @@ R"(#include <stdio.h>
...
+    char s[20];
+    sprintf(s, "%f", x);
+    return s;
+}
+
+char* _lpython_real_to_string(double x) {
+    char s[20];
+    sprintf(s, "%lf", x);
+    return s;
+}
```
czgdp1807, line 143: I am confused here. These functions are a part of `asr_to_c_cpp.h` and not the generated C code? How will these functions end up there?

Smit-create:
> Can you test this as well?

Yeah, I'm waiting for some other C PRs to be merged so that I can use that code.

Smit-create:
I haven't enabled the C integration because of precision issues while asserting float/str values in the llvm/C backend.

tests/tests.toml
```diff
@@ -313,6 +313,7 @@ asr = true
 [[test]]
 filename = "../integration_tests/test_builtin_str.py"
 asr = true
+c = true
```
certik, line 316: Does the example compile? If so, I would remove this line, and rather enable it in integration_tests. That's the ultimate test, and it doesn't slow down `./run_tests.py`. We should only use the `./run_tests.py` if we are testing some complicated feature where it matters how the C code is generated.

tests/reference/c-test_builtin_str-7e198f7.stdout
```diff
@@ -0,0 +1,139 @@
...
+    ASSERT(strcmp("123", _lfortran_int_to_str4(x)) == 0);
+    x = 12345;
+    ASSERT(strcmp("12345", _lfortran_int_to_str4(x)) == 0);
+    x = -12;
+    ASSERT(strcmp("-12", _lfortran_int_to_str4(x)) == 0);
+    x = -121212;
+    ASSERT(strcmp("-121212", _lfortran_int_to_str4(x)) == 0);
+    xx =   1.23222339999999999e+01;
+    ASSERT(strcmp(_lfortran_float_to_str4(xx), "12.322234") == 0);
+    yy =   1.23222339999999999e+01;
```
Smit-create, line 70: These tests will fail after compilation because of precision issues.

czgdp1807: Can you make them pass? What’s the blocker which’s failing them?

certik: Those are most probably bad tests if they compare floating point directly, as that will fail in general, as you found out. So the test should be improved to use `abs(x-y) < eps` approach.

tests/reference/c-test_builtin_str-7e198f7.stdout
```diff
@@ -0,0 +1,139 @@
...
+    s = _lfortran_int_to_str4(-567);
+    ASSERT(strcmp(s, "-567") == 0);
+    ASSERT(strcmp(_lfortran_int_to_str4(4), "4") == 0);
+    ASSERT(strcmp(_lfortran_int_to_str4(-5), "-5") == 0);
+    ASSERT(strcmp("", "") == 0);
+    ASSERT(strcmp("1234", "1234") == 0);
+    ASSERT(strcmp(false ? "True" : "False", "False") == 0);
+    ASSERT(strcmp(true ? "True" : "False", "True") == 0);
+    ASSERT(strcmp("just a str", "just a str") == 0);
+    ASSERT(strcmp(_lfortran_float_to_str8(  1.21234000000000002e+01), "12.1234") == 0);
```
Smit-create, line 107: The only failing test is this. The error comes from this line, `_lfortran_float_to_str8(  1.21234000000000002e+01)` gives: `"12.123400"` but RHS is `"12.1234"`. I think we have the same issue with LLVM in precision but it was not raised by LLVM backend because it visits the `x.m_value` which is not null in this case and so doesn't raise an error there.

certik: The original test is `assert str(12.1234) == "12.1234"` which is wrong. You can't test floating point numbers like that. You could (in this case) do it like this: `assert str(12.1234)[:7] == "12.1234"` and hopefully it would work.

tests/reference/c-test_builtin_str-7e198f7.stdout
czgdp1807: > So the test should be improved to use `abs(x-y) < eps` approach.

Correct. I agree.

Smit-create:
@certik @czgdp1807 Please have a look if this is good to go.

certik (approved):
I think this is fine. @czgdp1807 can you please review it?

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -793,6 +793,36 @@ R"(#include <stdio.h>
...
+            right = "0";
+            rig_present = "false";
+        }
+        if (x.m_step) {
+            self().visit_expr(*x.m_step);
+            step = src;
+        } else {
+            step = "1";
+        }
+        src = "_lfortran_str_slice(" + arg + ", " + left + ", " + right + ", " + \
```
czgdp1807, line 822: Do we need to handle the case where both `x.m_start`, `x.m_end` and `x.m_step` are `NULL`? Do we test this? I think there will be 8 cases to test with `nullptr` and `not nullptr` for each `x.m_start`, `x.m_end` and `x.m_end` giving 2 x 2 x 2 = 8 cases. Are we making sure we test each of these. If some of these are supposed to give semantic error then include them in `errors/` folder.

Smit-create: Done

czgdp1807 (approved):

Issue #938, opened 2022-08-08
Nice API to create loops and if statements in LLVM


certik:
Perhaps something with C++ lambda functions: if(<lambda condition>, <lamba then>, optional <lambda else>), and you just fill in the condition and then / else, and the if function generates the correct boilerplate.

Currently every time we need an if statement or a loop, we have to emit all the boilerplate instructions.

PR #1110, opened 2022-09-13
Implement If_stmt and loop API in the LLVM backend

Thirumalai-Shaktivel:
Closes: #938

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -277,6 +277,36 @@ public:
         builder->SetInsertPoint(bb);
     }
 
+    template <typename Cond, typename Body>
+    void create_loop(Cond condition, Body loop_body) {
+        dict_api_lp->set_iterators();
+        dict_api_sc->set_iterators();
+        llvm::BasicBlock *loophead = llvm::BasicBlock::Create(context, "loop.head");
+        llvm::BasicBlock *loopbody = llvm::BasicBlock::Create(context, "loop.body");
+        llvm::BasicBlock *loopend = llvm::BasicBlock::Create(context, "loop.end");
+        this->current_loophead = loophead;
+        this->current_loopend = loopend;
+
+        // head
+        start_new_block(loophead);
+        {
+            llvm::Value* cond = condition();
+            builder->CreateCondBr(cond, loopbody, loopend);
+        }
+
+        // body
+        start_new_block(loopbody);
+        {
+            loop_body();
+            builder->CreateBr(loophead);
+        }
+
+        // end
+        start_new_block(loopend);
+        dict_api_lp->reset_iterators();
+        dict_api_sc->reset_iterators();
+    }
+
     inline bool verify_dimensions_t(ASR::dimension_t* m_dims, int n_dims) {
         if( n_dims <= 0 ) {
             return false;
@@ -3662,31 +3692,17 @@ public:
     }
 
     void visit_WhileLoop(const ASR::WhileLoop_t &x) {
-        dict_api_lp->set_iterators();
-        dict_api_sc->set_iterators();
-        llvm::BasicBlock *loophead = llvm::BasicBlock::Create(context, "loop.head");
-        llvm::BasicBlock *loopbody = llvm::BasicBlock::Create(context, "loop.body");
-        llvm::BasicBlock *loopend = llvm::BasicBlock::Create(context, "loop.end");
-        this->current_loophead = loophead;
-        this->current_loopend = loopend;
-
-        // head
-        start_new_block(loophead);
-        this->visit_expr_wrapper(x.m_test, true);
-        llvm::Value *cond = tmp;
-        builder->CreateCondBr(cond, loopbody, loopend);
-
-        // body
-        start_new_block(loopbody);
-        for (size_t i=0; i<x.n_body; i++) {
-            this->visit_stmt(*x.m_body[i]);
-        }
-        builder->CreateBr(loophead);
-
-        // end
-        start_new_block(loopend);
-        dict_api_lp->reset_iterators();
-        dict_api_sc->reset_iterators();
+        create_loop(
+            [this, x]() {
+                this->visit_expr_wrapper(x.m_test, true);
+                return tmp;
+            },
+            [this, x]() {
+                for (size_t i=0; i<x.n_body; i++) {
+                    this->visit_stmt(*x.m_body[i]);
+                }
+            }
+        );
     }
 
     void visit_Exit(const ASR::Exit_t & /* x */) {
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -277,6 +277,31 @@ public:
         builder->SetInsertPoint(bb);
     }
 
+    // Note: `create_if_else` and `create_loop` are optional APIs
+    // that do not have to be used. Many times, for more complicated
+    // things, it might be more readable to just use the LLVM API
+    // without any extra layer on top. In some other cases, it might
+    // be more readable to use this abstraction.
+    template <typename IF, typename ELSE>
+    void create_if_else(llvm::Value * cond, IF if_block, ELSE else_block) {
+        llvm::Function *fn = builder->GetInsertBlock()->getParent();
+
+        llvm::BasicBlock *thenBB = llvm::BasicBlock::Create(context, "then", fn);
+        llvm::BasicBlock *elseBB = llvm::BasicBlock::Create(context, "else");
+        llvm::BasicBlock *mergeBB = llvm::BasicBlock::Create(context, "ifcont");
+
+        builder->CreateCondBr(cond, thenBB, elseBB);
+        builder->SetInsertPoint(thenBB); {
+            if_block();
+        }
+        builder->CreateBr(mergeBB);
+
+        start_new_block(elseBB); {
+            else_block();
+        }
+        start_new_block(mergeBB);
+    }
+
     template <typename Cond, typename Body>
     void create_loop(Cond condition, Body loop_body) {
         dict_api_lp->set_iterators();
@@ -288,15 +313,13 @@ public:
         this->current_loopend = loopend;
 
         // head
-        start_new_block(loophead);
-        {
+        start_new_block(loophead); {
             llvm::Value* cond = condition();
             builder->CreateCondBr(cond, loopbody, loopend);
         }
 
         // body
-        start_new_block(loopbody);
-        {
+        start_new_block(loopbody); {
             loop_body();
             builder->CreateBr(loophead);
         }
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3662,31 +3723,16 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+        auto test = [this, x]() {
+            this->visit_expr_wrapper(x.m_test, true);
+            return tmp;
+        };
+        auto body = [this, x]() {
+            for (size_t i=0; i<x.n_body; i++) {
+                this->visit_stmt(*x.m_body[i]);
+            }
+        };
+        loop_api(test, body);
```
certik, line 3735:
```suggestion
        loop_api([this, x]() {
            this->visit_expr_wrapper(x.m_test, true);
            return tmp;
        }, [this, x]() {
            for (size_t i=0; i<x.n_body; i++) {
                this->visit_stmt(*x.m_body[i]);
            }
        });
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -277,6 +277,77 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         builder->SetInsertPoint(bb);
     }
 
+    template <typename IF>
+    void if_api(llvm::Value * cond, IF if_block) {
```
certik, line 281:
```suggestion
    void create_if_stmt(llvm::Value * cond, IF if_block) {
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -277,6 +277,77 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            if_block();
+        }
+        builder->CreateBr(mergeBB);
+
+        start_new_block(elseBB);
+        start_new_block(mergeBB);
+    }
+
+    template <typename IF, typename ELSE>
+    void if_else_api(llvm::Value * cond, IF if_block, ELSE else_block) {
```
certik, line 300:
```suggestion
    void create_if_else(llvm::Value * cond, IF if_block, ELSE else_block) {
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -277,6 +277,77 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+
+        start_new_block(elseBB);
+        {
+            else_block();
+        }
+        start_new_block(mergeBB);
+    }
+
+    template <typename Cond, typename Body>
+    void loop_api(Cond condition, Body loop_body) {
```
certik, line 322:
```suggestion
    void create_loop(Cond condition, Body loop_body) {
```

certik (approved):
I think this looks great.

czgdp1807 (changes requested):
Please do the replacement of all loops, if-else blocks in https://github.com/lcompilers/lpython/blob/main/src/libasr/codegen/llvm_utils.cpp (where the actual benefit of this API will be visible) with the call to your APIs. The loops in `src/libasr/codegen/asr_to_llvm.cpp` are very simple apparently so the loops/if-else APIs won't be tested properly.

Thirumalai-Shaktivel:
Working on it!

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -920,24 +962,20 @@ namespace LFortran {
...
-                deepcopy_key_value_pair_linked_list(srci, desti, dest_key_value_pairs,
-                    src_capacity, dict_type, module);
-            }
-            builder->CreateBr(mergeBB);
-            llvm_utils->start_new_block(elseBB);
-            llvm_utils->start_new_block(mergeBB);
+            llvm_utils->create_if_else(
+                is_key_set,
+                [this, src_key_value_pairs, dest_key_value_pairs, itr,
+                        src_capacity, dict_type, &module]() {
```
Thirumalai-Shaktivel, line 971: - The above variables inside the square bracket are read-only.
- I can also replace it with the below suggestion (but it makes all the variables read-write):
```suggestion
                [&]() {
```
Both of them work fine, but the second one reduces the pain to define the variable inside the `[]`
Which one should I use, make the variables read-only or simply use `&`?
cc @certik, @czgdp1807

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1746,42 +1759,38 @@ namespace LFortran {
...
-            }
-            builder->CreateBr(mergeBB);
-
-            llvm_utils->start_new_block(elseBB);
-            llvm_utils->start_new_block(mergeBB);
+            llvm_utils->create_if_else(
+                is_key_set,
+                [this, key_list, idx, key_asr_type, value_list, value_asr_type,
+                        current_capacity, new_key_list, new_value_list,
+                        new_key_mask, &module]() {
```
Thirumalai-Shaktivel, line 1769: ..

Thirumalai-Shaktivel:
The PR has a huge change, Can you please review it carefully?
cc @czgdp1807

czgdp1807 (commented):
The amount of boilerplate for if-else and loop in LLVM is constant with respect to the size of the body and condition. So, having these APIs provides very slight benefit (at least according to me) when it comes to generating code for if-else and loop in LLVM as you are just replacing constant number of boilerplate lines with another constant number of API calls. The drawback of these APIs is that you have to think your logic in such a way that complies with the API (so flexibility is lost with very slight advantage). Hence, calling the API to create these if-else and loop is an overhead IMO. So, I am leaning towards not merging this. If other team members think that this should be merged then feel free to merge.

Thirumalai-Shaktivel:
Thanks for the review!

Thirumalai-Shaktivel:
> The drawback of these APIs is that you have to think your logic in such a way that complies with the API

```c++
create_if_else(
	cond,
	[]() {
		// if_block.
	},
	]() {} // empty else_block.
);
```
Here the API takes three parameters, among which 1st -> condition, 2nd and 3rd are if and else blocks. The `square bracket` is used to capture the variables within the block.
Here, the `if_block` can be written as we write code within the `if statement`.
I think this looks more readable than before. But I don't know whether it is difficult while writing the new code (future implementations).

czgdp1807:
It's not always the case that condition is a single LLVM API call. Sometimes condition can also be a block (it is actually the case with dict when you have to decide on the basis of key mask and matching keys so you need to extract the information inside the block to form the condition). Now tweaking the algorithm according to the API is not right IMO because somewhere we are taking the risk of sacrificing performance to comply with the API.

Also, note that its always at max 7 lines of boilerplate to generate code for if-else block and loops. Now we replacing those 7 lines with 1 call to the functions you have added. So a benefit of 6 lines which is very small compared to the overhead it brings in.

But again if the team wants to merge this then its okay.

Thirumalai-Shaktivel:
I see, now I understood the problem

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3620,73 +3668,53 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-        builder->CreateBr(mergeBB);
-
-        start_new_block(elseBB);
-        for (size_t i=0; i<x.n_orelse; i++) {
-            this->visit_stmt(*x.m_orelse[i]);
-        }
-
-        start_new_block(mergeBB);
+        create_if_else(
+            tmp,
```
certik, line 3672:
```suggestion
        create_if_else(tmp,
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3620,73 +3668,53 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                for (size_t i=0; i<x.n_body; i++) {
+                    this->visit_stmt(*x.m_body[i]);
+                }
+            },
+            [this, x]() {
+                for (size_t i=0; i<x.n_orelse; i++) {
+                    this->visit_stmt(*x.m_orelse[i]);
+                }
+            }
+        );
```
certik, line 3683: Let's see how short one can make this:
```suggestion
        create_if_else(tmp,
            [=]() {
                for (size_t i=0; i<x.n_body; i++) {
                    this->visit_stmt(*x.m_body[i]);
                }
            }, [=]() {
                for (size_t i=0; i<x.n_orelse; i++) {
                    this->visit_stmt(*x.m_orelse[i]);
                }
            }
        );
```

src/libasr/codegen/llvm_utils.h
```diff
@@ -96,6 +96,11 @@ namespace LFortran {
 
             void start_new_block(llvm::BasicBlock *bb);
 
+            template <typename IF, typename ELSE>
+            void create_if_else(llvm::Value * cond, IF if_block, ELSE else_block);
+            template <typename Cond, typename Body>
+            void create_loop(Cond condition, Body loop_body);
```
certik, line 102: Let's move all the implementation here.

And add documentation that this is an optional API, that does not have to be used, that many times for more complicated things it might be more readable to just use the LLVM API without any extra layer on top. In some other cases it might be more readable to use this abstraction. So this API is provided and can be optionally used, but it is not required.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -737,76 +754,59 @@ namespace LFortran {
...
-                LLVM::CreateStore(*builder,
-                    llvm::ConstantPointerNull::get(llvm::Type::getInt8PtrTy(context)),
-                    curr_dest_next_ptr
+        llvm_utils->create_loop(
+            [this]() {
+                return builder->CreateICmpNE(
+                    LLVM::CreateLoad(*builder, src_itr),
+                    llvm::ConstantPointerNull::get(llvm::Type::getInt8PtrTy(context))
+                );
+            },
```
Thirumalai-Shaktivel, line 763: @czgdp1807, the changes in the `asr_to_llvm.cpp` for if and loop API seem simplified. So, it looks good!
But, should the changes in the `llvm_utils.cpp` be reverted or keep it with the following changes?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3620,73 +3612,42 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-        }
-        builder->CreateBr(mergeBB);
-
-        start_new_block(elseBB);
-        for (size_t i=0; i<x.n_orelse; i++) {
-            this->visit_stmt(*x.m_orelse[i]);
-        }
-
-        start_new_block(mergeBB);
+        llvm_utils->create_if_else(tmp, [=]() {
```
Thirumalai-Shaktivel, line 3615: Using the `llvm_utils->create_if_else` causes `seg fault`. How do I fix this?
@certik @czgdp1807

src/libasr/codegen/llvm_utils.cpp
czgdp1807: I would say revert it because algorithms in `llvm_utils.cpp` shouldn't be bound by any API abstractions.

src/libasr/codegen/asr_to_llvm.cpp
czgdp1807: I don't know. You will have to debug. Add `std::cout` statements or use debugger to narrow down to the line which is causing seg fault.

Thirumalai-Shaktivel:
This PR is ready!

src/libasr/codegen/llvm_utils.cpp
Thirumalai-Shaktivel: Done!

Thirumalai-Shaktivel:
Thanks for the review!

src/libasr/codegen/asr_to_llvm.cpp
Thirumalai-Shaktivel: Yup, I tried to move `create_if_else` into `llvm_utils`, but it caused `seg fault`.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -277,6 +277,58 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+        {
+            loop_body();
+            builder->CreateBr(loophead);
+        }
+
+        // end
+        start_new_block(loopend);
+        dict_api_lp->reset_iterators();
+        dict_api_sc->reset_iterators();
+    }
```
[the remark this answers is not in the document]
Thirumalai-Shaktivel: Yup, I tried to move `create_if_else` into `llvm_utils`, but it caused `seg fault`.

Issue #1113, opened 2022-09-15
Supporting ``Enum`` and ``Union`` in LPython


czgdp1807:
As the title says, we need to figure out a good syntax for ``Enum`` and ``Union`` in LPython, especially in the C backend.

For syntax we can follow the same as of CPython,

- [x] ``enum`` - https://docs.python.org/3/library/enum.html - https://github.com/lcompilers/lpython/pull/1117
- [x] ``Union`` - Import from CTypes, see https://stackoverflow.com/a/45384034 - https://github.com/lcompilers/lpython/pull/1126

Implementation in C backend is fairly easy. For LLVM we can figure it out as well (just use structs or something else) but a low priority for LLVM backend.

czgdp1807:
@certik @rebcabin @ronnuriel Please let me know if you agree with this syntax. Then we can start implementing things in C backend.

rebcabin:
Regarding `enum`, the URL points to good candidate tests for the facility. So long as the same code works identically in CPython and LPython (including the import statements), I think it's good. Will comment on Union shortly

czgdp1807:
Cool I will start off with `enum` implementation then. I think the same syntax as CPython is fine.

rebcabin:
I approve of the `Union` proposal, too.

rebcabin:
remember the overriding tenet is 100% compatibility with ordinary CPython, a specific version x.y.z if necessary to limit it.

czgdp1807:
Sure. Makes sense. So I will go ahead and implement both Union and Enum in C according to our priorities discussed offline.

czgdp1807:
Can we close this? As the initial versions are already in. Please let me know. I am attaching `could close` label for now.

certik:
I think so. I am sure there will be tons of things to improve, but let's simply open up issues as we discover them. The initial implementation is done. Thanks @czgdp1807 !

PR #1117, opened 2022-09-16
Add support for `Enum` in C and LLVM backends

czgdp1807:
cc: @certik @rebcabin 

https://github.com/lcompilers/lpython/issues/1113


rebcabin:
I recommend a few more integration tests:
how about enums with negative numbers?
how about enums that start with zero?
how about enums that skip numbers, say 1, 3, 5? 
how about enums that have NO members?

rebcabin:
> Sure I will add more tests, once ASR.asdl changes receive approval.

I'm copying those changes to my independent test machine for ASR.

czgdp1807:
> Is there a sidetrack for debugging information, or are we even bothering with debug builds at this point?

All the heavy lifting is done in AST to ASR transition. So any semantic error will be thrown before start generating LLVM code. Once the ASR is generated it is supposed to be correct. And the ASR maintains all the Enum type and data as it is.

czgdp1807:
> user loses the symbolic rep of the enum

This wouldn't happen as ASR can always be seen using `--show-asr` and it stores Enums in their raw form so users can always see what happened to their Enum type. That's why I added `EnumRef`, `EnumTypeConstructor`, `Enum` nodes in that ASR. Waiting for @certik to comment on these changes.

rebcabin:
> Sure I will add more tests, once `ASR.asdl` changes receive approval. 

The ASR changes look good to my eye. Ondrej will be the final authority on them.

rebcabin:
> > Is there a sidetrack for debugging information, or are we even bothering with debug builds at this point?
> 
> All the heavy lifting is done in AST to ASR transition. So any semantic error will be thrown before we start generating LLVM code. Once the ASR is generated it is supposed to be correct. And the ASR maintains all the Enum type and symbolic data as it is, so users will receive errors if they access Enums incorrectly.

What about errors at run time, not in static semantics but in dynamic semantics? I passed BLUE = 42 instead of GREEN = 17 and my graphics engine raised an exception. Wouldn't I see "17 expected, received 42" instead of "GREEN expected, received BLUE?" 

What does clang do in debug builds for something like this?

czgdp1807:
> What does clang do in debug builds for something like this?

I don't know about debug builds for clang in this case. But for LPython its possible to show the exact name of Enum (instead of a number) for runtime errors related to Enums. If we need it urgently then I can I add it in this PR itself tomorrow or on Monday. If not then I can add it after completing Unions. Let me know what you think.

src/libasr/ASR.asdl
```diff
@@ -278,6 +281,7 @@ expr
 
     | BitCast(expr source, expr mold, expr? size, ttype type, expr? value)
     | DerivedRef(expr v, symbol m, ttype type, expr? value)
+    | EnumRef(symbol v, symbol? m, string property, ttype type, expr? value)
```
certik, line 284: What is the `property` used for?

I am going to setup documentation for ASR, so that we can document (in Markdown) each node and what it does and is meant to be used (for now I was thinking one file per node). We'll also soon have some tests and semantic definition of each ASR node thanks to the work that @rebcabin is doing.

certik: We might even use tangledown for this. ;)

certik: From reading the C++ code, it seems that `property` is always `value`? I would remove it.

What is the allowed result of `EnumRef` --- what type is it? Is it always the type of the given member?

It seems just the `m` argument is enough to specify. Also, why is it `symbol? m`? What is the meaning if `m==nullptr`?

src/libasr/ASR.asdl
```diff
@@ -97,6 +97,8 @@ symbol
         identifier original_name, access access)
     | DerivedType(symbol_table symtab, identifier name, identifier* members,
         abi abi, access access, symbol? parent)
+    | EnumType(symbol_table symtab, identifier name, identifier* members,
+        abi abi, access access, ttype type, symbol? parent)
```
certik, line 101: Each `EnumType` has one underlying type, say `i32`, to represent each enumerator, and this type is stored in `type`?

certik:
Python also allows two enumerators (blue and red) to have the same number (3), unless the `@unique` decorator is used (https://docs.python.org/3/library/enum.html#ensuring-unique-enumeration-values), so we should test that.

We also need to allow different than integer values: floats, strings and tuples to start. Here is one example from Python's docs:
```python
>>> class Planet(Enum):
...     MERCURY = (3.303e+23, 2.4397e6)
...     VENUS   = (4.869e+24, 6.0518e6)
...     EARTH   = (5.976e+24, 6.37814e6)
...     MARS    = (6.421e+23, 3.3972e6)
...     JUPITER = (1.9e+27,   7.1492e7)
...     SATURN  = (5.688e+26, 6.0268e7)
...     URANUS  = (8.686e+25, 2.5559e7)
...     NEPTUNE = (1.024e+26, 2.4746e7)
```

Other than that and my comments above, it seems the ASR design is sufficiently general for now. Fortran is also getting enumerations, and I think they will fit. The C and C++ enumerations should also fit into this.

Finally, the Swift enumeration (https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html) allows values to be of different types:
```swift
    enum Barcode {
        case upc(Int, Int, Int, Int)
        case qrCode(String)
    }
```
Python also seems to allow that:
```python
from enum import Enum

class A(Enum):
    x = 1
    y = 2
    z = 3
    p = "ss"

a = A(3)

print(A(3))
print(A("ss"))
```
prints:
```
A.z
A.p
```
So I think we should allow that. That would cover all cases I think.

Internally in LLVM the enum can still be represented by an integer I think. When you access it, we can require to know the enumerator at compile time for now, then the type is also known at compile time (for different type enum).


Regarding debugging, we currently do not store any debugging information in LLVM, but we should, however that's a separate issue.

I think using the actual value in LLVM is good.

We should be careful with some of the more general cases to ensure we don't accidentally lose performance.

src/libasr/ASR.asdl
czgdp1807: > What is the `property` used for?

It can be either `name` or `value`. For now I only added support for `value` but in follow up PRs I will add support for `name` as well.



> It seems just the `m` argument is enough to specify. Also, why is it `symbol? m`? What is the meaning if `m==nullptr`?

There are two cases,

1. `Color.RED.value` - Here we are accessing `RED` via `Color` class directly. So `EnumRef(Color, RED, value)` is created.
2. `selected_color.value` - Here `selected_color` is an object (from CPython's perspective). So in LPython we create `EnumRef(selected_color, nullptr, value)` (see how `m` is nullptr now).

src/libasr/ASR.asdl
czgdp1807: Yes. Correct.

czgdp1807:
> Python also allows two enumerators (blue and red) to have the same number (3), unless the `@unique` decorator is used (https://docs.python.org/3/library/enum.html#ensuring-unique-enumeration-values), so we should test that.

Makes sense. Will do that.

> We also need to allow different than integer values: floats, strings and tuples to start.

In ASR and LLVM they are easily possible and I think I implemented LLVM in such a way that it already should be supporting values other than integers.. But C backend only supports integer values, so not sure how we will handle `float`, `string`, etc in C backend.



> Python also seems to allow that:
> 
> ```python
> from enum import Enum
> 
> class A(Enum):
>     x = 1
>     y = 2
>     z = 3
>     p = "ss"
> 
> a = A(3)
> 
> print(A(3))
> print(A("ss"))
> ```
> 
> prints:
> 
> ```
> A.z
> A.p
> ```
> 
> So I think we should allow that. That would cover all cases I think.

I think for compile time values of `some_argument` figuring out the type of `A(some_argument)` is possible. But for runtime values, we need to figure out the type at runtime which will complicate the situation. In addition, C backend doesn't support anything other than integers as values Enum members. That's why I didn't support heterogeneous typed members of an Enum. **I don't see heterogeneous typed members of an Enum as immediate requirement so I don't plan to support it here**. Do you agree? Please let me know.

czgdp1807:
> When you access it, we can require to know the enumerator at compile time for now, then the type is also known at compile time (for different type enum).

Yes, correctly pointed out. That's the main reason why I didn't support heterogeneous typed Enum.

> This is slower, so I would not allow that at all for now. Later we can braintorm if and how to do that. It's a "slow" feature, with a danger of users using it by mistake and accidentally slowing their codes down, so we have to be very careful with these. Ideally they should be designed in such a way to make it clear that it is "slow". We want all default features to be "fast".

Agreed.

So from what I can figure out,

1. Supporting floats, strings is possible only for LLVM backend. For C backend its not possible so I would raise CodeGen error for C backend for enum of type other than integers.
2. Heterogeneous typed feature are a thing for the future so not to be supported here.
3. Add more tests as pointed out by @rebcabin in https://github.com/lcompilers/lpython/pull/1117#issuecomment-1249396411.

rebcabin:
> > What does clang do in debug builds for something like this?
> 
> I don't know about debug builds for clang in this case. But for LPython its possible to show the exact name of Enum (instead of a number) for runtime errors related to Enums. If we need it urgently then I can I add it in this PR itself tomorrow or on Monday. If not then I can add it after completing Unions. Let me know what you think.

I'm just bringing it up as a potentially neglected issue. Debugging at run time will be demanded. It might be easier to plumb it in now incrementally, or it might be easier to do it later in a big "sprint," too close for me to call.

rebcabin:
> The general case of runtime only enumerators is something like "reflection", which we can do later, I think it is similar to calling methods or functions at runtime only knowing their name as a runtime "string". This is slower, so I would not allow that at all for now. Later we can braintorm if and how to do that. It's a "slow" feature, with a danger of users using it by mistake and accidentally slowing their codes down, so we have to be very careful with these. Ideally they should be designed in such a way to make it clear that it is "slow". We want all default features to be "fast".

One "term of art" for this topic is "late binding." You have a symbol, but you don't know the actual code it's bound to until run time. It is slow. It also gets wrapped up in the type system. You have not only a symbol, but an interface spec (parameter types, return type, generics). It can grow to overwhelming complexity. Microsoft had a whole Universe called COM and OLE Automation, kind of a side-car to every language.

src/libasr/ASR.asdl
certik: I would distinguish these two cases, it seems they are unrelated.

`Color.RED.value` -> Just use the compile time value of the symbol RED. The symbol `RED` will have a parentsymtab equal to the Enum's symtab, so we can have a function "is_enumerator" that will determine if a given symbol is an enumerator. If we need a special node for this, we could do `EnumeratorValue(RED)`, but given the fact that the implementation just does `RED->m_value` it seems we do not need a node for this?

`a = Color.RED; a.value` -> `(Assign a EnumItem(Color, RED))  (a->m_value)`

certik:
Great point, it's late binding: complex, slow, etc. If we later decide to do it at all, we should design it in a way to make it clear it is late binding, and thus slow. For now let's see how far we can get by early binding, knowing at compile time what each symbol's code is.

@czgdp1807 I would allow heterogeneous types in Python source code and in ASR. That way everything is done regarding the ASR design and it will support the most general case in all languages. In the backends just say "not implemented yet". In C backend only support integers (to support other types or heterogeneous types, we would have to store some list/vector of those values, one vector for each type; or we would just substitute the actual values). In the LLVM let's do also floats (and optionally strings if it is not too complicated), tuples we can do later.

certik:
Ok, it might be that heterogeneous types are late binding also (and thus we should not do it). Is there a way to design heterogeneous types to be early binded? Consider this simple use case:
```python
class Color(Enum):
    RED = 1
    GREEN = 2
    BLUE = "str"

def test_selected_color(color: Color):
    print(color)
    print(color.value)

test_selected_color(Color.RED)
test_selected_color(Color.BLUE)
```
Here we do not know the type of `color.value` at compile time, which is a problem. This seems similar to having lists of items of different type. We do not allow that, because that is "late binding".

Ok, I would say that an ASR requirement for enumeration is:

* All enumerators are of single type (integer, float, string, tuple, etc.)
 
How would a type float be implemented in the backend? Say:
```python
def test_selected_color(color: Color):
    v: f64 = color.value
```
The `color` argument can either be passed as `f64` (then it would be the "value" of the enumeration), or it can be passed as integer (in which case we need a lookup for the float).

It seems it would be easier and faster to simply pass the Enum as the type of the "value".

However, the next thing we need: Fortran also allows "next()" and "previous()" as well as "first()" and "last()" functions on enumeration. This is also needed if you want to loop over the enumeration. How would these be implemented? Well, even for integers, if they are not consecutive, it seems that you need a runtime list of all possible values (in order), and then use that to lookup the "current" integer (or float)'s index, and then increment or decrement it.

Here is an example that would fail:
```python
from ltypes import i32
from enum import Enum

class Color(Enum):
    RED: i32 = 1
    GREEN: i32 = 10
    BLUE: i32 = 11

def test_selected_color(selected_color: Color):
    color: Color
    color = Color(selected_color.value + 1)
    assert color == Color.GREEN
    print(color)

test_selected_color(Color.RED)
```
Where we are effectively doing `Color(2)` which is not defined. CPython gives "ValueError: 2 is not a valid Color".

This works:
```python
from enum import Enum

class Color(Enum):
    RED = 1
    GREEN = 10
    BLUE = 11

def test_selected_color(selected_color: Color):
    color: Color
    color = Color(selected_color.value + 9)
    assert color == Color.GREEN
    print(color)

test_selected_color(Color.RED)
```

How would we implement this? For this to work, it seems that you need a runtime mapping of value -> enumerator.

So perhaps for now, we should only allow using "enumerator", so `Color(RED)` is allowed, but not `Color(10)`.

certik:
But we can support this:
```python
from enum import Enum

class Color(Enum):
    RED = 1
    GREEN = 10
    BLUE = 11

for color in Color:
    print(color, color.value)
```
which prints:
```
Color.RED 1
Color.GREEN 10
Color.BLUE 11
```
Because the order and names and values are all known at compile time.

src/libasr/ASR.asdl
certik: The tough runtime use case is:
```python
def f(color: Color):
    print(color, color.value)
```
So let's consider:
```
EnumName(symbol v, ttype type, expr? value) # type = string
EnumValue(symbol v, ttype type, expr? value) # type = Enum value type (int, float, str, ...)
```
and the above would be represented by: `EnumName(color, string, nullptr)` and `EnumValue(color, i32, nullptr)`.
The `value` argument is either nullptr (runtime case, such as above), or it contains the compile time result. For example, if you do:
```python
print(Color.RED, Color.RED.value)
```
then you can set `EnumName(color, string, "RED")` and `EnumValue(color, i32, 1)`.

certik: It's clear how to implement `EnumValue`, you already did it in this PR. But how would you implement `EnumName` for the runtime case? That is, `EnumName(color, string, "RED")` is clear, but how would you implement `EnumName(color, string, nullptr)`?

certik: It seems to implement `EnumName(color, string, nullptr)`, say if Color is of type float, then we have two options:

1. Either we represent Enum as an integer which is an index into two lists: one with names, the other with values.
2. Represent Enum with its type, say f64, then value is immediate, but to get the name, you need to have a runtime hashtable of f64-> str.

Each approach has pros and cons. If we want fast implementation of EnumName, then we should use the first approach (but EnumValue will be slower). If we want fast EnumValue, then the second approach is better, but then the name lookup will be slightly slower.

For iterating, the first approach is better.

I would argue the second approach is probably better, since a very fast lookup of the value is more important. Name lookup as well as iteration is probably secondary. I believe in C and C++, you cannot even lookup the name, nor iterate. But the value retrieval is immediate (of non-consecutive integers).

certik: So if we go with the second approach, then I would not even implement EnumName now, we can do that later if needed. I would just implement EnumValue for now.

certik:
After thinking through the issues above, I think the requirements for ASR enumerations should be:

* Support retrieving the name of the enumerator as a string, and the value
* Values should all be of the same type and kind, say, f64, or i32 (consistent with our choices of lists of the same type, etc.; for efficiency and general user friendliness of knowing at compile time what the type of a value is)
* Support iterating over the enumerators (`for color in Color:`)
* Support accessing the next and previous enumerator
* Regarding performance: in general, assigning values to enumerators (say floating point or string values) is to be used at a higher level, not in performance critical parts of the code; It is reasonable (if we decide to go that route) to say that enumerators are slightly slower than the raw "value type" (say f64), and so for performance critical code, one just passes in the value as a type (f64), but at the higher level one can use an Enum type. If enumerators are used without values, and always comparing them directly (say for a boundary condition or a solver type) using a switch statement or if statement, then it must be as performing as an integer.

There are two possible implementations.

1. Enum is implemented as an integer index 0..n-1 into an array of tuples ("enumerator name", value). The index is used to lookup the enumerator name, as well as its value. For direct comparison as well as iterating and previous/next, just the index is used.
2. Enum is implemented as its value. So value is retrieved immediately. To lookup the enumerator name, to iterate and previous/next one must use a hash table mapping of value->index; And have an array index->"enumerator name".


If we go the route 1., the Enum ASR node can have a flag "index_value" which if true would mean that the index 0..n-1 is equal to the values, so that one can use the index itself to obtain the value and vice versa, as an optimization to avoid the lookup.

We can also both approaches 1. and 2., depending on the value type, or even expose it as a flag to the user to select.

Last thing to consider is interoperation with C. It seems that C only allows integers as values, but in general they can be in any order and even duplicate. C does not allow lookup the name of the enumerator. It seems the approach 2. would be compatible with C.

src/libasr/ASR.asdl
czgdp1807: > So if we go with the second approach, then I would not even implement EnumName now, we can do that later if needed. I would just implement EnumValue for now.

Yes that's what I did here. So I think we just need to implement `EnumValue` for now in LLVM and C backend. We will do `EnumName` later.

czgdp1807:
I would like you to remind that supporting `f64` values in enumerations is exactly the same as allowing `f64` typed keys in hash maps. Due to fuzzy nature of `f64` comparisons, its impossible to determine the exact name/value which the user wants in the enumeration. So I am not sure if you really want to support `f64` values in enumerations.

So I need some clarity,

1. We need all the ASR requirements (listed in https://github.com/lcompilers/lpython/pull/1117#issuecomment-1249921032) to be implemented in this PR, right?
2. I support 2nd approach i.e., implement enumerations as value in LLVM backend and name lookups can be easily done via hashmaps. So, for now in this PR we should only support value access in LLVM backend and name lookups can be added later, right?

Let me know your thoughts so that I can go ahead and complete this PR.

certik:
Yes, I am aware of the issues with f64. One solution is the first approach, which would work. The second approach is to cast f64 to i64, and use i64 keys. The floats do not have fuziness if we just use the data and copy it, which I think is all that we would be doing.

I think you don't need to implement everything, as long as you do not close doors to implement everything. I think the 2nd approach is fine, if you want to go ahead with it. I think using a hashmap to lookup the name is fine.

One caveat is that enumerators that have the same value would then effectively be equivalent. I don't know if there is an issue with that or not.

So I would just do what you said. Let me know if you have any other questions or doubts.

czgdp1807:
Awesome. I will implement all the features in ASR and support limited features in C and LLVM backend. This PR should be ready by Monday EOD.

certik:
I keep thinking about it, and there are multiple competing use cases. Take the use case of "next one". It will fail if multiple enumerators have the same value. Consider a simplified case:
```python
class Color(Enum):
  red = True
  blue = False
  green = True
  orange = False
  yellow = False
...
```
If we represent the enum by its value, we can only represent two enumerators, not all of them.

Consider another use case:
```python
>>> class Planet(Enum):
...     MERCURY = (3.303e+23, 2.4397e6)
...     VENUS   = (4.869e+24, 6.0518e6)
...     EARTH   = (5.976e+24, 6.37814e6)
...     MARS    = (6.421e+23, 3.3972e6)
...     JUPITER = (1.9e+27,   7.1492e7)
...     SATURN  = (5.688e+26, 6.0268e7)
...     URANUS  = (8.686e+25, 2.5559e7)
...     NEPTUNE = (1.024e+26, 2.4746e7)
```
Here the value is quite complicated, probably should not be used as the "index". What if two enumerators happen to have the same value?

What about this:
```python
class MolecularMass(Enum):
    water = 18
    methane = 16
    ammonia = 17
    oxygen = 16
...
```
One could use floats and get (hopefully?) distinct numbers, but one could also need these integer mass numbers, number of nucleons, say for some teaching application. (I remember these, but do not remember the floating point numbers based on the isotopes frequency.)

What about string values:
```python
class ErrorMessages(Enum):
  E1 = "Type mismatch"
  E2 = "Redeclaration of a variable"
  E3 = "Assignment types must match"
  # E4 was used in an earlier version of the compiler, but removed
  E5 = "Variable is not declared"
```
I don't think we should be using the string as an index.

For these applications, the value really is not an index. It's just a useful value associated with the enumerator. I don't think we should be using the value as an index.

For non integer values, I think we should be using the implementation 1. that is, enum is an index, and we have global arrays for enumerator name + values. I think we should also use this for integer values that are not unique, such as the MolecularMass example above, since we want things like "next" and "last" to work, so "last(oxygen)" would be True, but "last(methane)" would be False, even though both values are 16.

For integer values that are equal to the index (0..N-1) we should avoid the extra lookup. This I think takes care of most use cases with interfacing with C and keeping the same speed as C.

There is one last case not covered: integer values that are unique, but not consecutive and/or starting from 0. One example is our parser in `parser.tab.hh`:
```c
  enum yytokentype     
  {                
    END_OF_FILE = 0,     
    TK_NEWLINE = 258,   
    TK_NAME = 259,       
    TK_DEF_OP = 260,    
    TK_INTEGER = 261, 
...
    KW_WHERE = 493,
    KW_WHILE = 494,
    KW_WRITE = 495,
    UMINUS = 496
  };
```
For this, I think initially we should use the implementation 1., which should work. Later, we can possibly optimize it more.

czgdp1807:
Well if there are several use cases which should be implemented using approach 1 then covering only integer unique enum values using approach 2 should be done or not? Let me know what you think.

certik:
Based on the above, here seems to be the design:

| Case | enum type | implementation | C interoperation |
| ---------- | ---------- | ------------ | --- | 
| a) | Non integer values (str, float, tuple, arrays) unique or not | index 0..N-1 into global arrays | No |
| b) | integer values 0..N-1 | use values as index | Yes |
| c) | non-consecutive but unique integers | use values as index, use hash table | Yes |
| d) | non-unique integers | index 0..N-1 into global arrays | No |

C only allows integers and we need this type to be interoperable with C. 

So case a) is not interoperable and there is no problem.  Cases b), c) and d) must all interoperate with C, so we need to figure out a way. It turns out the case d) should not interoperate and we should use c) for that, see below.

Case b) is the most common and we must make it interoperable, and also for performance, we just use the value as an index. This case b) is also what we immediately need to deliver soon I think, including interoperation with C. I think there is no problem there.

The case c) is our parser. In order to interoperate, the easiest is to just treat the value as an index. But the index then has to be into a hashtable, not a linear array, to retrieve the name of the enumerator if needed, as well as next/prior/first/last.

The case d) is unclear to me right now. In C, I think you never go from the index back to the enumerator. In Python you do. How does Python handle the non-unique case of value->enumerator mapping? Does it pick the first enumerator that matches? The issue of using a value as an index here is that first/last will not work correctly. The fundamental issue is that our enumerator is semantically richer than the C enum: our enum is defined by the symbolic enumerators, while in C the enum is defined by their integer values, and you always go just one way: enumerator->value, never the other way. You do not loop over the enum, you do not have next/prior/first/last. This case is very rare, and we do not need to resolve everything here, we just need some design that does not close doors here. It seems that C just gives us the value integer, which however does not have a unique mapping into the enumerator. So one way to interoperate is to simply choose a unique mapping for a type that has to interoperate. For example, if in C we have "red = 1, green = 2, blue = 1", in ASR we represent this either with "red=1, green=2", or with "green=2, blue=1", both would work, and in both cases we have converted this case to c). 


@czgdp1807, let me know what you think.

If you agree, then from this we can then figure out a good ASR representation of this, to make it easy on the backends.

czgdp1807:
> If you agree, then from this we can then figure out a good ASR representation of this

I think ASR design is not a problem, its already fine and storing information as it is from the syntax. The thing we need to figure out is implementation at the backend level. I think we should start off with linear array implementation as it stores the Enum in a simple way and is easily applicable across all the cases you have listed. If it is slow then we can optimise it by using advanced data structure like hash map. 

So my recommendation now - Use simple approach like using a linear array of tuples (enum name, enum value) which works for all the cases. Then in future optimise for use cases where this linear array approach seems slow. Doing too much right now might be pre-mature. Whether we use hash map or linear array, I don't see any change in ASR design anyways. Let me know what you think. I will try to complete this PR by tomorrow so that we can deliver initial version of the feature and then optimisations can be performed later on based on our experiences.

certik:
Good idea, so let's see how it would work:

| Case | enum type | implementation | C interoperation |
| ---------- | ---------- | ------------ | --- | 
| a) | Non integer values (str, float, tuple, arrays) unique or not | index 0..N-1 into global linear arrays for enum strings and values | No |
| b) | integer values 0..N-1 | index 0..N-1 into global linear arrays for enum strings and values | Yes |
| c) | non-consecutive but unique integers | NOT IMPLEMENTED | Yes |
| d) | non-unique integers | index 0..N-1 into global linear arrays for enum strings and values | No |

That would work. We could also do:

| Case | enum type | implementation | C interoperation |
| ---------- | ---------- | ------------ | --- | 
| a) | Non integer values (str, float, tuple, arrays) unique or not | index 0..N-1 into global linear arrays for enum strings and values | No |
| b) | integer values 0..N-1 | index 0..N-1 into global linear arrays for enum strings and values | Yes |
| c) | non-consecutive but unique integers | index 0..N-1 into global linear arrays for enum strings and values | NO |
| d) | non-unique integers | index 0..N-1 into global linear arrays for enum strings and values | No |

None of that will work in all cases. We are not talking about optimizations, just satisfying use cases. The issue is just with the case c), which needs to be interoperable with C, and so we need to use the actual value as an index. The index can be either to a hashmap, or a linear array (the only downside is that one might run out of memory easily if the value is too large), either is fine with me.

czgdp1807:
So should I go ahead and implement the global array approach?

certik:
I think so, I think we need it for a), b) and d).

The ASR looks pretty good. I would change EnumRef to EnumName and EnumValue.

We can iterate on the details as we go.

czgdp1807:
Okay. For `c` we can make changes later (like implement optimal approach for it so that we don't consume too much memory).

certik:
Yes, for C interoperability, the ASR's `abi` member would be `BindC`, which is telling the backend to change the internal representation of the given node (in this case `Enum`) to be binary compatible with C. So we would then do whatever needs to be done to make it work.

czgdp1807:
I have updated the implementation and the ASR design according to the discussions here. I have tried my best to form the base of enums correctly. I will add subsequent features (support for tuples, strings enum values) in subsequent PRs. Feel free to push commits if you want to make any changes.

src/libasr/asr_verify.cpp
```diff
@@ -279,6 +280,14 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
         current_symtab = parent_symtab;
     }
 
+    void visit_DerivedType(const DerivedType_t& x) {
+        visit_DerivedTypeEnumType(x);
+    }
+
+    void visit_EnumType(const EnumType_t& x) {
+        visit_DerivedTypeEnumType(x);
+    }
```
certik, line 289: Let's add more checks here. 

* We require all members of `Enum` to have the `value` (?) to be set.
* We require all `value` to be of the same type, and that type to be equal to the `type` in `Enum`.

I think we do not have any other restrictions at this point. Later we will.

src/libasr/codegen/asr_to_c.cpp
```diff
@@ -490,6 +497,84 @@ R"(
...
+        int64_t min_value = INT64_MAX;
+        int64_t max_value = INT64_MIN;
+        size_t max_name_len = 0;
+        for( size_t i = 0; i < x.n_members; i++ ) {
+            ASR::symbol_t* member = x.m_symtab->get_symbol(x.m_members[i]);
+            LFORTRAN_ASSERT(ASR::is_a<ASR::Variable_t>(*member));
+            ASR::Variable_t* member_var = ASR::down_cast<ASR::Variable_t>(member);
+            ASR::expr_t* value = ASRUtils::expr_value(member_var->m_symbolic_value);
+            int64_t value_int64 = -1;
+            ASRUtils::extract_value(value, value_int64);
```
certik, line 516: This will fail if the value is not Integer, correct?

We need to give an error message here. 

I think to do this cleanly, why not store the `enum_type` in ASR directly, and in here, only support `IntegerConsecutiveFrom0` and `IntegerUnique`, and for everything else give an error message "not implemented yet"?

src/libasr/asr_verify.cpp
czgdp1807: > * We require all `value` to be of the same type, and that type to be equal to the `type` in `Enum`.

This is already done in AST to ASR transition as we cannot create EnumType without this condition being satisfied. But anyways I will add the checks here as well.

> We require all members of Enum to have the value (?) to be set.

Correct. Will add the checks.

src/libasr/codegen/asr_to_c.cpp
czgdp1807: > This will fail if the value is not Integer, correct?
> 
> We need to give an error message here.

Yes. Will do.



> I think to do this cleanly, why not store the `enum_type` in ASR directly, and in here, only support `IntegerConsecutiveFrom0` and `IntegerUnique`, and for everything else give an error message "not implemented yet"?

Yes we can but I think its too early to create an `enum_type`. I will wait for all the other cases to be supported and then do this.

czgdp1807: Done. Added error messages for both non-integer and non-unique integer cases.

src/libasr/asr_verify.cpp
czgdp1807: Done.

czgdp1807:
@certik I have addressed your concerns. Please let me know if anything else is to be done to get the initial version of Enums delivered.

src/libasr/asr_verify.cpp
certik: Thanks. The idea of verify is that it checks all conditions that we require ASR to satisfy. Of course, the idea is that the frontend satisfies them all, so that in release mode we do not need to run verify, for performance reasons. But for development and debugging we want to do that.

src/libasr/codegen/asr_to_c.cpp
certik: Perfect.

czgdp1807:
@certik So, should we go ahead and merge this after cleaning up git history?

certik (approved):
I think this is good to merge. Thanks @czgdp1807 for implementing this!

We can iteratively improve upon this once it is merged.

src/libasr/ASR.asdl
```diff
@@ -278,6 +281,8 @@ expr
 
     | BitCast(expr source, expr mold, expr? size, ttype type, expr? value)
     | DerivedRef(expr v, symbol m, ttype type, expr? value)
+    | EnumName(symbol v, ttype enum_type, ttype type, expr? value)
```
certik, line 284: Isn't `enum_type` the same as `symbol_type(v)`?

czgdp1807: No. `enum_type` is the original `EnumType` wrapped under `Enum` node. Type of `v` is same as member variable's type. `enum_type` is needed to fetch the global array in C and LLVM backend to lookup the name and values for non-integers.

czgdp1807:
Initial version is in. Subsequent additions, bug fixes will keep happening.

src/libasr/ASR.asdl
certik: `enum_type` seems to be the same as `symbol_symtab(v)->asr_owner`.

certik:
Awesome, thanks @czgdp1807 for merging this!

src/libasr/ASR.asdl
czgdp1807: Its not always the case. For the following test for example,

https://github.com/lcompilers/lpython/blob/2d52f5181287ab5553cda0333aabfac50e623f28/integration_tests/enum_01.py#L15-L19

ASR owner of `color` variable is the function, `test_selected_color` and not `Color` enumeration. I can distinguish between the two in the backend i.e., when the variable is of Enum type then extract EnumType from it otherwise do `symbol_symtab(v)->asr_owner` but then it would be boilerplate for the backend which could be easily avoided by storing `enum_type` in the ASR node itself. Let me know what you think.

certik: It seems there are two use cases `EnumName(color, ...)` and `EnumName(Color.RED, ...)`. The `enum_type` should be `Color` in both, correct?

Isn't `enum_type` just equal to `symbol_type(v)`? If not, there must be a way to get to it, just try to figure it out.

czgdp1807: As you have pointed there are two use cases,

1. `EnumName(color, ...)` - Here `enum_type = symbol_type(color)`, so not a problem.
2. `EnumName(Color.RED, ...)` - Here `enum_type != symbol_type(Color.RED)`, because `Color.RED` is of `i32` type (the common member type of an Enum). `Color.RED` is a `Variable` which has `m_parent_symtab` which is the scope of `Enum`. But `Variable.m_parent_symtab->asr_owner` should be the `EnumType`.

So clearly you will have to do,

```
if( ASR::is_a<ASR::Enum_t>(*ASRUtils::symbol_type(x.m_v) ) {
    // Extract EnumType from Enum_t.
} else {
   // Do Variable.m_parent_symtab->asr_owner
}
```

The above if-else check will happen whenever you want to access the global enum values and name array. That's why I think we shouldn't use this if-else check but instead store it in `EnumName` and `EnumValue` node as `ttype enum_type`. What do you think? Do you want me to use the above if-else check instead?

czgdp1807: Opened https://github.com/lcompilers/lpython/pull/1133. Seems like the if-else check is simple enough, encapsulated it in a function and called it from backends. So I am going to merge the PR anyways.

certik: I see. It's good enough for now. I am still not sure about this, it seems there is a better way, but it's good enough for now.

PR #1118, opened 2022-09-16
WASM: Improvements for mandelbrot

Shaikh-Ubaid:
This `PR` adds improvements to the `wasm` backend to support the `mandelbrot` example at https://www.ubaidshaikh.me/lcompilers_web_frontend/

### DIFF
--- a/src/libasr/codegen/asr_to_wasm.cpp
+++ b/src/libasr/codegen/asr_to_wasm.cpp
@@ -1411,7 +1411,7 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
         for (uint32_t i = 0; i < x.n_args; i++) {
             if (x.m_args[i].m_right) {
                 this->visit_expr(*x.m_args[i].m_right);
-                wasm::emit_i32_const(m_code_section, m_al, 1);
+                this->visit_expr(*m_dims[i].m_start);
                 wasm::emit_i32_sub(m_code_section, m_al);
                 size_t jmin, jmax;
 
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/asr_to_wasm.cpp
```diff
@@ -1411,12 +1422,12 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
...
             if (x.m_args[i].m_right) {
                 this->visit_expr(*x.m_args[i].m_right);
-                wasm::emit_i32_const(m_code_section, m_al, 1);
+                this->visit_expr(*m_dims[i].m_start);
                 wasm::emit_i32_sub(m_code_section, m_al);
                 size_t jmin, jmax;
 
                 // TODO: add this flag to ASR for each array:
-                bool column_major = true;
+                bool column_major = false;
```
Shaikh-Ubaid, line 1430: This is a temporary change. We need to support the array storage ordering (row/col major) in the `ASR` soon.

Shaikh-Ubaid:
This is ready. Please possibly review and please share  feedback.

Shaikh-Ubaid:
https://www.ubaidshaikh.me/lcompilers_web_frontend/ now uses `mandelbrot` example as default example.

src/libasr/codegen/asr_to_wasm.cpp
certik: This change will break LFortran I think. That is fine, we can diverge for a bit, we have done that in the past.

@czgdp1807 when syncing with LFortran, avoid this change. We will create a general solution soon.

certik (approved):
Amazing work!

Issue #723, opened 2022-07-02
Update lpython to use optimization passes


Smit-create:
LFortran supports optimization passes like `loop_unroll`, `fma`, etc. which are already supported in `libasr`. We should update LPython to make use of these optimization passes.

cc @czgdp1807

certik:
The new pass manager will allow it, we just have to integrate it.

czgdp1807:
LPython already uses the passes which you have mentioned. The interface to input the list of passes is exactly the same for LPython and LFortran. Use command line argument, `—pass`.

czgdp1807:
In addition also see, `asr_to_llvm.cpp` in both the compilers. They both call the pass manager to apply same set of passes.

Smit-create:
On the main branch, I get:
```
% lpython --show-asr expr.py --indent --pass=fma
Pass must be one of: do_loops, global_stmts, implied_do_loops, array_op, class_constructor, print_arr, arr_slice, unused_functions
```

On updating the `bin/lpython` to use fma, I get:
```
LFortranException: Module 'lfortran_intrinsic_optimization' not declared in the current source and the modfile was not found
```

czgdp1807:
Here's what I get,

```bash
(lp) 16:48:31:~/lpython_project/lpython % lpython --pass=fma --show-asr tests/expr16.py
[...]

PR #1119, opened 2022-09-17
Implement list_concat ASR pass

Thirumalai-Shaktivel:


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2237,6 +2237,59 @@ namespace LFortran {
...
+                        llvm::ConstantInt::get(context, llvm::APInt(32, 1)));
+            LLVM::CreateStore(*builder, tmp, pos_ptr);
+        }
+        builder->CreateBr(loophead);
+
+        // end
+        llvm_utils->start_new_block(loopend);
+
+        return new_list;
+    }
```
Thirumalai-Shaktivel, line 2291: Logic: I create a new list, deep copy left list, and append all the elements of the right to the new list.

czgdp1807: Following approaches exist,

1. ASR pass - List concatenation can be easily implemented using Python's syntax. So we can use ASR APIs to replace ListConcat with a temporary result variable. This result variable can be filled using two do loops one which iterates over first list and the other which iterates over second list. Can be easily optimised by tracking if we need a temporary variable or the result can go into some variable on the left hand side of an assignment statement. Pros - Can be shared across all backends at once. Cons - Re-filling the result list using append but not that big of a deal because as the list grows time complexity averages out. For small lists this is not a problem at all.
2. Implementing ListConcat in each backend as you have done for LLVM - We will be able to use `memcpy` and hence this approach might be faster. But that will only work for simple lists of primitive types. For types like `list[str]`, `list[tuple[...]]`, benefits cannot be availed.

I am leaning towards ASR pass approach because our `append` is already fast for large lists, so benefits of second approach should be marginal.

Thirumalai-Shaktivel: I see, then let's work on `ASR pass` for `List Concat`.
I didn't understand properly, how to implement the `ASR pass`?

Thirumalai-Shaktivel: You meant is that to create a function is Python syntax, something like:
```
def list(left: list[i32], right: list[i32]) -> list[i32]:
    l: list[i32] = []
    i: i32
    for i in left:
        l.append(i)
    for i in right:
        l.append(i)
    return l
```
and link this to LLVM backend?

certik: I think @czgdp1807 means to create ASR->ASR pass that takes ListConcat ASR node and replaces it with a function call `_lcompilers_list_concat` which implements the list concatenation in Python itself. Then all backends will automatically work with this.

One problem is that we need to figure out a way to make this pass work with LFortran also. But I thin we should figure it out, so that we can share more functions in ASR across compilers and so that we can write more of such passes.

Thirumalai-Shaktivel: Got it, thank you!

czgdp1807: > I think @czgdp1807 means to create ASR->ASR pass that takes ListConcat ASR node and replaces it with a function call `_lcompilers_list_concat` which implements the list concatenation in Python itself. Then all backends will automatically work with this.

And for that we need generics. We already have those, right? Other way is to do pragmatically using ASR APIs, then you already know what type the lists are and hence generate ASR in the ASR->ASR pass directly.

czgdp1807: > One problem is that we need to figure out a way to make this pass work with LFortran also.

LFortran has lists? I am not sure. If Fortran doesn't has lists or anything similar then this pass won't have any effect on Fortran code.

certik: People keep asking for lists, so we might provide it as an extension.

Also I was thinking providing an optimization that recognizes array usage that would be faster with lists (appending to the end and reallocating), and replace it with a list.

All ASR passes must work with all compilers.

Equivalently: nothing in libasr should be dependent on any one frontend.

We need to design things to work with all frontends.

czgdp1807: > I think @czgdp1807 means to create ASR->ASR pass that takes ListConcat ASR node and replaces it with a function call `_lcompilers_list_concat` which implements the list concatenation in Python itself. Then all backends will automatically work with this.

I will generate the code for `_lcompilers_list_concat` using ASR APIs, the way we do for vectorisation optimisation. We faced a similar problem with optimisation implemented in LFortran (I listed multiple approaches in https://github.com/lcompilers/lpython/issues/723#issuecomment-1172909436) and we agreed to use ASR APIs (I commented our agreed solution in https://github.com/lcompilers/lpython/issues/723#issuecomment-1181815465) to generate the fallback code so that they can be shared across all the LCompilers. Its the same case here in this PR so I will use the same approach we agreed upon in the past. Do you still want to follow the idea in https://github.com/lcompilers/lpython/issues/723#issuecomment-1181815465 or is there any other approach that can be better?

certik: I think that approach still works. I opened up a dedicated issue for this: https://github.com/lcompilers/lpython/issues/1150

czgdp1807: Cool. Let's try it out here as an experiment (it will work most probably but we will gain more experience if we start earlier). @Thirumalai-Shaktivel Would you like to try it out here? Feel free to let me know if you get stuck anywhere. For replacing expressions you might need to follow the pattern in. Replacing function calls is same as replacing ListConcat as both expression nodes.

https://github.com/lcompilers/lpython/blob/91cb772ef86a58d8b5020a8dde864754bc06d06a/src/libasr/pass/pass_array_by_data.cpp#L303

Thirumalai-Shaktivel: Sure, I will give it a try tomorrow.

Thirumalai-Shaktivel: Sorry, I was working on the parser issue (match statement). I start working tomorrow and report back.

czgdp1807: No worries. Its not a critical feature IMO, so take as much time as you need to come up with a nice solution.

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+        Allocator& al;
+
+    public:
+        ReplaceListConcat(Allocator &al_) : al(al_)
+        { }
+
+        void replace_ListConcat(const ASR::ListConcat_t* x) {
+            Location loc = x->base.base.loc;
+            *current_expr = nullptr;
```
Thirumalai-Shaktivel, line 41: @czgdp1807 I got questions:
- What is the use of `replace_<node>()`?
- If I place `nullptr` for the `current_expr`, will it remove the `ListConcat` node from ASR?
- Can you please explain the `replacer`?


Does this template look good to start?

czgdp1807: > What is the use of replace_<node>()?

Well its the method where you have to implement the replacement logic according to the purpose at hand. For example in `replace_ListConcat` you will implement the logic to replace it with the call to the function which is supposed to implement list concatenation.

> * If I place `nullptr` for the `current_expr`, will it remove the `ListConcat` node from ASR?

Well it will replace `ListConcat` with `nullptr`. Instead you should replace it with a `FunctionCall` node.

> Can you please explain the replacer?

Done above.

> How do I verify the asr output? using --show-asr, right?

Only `--show-asr` will give you original ASR. You should also provide `--pass=name_of_the_pass` to get the modified ASR once the pass is applied.

Thirumalai-Shaktivel: I got it, thanks.

src/libasr/pass/pass_manager.h
```diff
@@ -74,6 +75,7 @@ namespace LCompilers {
             {"select_case", &LFortran::pass_replace_select_case},
             {"loop_vectorise", &LFortran::pass_loop_vectorise},
             {"array_dim_intrinsics_update", &LFortran::pass_update_array_dim_intrinsic_calls},
+            {"pass_list_concat", &LFortran::pass_list_concat},
```
Thirumalai-Shaktivel, line 78: Why does it generate an error?
```
error: could not convert ‘{{"do_loops", LFortran::pass_replace_do_loops}, {"global_stmts", LFortran::pass_wrap_global_stmts_into_function}, {"implied_do_loops", LFortran::pass_replace_implied_do_loops}, {"array_op", LFortran::pass_replace_array_op}, {"arr_slice", LFortran::pass_replace_arr_slice}, {"print_arr", LFortran::pass_replace_print_arr}, {"print_list", LFortran::pass_replace_print_list}, {"class_constructor", LFortran::pass_replace_class_constructor}, {"unused_functions", LFortran::pass_unused_functions}, {"flip_sign", LFortran::pass_replace_flip_sign}, {"div_to_mul", LFortran::pass_replace_div_to_mul}, {"fma", LFortran::pass_replace_fma}, {"sign_from_value", LFortran::pass_replace_sign_from_value}, {"inline_function_calls", LFortran::pass_inline_function_calls}, {"loop_unroll", LFortran::pass_loop_unroll}, {"dead_code_removal", LFortran::pass_dead_code_removal}, {"forall", LFortran::pass_replace_forall}, {"select_case", LFortran::pass_replace_select_case}, {"loop_vectorise", LFortran::pass_loop_vectorise}, {"array_dim_intrinsics_update", LFortran::pass_update_array_dim_intrinsic_calls}, {"pass_list_concat", LFortran::pass_list_concat}, {"pass_array_by_data", LFortran::pass_array_by_data}}’ from ‘<brace-enclosed initializer list>’ to ‘std::map<std::__cxx11::basic_string<char>, void (*)(Allocator&, LFortran::ASR::TranslationUnit_t&, const LCompilers::PassOptions&)>’
   80 |         };
      |         ^
      |         |
      |         <brace-enclosed initializer list>
```

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,70 @@
...
+        ListConcatVisitor(Allocator& al_) : replacer(al_) {}
+
+        void call_replacer() {
+            replacer.current_expr = current_expr;
+            replacer.replace_expr(*current_expr);
+        }
+
+};
+
+void pass_list_concat(Allocator &al, ASR::TranslationUnit_t &unit) {
```
czgdp1807, line 63: You are not accepting PassOptions. See how other passes are declared and defined. Follow the same pattern here.

src/libasr/pass/pass_manager.h
czgdp1807: https://github.com/lcompilers/lpython/pull/1119/files#r983134603

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,114 @@
...
+            /* a_return_var */ return_var_ref,
+            ASR::abiType::BindC,
+            ASR::Public, ASR::Implementation,
+            nullptr,
+            false, false, false, false, false,
+            nullptr, 0,
+            nullptr, 0,
+            false);
+            unit.m_global_scope->add_symbol(fn_name, down_cast<ASR::symbol_t>(fn));
+        }
```
Thirumalai-Shaktivel, line 68: > I will generate the code for `_lcompilers_list_concat` using ASR APIs

@czgdp1807, you meant to do something like this, right?

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,114 @@
...
+            ASR::call_arg_t left_list, right_list;
+            left_list.loc = x->m_left->base.loc;
+            left_list.m_value = x->m_left;
+            a_args.push_back(al, left_list);
+            right_list.loc = x->m_right->base.loc;
+            right_list.m_value = x->m_right;
+            a_args.push_back(al, right_list);
+            create_concat_function();
+            *current_expr = ASRUtils::EXPR(ASR::make_FunctionCall_t(al, loc,
+                                           nullptr, nullptr, a_args.p, 2,
```
Thirumalai-Shaktivel, line 83: Later, use that function `_lcompilers_list_concat` here to concat the list, right?

src/libasr/pass/pass_list_concat.cpp
Thirumalai-Shaktivel: Or 
```py
def _lcompilers_list_concat(left: list[i32], right: list[i32]) -> list[i32]:
    l: list[i32] = []
    i: i32
    for i in left:
        l.append(i)
    for i in right:
        l.append(i)
    return l
```
Add the above to function to `src/runtime/lpython_builtin.py`(or something similar) and call that func in this pass

Thirumalai-Shaktivel:
`ASR verify failed` is fixed.

Thirumalai-Shaktivel:
This is ready for review!

czgdp1807:
The tests are failing.

examples/expr2.py
```diff
@@ -1,10 +1,6 @@
 def main0():
-    x: i32
-    x = (2+3)*5
+    i: list[i32] = [1, 2, 3]
+    j: list[i32] = [4, 5, 6]
+    x: list[i32] = j + i
     print(x)
```
Thirumalai-Shaktivel, line 5: This example fail in the llvm with `AssertFailed: std::find(nested_globals.begin(), nested_globals.end(), x_h) != nested_globals.end()`

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,222 @@
...
+            /* a_return_var */ arg_expr,
+            ASR::abiType::Source,
+            ASR::accessType::Public, ASR::deftypeType::Implementation,
+            nullptr,
+            false, false, false, false, false,
+            nullptr, 0,
+            nullptr, 0,
+            false);
+            global_scope->add_symbol(fn_name, down_cast<ASR::symbol_t>(fn));
+            return ASR::down_cast<ASR::symbol_t>(fn);
```
Thirumalai-Shaktivel, line 174: Is this the correct approach to implement this?

integration_tests/test_list_09.py
```diff
@@ -0,0 +1,12 @@
+from ltypes import i32
+
+def test_list_concat():
+    x: list[i32] = [1, 2, 3]
+    y: list[i32] = [4, 5, 6]
+    z: list[i32] = x + y
+
+    i: i32
+    for i in range(1, 7):
+        assert z[i-1] == i
```
czgdp1807, line 10: Please increase the strength of the tests. Cover the following cases,

1. Either list is empty.
2. Both lists are empty.
3. Both lists are not of the same length (first one small, then second one small).
4. More than two lists are getting concatenated. (say x + y + z).
5. All operations are deepcopy by default. Change `x` or `y` and `z` should remain the same.

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+            // Empty List
+            ASR::expr_t* value = ASRUtils::EXPR(ASR::make_ListConstant_t(
+                al, loc, nullptr, 0, list_type));
+            // Initialize `result_list` with `empty value`
+            ASR::stmt_t* list_concat_stmt = ASRUtils::STMT(ASR::make_Assignment_t(
+                al, loc, arg_expr, value, nullptr));
+            body.push_back(al, list_concat_stmt);
+
+            // Declare `i` for iterations
```
czgdp1807, line 105: Please use `void PassUtils::create_idx_vars(Vec<ASR::expr_t*>& idx_vars, int n_dims, const Location& loc,
                             Allocator& al, SymbolTable*& current_scope, std::string suffix="_k");` for creating auxiliary index variables.

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,222 @@
...
+                    result_list = []
+                    i: i32
+                    for i in range(len(left_list)):
+                        result_list.append(left_list[i])
+                    for i in range(len(right_list)):
+                        result_list.append(right_list[i])
+                    return result_list
+            */
+            SymbolTable* list_concat_symtab = al.make_new<SymbolTable>(global_scope);
+            std::string fn_name = "_lcompilers_list_concat";
```
czgdp1807, line 56: Do, `global_scope->get_unique_name("_lcompilers_list_concat")`.

czgdp1807 (changes requested):

src/libasr/pass/pass_list_concat.cpp
czgdp1807: Seems like it. Would comment once you increase the strength of your tests.

integration_tests/test_list_09.py
Thirumalai-Shaktivel: In order to test this, `LLVM` should be working, right?

Thirumalai-Shaktivel: yup, Working on it.

czgdp1807: Correct. I think LLVM should be working to consider this feature as complete.

src/libasr/pass/pass_list_concat.cpp
Thirumalai-Shaktivel: I got it! Done.

src/libasr/pass/pass_list_concat.cpp
Thirumalai-Shaktivel: Done

integration_tests/test_list_09.py
Thirumalai-Shaktivel: I didn't understand the 5th point:
```
x = [1, 2, 3]
y = [3, 4, 5]
z = x + y
x = [10, 20, 30]
```
you mean the above test, right? After `x` is altered, `z` should be `[1, 2, 3, 4, 5, 6]`, right?

czgdp1807: Also,
```
x = [1, 2, 3]
y = [3, 4, 5]
z = x + y
x[0] = 6
```

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,220 @@
...
+            /* a_return_var */ arg_expr,
+            ASR::abiType::Source,
+            ASR::accessType::Public, ASR::deftypeType::Implementation,
+            nullptr,
+            false, false, false, false, false,
+            nullptr, 0,
+            nullptr, 0,
+            false);
+            global_scope->add_symbol(fn_name, down_cast<ASR::symbol_t>(fn));
+            return ASR::down_cast<ASR::symbol_t>(fn);
```
Thirumalai-Shaktivel, line 172: After applying the suggestions. I get `LCompilersException: visit_DoLoop() not implemented`. So, on the global scope, `DoLoop` is not implemented. So, in this case, we have to create a function scope, right?

czgdp1807: I think the `do_loop` pass is not getting applied to your function. But it gets applied to vector copy intrinsic optimisation fallback. So probably there is something incorrect here. Can you try to figure it out?

Thirumalai-Shaktivel: Got it, I will try to debug it.

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,220 @@
...
+            args.reserve(al, 2);
+            ASR::call_arg_t left_list, right_list;
+            left_list.loc = x->m_left->base.loc;
+            left_list.m_value = x->m_left;
+            args.push_back(al, left_list);
+            right_list.loc = x->m_right->base.loc;
+            right_list.m_value = x->m_right;
+            args.push_back(al, right_list);
+            ASR::symbol_t* fn_name = create_concat_function(unit.base.base.loc,
+                unit.m_global_scope, x->m_type);
```
Thirumalai-Shaktivel, line 187: Here, I pass `unit.m_global_scope` which I assume is a global_scope.
But in vector copy intrinsic optimization fallback, you pass `current_scope`. So that is the reason for the failure, I think.

integration_tests/test_list_09.py
```diff
@@ -1,12 +1,43 @@
...
+    print(z)
+
+    x = [1, 2, 3, 4, 5]
+    z = x + y
+    for i in range(1, 9):
         assert z[i-1] == i
+    print(z)
+
+    x = [9, 10]
+    y = [11, 12]
```
czgdp1807, line 32: Please generate `x` and `y` using loops. The size should be at least 50ish so that we know things are not hard coded in the implementation.

czgdp1807 (changes requested):

src/libasr/pass/pass_list_concat.cpp
Thirumalai-Shaktivel: The `current_scope` is availabe in `PassVisitor` class. I need to assess it somehow, right? Or is there any other way?

Thirumalai-Shaktivel: How do I get the scope in which the `ListConcat` is available?

Thirumalai-Shaktivel: Okay, So I should not worry about the `ListConcat` scope. Why? Because we have to create the function (_lcompilers_list_concat) in the `global_scope`. Then create a new function scope and declare all the variable using that scope. I think this might work

Thirumalai-Shaktivel: I don't know, how to fix this.

Thirumalai-Shaktivel: I got it! So, Our DoLoop is not visited in the BodyVisitor. Instead as @czgdp1807 said it is taken care by the `pass`.

integration_tests/test_list_09.py
```diff
@@ -0,0 +1,38 @@
+from ltypes import i32
+
+def test_list_concat():
+    x: list[i32] = []
+    y: list[i32] = []
+    z: list[i32] = x + y
```
Thirumalai-Shaktivel, line 6: `print(z)` throws `LCompilersException: Printing support is not available for list type.`

Thirumalai-Shaktivel: If I replace it with
```py
for i in range(len(z)):
	print(z[i])
```
it throws: `LCompilersException: visit_DoLoop() not implemented`

Thirumalai-Shaktivel: Sorry for the confusion. I didn't use the `print_list` and `do_loops` passes. After passing the passes it works.

Thirumalai-Shaktivel:
This PR is ready!
@czgdp1807, please review this PR.
If everything looks good let me know. I will clean up the commits.

src/libasr/pass/pass_manager.h
```diff
@@ -96,6 +98,7 @@ namespace LCompilers {
                 "class_constructor",
                 "implied_do_loops",
                 "pass_array_by_data",
+                "pass_list_concat",
```
Thirumalai-Shaktivel, line 101: Should I add "pass_list_concat" to `_with_optimization_passes`?

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,234 @@
...
+            al, loc,
+            /* a_symtab */ list_concat_symtab,
+            /* a_name */ s2c(al, fn_name),
+            /* a_args */ arg_exprs.p,
+            /* n_args */ arg_exprs.n,
+            /* a_body */ body.p,
+            /* n_body */ body.n,
+            /* a_return_var */ arg_expr,
+            ASR::abiType::Source,
+            ASR::accessType::Public, ASR::deftypeType::Implementation,
```
Thirumalai-Shaktivel, line 177: Is this correct?

czgdp1807 (approved):
Other than my only comment the code and the tests seem fine to me.

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,234 @@
...
+        do_loop_head.loc = loc;
+        Vec<ASR::stmt_t*> loop_body;
+        loop_body.reserve(al, 1);
+        ASR::stmt_t* list_append = ASRUtils::STMT(ASR::make_ListAppend_t(
+            al, loc, arg_expr, ASRUtils::EXPR(ASR::make_ListItem_t(
+                al, loc, arg_exprs[0], do_loop_head.m_v, item_type, nullptr))));
+        loop_body.push_back(al, list_append);
+        list_concat_stmt = ASRUtils::STMT(ASR::make_DoLoop_t(
+            al, loc, do_loop_head, loop_body.p, loop_body.n));
+        Vec<ASR::stmt_t*> list_while_loop = PassUtils::replace_doloop(
```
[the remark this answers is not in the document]
Thirumalai-Shaktivel: Perfect, I'm working on it.

Thirumalai-Shaktivel:
This PR is ready!

src/libasr/pass/pass_list_concat.cpp
czgdp1807: Okay. Wait. So for every list concatenation operation the, `create_list_concat` function is called and hence everytime `global_scope->add_symbol` will be called. I think we need to add it only once. I think you can add a flag in this class and if its false then only execute the body of this execution otherwise just do return. Does that make sense? Or you already handle this?

Thirumalai-Shaktivel: Yup, it make sense. Sorry, I totally forgot it.
Also, instead of using flag. Why not check the symtab whether the function is created. If so, we pass or else create one. Is it possible to do that?

czgdp1807: So here are the cases to consider,

1. `_lcompilers_list_concat` is already defined by the user and is present in the `global_scope`. In this case `global_scope->resolve_symbol` will always give non-`nullptr` result. It will be hard to distinguish between user's function and what we want to add.
2. `_lcompilers_list_concat` is not defined by the user. In this case it is possible to just query the `global_scope` and get the correct result.

So if we receive non-`nullptr` result we never know whether its user's function or our own function. 

Though note that I am not sure if its even possible for the users to add a function to global scope. But I would play safe since if its not possible for the users to add functions to global scope, it might become possible in future and then it will fail.

Thirumalai-Shaktivel: I agree.

Thirumalai-Shaktivel: For every `ListConcat`, this class will be initialize. Then, how do I store the flag?

czgdp1807: I think the flag can be stored in `ListConcatVisitor` and while initialising `ReplaceListConcat` pass it into its constructor. So we won't depend on static variables and initialise the list concat function only once. Make sense?

czgdp1807: > For every `ListConcat`, this class will be initialize. Then, how do I store the flag?

I don't think. The replacer is initialised only once along with list concat visitor. Anyways its still possible to use the flag approach, just keep at the right level of control flow.

Thirumalai-Shaktivel: I applied your suggestions

src/libasr/pass/pass_list_concat.cpp
```diff
@@ -0,0 +1,247 @@
...
+public:
+    ReplaceListConcat(Allocator &al_, ASR::TranslationUnit_t &unit_,
+                      bool is_list_concat_defined_) :
+        al(al_), unit(unit_), is_list_concat_defined(is_list_concat_defined_)
+        { }
+
+    ASR::symbol_t* create_concat_function(Location& loc,
+            SymbolTable*& global_scope, ASR::ttype_t* list_type) {
+        if (is_list_concat_defined) {
+            return global_scope->get_symbol("_lcompilers_list_concat");
```
Thirumalai-Shaktivel, line 47: What if the user created the function "_lcompilers_list_concat"

Thirumalai-Shaktivel: In that case, I think instead of using the flag, we should store the function name. This might helo us.

Thirumalai-Shaktivel: Done.

Thirumalai-Shaktivel:
@czgdp1807 Does this look good to you?
[`45a7e01` (#1119)](https://github.com/lcompilers/lpython/pull/1119/commits/45a7e01be653132b3804da689157ac7ce97ddf79)

czgdp1807 (approved):

Thirumalai-Shaktivel:
Thank you for the review and guidance.

PR #1123, opened 2022-09-17
PASS: Improve print tuple

Shaikh-Ubaid:
towards #882.

This PR adds support of printing tuples to the `ASR` pass `print_list` (this pass is renamed to `print_list_tuple` in this `PR`).


Shaikh-Ubaid:

czgdp1807 (changes requested):
So here's the problem, `src/libasr/pass/print_list.cpp` moved to `src/libasr/pass/print_list_tuple.cpp` which means printing of `list` and `tuple` is tightly coupled now. But they can be easily de-coupled. Printing of lists ASR pass shouldn't depend upon on the list's element's type.  So consider this example,

```python
x: list[tuple[f64, i32, str]]
print(x)
```

First printing of list pass should be applied which will give,

```python
x: list[tuple[f64, i32, str]]
print("[")
for i in range(len(x)):
    print(x[i], end="")
    if i < len(x) - 1:
        print(", ")
print("]")
```

Now printing tuples should be applied on the above result,

```python
x: list[tuple[f64, i32, str]]
print("[")
for i in range(len(x)):
    print(x[i][0], x[i][1], x[i][2] end="")
    if i < len(x) - 1:
        print(", ")
print("]")
```

Similarly the scene would be reversed if I create `tuple` of `list` i.e., `tuple[list[i32], list[f64]]`. Basically, aggregate data structure printing is tricky, you need to keep applying these in a group until ASR stops changing.

Shaikh-Ubaid:
>Basically, aggregate data structure printing is tricky, you need to keep applying these in a group until ASR stops changing.

Please, could you possibly share if you mean to apply two `passes print_list` and `print_tuple` alternatively untill the ASR stops changing? I am concerned if it would be efficient to apply the passes multiple times.

czgdp1807:
Well, I would say try thinking of an approach that would print the following types easily,

1. `list[dict[i32, tuple[i32, f64]]]`
2. `tuple[list[i32], dict[i32, f64]]`
3. `dict[tuple[i32, str], list[dict[i32, i64]]]`
4. .
5. .
6. .
7. Any combination of aggregate data types like `list`, `tuple`, `dict` and so on.

Shaikh-Ubaid:
Also, I think the pass currently does not support `tuple` of lists (example: `tuple[list[i32], list[f64]]`). I assumed that since a tuple is immutable, it could not contain a list (because lists are mutable). I checked with `CPython` and it seems tuple of lists is allowed.

czgdp1807:
See, 

```python
Python 3.9.6 (default, Aug  5 2022, 15:21:02) 
[Clang 14.0.0 (clang-1400.0.29.102)] on darwin
>>> l = [0, 1, 2]
>>> m = [3, 4, 5]
>>> t = (l, m)
>>> t[0]
[0, 1, 2]
>>> t[1]
[3, 4, 5]
>>> t[0][0] = 6
>>> t
([6, 1, 2], [3, 4, 5])
```

Shaikh-Ubaid:
>Well, I would say try thinking of an approach that would print the following types easily (doesn't matter what the solution will be, list all and figure out trade offs between them),

Got it. I will work on it and report back.

Also, I had a concern. It seems different langauges could prefer to print different types in their own/respective format. For example, `Python` prints lists using `[]` (square brackets), where as there could be another language that could prefer to print its lists using `()` (parentheses notation) (I could not find a langauge but I guess there is a possibility). Or there could be a langauge that prefers to print its set using `()` (Python uses `{}`).

Is it possible to implement the logic to print `lists`, `sets`, `tuples` and `dict` in the respective frontend languge. For example, in case of `lpython`, can we implement it as a part of `src/runtime/lpython_builtin.py`?

czgdp1807:
I think implementing in frontend language will be better as you can leverage recursion. But then you will loose the `print` statement itself as you will be required to call the frontend language function. cc: @certik

certik:
Print is a tough one, that we should probably meet and brainstorm.  We need a solid Print node (I think) that works across languages. Let's evaluate the approaches, pros/cons and then choose a good design.

src/libasr/pass/print_list_tuple.cpp
```diff
@@ -0,0 +1,277 @@
...
+namespace LFortran {
+
+/*
+This ASR pass replaces print list with print every value,
+comma_space, brackets and newline. The function
+`pass_replace_print_list_tuple` transforms the ASR tree in-place.
+
+Converts:
+
+    print(l, sep="pqr", end="xyz") # l is a list
```
certik, line 18: The pass could be transformed to take `StringFormatPython1("%s", l)` and transform it into manual string creation.

certik:
The advantage of implementing the string formatting as  a pass is that it works for all backends and it simplifies the backends. At the same time, the pass is optional, so if a given backend decides to implement it directly, it does not need to run the pass.

czgdp1807:
Do you want to implement a generic solution for the cases listed in https://github.com/lcompilers/lpython/pull/1123#issuecomment-1250084311 by using the approach in #1129.

Shaikh-Ubaid:
>Do you want to implement a generic solution for the cases listed in https://github.com/lcompilers/lpython/pull/1123#issuecomment-1250084311 by using the approach in https://github.com/lcompilers/lpython/issues/1129.

I would love to.

Smit-create:
@Shaikh-Ubaid With the current main branch and a slight modification of the added [test here](https://github.com/lcompilers/lpython/pull/1123/files#diff-4052c87548ed1df7433b7cb2ff5acb09b684cfc23d38e1bea3dd33eac8a09a66), I'm able to produce correct results with `lpython`.

```diff
>     assert (t[1] - f64(size//2 - 1)) < 1e-10
---
<     assert t[1] == size//2 - 1
```

Can you please just reset this branch and add a test with the above diff to make this ready to go in?

Shaikh-Ubaid:
>Can you please just reset this branch and add a test with the above diff to make this ready to go in?

Done.

Shaikh-Ubaid:

czgdp1807 (approved):

Smit-create (approved):
Thank you, LGTM 👍

certik (approved):

PR #1125, opened 2022-09-20
User-defined restrictions (clean history)

ansharlubis:
The cleaned version of the PR #1048.


ansharlubis:
Notes from @certik:
In subsequent PRs we can also then provide some explicit functions (add/zero/etc) for integers, and use the general mechanism even for integers, and then remove the special casing for integers from the instantiation code.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -669,6 +670,49 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                            std::string msg = "The restriction " + keyword_arg + 
+                                " can't be assigned with a non-function " + arg_id;
+                            throw SemanticError(msg, loc);
+                        }
+                    }
+                }
+            }
+        }
+    }
+    */
```
certik, line 714: What is the purpose of this function? It does not seem to be used anywhere.

ansharlubis: Sorry, I forgot to remove the function after merging.

certik (approved):
I reviewed this, locally as well, it's a huge change and I think we can merge it. Some things might have broke, I am not 100% sure, but I think the big issues are done, so let's merge it, and we can fix up later if anything broke.

certik:
I merged it. Thank you @ansharlubis for this amazing work. This lays the foundation to generics. 

@czgdp1807, if you notice anything that broke due to this PR, just report it, and we'll fix it. There might be some small things.

@Oshanath, we can now use this in LFortran, it does the heavy lifting.

PR #1126, opened 2022-09-20
Supporting Union in C and LLVM backend

czgdp1807:
https://github.com/lcompilers/lpython/issues/1113


integration_tests/union_01.py
```diff
@@ -0,0 +1,17 @@
+from ltypes import Union, i32, i64, f64, f32, ctype
+
+class u_type(Union):
+    _fields_ = ("integer32", ctype(i32)), ("real32", ctype(f32)), ("real64", ctype(f64)), ("integer64", ctype(i64))
```
certik, line 4: This is just syntax (so the least important part of the PR), but isn't this more readable:
```suggestion
@ccall
class u_type(Union):
    integer32: i32
    real32: f32
    real64: f64
    integer64: i64
```
The `@ccall` is telling LPyhon that this `u_type` Union has to be interoperable with C. Without the decorator it does not have to be.

Then Union I think doesn't care about the order of the elements (but if it does, we'll use the order as written, just like for a struct).

certik: I think this might be using some CTypes style. I would use our own simpler style, just like with the `@ccall` decorator, where we use our own natural style using Python annotations, but then underneath we implement it using ctypes.

czgdp1807: LPython can work with any syntax. We need it to work with CPython, so if `ccall` syntax works with CPython I will also prefer it.

czgdp1807: I updated the syntax and `ltypes.py` to make CPython work with this new syntax.

integration_tests/union_02.py
```diff
@@ -0,0 +1,48 @@
...
+class C:
+    cx: i64
+    cy: f64
+    cz: f64
+
+@ccall
+class D(Union):
+    a: A
+    b: B
+    c: C
```
certik, line 23: Let's also test (probably in a new test file) a Union that is not c interoperable:
```python
class D(Union):
    dx: i64
    dy: f64
    dz: f64
```
Btw, I think this union feature can also be used for common blocks and equivalence in LFortran, see https://github.com/lfortran/lfortran/issues/718.

integration_tests/union_01.py
certik: I think the new syntax looks much better, thanks!

src/libasr/codegen/asr_to_c.cpp
```diff
@@ -282,7 +291,34 @@ class ASRToCVisitor : public BaseCCPPVisitor<ASRToCVisitor>
...
+                        ptr_char.clear();
+                    }
+                    sub = format_type_c(dims, "struct " + der_type_name + ptr_char,
+                                        v.m_name, use_ref, dummy);
+                }
+            } else if (ASR::is_a<ASR::Union_t>(*v.m_type)) {
+                std::string indent(indentation_level*indentation_spaces, ' ');
+                ASR::Union_t *t = ASR::down_cast<ASR::Union_t>(v.m_type);
+                std::string der_type_name = ASRUtils::symbol_name(t->m_union_type);
+                 if( is_array ) {
```
certik, line 305:
```suggestion
                if( is_array ) {
```

certik:
Great job, I think this looks good. We might need to update / change the syntax if needed, but to the core functionality it seems it should work.

How do you create a Union that is not c interoperable? Or is this mean to be C interoperable? 

It seems, we might need to use `ctypes` underneath to emulate unions even in Python, even if they are not c interoperable. If so, then I would change the syntax to the following:
```python
# Regular pure Python Union (underneath the `@union` decorator might need to use ctypes to implement this)
@union
class D:
    dx: i64
    dy: f64
    dz: f64

# C interoperable union, can be passed to a function implemented in C
@ccall
@union
class D:
    dx: i64
    dy: f64
    dz: f64
```

czgdp1807:
Makes sense. Will update tomorrow.

czgdp1807:
> Makes sense. Will update tomorrow.

@certik Done.

src/libasr/asr_verify.cpp
```diff
@@ -303,6 +303,10 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
         }
     }
 
+    void visit_UnionType(const UnionType_t& x) {
+        visit_DerivedTypeEnumTypeUnionType(x);
+    }
```
certik, line 308: Are there any requirements on Union besides the struct?

src/libasr/codegen/asr_to_c.cpp
certik: @czgdp1807 I think you forgot this one.

certik (approved):
Otherwise good.

src/libasr/codegen/asr_to_c.cpp
czgdp1807: Ah! yes.

src/libasr/asr_verify.cpp
czgdp1807: I don't think. On the syntax and ASR level, DerivedType and Union are exactly the same. Its in the backend level where the difference comes into play, DerivedType stores all the members in memory but Union stores only the maximum sized element.

src/libasr/codegen/asr_to_c.cpp
czgdp1807: I will resolve this tomorrow Morning, clean up the git history and merge.

certik: Good to merge, this is just a tiny thing anyway.

konradha:
Oh that's a great addition!

PR #1134, opened 2022-09-22
Support automatic value assignment to Enum members

czgdp1807:


integration_tests/enum_04.py
```diff
@@ -0,0 +1,71 @@
...
+    GREY: i32 = auto()
+
+class Integers(Enum):
+    a: i32 = -5
+    b: i32 = auto()
+    c: i32 = auto()
+    d: i32 = 0
+    e: i32 = auto()
+    f: i32 = 7
+    g: i32 = auto()
```
certik, line 20: Python does not have the best syntax here, but since this is what they do, we'll do it too for now. We can later fix this to something like:
```python
@enum
class Color:
    RED: i32
    GREEN: i32
    BLUE: i32 = 7
    YELLOW: i32
    WHITE: i32
    PINK: i32 = 15
    GREY: i32
```
which seems nicer.

certik (approved):
I think that this is fine.

PR #1135, opened 2022-09-22
Support C-interoperability in Enums

czgdp1807:


czgdp1807:
@certik Initial version of C-interoperable Enums is ready. Let me know if things look good to you. I will add more test cases then.

certik (approved):
Nice, much cleaner!

In subsequent PRs, can you make the verify() pass check the `enum_value_type` that it actually corresponds to the enumerators correctly?

certik:
From now on, every time we add a new semantic feature into ASR, we should check it in verify(). It will pay off in the long run big time.

czgdp1807:
> In subsequent PRs, can you make the verify() pass check the enum_value_type that it actually corresponds to the enumerators correctly?

AFAICT, you mean to make sure that the value of the enum members follow the same properties as specified by the `enum_value_type`, right?

> From now on, every time we add a new semantic feature into ASR, we should check it in verify(). It will pay off in the long run big time.

We can enforce it by making the parent's class methods abstract so that one has to implement the verification checks in the `asr_verify.cpp` in order to compile the project. Not sure if we should do this.

czgdp1807:
> In subsequent PRs, can you make the verify() pass check the `enum_value_type` that it actually corresponds to the enumerators correctly?

Done to the best of my understanding. I have enable auto-merge. If any changes are to be made please let me know I will make them in a subsequent PR.

certik:
Perfect, thanks for adding the verify().

As to how to enforce it, the most important part is not to add the verify methods, but how they are implemented to exhaustively check everything that is not allowed. That we have to tackle with a more general test framework I think.

PR #1138, opened 2022-09-23
C: Fix negative string index

### LINKED ISSUE #1124: visit_StringItem bug -C backend
C backend bug when trying to access to a negative position in string
can be fund in file test: integration_tests/test_str_01.py in function: test_str_index

Example : 
```
    a: str
    a = "012345"
     assert a[-1] == "5"
```
Quick fix offer will be : 
   ```
void visit_StringItem(const ASR::StringItem_t& x) {
        this->visit_expr(*x.m_idx);
        std::cout << "LETF IS = " + src << std::endl;
        std::string new_idx = src;
        //removes "  + 1"
        new_idx = new_idx.pop_back();
        new_idx = new_idx.pop_back();
        new_idx = new_idx.pop_back();
        new_idx = new_idx.pop_back();

        if(std::stoi(new_idx) >0 ) {
            new_idx = src;
        } else {
            new_idx = "len(+" + new_idx + ") + 1";
        }
```
But please check ASR

### PR DESCRIPTION
**Pull request by Smit-create**
Fixes #1124

### DIFF
--- a/integration_tests/test_str_03.py
+++ b/integration_tests/test_str_03.py
@@ -10,5 +10,16 @@ def test_int():
     l: i64 = 4
     print("abc:", i, j, k, l)
 
+
+def test_issue_1124():
+    a: str
+    a = "012345"
+    assert a[-1] == "5"
+    assert a[-1] == a[5]
+    assert a[-2] == a[4]
+    assert a[-4] == "2"
+
+
 test_new_line()
 test_int()
+test_issue_1124()
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -2586,6 +2586,19 @@ public:
             }
             ai.m_right = index;
             if (ASRUtils::is_character(*type)) {
+                ASR::expr_t* val = ASRUtils::expr_value(index);
+                if (val && ASR::is_a<ASR::IntegerConstant_t>(*val)) {
+                    if (ASR::down_cast<ASR::IntegerConstant_t>(val)->m_n < 0) {
+                        // Replace `x[-1]` to `x[len(x)+(-1)]`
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: integration_tests/test_str_03.py [hunk 1]
```diff
@@ -10,5 +10,21 @@ def test_int():
...
+    a = "012345"
+    assert a[-1] == "5"
+    assert a[-1] == a[5]
+    assert a[-2] == a[4]
+    assert a[-4] == "2"
+
+    # runtime tests
+    k: i32 = len(a); i: i32
+    for i in range(k):
+        assert a[-i-1] == a[k-i-1]
```
certik, line 25: For the same reason as discussed in the PR for negative list indices, we should not allow runtime negative string indices, because they can't be implemented in LLVM/C with high performance (that is, without an extra check for every index access).

We can however allow compile time negative value.

src/libasr/codegen/asr_to_c.cpp
```diff
@@ -806,8 +806,9 @@ R"(
         std::string idx = std::move(src);
         this->visit_expr(*x.m_arg);
         std::string str = std::move(src);
+        idx = "((" + idx + " - 1) >= 0 ? (" + idx + " - 1) : (strlen(" + str + ") + " + idx + " - 1))";
```
certik, line 809: This change should not be done, as it slows down every index access, see above.

integration_tests/test_str_03.py
Smit-create: Oh okay, get it. I'll revert the last two commits.

src/libasr/codegen/asr_to_c.cpp
Smit-create: Removed this. Thanks!

czgdp1807 (approved):

certik (approved):
Looks good!

PR #1141, opened 2022-09-23
Fix some parser related issues

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
I compared all the files (LPython AST vs CPython AST) in the LPython repository. These are the fixes for those which didn't match.

certik (approved):

PR #1147, opened 2022-09-23
Switch to the new parser

certik:


certik:
I think everything passes, so I am going to merge this now.

We can then polish things with subsequent PRs.

Issue #1022, opened 2022-08-24
Match statement is not recognized 


Thirumalai-Shaktivel:
Example:
```py
def _get_slots(cls):
    match cls.__dict__.get('__slots__'):
        case None:
            return
        case str(slot):
            yield slot
        # Slots may be any iterable, but we cannot handle an iterator
        # because it will already be (partially) consumed.
        case iterable if not hasattr(iterable, '__next__'):
            yield from iterable
        case _:
            raise TypeError(f"Slots of '{cls.__name__}' cannot be determined")
```
Error:
```py
syntax error: Token 'cls' (of type 'identifier') is unexpected here
    --> ./Lib/dataclasses.py:1130:11
     |
1130 |     match cls.__dict__.get('__slots__'):
     |           ^^^
```

Thirumalai-Shaktivel:
Here, match and case are soft keywords: https://docs.python.org/3/reference/compound_stmts.html#the-match-statement
i.e., 'match' can be used as an identifier!
@certik how do we handle this?

certik:
So `match` is a keyword that is not reserved. See https://docs.python.org/3/reference/lexical_analysis.html#soft-keywords, they call it a *soft keyword*. In our tokenizer.re, we have the following sets of *reserved keywords*:
```
"as"      
"assert"  
"async"   
"await"   
"break"   
"class"   
"continue"
"def"     
"del"     
"elif"    
"else"    
"except"  
"finally" 
"for"     
"from"    
"global"  
"if"      
"import"  
"in"      
"is"      
"lambda"  
"None"    
"nonlocal"
"pass"    
"raise"   
"return"  
"try"     
"while"   
"with"    
"yield"   
"yield"
```
These cannot be used as variable names such as in `with = 5`.

`match`, `case` are different (perhaps also `_` as the docs say, not sure about how we handle that). We have to allow them as variable names.

certik:
We have several options how to tackle the `match` and `case` non-reserved keywords (each has pros/cons):

* we can write a "strong tokenizer" like the fixed-form LFortran tokenizer, which is a recursive descent parser that recognizes the context that we are in, and based on the context it will know if `match` and `case` are keywords, or just a name (such as a variable).
* we can enable Bison GLR, just like we use in the free-form, so the tokenizer always returns `match` as a keyword, and in Bison we add an `id` rule that is a "TK_NAME | KW_MATCH | KW_CASE", this creates conflicts, but Bison GLR can take care of it
[...]

PR #1148, opened 2022-09-23
Recognize Match Statement

Thirumalai-Shaktivel:
Fixes: #1022


certik:
Here are some more test cases to try:
```
def case(x):
    pass

def match(x):
    pass
```

src/lpython/parser/tokenizer.re
```diff
@@ -397,6 +397,10 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
             "yield"    { KW(YIELD) }
             "yield" whitespace "from" whitespace { KW(YIELD_FROM) }
 
+            // Soft Keywords
+            "match" / [^\n\x00]+ ":" newline { KW(MATCH) }
+            "case" / [^\n\x00]+ ":" newline { KW(CASE) }
```
certik, line 402: I think to fix the test cases that I posted above, you need to check here that the last token was TK_NEWLINE, if so, it's a KW_MATCH, otherwise it's a TK_NAME.

certik:
For `case` here is a good one:
```python
case = {"ok": 5}
a = {
case ["ok"]:
  6}
print(a)
```
I don't think you can even distinguish this at the token level.

certik:
Let's try this:

* Must be immediately preceded by TK_NEWLINE and optional indent (for match) or required indent (for case)
* the line must end with : and \n
* parenlevel must be 0

That might do it.

certik:
Hm, but what about this:
```
>>> def file_handler_v1(command):
...     match command.split():
...         case ['show']:
...             print('List all files and directories: ')
...             # code to list files
...         case ['remove',
   *files]:
...             print('Removing files: {}'.format(files))
...             # code to remove files
...         case _:
...             print('Command not recognized')
```
The second `case` does not end with `:`...

Thirumalai-Shaktivel:
Exactly, I'm working on it.

tests/parser/match_stmt1.py
```diff
@@ -0,0 +1,80 @@
...
+        case [*_]:
+            return "seq"
+
+def file_handler_v1(command):
+    match command.split():
+        case ['show']:
+            print('List all files and directories: ')
+            # code to list files
+        case ['remove',
+              *files]:
```
Thirumalai-Shaktivel, line 33: Fixed all the cases, but this is tricky. Here, the parentheses value is not recorded, simply put the `parenlevel = 0`.

certik: Why is it not recorded?

certik:
Another use case:
```python
match = """some string:
  surprise!"""
```

tests/parser/match_stmt1.py
certik: Ah I see, it's not recorded in the lookahead part. 

Yes, indeed, that's because regex cannot parse nested parentheses.

certik:
I think I would try the following. When you see "match space character" (and you are not inside a string), then it must be a keyword match, isn't it?

If so, when you see this, call a function `lex_match()`. Inside `lex_match`, finishing tokenizing the rest of the line, until a TK_NEWLINE. Then check on the new line that it is indented and it starts with "case". If so, call lex_case() and repeat. If it does not start with "case", that's the end of match, so return.

We can iterate on the exact details of lex_match() and lex_case(), but the overall structure seems very robust.

Thirumalai-Shaktivel:
I will try to implement this, thank you.

Thirumalai-Shaktivel:
I'm confused here:
- the `lex_match()` and `lex_case()` are a mini-tokenizer which is used to recognize the match statement, right?
- if so, how would I return each recognized tokens to `tokens()` function?
- Also, should I re-define everything like `digits`, `integer`, `real`, `string`... ?
- can `lex()` function be called inside `lex_match()`?
- Everything works but only the `strings` and `parentheses` to be fixed.

certik:
Let's do the following. Your current design is good enough (except strings and parentheses), so let's write test, polish this up and merge it.

In a subsequent PRs we'll try to fix the remaining case of strings and parentheses. Yes, it's tough to return the tokens from `lex_match()`. One idea is to only make `lex_match/lex_case()` "parse" (say using re2c) but without returning tokens, just to recognize robustly if it is indeed a "match" case. Then come back to the main tokenizer and return KW_MATCH and continue matching as usual, until a "case" is encountered. We can then use a "match state" and beginning of the line / indent to robustly ensure we have a "case" (I think). Or something like that.

Thirumalai-Shaktivel:
I got it, thanks.

Thirumalai-Shaktivel:
This PR is ready to go in.

certik (approved):
I think this is excellent, thanks for getting most of it done. The tests look thorough. There are some corner cases (commented out) that doesn't work, but this should be good enough to get started. 

Can you open up an issue for the cases that don't work yet?

Thirumalai-Shaktivel:
Thank you for reviewing and merging this PR.
Yup, I will create an issue for it.

PR #1159, opened 2022-09-26
Fix For Loop

### LINKED ISSUE #1153: Bug in `for` with negative step
Python code:
```py
def f():
    i: i32
    for i in range(5, 0, -1):
        print(i)

f()
```

LPython Output:
```
% lpython a.py && ./a.out         
5
4
3
2
1
0
-1
```

CPython:
```
% python a.py                    
5
4
3
2
1
```

Smit-create:
Fixes #1153


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3592,6 +3590,46 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+        } else {
+            ASR::expr_t *constant_one = ASR::down_cast<ASR::expr_t>(ASR::make_IntegerConstant_t(
+                                                al, loc, 1, a_type));
+            head.m_increment = constant_one;
+        }
+        // Loop end depends upon the sign of m_increment.
+        // if inc > 0 then: loop_end -=1 else loop_end += 1
+
+        // Compute the sign value of m_increment
+        ASR::symbol_t *fn_sign = resolve_intrinsic_function(loc, "_find_sign");
```
Smit-create, line 3614: This approach will require us to replace all the `for` loops in `_lpython_builtin.py` to be replaced by `while` loops in order to avoid an infinite loop.

Smit-create: @certik @czgdp1807 Should we move forward with this approach?

czgdp1807: I would like to see all the 216 cases tested first. It should be easy to do it if you use loops. I can't say for sure (without looking at the tests) whether any approach will work or not. But yes its making sense to me if I read your code. Branching can be avoided in `_find_sign`'s implementation?

Smit-create: > This approach will require us to replace all the `for` loops in `_lpython_builtin.py` by `while` loops in order to avoid an infinite loop.

For testing, we will require this change. If the approach looks good, it will indeed run for all 216 cases.

czgdp1807: Yes. Please test it. It would be great to see this approach working.

czgdp1807: If compile time values are hard to test, you can also test for runtime values only which is easy. Just create 3 lists for start, stop and end and send each combination into a function which tests these three in a loop. If it works, we will know everything is correct for runtime values.

czgdp1807:
The tests look okay. Though CI seems to fail due to out of date reference tests?

integration_tests/test_for_loop.py
```diff
@@ -0,0 +1,37 @@
...
+            for j in range(start[a], stop[b], 2):
+                count += 1
+            assert count == max(0, (stop[b] - start[a]) // 2)
+
+    for a in range(len(start)):
+        for b in range(len(stop)):
+            count = 0
+            for j in range(start[a], stop[b], 1):
+                count += 1
+            assert count == max(0, stop[b] - start[a])
```
certik, line 33: All these just test use compile time increment, so you don't need any runtime function.

certik:
I would recommend to allow using for loops in our runtime library code.

For now in this PR, just implement a compile time increment, which is what you test for anyway so far.

Regarding runtime increment, we should work on that in a separate PR, I can't remember if Fortran even allows it. If not, then we might just disallow it as well and one can use while loop if one needs it.

czgdp1807 (approved):
Please rebase on top of main before merging.

Smit-create:
> Please rebase on top of main before merging.

I have rebased it.

PR #1160, opened 2022-09-26
Enabling array programming features in C backend

czgdp1807:


certik (approved):
I think this is fine. We can refine as we go.

PR #1161, opened 2022-09-27
Replace ``sprintf`` -> ``snprintf``

czgdp1807:
I got several of the following security warning after recent macOS update. Hence the replacement.

```bash
/Users/czgdp1807/lpython_project/lpython/src/libasr/containers.h:188:10: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
    std::sprintf(&output[0], format.c_str(), std::forward<Args>(args)...);
         ^
/Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/stdio.h:188:1: note: 'sprintf' has been explicitly marked deprecated here
__deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")
^
/Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/sys/cdefs.h:215:48: note: expanded from macro '__deprecated_msg'
        #define __deprecated_msg(_msg) __attribute__((__deprecated__(_msg)))
                                                      ^
1 warning generated.
```

### DIFF
--- a/src/libasr/containers.h
+++ b/src/libasr/containers.h
@@ -185,7 +185,7 @@ std::string string_format(const std::string& format, Args && ...args)
 {
     auto size = std::snprintf(nullptr, 0, format.c_str(), std::forward<Args>(args)...);
     std::string output(size, '\0');
-    std::sprintf(&output[0], format.c_str(), std::forward<Args>(args)...);
+    std::snprintf(&output[0], size + 1, format.c_str(), std::forward<Args>(args)...);
     return output;
 }
 


src/libasr/containers.h
```diff
@@ -185,7 +185,7 @@ std::string string_format(const std::string& format, Args && ...args)
 {
     auto size = std::snprintf(nullptr, 0, format.c_str(), std::forward<Args>(args)...);
     std::string output(size, '\0');
-    std::sprintf(&output[0], format.c_str(), std::forward<Args>(args)...);
+    std::snprintf(&output[0], size + 1, format.c_str(), std::forward<Args>(args)...);
```
czgdp1807, line 188: @certik Is this change correct. I think size of the buffer should be ` + 1` because it wills only till second last position. Am I correct? Locally all the tests pass so seems like it. Let me know. I will merge this if you agree.

czgdp1807: Without ` + 1` I get incorrect updates in reference tests. So I think the above change is fine. Will merge once this test pass and no objections happen.

certik:
Thanks!

PR #1163, opened 2022-09-28
Implement ``goto`` in LPython

ronnuriel:


src/runtime/ltypes/goto.py
```diff
@@ -0,0 +1,235 @@
...
+    @property
+    def argument_bits(self):
+        return self.argument.size * 8
+
+
+_BYTECODE = _Bytecode()
+
+
+def _make_code(code, codestring):
+    return code.replace(co_code=codestring)
```
czgdp1807, line 46: Instead of manually calling `types.Code(*args)` as in [the original source code]() I am using `.replace` method and passing the updated code. This makes it work with Python > 3.8 robustly since we don't depend on the internal order of arguments to `PyCode_New`.

czgdp1807:
@certik This is ready and works for CPython, LLVM and C backend. The file [src/runtime/ltypes/goto.py](https://github.com/lcompilers/lpython/pull/1163/files#diff-84ce6cdc88619340323598c586c339b826ac4b95d0776695790e3eff4f34dde8) is taken from [snoack/python-goto](https://github.com/snoack/python-goto/blob/master/goto.py). I have the following questions,

1. How should I give credit to [python-goto]() developers? They use [UNLICENSE](https://github.com/snoack/python-goto/blob/master/LICENSE) which doesn't impose any legal restrictions on us. Should I add the commit we are using in a comment at the top of `src/runtime/ltypes/goto.py` (our goto.py file)?
2. I figured out a fix (see https://github.com/lcompilers/lpython/pull/1163#discussion_r983345806). My question is should I send a PR to the original project or is it not necessary because we are not facing the burden of supporting very old Python versions.

Please let me know. Other than the above questions `goto` feature is fully ready from my side.

rebcabin:
Is "ULICENSE" equivalent to "PUBLIC DOMAIN?" If we have any other "derivative works" in our sources, how are we giving credit to the authors of the originals?

rebcabin:
> 2. I figured out a fix (see [Goto #1163 (comment)](https://github.com/lcompilers/lpython/pull/1163#discussion_r983345806)). My question is should I send a PR to the original project or is it not necessary because we are not facing the burden of supporting very old Python versions.

I would say we do not need the original project to pull any of our changes because we will never support earlier versions of Python. Is there any residual benefit to them? What's our version floor? 3.7, 3.9, ... ?

rebcabin:
> > 3\. I figured out a fix (see [Goto #1163 (comment)](https://github.com/lcompilers/lpython/pull/1163#discussion_r983345806)). My question is should I send a PR to the original project or is it not necessary because we are not facing the burden of supporting very old Python versions.
> 
> I would say we do not need the original project to pull any of our changes because we will never support earlier versions of Python. Is there any residual benefit to them? What's our version floor? 3.7, 3.9, ... ?

Looks like we're 3.9 and greater for LPython.

src/libasr/ASR.asdl
```diff
@@ -173,10 +173,10 @@ stmt
         -- GoTo points to a GoToTarget with the corresponding target_id within
         -- the same procedure. We currently use `int` IDs to link GoTo with
         -- GoToTarget to avoid issues with serialization.
-    | GoTo(int target_id)
+    | GoTo(int target_id, identifier name)
```
certik, line 176: I think GoTo should just reference using an ID. The name should go into GoToTarget.
```suggestion
    | GoTo(int target_id)
```
Or am I missing something?

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4543,8 +4586,14 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
-        ASRUtils::EXPR(tmp);
-        tmp = nullptr;
+
+        // If tmp is a statement and not an expression
+        // never cast into expression using ASRUtils::EXPR
+        // Just ignore and exit the function naturally.
+        if( !ASR::is_a<ASR::stmt_t>(*tmp) ) {
+            ASRUtils::EXPR(tmp);
+            tmp = nullptr;
+        }
```
certik, line 4596: What is the purpose of these lines, both the original and the new ones?

Are we checking that the result is an expression? If so, you can use an assert for `ASR::is_a` (or perhaps `down_cast`, since `is_a` might require a particular class).

certik (approved):
I think this is good enough to merge as is. But there are issues to resolve. One is the ASR relationship of GoTo and GoToTarget. It looks like what is needed is a direct link (in C++) from goto to the target, so that one can read off the name if needed easily. 

It seems the ASR.asdl structure does not easily allow a way to do it. Storing both the number and a string is redundant.

certik:
Regarding where to maintain the CPython implementation, yes, we should maintain it ourselves and ensure it actually works in all modern Python versions.

src/libasr/ASR.asdl
czgdp1807: Well in C backend we need the name to produce, `goto name_of_the_label`. So we need the name to be stored in both GoTo and GoToTarget.

src/lpython/semantics/python_ast_to_asr.cpp
czgdp1807: So the scene is that `goto.end` is represented as `[(Expr (Attribute (Name goto Load) end Load))]` at AST level. So, now when we will complete visiting `Attribute`, `tmp` will either point to a ASR expression or a ASR statement (because `GoTo` in ASR is a statement). So in case we get something like, `goto.label` then `tmp` will point to a statement in which case we should exit the function naturally and the `transform_stmts` before this node will include that GoTo statement. However if `tmp` is not pointing to a statement then I replaced `ASRUtils::EXPR(tmp)` with an `LFORTRAN_ASSERT`.

czgdp1807: > If so, you can use an assert for `ASR::is_a` (or perhaps `down_cast`, since `is_a` might require a particular class).

Assert with `ASR::is_a` works fine.

czgdp1807:
> It looks like what is needed is a direct link (in C++) from goto to the target, so that one can read off the name if needed easily.
> 
> It seems the ASR.asdl structure does not easily allow a way to do it. Storing both the number and a string is redundant.

There are two ways in my mind to deal with this,

1. Store only the name and generate IDs in the backend. This goes against the principle that we backend shouldn't need to track anything.
2. Change `GoTo(int target_id, identifier name)` to `GoTo(stmt_t goto_target)`. Then one can do `ASR::down_cast<ASR::GoToTarget>(goto_object->goto_target)->m_name`. One extra indirection but it should removes the redundancy.

certik:
> Change `GoTo(int target_id, identifier name)` to `GoTo(stmt_t goto_target)`. Then one can do `ASR::down_cast<ASR::GoToTarget>(goto_object->goto_target)->m_name`. One extra indirection but it should removes the redundancy.

I like this idea, so I opened up a dedicated issue for this: https://github.com/lcompilers/lpython/issues/1167

src/lpython/semantics/python_ast_to_asr.cpp
certik: Can you please document this as a comment? Next person to see that code will have the same question, including me tomorrow.

PR #1165, opened 2022-09-30
C: Fix a bug which has `main` function in python

Smit-create:


czgdp1807 (approved):

certik (commented):
I think this is fine.

Issue #1130, opened 2022-09-21
Change "lpython examples/expr2.py" to run the script, after compiling


certik:
Currently `lpython examples/expr2.py` just creates an executable `a.out`. However, coming from Python people would rather expect LPython to run the script after compiling. So we should change the behavior of the driver to do this. When `-c` or `-o` options are passed, then it would compile to a binary and not run it.

PR #1166, opened 2022-09-30
Execute `a.out` after successful linking of the executable

Thirumalai-Shaktivel:
Fixes: #1130

### DIFF
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -792,6 +792,13 @@ int link_executable(const std::vector<std::string> &infiles,
                 std::cout << "The command '" + cmd + "' failed." << std::endl;
                 return 10;
             }
+            if (outfile == "a.out") {
+                err = system("a.out");
+                if (err) {
+                    std::cout << "The command 'a.out' failed." << std::endl;
+                    return 10;
+                }
+            }
         } else {
             std::string CC = "cc";
             char *env_CC = std::getenv("LFORTRAN_CC");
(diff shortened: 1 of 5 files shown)


src/bin/lpython.cpp
```diff
@@ -817,6 +817,13 @@ int link_executable(const std::vector<std::string> &infiles,
                 std::cout << "The command '" + cmd + "' failed." << std::endl;
                 return 10;
             }
+            if (outfile == "a.out") {
+                err = system("./a.out");
+                if (err) {
+                    std::cout << "The command './a.out' failed." << std::endl;
+                    return 10;
```
certik, line 824: Does this work on Windows as well?

Thirumalai-Shaktivel: This doesn't work on Windows. I wonder why?

Thirumalai-Shaktivel: Updated, Now it works on windows as well.

Thirumalai-Shaktivel:
What is the command used to switch the backend?
Following doesn't work on windows:
```py
$ src\bin\lpython.exe --backend cpp examples\expr2.py
LCompilersException: Unsupported backend.
```

Thirumalai-Shaktivel:
This might be reason: 
https://github.com/lcompilers/lpython/blob/7a07cb98a35cce504b3a5eba4ec6efa9b18a6ab3/src/bin/lpython.cpp#L1375-L1384

certik (approved):
I think this is fine, this will give the expected outcome.

I don't know yet if we need if we want to even generate `a.out` or not, but this PR is iterative, so we used to generate it and we still do, we just run it as well.

certik:
Feel free to merge when ready.

Thirumalai-Shaktivel:
Thanks for the review.
Yup, this is ready. Let's merge it.

PR #1169, opened 2022-10-02
C: Implement nested Lists

Smit-create:


Smit-create:
I'm trying to compile the generated C code from `integration_tests/test_list_06.py`. The generated code using this PR has some errors like:
```c
int list_find_item_list_r64_(struct list_list_r64_* x, struct list_r64 element) {
    int el_pos = 0;
    while(x->current_end_point > el_pos) {
        if (x->data[el_pos] == element) return el_pos;
        el_pos++;
    }
    return -1;
}
```

`C` does not allow operator overloading and so there isn't any direct way of comparing two structs as in this case `if (x->data[el_pos] == element) return el_pos;`. Do we have any other approach to fix this?

czgdp1807:
> Do we have any other approach to fix this?

Yes. You can use recursive approach to compare aggregate data structures. Follow the same pattern as the `deepcopy` for something like, `compare`,

https://github.com/lcompilers/lpython/blob/16873d03cfab100232602292a5ba8c78ae0e9b27/src/libasr/codegen/asr_to_c_cpp.h#L71

czgdp1807 (approved):

PR #1173, opened 2022-10-04
Accommodate expressions in EnumValue, EnumName

czgdp1807:
As you can see in [this test file](https://github.com/czgdp1807/lpython/blob/459f716b23dd8ed3facbeb9b6af89bf5a4f6a85f/integration_tests/enum_06.py) an enum's value is accessed by first accessing the member of a derived type. The most natural approach is to transform `EnumValue(symbol v, ttype type, expr? value)` to `EnumValue(expr v, ttype enum_type, ttype type, expr? value)`. In words this means changing the type of `v` from `symbol` to `expr`. But the problem with this approach is `Color.RED.value` is represented as `EnumValue(Var(RED), ...)` and hence `visit_Var` in `asr_verify.cpp` fails because of [this error](https://github.com/czgdp1807/lpython/blob/459f716b23dd8ed3facbeb9b6af89bf5a4f6a85f/src/libasr/asr_verify.cpp#L399-L400) (which I commented out for highlighting).

I list down some alternative approaches,

1. **Use different nodes for Enum Objects and Enum Classes** - Here I mean when we use an enum object then use `EnumObjectValue(expr v, …)` because its in this case where we will encounter expressions and for direct access of enum members via the enum class use, `EnumClassValue(symbol v, …)` because here we will do `Color.RED.value` and hence symbol `RED` is accessible and is not wrapped under an unknown expression.
2. **Store one of expression and symbol and keep the other as nullptr** - `EnumValue(expr? vexpr, symbol? vsym, …, ttype enum_type, …)`. This means that both are optional but ASR verify will make sure that exactly one of them is present. So for cases like, `Color.RED` we can fill `vsym` and for cases like, `DerivedRef(car1, color)` we fill `vexpr`.
3. **Combine EnumValue and EnumName into a single node EnumRef** - This means replace `EnumValue`, `EnumObject` with `EnumValueRef(expr v, symbol? m, …)`. Now here `m` is optional and it will be nullptr for cases like, `DerivedRef(car1, color)` but for cases like, `Color.RED.value` `v` will be `Color` and `m` will be `RED`. Basically its the same approach as `2` but makes `v` compulsory.

Note that when we use `expr v` in `EnumValue` we have to store the original `enum_type` in the node itself because the underlying symbol is unknown in this case.

@certik Let me know what you think.


certik:
So let's clean this up.

* `DerivedRef` -> `StructInstanceMember`. Example: `car1: Car = ...; car1.color` would be `StructInstanceMember(car1, color)`
* `StructMember`. Example: `Car.some_static_variable`. Note that if you do `car1.some_static_variable` that would be `StructInstanceMember(car1, some_static_variable)`.
* Similarly: `EnumInstanceMember` and `EnumMember`

We can possibly unify Enum with Struct later, that is much easier than separating things out.

czgdp1807:
@certik  I have renamed `Derived` to `Struct` in ASR and updated `EnumValue`, `EnumName` according to what we discussed in our meeting.

src/libasr/ASR.asdl
```diff
@@ -283,10 +283,11 @@ expr
     | ArrayReshape(expr array, expr shape, ttype type, expr? value)
 
     | BitCast(expr source, expr mold, expr? size, ttype type, expr? value)
-    | DerivedRef(expr v, symbol m, ttype type, expr? value)
+    | StructInstanceMember(expr v, symbol m, ttype type, expr? value)
+    | EnumMember(expr v, symbol m, ttype type, expr? value)
```
certik, line 287: I would still do `EnumClassMember`. I think the `class` and `instance` relationship is well established and people understand it. I had to lookup our discussion in this PR to figure out what `EnumMember` means.

Essentially, Struct, Enum, Union, etc. they all have a "class" and "instance" and one can access both (with different semantics and consequences). If you don't like the term "class" then let's figure out a different word, to make this clear.

This can be done in a subsequent PR.

certik (approved):

src/libasr/ASR.asdl
certik: I opened up an issue about this at https://github.com/lcompilers/lpython/issues/1183.

PR #1178, opened 2022-10-04
Added ``dependencies`` to Function

czgdp1807:
This means the dependencies array stores the functions (ideally symbols but for now functions) on which the current function depends.

Implements https://github.com/lfortran/lfortran/issues/802#issuecomment-1263994992 for functions.


czgdp1807:
Will implement the exact thing tomorrow.

czgdp1807:
Okay so I figured its better to store the functions on which the current function depends in the dependency list and not the other way around. That way we only need to scan the body and update the dependency list of current function without modifying the functions being called. So I will do that on Sunday or Monday. I am out of office on 6-th and 7-th October. Will come back and complete this. The approach is pretty much clear. cc: @certik

certik:
> Okay so I figured its better to store the functions on which the current function depends in the dependency list and not the other way around.

So you are saying that it's better to store dependencies, as opposed to storing dependents? Yes, I agree, it seems natural to always do it like that, because you keep adding new functions and it's a local operation (`O(n)` in the number of functions), while if you store dependents, you need to update the whole tree, so most likely a `O(n^2)` operation, so not a good design.

czgdp1807:
> Yes, I agree, it seems natural to always do it like that, because you keep adding new functions and it's a local operation (`O(n)` in the number of functions), while if you store dependents, you need to update the whole tree, so most likely a `O(n^2)` operation, so not a good design.

Correct. One of the reasons why I went for the approach I explained above.

czgdp1807:
@certik This is ready for review. I implemented the verify checks as discussed in our meeting.  Please let me know if everything looks good. I will clean up the git history in that case.

czgdp1807:
> such as requiring that the dependencies do not contain names that are actually not dependencies

This is already being done. `verify` will raise an error if it finds something in dependency list of a function which is actually not a dependency of that function.

> checking more than just functions

Yes we need to do this. Extending the dependencies to variables, derived types, etc.

Issue #1128, opened 2022-09-20
Test runtime errors


certik:
The way to test the runtime errors might be to just expand our `run_tests.py`, we would put the code that should fail into `tests/errors/` and then register it in `tests.toml`, and it would compile and run the code, it would capture stderr and stdout, compare/update the reference results, and also it would test the exit code.

Smit-create:
> expand our `run_tests.py`,

Then we should also modify `run_tests.py` to capture some flag like `run_llvm` or `run_cpp` so that it compiles and runs the file and capture the stderr?

certik:
We probably should, since we want to check the runtime error messages in different backends.

certik:
Related issue: https://github.com/lcompilers/lpython/issues/772

Thirumalai-Shaktivel:
This diff wouldn't be right fix, but:
```diff
diff --git a/run_tests.py b/run_tests.py
index 06e9a5e78..2fa01bfc2 100755
--- a/run_tests.py
+++ b/run_tests.py
@@ -23,6 +23,7 @@ def single_test(test, specific_test, verbose, no_llvm, update_reference):
     c = test.get("c", False)
     wat = test.get("wat", False)
     pass_ = test.get("pass", None)
+    runtime_error = test.get("runtime_error", False)
     optimization_passes = ["flip_sign", "div_to_mul", "fma", "sign_from_value",
                            "inline_function_calls", "loop_unroll",
                            "dead_code_removal", "loop_vectorise", "print_list"]
@@ -99,6 +100,10 @@ def single_test(test, specific_test, verbose, no_llvm, update_reference):
         run_test(filename, "wat", "lpython --no-color --show-wat {infile}",
                  filename, update_reference, extra_args)
 
+    if runtime_error:
+        run_test(filename, "runtime_error", "lpython {infile} && ./a.out",
+                 filename, update_reference, extra_args)
+
 
 
 if __name__ == "__main__":
diff --git a/tests/tests.toml b/tests/tests.toml
index d1ae0ab8d..1108621dc 100644
--- a/tests/tests.toml
+++ b/tests/tests.toml
@@ -4,6 +4,7 @@
 # asr ... run the Semantics and output ASR, compare against reference version
 # llvm ... run the Semantics and output LLVM, compare against reference version
 # tokens ... output Tokens, compare against reference version
+# runtime_error ... Check runtime errors
 
 # tests
 
@@ -637,6 +638,10 @@ asr = true
[...]

PR #1180, opened 2022-10-05
Add runtime tests support

Thirumalai-Shaktivel:
Related: #1128


tests/runtime_errors/test_dictionary_01.py
```diff
@@ -0,0 +1,8 @@
+from ltypes import i32
+
+def test_dictionary():
+    x: dict[i32, str]
+    x = {1: "one", 2: "two"}
+    print(x[10])
```
Thirumalai-Shaktivel, line 6: For the following condition:
https://github.com/lcompilers/lpython/blob/16873d03cfab100232602292a5ba8c78ae0e9b27/src/libasr/codegen/llvm_utils.cpp#L1495-L1522
is this the correct test?
Or 
How should I test KeyError?
@czgdp1807

Thirumalai-Shaktivel:
This is weird, In MacOS the output is captured as *.stderr, but in Ubuntu it is captured as *.stdout

tests/runtime_errors/test_dictionary_01.py
Thirumalai-Shaktivel: But it prints:
```console
$ lpython tests/runtime_errors/test_dictionary_01.py
(null)
```
Is there any other tests?

certik:
This is a bug in LPython. This file:
```
from ltypes import i32

def test_dictionary():
    x: dict[i32, str]
    x = {1: "one", 2: "two"}
    print(x[10])

test_dictionary()
```
gives for me:
```console
$ lpython a.py -o a       
$ ./a
(null)
$ PYTHONPATH=src/runtime/ltypes  python a.py
  File "/Users/ondrej/repos/lpython/a.py", line 8, in <module>
    test_dictionary()
  File "/Users/ondrej/repos/lpython/a.py", line 6, in test_dictionary
    print(x[10])
KeyError: 10
```
So the runtime bounds check does not work in this case. So we have to debug this and fix it.

czgdp1807:
Should we enable this runtime check only in Debug mode only. Should ReleaseSafe also contain it? ReleaseFast shouldn't do this runtime check I think. What do you say?

certik:
Yes, I think ReleaseFast should not have this check, but ReleaseSafe and Debug modes should have it. We can implement this later as well.

Thirumalai-Shaktivel:
I thought *.stderr file will be generated based on the exit_code. But, it doesn't!
> That might be an actual bug that we need to fix in LPython itself.

We have to debug the bug.

Thirumalai-Shaktivel:
The reason can also be that we print the error message to stdout.

Thirumalai-Shaktivel:
Here it is: https://github.com/lcompilers/lpython/blob/4d4b8e682aeca84b371f6526cbdc5e9ba32f08d8/src/libasr/runtime/lfortran_intrinsics.c#L54-L55

Thirumalai-Shaktivel:
@czgdp1807 What do you think, is this the correct fix?

Thirumalai-Shaktivel:
Now, Stderr_hash differs in MacOS!

Thirumalai-Shaktivel:
Hello @certik,
Everything works properly here!
The error output is printed as
```
ld: warning: dylib (/Users/runner/work/lpython/lpython/src/runtime/liblpython_runtime.dylib) was built for newer macOS version (11.7) than being linked (11.0)
ValueError: The list does not contain the element
```
in MacOS
Screenshot from 2022-10-14 19-33-58
This is the reason for the failure!
Why does it prin this warning?

certik:
The macOS failure is due to the extra line in the output:
```
> ld: warning: dylib ($DIR/src/runtime/liblpython_runtime.dylib) was built for newer macOS version (11.7) than being linked (11.0)
```
Which is caused by the `liblpython_runtime.dylib` being compiled for a different mac deployment target than the actual file that LPython itself links (via Clang I think). I think it has to do something with the `MACOSX_DEPLOYMENT_TARGET` variable.

.github/workflows/CI.yml
```diff
@@ -7,7 +7,8 @@ on:
   pull_request:
     branches:
       - main
-
+env:
+  MACOSX_DEPLOYMENT_TARGET: 11.0
```
Thirumalai-Shaktivel, line 11: This fixed the MacOS CI Failure.

certik (approved):
This looks good.

PR #1185, opened 2022-10-07
Compile Simple Python package

Thirumalai-Shaktivel:
Related: #992


test_a.py
```diff
@@ -0,0 +1,10 @@
+import a
+from a import print_a
+
+print(print_a())
+print(a.print_a())
+print(a.print_b())
+print(a.print_c())
```
Thirumalai-Shaktivel, line 7: The above code works, Can anyone please review and provide feedback?

Thirumalai-Shaktivel: I think the design is not correct
Consider the example
**test_a.py**
```py
import a
```
**ASR**
```py
(TranslationUnit 
    (SymbolTable
        1
        {
            a:
                (Module 
                    (SymbolTable
                        3
                        {
                            print_a:
                                (ExternalSymbol 
                                    3 
                                    print_a 
                                    5 print_a 
                                    b 
                                    [] 
                                    print_a 
                                    Public
                                ), 
                            print_b:
                                (ExternalSymbol 
                                    3 
                                    print_b 
                                    5 print_b 
                                    b 
                                    [] 
                                    print_b 
                                    Public
                                ), 
                            print_c:
                                (ExternalSymbol 
                                    3 
                                    print_c 
                                    9 print_c 
                                    c 
                                    [] 
                                    print_c 
                                    Public
                                )
                            
                        }) 
                    a 
                    [b
                    c] 
                    .false. 
                    .false.
                ), 
            b:
                (Module 
                    (SymbolTable
                        5
                        {
                            print_a:
                                (Function 
                                    (SymbolTable
                                        6
                                        {
                                            _lpython_return_variable:
                                                (Variable 
                                                    6 
                                                    _lpython_return_variable 
                                                    ReturnVar 
                                                    () 
                                                    () 
                                                    Default 
                                                    (Character 1 -2 () []) 
                                                    Source 
                                                    Public 
                                                    Required 
                                                    .false.
                                                )
                                            
                                        }) 
                                    print_a 
                                    [] 
                                    [(= 
                                        (Var 6 _lpython_return_variable) 
                                        (StringConstant 
                                            "A" 
                                            (Character 1 1 () [])
                                        ) 
                                        ()
                                    )
                                    (Return)] 
                                    (Var 6 _lpython_return_variable) 
                                    Source 
                                    Public 
                                    Implementation 
                                    () 
                                    .false. 
                                    .false. 
                                    .false. 
                                    .false. 
                                    .false. 
                                    [] 
                                    [] 
                                    .false.
                                ), 
                            print_b:
                                (Function 
                                    (SymbolTable
                                        7
                                        {
                                            _lpython_return_variable:
                                                (Variable 
                                                    7 
                                                    _lpython_return_variable 
                                                    ReturnVar 
                                                    () 
                                                    () 
                                                    Default 
                                                    (Character 1 -2 () []) 
                                                    Source 
                                                    Public 
                                                    Required 
                                                    .false.
                                                )
                                            
                                        }) 
                                    print_b 
                                    [] 
                                    [(= 
                                        (Var 7 _lpython_return_variable) 
                                        (StringConstant 
                                            "B" 
                                            (Character 1 1 () [])
                                        ) 
                                        ()
                                    )
                                    (Return)] 
                                    (Var 7 _lpython_return_variable) 
                                    Source 
                                    Public 
                                    Implementation 
                                    () 
                                    .false. 
                                    .false. 
                                    .false. 
                                    .false. 
                                    .false. 
                                    [] 
                                    [] 
                                    .false.
                                )
                            
                        }) 
                    b 
                    [] 
                    .false. 
                    .false.
                ), 
            c:
                (Module 
                    (SymbolTable
                        9
                        {
                            print_c:
                                (Function 
                                    (SymbolTable
                                        10
                                        {
                                            _lpython_return_variable:
                                                (Variable 
                                                    10 
                                                    _lpython_return_variable 
                                                    ReturnVar 
                                                    () 
                                                    () 
                                                    Default 
                                                    (Character 1 -2 () []) 
                                                    Source 
                                                    Public 
                                                    Required 
                                                    .false.
                                                )
                                            
                                        }) 
                                    print_c 
                                    [] 
                                    [(= 
                                        (Var 10 _lpython_return_variable) 
                                        (StringConstant 
                                            "C" 
                                            (Character 1 1 () [])
                                        ) 
                                        ()
                                    )
                                    (Return)] 
                                    (Var 10 _lpython_return_variable) 
                                    Source 
                                    Public 
                                    Implementation 
                                    () 
                                    .false. 
                                    .false. 
                                    .false. 
                                    .false. 
                                    .false. 
                                    [] 
                                    [] 
                                    .false.
                                )
                            
                        }) 
                    c 
                    [] 
                    .false. 
                    .false.
                ), 
            main_program:
                (Program 
                    (SymbolTable
                        11
                        {
                            
                        }) 
                    main_program 
                    [] 
                    []
                )
            
        }) 
    []
)
```
Instead of creating `ExternalSymbol` in the module `a`, we have to replace it with the actual Function, variables, ...
Also, we should not create the module `b` and `c` symtab.
**Reason:**
1.
```py
from a import b
```
Here `from a` will import the module `b`, so `import b` throws `b already defined` (Solution: we have to skip the `import b`)
2.
```py
import a
print(b.print_a())
```
The above code should throw an error. but it doesn't, as module `a` import module `b` also.

What are your thoughts? @certik @czgdp1807

czgdp1807:
I think we first need to complete https://github.com/lcompilers/lpython/pull/1065. I have assigned @Shaikh-Ubaid to try completing it.

certik:
@czgdp1807 I think the #1065 is needed, but that is the next step, an optimization. I think at first all we need to do when we see any import statement is to see if the module is already loaded in ASR, and if so, use it; otherwise find the file on the disk and load it to ASR and use it. When compiling a program (application) that depends on 100 dependencies, this approach will work.

The #1065 is for the use case when you want to precompile a library ahead of time. I think the ASR (without optimizations) is platform independent, so a given library can be fully precompiled even into a single ASR binary file, and then when you need anything from the library, you just quickly load that file. We definitely want that, but we can simply slowly do this as an optimization.

certik:
This looks great as a start, so I suggest we wrap this PR up (it's not going to be perfect), and then we'll send more PRs to iteratively improve various aspects of this.

TODO before merging:

* move the tests into integration_tests
* polish everything, I think no new features need to be added, just the code needs to be maintainable enough

czgdp1807:
> I think at first all we need to do when we see any import statement is to see if the module is already loaded in ASR, and if so, use it; otherwise find the file on the disk and load it to ASR and use it

This step is auto-compiling. One extra thing which #1065 is doing is saving the auto-compiled output in a `pyc` file. Correct me if I am wrong, so you are saying auto-compile a module if needed and include in the ASR in the TranslationUnit's global scope. Please let me know.

certik:
The bigger feature we'll have to implement (later!) is nesting modules. We need to design this well at the ASR level and then make it work. We'll work towards it iteratively.

certik:
@czgdp1807 I think so. I think we just need to get something working first, then improve iteratively. It seems the saving to `.pyc` files is not necessary for the first implementation, it is a performance improvement (that we obviously need long term), so we can work on the `.pyc` file compiling separately.

czgdp1807:
Sounds good to me. Auto-compile can be enabled easily and then including the module in global scope should also be simple. (AFAICT). So this seems doable.

Thirumalai-Shaktivel:
This PR is ready to go in.
Known issues to fix:
1. This should not work
```console
$ lpython -I /home/thirumalai/Open_Source/lpython integration_tests/test_import_01.py 
A
B
C
```
Instead do:
```console
$ lpython -I /home/thirumalai/Open_Source/lpython/integration_tests integration_tests/test_import_01.py
```
2.
```console
$ cd integration_tests
$ lpython test_import_01.py
[1]    22149 segmentation fault (core dumped)  lpython test_import_01.py
$ lpython -I /home/thirumalai/Open_Source/lpython/integration_tests test_import_01.py
$
```

Thirumalai-Shaktivel:
I got questions:
1. When to use `-I`? This feature is not available in CPython (Users do https://stackoverflow.com/questions/67631/how-do-i-import-a-module-given-the-full-path to import packages.)
2. is `-I` intented to do `lpython -I /home/thirumalai/Open_Source/lpython/integration_tests test_import_01.py`. If so, how do we test this in comparison with CPython?
P.S. using this PR `lpython integration_tests/test_import_01.py` works. Here lpython just look for packages using the parent_dir path (i.e., integration_tests)

src/libasr/utils.h
```diff
@@ -37,6 +37,7 @@ struct CompilerOptions {
     bool implicit_typing = false;
     bool implicit_interface = false;
     std::string target = "";
+    std::string import_path = "";
```
certik, line 40: This will probably later be `std::vector<std::string>`, but for now that's fine.

src/lpython/utils.h
```diff
@@ -4,11 +4,26 @@
...
+#ifndef _WIN32
+    #include <unistd.h>
+#endif
+
+#ifdef _WIN32
+    #define stat _stat
+    #if !defined S_ISDIR
+        #define S_ISDIR(m) (((m) & _S_IFDIR) == _S_IFDIR)
+    #endif
+#endif
```
certik, line 18: This needs to be moved to the `.cpp` file and later we'll move it to a dedicated file, since we want most of our C++ files to be platform agnostic.

certik (approved):
I think in general this looks good, I left some comments. We can later simply pass `CompilerOptions` into `python_ast_to_asr`, to avoid the long argument list.

@czgdp1807 go ahead and review this as well.

src/lpython/utils.h
Thirumalai-Shaktivel: Yup, I will move this to `utils.cpp` file.

certik (approved):
Looks great!

PR #1186, opened 2022-10-07
[DO NOT MERGE] Create a workflow for building using compilers from conda-forge

xmnlab:
This PR is a work in progress about creating a workflow for building lpython with compilers from conda-forge.

it is a proof-of-concept, any comment would be really appreciated.

currently, it is almost 100% (no documentation updated yet) but it is struggling in the linking step with binutils

last binutils from conda-forge is 2.36 .. on ubuntu 22.04 (my current system), the binutils version is 2.38 ... maybe I would need to work on the binutils on conda-forge ... although not sure if I would be able to do that .. at least it didn't work here locally.

I will keep this open in order to follow my investigation

### DIFF
--- a/build1.sh
+++ b/build1.sh
@@ -3,6 +3,29 @@
 set -e
 set -x
 
+if [[ -n "${CC}" ]]; then
+  CC=$(which "${CC}")
+fi
+
+if [[ -n "${CXX}" ]]; then
+  CXX=$(which "${CXX}")
+fi
+
+if [[ -n "${CONDA_PREFIX}" ]] && [[ "$(echo ${CC} | grep -m 1 -c ${CONDA_PREFIX})" == 1 ]]; then
+    GCC_INCLUDE_PATH=$(dirname $(find ${CONDA_PREFIX} -name cxxabi.h))
+    GCC_ARCH_INCLUDE_PATH=$(dirname $(dirname $(find ${GCC_INCLUDE_PATH} -name c++config.h)))
+
+    # ref: https://github.com/zdevito/ATen/issues/183
+    EXTRA_FLAGS="-fPIC -no-pie -D__STDC_FORMAT_MACROS"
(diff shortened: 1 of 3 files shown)


czgdp1807 (commented):
LGTM. @xmnlab Is this PR ready for review? If so please change the status of this PR from Draft to Open.

xmnlab:
thanks @czgdp1807 ! this PR is ready for review.

xmnlab:
btw, after we have this PR merged, if it is ok, I can work on a conda-forge package for lpython.

build1.sh
```diff
@@ -3,14 +3,41 @@
...
+  CC=$(which "${CC}")
+fi
+
+if [[ -n "${CXX}" ]]; then
+  CXX=$(which "${CXX}")
+fi
+
+if [[ -n "${CONDA_PREFIX}" ]] && [[ "$(echo ${CC} | grep -m 1 -c ${CONDA_PREFIX})" == 1 ]]; then
+    GCC_INCLUDE_PATH=$(dirname $(find ${CONDA_PREFIX} -name cxxabi.h))
+    GCC_ARCH_INCLUDE_PATH=$(dirname $(dirname $(find ${GCC_INCLUDE_PATH} -name c++config.h)))
```
certik, line 18: How will this work when I use Conda, but not compilers from it?

certik: I don't like the dirname, find and searching for `cxxabi.h`. It seems very fragile. Ideally we would just use cmake, and have only minimal bash script to show how to use it.

Can this logic be moved inside cmake with some more clean approach?

xmnlab: Sounds fair enough. I am not an expert on cmake, but I will give it a shot. Thanks for the review

certik:
I am marking it as Draft. If you are ready for another review, just mark it as ready.

xmnlab:
thanks, that makes sense. I am planning to work on that next week. I was on vacation .. and this week I am recovering from a health issue.
thanks for the review.

PR #1188, opened 2022-10-10
Sync `libasr` with LFortran

czgdp1807:
@certik @Shaikh-Ubaid @lucifer1004 @konradha @Pranavchiku @Smit-create @Thirumalai-Shaktivel Please go through the diff and let me know if there's any change that got reverted. I will restore it. I will wait for an hour or so before I create sync PR in lcompilers/libasr repository. Thanks.


Shaikh-Ubaid (approved):
It looks good to me, thank you.

src/libasr/ASR.asdl
```diff
@@ -360,6 +365,7 @@ cast_kind
     | IntegerToReal
     | LogicalToReal
     | RealToReal
+    | TemplateToReal
```
certik, line 368: This should not be here.

src/libasr/ASR.asdl
```diff
@@ -341,7 +342,11 @@ ttype
     | Dict(ttype key_type, ttype value_type)
     | Pointer(ttype type)
     | CPtr()
-    | TypeParameter(identifier param, dimension* dims)
+    | TypeParameter(identifier param, dimension* dims, restriction* rt)
+
+restriction = Restriction(trait rt)
+
+trait = SupportsZero | SupportsPlus | Divisible | Any
```
certik, line 349: This should be reverted as well.

certik (commented):
Some regressions crept in from LFortran, the template stuff should be reverted, since LPython had the latest version, we now regressed back to the older design.

czgdp1807:
Makes sense. Will revert the changes.

certik:
Thanks!

PR #1189, opened 2022-10-10
Modified generic function instantiation to allow nested generic function call

ansharlubis:
I modified the generic function instantiation so that a generic function such as
````
def sum(x: list[T], **kwargs) -> T:
    k: i32 = len(x)
    res: T
    res = zero(x[0])
    i: i32
    for i in range(k):
        res = add(res, x[i])
    return res

def mean(x: list[T], **kwargs) -> f64:
    s: T = sum(x)
    k: i32 = len(x)
    if k == 0:
        return 0.0
    return div(s, k)
````
is properly instantiated.

### CODE REVIEW

File: src/libasr/pass/instantiate_template.cpp [hunk 1]
```diff
@@ -10,25 +10,26 @@ namespace LFortran {
 class FunctionInstantiator : public ASR::BaseExprStmtDuplicator<FunctionInstantiator>
 {
 public:
-    SymbolTable *current_scope;
+    SymbolTable *func_scope;
+    SymbolTable *new_scope;
     std::map<std::string, ASR::ttype_t*> subs;
     std::map<std::string, ASR::symbol_t*> rt_subs;
-    std::string new_func_name;
+    // std::string new_func_name;
```
certik, line 17: If we don't use it, we can remove it:
```suggestion
```

certik (approved):
I think this looks good.

Conflicts have to be resolved.

certik (approved):
I think the changes look great. Go ahead and polish the git history and we can merge.

Issue #1174, opened 2022-10-04
Supporting packed structures in LPython


czgdp1807:
We need to support packed structures in LPython. The syntax can look like this,

```python
@packed(aligned=1)
class sSampleStruct:
     Data1: i8
     Data2: i32
     Data3: i8
     Data4: i8
```

should result in,

```c
typedef struct
{
     char Data1;
     int Data2;
     unsigned short Data3;
     char Data4;

} __attribute__((packed, aligned(1))) sSampleStruct;
```

Reference - https://stackoverflow.com/a/11772340

PR #1192, opened 2022-10-11
Supporting packed structs in LLVM/C backend

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1174


src/libasr/ASR.asdl
```diff
@@ -96,7 +96,7 @@ symbol
         symbol external, identifier module_name, identifier* scope_names,
         identifier original_name, access access)
     | StructType(symbol_table symtab, identifier name, identifier* members,
-        abi abi, access access, symbol? parent)
+        abi abi, access access, bool is_packed, expr? alignment, symbol? parent)
```
certik, line 99: Let's enumerate the options here:

* is_packed=true, alignment=nullptr
* is_packed=true, alignment=expr
* is_packed=false, alignment=nullptr
* is_packed=false, alignment=expr

I can see the following test kinds, which I annotated with the ASR case above:
```
# is_packed=true, alignment=expr
@packed(aligned=2)
@dataclass
class PackedStructAligned2:
    Data1: i8
    Data2: i32
    Data3: i8
    Data4: i8

# is_packed=true, alignment=nullptr
@packed
@dataclass
class PackedStructUnaligned:
    Data1: i8
    Data2: i32
    Data3: i8
    Data4: i8

# is_packed=false, alignment=nullptr
@dataclass
class A:
    y: f64
    x: i32
```

So we are missing a test for `is_packed=false, alignment=expr`, or this case must be excluded in verify().

What does alignment mean --- the alignment of the whole struct, or the alignment of each member of the struct?

certik (approved):
I think this looks good overall, I left a comment above that can be addressed in subsequent PRs.

src/libasr/ASR.asdl
czgdp1807: > What does alignment mean --- the alignment of the whole struct, or the alignment of each member of the struct?

https://stackoverflow.com/a/11772340 explains well. Seems like `aligned(X)` is the alignment of each member of the struct.

certik: Can you create an issue for the remaining tasks so that we do not forget?

certik: Here: https://github.com/lcompilers/lpython/issues/1201

PR #875, opened 2022-08-02
Add bit_length() for integers

Abdelrahman-Kh-Fouad:
Related to #281

Issue #281, opened 2022-03-24
Implement a set of built-in methods to be used on strings/integers/....


namannimmo10:
Since we now have the [`Attribute` visitor](https://github.com/lcompilers/lpython/blob/main/src/lpython/semantics/python_ast_to_asr.cpp#L1844) in AST->ASR, we should start implementing the set of built-in methods to be used on strings, integers, lists, dictionaries, etc.

Here's a list of those operating on strings: https://www.w3schools.com/python/python_ref_string.asp
And one can navigate on the left for other types.

Let's start with something like `bit_length` for integers:
```python
>>> num = 9
>>> num.bit_length()
4
```

certik:
Yes. Implement them as `_lpython_bit_length(i: i32)`, make the AST to ASR insert these in. Those that can be implemented in Python should be. Otherwise just do the declaration and the backend (ASR->LLVM) has to implement them.

namannimmo10:
We can try this:
```python
>>> def _lpython_bit_length(i: int) -> int:
...   x: str
...   x = bin(i)
...   if i == 0: return 0
...   elif i < 0: x = x[3:]
...   else: x = x[2:]
...
...   return len(x)
...
>>> all(_lpython_bit_length(i) == i.bit_length() for i in range(-1000, 1000))
True
>>>
```

Smit-create:
Just to make the design cleaner and not just put everything into `Attribute`, I suggest creating a map something similar to intrinsic procedures, where we can directly look for the methods (for example, maybe we can map `list.append` to a string `list%append` and use this converted string to find the appropriate function.)

certik:
We should try to systematically manage these methods, and at the same time keep the complexity low at the C++ level. Having some kind of a map like we do for `comptime_eval` is fine I think.

Shaikh-Ubaid:
Please, could someone possibly share if these built in methods need to be implemented as compile time functions (like [here](https://github.com/lcompilers/lpython/blob/64e0c33ca22e4cf0bf21c56fd78286ee1721e51d/src/lpython/semantics/python_comptime_eval.h#L100) ) or runtime functions (like [here](https://github.com/lcompilers/lpython/blob/64e0c33ca22e4cf0bf21c56fd78286ee1721e51d/src/runtime/lpython_builtin.py#L52)) or both?

certik:
[...]

PR #1196, opened 2022-10-12
Add builtins

zeproffesor:
I have implemented the compile time function (mostly by monkey-ing the changes made in #875) that will be called when the `int.to_bytes()` method will be called.

As per discussion (#281) the implementation currently returns a string only and NOT an instance of the `class bytes` which is yet to be implemented.

Creating a draft PR so that maintainers can keep an eye on the progress.

Please feel free to leave comments because this is my first PR on any open source project so it will be very helpful! Thanks!

Related to #281

### DIFF
--- a/src/lpython/semantics/python_attribute_eval.h
+++ b/src/lpython/semantics/python_attribute_eval.h
@@ -18,6 +18,7 @@ struct AttributeHandler {
 
     AttributeHandler() {
         attribute_map = {
+            {"int@to_bytes", &eval_int_to_bytes},
             {"int@bit_length", &eval_int_bit_length},
             {"list@append", &eval_list_append},
             {"list@remove", &eval_list_remove},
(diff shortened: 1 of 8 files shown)


certik:
Thanks! In general we try to avoid adding things to ASR unless necessary. There might be a way to implement this in our regular runtime library, in which case we should do that.

zeproffesor:
> Thanks! In general we try to avoid adding things to ASR unless necessary. There might be a way to implement this in our regular runtime library, in which case we should do that.

Okay, I was following what was done to add the int.bit_length() method, which had both a runtime as well as a compile time implementation. I'll work on the run time implementation for now and we can figure out what we want to keep or remove later. Thanks for the feedback!

zeproffesor:
Hello @certik can we please meet over video so that you can explain your idea about going about creating the required function without adding a new ASR node?

certik:
Yes, please contact me on Zulip (https://lfortran.zulipchat.com/) or email and let's schedule it.

certik:
So I think the current approach of returning the string representation of bytes is a good one. Currently it is implemented at compile time, which is fine.  Regarding runtime implementation, there are multiple approaches:

* Reuse the `BitCast` node: https://github.com/lcompilers/lpython/issues/1268
* Add a new node `IntegerToBytes` (currently in this PR)
* Call a runtime function (no new ASR node)

src/lpython/semantics/python_attribute_eval.h
```diff
@@ -74,6 +75,17 @@ struct AttributeHandler {
         return ASR::make_IntegerBitLen_t(al, loc, s, int_type, nullptr);
     }
 
+    static ASR::asr_t* eval_int_to_bytes(ASR::expr_t *s, Allocator &al, const Location &loc,
+            Vec<ASR::expr_t*> &args, diag::Diagnostics &/*diag*/) {
+        if (args.size() != 0) {
+            throw SemanticError("int.bit_length() takes no arguments", loc);
```
certik, line 81:
```suggestion
            throw SemanticError("int.to_bytes() takes no arguments", loc);
```

src/lpython/semantics/python_attribute_eval.h
```diff
@@ -74,6 +75,16 @@ struct AttributeHandler {
...
     }
 
+    static ASR::asr_t* eval_int_to_bytes(ASR::expr_t *s, Allocator &al, const Location &loc,
+            Vec<ASR::expr_t*> &args, diag::Diagnostics &/*diag*/) {
+        if (args.size() != 2) {
+            throw SemanticError("int.to_bytes(i, s) takes exactly two arguments", loc);
+        }
+        ASR::ttype_t *char_type = ASRUtils::TYPE(ASR::make_Character_t(al, loc,
+                                        1, -1, nullptr, nullptr, 0));
+        return ASR::make_IntegerToBytes_t(al, loc, s, char_type, nullptr);
```
certik, line 85: The code from `python_ast_to_asr` that computes the compile time evaluation should be moved here and it would be returned in the last argument.

certik:
TODO:
* [ ] finish the actual implementation in `src/libasr/runtime/lfortran_intrinsics.c`
* [ ] update the integration test to only test the correct answer

PR #1197, opened 2022-10-13
Add missing operators in AugAssign

Smit-create:


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2335,18 +2335,17 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
-        args.reserve(al, 2);
-        ASR::call_arg_t arg1, arg2;
-        arg1.loc = left->base.loc;
-        arg1.m_value = left;
-        args.push_back(al, arg1);
-        arg2.loc = right->base.loc;
-        arg2.m_value = right;
-        args.push_back(al, arg2);
-        Vec<ASR::restriction_arg_t*> rt_args;
-        rt_args.reserve(al, 1);
```
czgdp1807, line 2348: I am not sure if this is the right thing to do. Would confirm from @certik @ansharlubis

czgdp1807 (commented):
Other than the code deletion everything looks okay to me.

src/lpython/semantics/python_ast_to_asr.cpp
czgdp1807: @Smit-create What would happen if you restore this piece of code? Will your PR stop working?

Smit-create: Nope, it was an unused variable so I removed that.

czgdp1807: I see. If the presence of this code generates a compiler warning then keep it removed, otherwise keep it.

Smit-create: > I don't see any unused variable warning with the presence of this code

That is because we have a line `rt_args.reserve(al, 1);`.

czgdp1807: I see. Makes sense.

czgdp1807 (approved):

PR #1203, opened 2022-10-15
Support `Const[...]` type annotation in LPython syntax (LLVM/C)

### LINKED ISSUE #1175: Supporting `Const[...]` annotated variables
As the title says some syntax examples,

1. `x: Const[i32]` is a constant 32 bit integer variable.
2. `x: Const[CPtr]` is constant C pointer (`void*`) variable.

Any such variable declared in global scope should become macros in C backend. For example, `x: Const[i32] = 0` should become `#define x 0` in C backend.

Posting the link to Zulip chat discussion on the same topic - https://lfortran.zulipchat.com/#narrow/stream/311866-LPython/topic/Constant.20Variables

czgdp1807:
Closes #1175

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -807,6 +807,13 @@ static inline std::string get_type_code(const ASR::ttype_t *t, bool use_undersco
             }
             return "Pointer[" + get_type_code(p->m_type, use_underscore_sep, encode_dimensions_) + "]";
         }
+        case ASR::ttypeType::Const: {
+            ASR::Const_t* p = ASR::down_cast<ASR::Const_t>(t);
+            if( use_underscore_sep ) {
+                return "Const_" + get_type_code(p->m_type, use_underscore_sep, encode_dimensions_) + "_";
+            }
+            return "Const[" + get_type_code(p->m_type, use_underscore_sep, encode_dimensions_) + "]";
+        }
         default: {
             throw LCompilersException("Type encoding not implemented for "
                                       + std::to_string(t->type));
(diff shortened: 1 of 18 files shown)


src/libasr/asr_utils.h
```diff
@@ -105,6 +105,33 @@ static inline ASR::ttype_t* symbol_type(const ASR::symbol_t *f)
...
+            ASR::Pointer_t* pointer_asr = ASR::down_cast<ASR::Pointer_t>(asr_type);
+            return pointer_asr->m_type;
+        }
+        case ASR::ttypeType::Const: {
+            ASR::Const_t* const_asr = ASR::down_cast<ASR::Const_t>(asr_type);
+            return const_asr->m_type;
+        }
+        default: {
+            return asr_type;
+        }
```
certik, line 131: What if it is a `Const[Pointer[i32]]`?

certik: For now, let's add (in subsequent PRs) tests to verify() that only single level of nesting is allowed for Const and Pointer.

certik (approved):
Excellent, very good!

Good to merge as is, after polishing git history.

I left some comments how to robustify verify(), please either do it or open up an issue.

src/libasr/asr_utils.h
czgdp1807: > What if it is a `Const[Pointer[i32]]`?

Well, in that case we will receive, `Pointer[i32]` which is what we need? Is there any case where we need `i32` out of `Const[Pointer[i32]]`. In fact we can specify the depth upto which we want to unwrap a given type. But yeah that should be done when we find the need for it via a concrete case. I will open an issue for this.

PR #1204, opened 2022-10-17
Modified generic instantiation to allow nested generic calls

ansharlubis:


src/libasr/asr_utils.h
```diff
@@ -1081,6 +1081,27 @@ static inline bool is_generic(ASR::ttype_t &x) {
...
 
+static inline bool is_generic_function(ASR::symbol_t *x) {
+    switch (x->type) {
+        case ASR::symbolType::Function: {
+            ASR::Function_t *func_sym = ASR::down_cast<ASR::Function_t>(x);
+            return func_sym->n_type_params > 0;
+        }
+        default: return false;
+    }
+}
```
czgdp1807, line 1092: Do we need a switch for this? I think if-else will fit better as only `Function` is to be checked and for all other cases the result would be `false`.

czgdp1807 (commented):
Rest looks fine to me.

src/libasr/pass/instantiate_template.cpp
```diff
@@ -401,14 +404,54 @@ class FunctionInstantiator : public ASR::BaseExprStmtDuplicator<FunctionInstanti
...
+        generic_func_nums[func_name] = new_function_num + 1;
+        std::string new_func_name = "__lpython_generic_" + func_name + "_" + std::to_string(new_function_num);
+        generic_func_subs[new_func_name] = subs;
+        t = ASR::down_cast<ASR::symbol_t>(this->instantiate_Function(func, new_func_name));
+        return t;
+    }
 };
 
+std::map<std::string, int> FunctionInstantiator::generic_func_nums;
+std::map<std::string, std::map<std::string, ASR::ttype_t*>> FunctionInstantiator::generic_func_subs;
```
certik, line 449: Are these global variables? I think that is asking for trouble regarding parallel compilation and other future improvements. I would not keep any global state.

ansharlubis: Got it, I was trying to reuse already generated functions. But I think it can easily be done by tracing back to the parent scope.

ansharlubis:
I seem to find problems when testing `test_import_02.c` probably because of some encoding issue.

src/libasr/asr_utils.h
ansharlubis: Got it. Modified it.

ansharlubis:
Okay, I'll make a separate PR to make LPython's implementation work with LFortran's `instantiate_template.cpp`.

PR #1205, opened 2022-10-17
CPP: Implement list

Smit-create:
@certik @czgdp1807 

1. The current implementation uses `list_api` which is the same as in `C`. Should also add an option to use STL libraries in CPP like `vector` which might provide cleaner code just from developer's point of view. I'm not sure about the performance difference and would try the benchmarks using the current `list_api` and `std::vector`.
2. One more question is should we keep using `Kokkos`?
3. Once we figure out the above 2 solutions we can also start testing the CPP backend in integration tests?

### DIFF
--- a/src/libasr/codegen/asr_to_cpp.cpp
+++ b/src/libasr/codegen/asr_to_cpp.cpp
@@ -341,6 +341,12 @@ public:
                 } else {
                     sub = format_type(dims, "struct", v.m_name, use_ref, dummy);
                 }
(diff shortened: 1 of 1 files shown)


certik (approved):

czgdp1807:
Seems like `list_api` is faster?

Smit-create:
> Seems like `list_api` is faster?

Yes

czgdp1807:
Try with O3 optimisation for clang in both the cases you have considered.

PR #1207, opened 2022-10-17
C: Add Tuple support

Smit-create:


Smit-create:
I have refactored the CPPList and other common functions in a new utils file. I have started implementing tuples with the current approach, I am just creating a struct with all tuple element types:
```c
// tuple[i32, f64]
struct tuple_c{
    int32_t element_1;
    double element_2;
};
```

The current issue is that we need to figure out the best approach to implement a tuple so that we can also easily access the elements (like `tuple[1]`) based on the position and also figure out the type at runtime.

czgdp1807 (approved):

PR #1211, opened 2022-10-19
Adding dependencies in `StructType`, `UnionType` and `EnumType`

czgdp1807:


czgdp1807:
@certik This is ready. I will add dependencies for Variable symbol later. For now dependencies in aggregate types is the immediate requirement.

certik (approved):
The PR looks good. I can see you construct the dependencies and check them in verify.

But I don't see you using the dependencies in the LLVM backend -- is it not needed there? I can see you are using them in the C backend.

czgdp1807:
> But I don't see you using the dependencies in the LLVM backend -- is it not needed there

I think so. The test I added works fine in LLVM already. C backend needs those dependencies because the order of declaration matters there.

certik:
Makes sense.

PR #1213, opened 2022-10-20
[Sprint] Feature addition in ASR and C/LLVM backends

czgdp1807:

PR #1214, opened 2022-10-20
Allowing using global constants to initialise other global constants

czgdp1807:
Taken from https://github.com/lcompilers/lpython/pull/1213/files

### DIFF
--- a/integration_tests/const_03.py
+++ b/integration_tests/const_03.py
@@ -1,0 +1,12 @@
+from ltypes import i64, f64, Const
+
+CONST_1: Const[f64] = 32.0
+CONST_2: Const[f64] = CONST_1 * 2.0
+CONST_3: Const[i64] = CONST_1 + CONST_2
+
+def test_global_consts():
+    assert CONST_1 == 32
+    assert CONST_2 == 64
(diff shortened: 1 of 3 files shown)


integration_tests/const_03.py
```diff
@@ -0,0 +1,12 @@
+from ltypes import i64, f64, Const
+
+CONST_1: Const[f64] = 32.0
+CONST_2: Const[f64] = CONST_1 * 2.0
+CONST_3: Const[i64] = CONST_1 + CONST_2
```
certik, line 5: Let's flip the numbers, so that they are ordered (by name?) and ensure the dependencies actually work:
```suggestion
CONST_3: Const[f64] = 32.0
CONST_2: Const[f64] = CONST_3 * 2.0
CONST_1: Const[i64] = CONST_3 + CONST_2
```

certik (approved):
Looks great, thanks!

The suggestion can be done in a subsequent PR.

integration_tests/const_03.py
czgdp1807: Variables don't have dependencies right now. But when I will add dependencies to variables I will use this as a test.

PR #1236, opened 2022-10-27
Always declare global constants in LLVM backend

czgdp1807:
We shouldn't directly use the values of global constants in LLVM backend. Because by doing so we skip the declaration of these global constants and in turn they become un-usable in the rest of the program (like passing them as an argument to a function). The other way is to always check whether a variable is a global constant or not and if its a global constant then use the compile time value. However IMO, that's a very bug prone path as we will almost always forget to do such checks while making use of `ASR::Variable_t` in LLVM backend. So, declaring it is a decent and robust solution. We can optimise by directly using values on a case by case basis (like if we find the need to do it in a particular then do it otherwise fallback global variable is always present).

### DIFF
--- a/integration_tests/const_03.py
+++ b/integration_tests/const_03.py
@@ -1,10 +1,19 @@
-from ltypes import i64, f64, Const
+from ltypes import i64, f64, i32, Const
 
 CONST_1: Const[f64] = 32.0
 CONST_2: Const[f64] = CONST_1 * 2.0
 CONST_3: Const[i64] = CONST_1 + CONST_2
 
+CONST_11: Const[i32] = 32
+CONST_12: Const[i32] = CONST_11
+
+def print_value(value: Const[f64]):
+    print(value)
+
 def test_global_consts():
+    print_value(CONST_1)
+    print_value(CONST_2)
+    print(CONST_12)
     assert CONST_1 == 32
     assert CONST_2 == 64
     assert abs(CONST_3 - 96.0) < 1e-12
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -1183,6 +1183,9 @@ static inline int extract_kind_from_ttype_t(const ASR::ttype_t* type) {
         case ASR::ttypeType::Pointer: {
             return extract_kind_from_ttype_t(ASR::down_cast<ASR::Pointer_t>(type)->m_type);
         }
+        case ASR::ttypeType::Const: {
+             return extract_kind_from_ttype_t(ASR::down_cast<ASR::Const_t>(type)->m_type);
+         }
         default : {
             return -1;
         }
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -5668,11 +5668,24 @@ public:
                         } else {
                             auto finder = std::find(nested_globals.begin(),
                                     nested_globals.end(), h);
-                            LFORTRAN_ASSERT(finder != nested_globals.end());
-                            llvm::Value* ptr = module->getOrInsertGlobal(nested_desc_name,
-                                nested_global_struct);
-                            int idx = std::distance(nested_globals.begin(), finder);
-                            tmp = CreateLoad(llvm_utils->create_gep(ptr, idx));
+                            if (finder == nested_globals.end()) {
+                                if (arg->m_value == nullptr) {
+                                    throw CodeGenError(std::string(arg->m_name) + " isn't defined in any scope.");
+                                }
+                                this->visit_expr_wrapper(arg->m_value, true);
+                                llvm::BasicBlock &entry_block = builder->GetInsertBlock()->getParent()->getEntryBlock();
+                                llvm::IRBuilder<> builder0(context);
+                                builder0.SetInsertPoint(&entry_block, entry_block.getFirstInsertionPt());
+                                llvm::AllocaInst *target = builder0.CreateAlloca(
+                                    get_type_from_ttype_t_util(arg->m_type), nullptr, "call_arg_value");
+                                builder->CreateStore(tmp, target);
+                                tmp = target;
+                            } else {
+                                llvm::Value* ptr = module->getOrInsertGlobal(nested_desc_name,
+                                    nested_global_struct);
+                                int idx = std::distance(nested_globals.begin(), finder);
+                                tmp = CreateLoad(llvm_utils->create_gep(ptr, idx));
+                            }
                         }
                     } else if (is_a<ASR::Function_t>(*symbol_get_past_external(
                         ASR::down_cast<ASR::Var_t>(x.m_args[i].m_value)->m_v))) {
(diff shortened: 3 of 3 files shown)


certik:
Can you give an example when the constant needs to be declared? Why not always using the value?

czgdp1807:
https://github.com/lcompilers/lpython/blob/d4eb9e683d5d3f3bdf55e8ff718ca5d343f2f625/integration_tests/const_03.py itself is an example here.

You can see `print_value(CONST_1)` call. If we don't declare them then the following check fails,

https://github.com/lcompilers/lpython/blob/de3337a48bd6ecadc30ddee7a81533eee992d3d9/src/libasr/codegen/asr_to_llvm.cpp#L5669

czgdp1807:
We can say that always check for constant-ness of global variables and use its value if needed. I can figure out a robust solution for that I think. But still I feel always declaring global variables anyways will ensure that the user's code will always work. Using value of global constants is an optimisation so should be done on a case by case basis.

czgdp1807:
> The first option is much faster and easier it seems. So let's explore it a bit more.

I don't think so AFAICT. 

> I looked at the example and I don't see why we cannot simply use the value of the constant in the print statement.

In print statement definitely you can use the value of the global constant. My fix is for passing global constants as arguments to user defined functions such as `print_value`. See the generated LLVM code,

You can see the signature of `print_value` (which is user defined) is `define void @print_value(double* %value)` and the call to `print_value` is, `call void @print_value(double* @CONST_1)`. Now if you want to pass the value of `@CONST_1` directly to `print_value` then there are two options,

1. **Change the signature of `print_value` to, `define void @print_value(double %value)`**. But then you won't be able to pass variables to `print_value` because you will always have to load them. In addition to that at the LLVM level you will have to always check whether the variable is a constant if so always load it, a very tricky and bug prone change to make.
2. Declare global constants and use those as arguments to user defined functions. This is what I am doing in this PR.

We can go for option 1 in future but for now option 2 is the fastest way to move things forward. Let me know how should we proceed.


```llvm
; ModuleID = 'LFortran'
source_filename = "LFortran"

%complex_8 = type { double, double }

@CONST_1 = global double 3.200000e+01
@CONST_11 = global i32 32
@CONST_12 = global i32 32
@CONST_2 = global double 6.400000e+01
@CONST_3 = global i64 96
@0 = private unnamed_addr constant [2 x i8] c" \00", align 1
@1 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@2 = private unnamed_addr constant [10 x i8] c"%23.17f%s\00", align 1
@3 = private unnamed_addr constant [2 x i8] c" \00", align 1
@4 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@5 = private unnamed_addr constant [5 x i8] c"%d%s\00", align 1
@6 = private unnamed_addr constant [16 x i8] c"AssertionError\0A\00", align 1
@7 = private unnamed_addr constant [16 x i8] c"AssertionError\0A\00", align 1
@8 = private unnamed_addr constant [16 x i8] c"AssertionError\0A\00", align 1

define void @_lpython_main_program() {
.entry:
  call void @test_global_consts()
  br label %return

  ret void
}

define void @print_value(double* %value) {
.entry:
  %0 = load double, double* %value, align 8
  call void (i8*, ...) @_lfortran_printf(i8* getelementptr inbounds ([10 x i8], [10 x i8]* @2, i32 0, i32 0), double %0, i8* getelementptr inbounds ([2 x i8], [2 x i8]* @1, i32 0, i32 0))
  br label %return

  ret void
}

define void @test_global_consts() {
.entry:
  %call_arg_value = alloca double, align 8
  call void @print_value(double* @CONST_1)
  call void @print_value(double* @CONST_2)
  call void (i8*, ...) @_lfortran_printf(i8* getelementptr inbounds ([5 x i8], [5 x i8]* @5, i32 0, i32 0), i32 32, i8* getelementptr inbounds ([2 x i8], [2 x i8]* @4, i32 0, i32 0))
  br i1 true, label %then, label %else

then:                                             ; preds = %.entry
  br label %ifcont

else:                                             ; preds = %.entry
  call void (i8*, ...) @_lcompilers_print_error(i8* getelementptr inbounds ([16 x i8], [16 x i8]* @6, i32 0, i32 0))
  call void @exit(i32 1)
  br label %ifcont

ifcont:                                           ; preds = %else, %then
  br i1 true, label %then1, label %else2

then1:                                            ; preds = %ifcont
  br label %ifcont3

else2:                                            ; preds = %ifcont
  call void (i8*, ...) @_lcompilers_print_error(i8* getelementptr inbounds ([16 x i8], [16 x i8]* @7, i32 0, i32 0))
  call void @exit(i32 1)
  br label %ifcont3

ifcont3:                                          ; preds = %else2, %then1
  store double 0.000000e+00, double* %call_arg_value, align 8
  %0 = call double @__module_lpython_builtin___lpython_overloaded_0__abs(double* %call_arg_value)
  %1 = fcmp ult double %0, 0x3D719799812DEA11
  br i1 %1, label %then4, label %else5

then4:                                            ; preds = %ifcont3
  br label %ifcont6

else5:                                            ; preds = %ifcont3
  call void (i8*, ...) @_lcompilers_print_error(i8* getelementptr inbounds ([16 x i8], [16 x i8]* @8, i32 0, i32 0))
  call void @exit(i32 1)
  br label %ifcont6

ifcont6:                                          ; preds = %else5, %then4
  br label %return

return:                                           ; preds = %ifcont6
  ret void
}

define double @__module_lpython_builtin___lpython_overloaded_0__abs(double* %x) {
.entry:
  %_lpython_return_variable = alloca double, align 8
  %result = alloca double, align 8
  %0 = load double, double* %x, align 8
  %1 = fcmp uge double %0, 0.000000e+00
  br i1 %1, label %then, label %else

then:                                             ; preds = %.entry
  %2 = load double, double* %x, align 8
  store double %2, double* %result, align 8
  br label %ifcont

else:                                             ; preds = %.entry
  %3 = load double, double* %x, align 8
  %4 = fsub double 0.000000e+00, %3
  store double %4, double* %result, align 8
  br label %ifcont

ifcont:                                           ; preds = %else, %then
  %5 = load double, double* %result, align 8
  store double %5, double* %_lpython_return_variable, align 8
  br label %return

unreachable_after_return:                         ; No predecessors!
  br label %return

return:                                           ; preds = %unreachable_after_return, %ifcont
  %6 = load double, double* %_lpython_return_variable, align 8
  ret double %6
}

declare float @_lfortran_caimag([2 x float])

declare double @_lfortran_zaimag(%complex_8)

declare void @_lcompilers_print_error(i8*, ...)

declare void @exit(i32)

define i32 @main() {
.entry:
  call void @_lpython_main_program()
  ret i32 0
}
```

certik:
I think we have exactly the same issue if you call the function with a value like `f(5)`. I think we pass it by reference. So we simply can pass the constant in exactly the same way.

czgdp1807:
> I think we have exactly the same issue if you call the function with a value like `f(5)`. I think we pass it by reference. So we simply can pass the constant in exactly the same way.

Makes sense. Will try to follow the same pattern here. Will update you shortly.

certik (approved):
I think that this is fine. Isn't this change simpler than the previous approach?

I think later we can figure out some improvements to ASR to simplify the backend handling of these "parent variables". I suggest we tackle that later, after implementing nested functions in all combinations (closures). I know we have some bugs there.

PR #1243, opened 2022-10-30
ASR: Allow variable access from module using dot

### LINKED ISSUE #1242: ASR: [BUG] Error while using dot to access module variable
```py
% cat a.py
import math

def f():
    print(math.pi, math.e)

f()
```

LPython Output:
```
% lpython a.py 
semantic error: Only Variable type is supported for now in Attribute
 --> a.py:4:11
  |
4 |     print(math.pi, math.e)
  |           ^^^^^^^ 
```

CPython:
```
% python a.py 
3.141592653589793 2.718281828459045
```

Smit-create:
Fixes #1242

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -395,14 +395,15 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
 
 ASR::symbol_t* import_from_module(Allocator &al, ASR::Module_t *m, SymbolTable *current_scope,
                 std::string mname, std::string cur_sym_name, std::string new_sym_name,
-                const Location &loc) {
+                const Location &loc, bool skip_current_scope_check=false) {
     ASR::symbol_t *t = m->m_symtab->resolve_symbol(cur_sym_name);
     if (!t) {
         throw SemanticError("The symbol '" + cur_sym_name + "' not found in the module '" + mname + "'",
                 loc);
     }
-    if (current_scope->get_scope().find(cur_sym_name) != current_scope->get_scope().end()) {
-        throw SemanticError(cur_sym_name + " already defined", loc);
+    if (!skip_current_scope_check &&
+        current_scope->get_scope().find(new_sym_name) != current_scope->get_scope().end()) {
+        throw SemanticError(new_sym_name + " already defined", loc);
     }
     if (ASR::is_a<ASR::Function_t>(*t)) {
         ASR::Function_t *mfn = ASR::down_cast<ASR::Function_t>(t);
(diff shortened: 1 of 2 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4391,6 +4391,13 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                 tmp = ASR::make_EnumValue_t(al, x.base.base.loc, enum_member_var,
                         enum_t, enum_member_variable->m_type,
                         ASRUtils::expr_value(enum_member_variable->m_symbolic_value));
+            } else if (ASR::is_a<ASR::Module_t>(*t)) {
+                ASR::Module_t *m = ASR::down_cast<ASR::Module_t>(t);
+                ASR::symbol_t *sym = import_from_module(al, m, current_scope, value,
+                                    x.m_attr, x.m_attr, x.base.base.loc);
+                LFORTRAN_ASSERT(ASR::is_a<ASR::ExternalSymbol_t>(*sym));
+                current_scope->add_symbol(x.m_attr, sym);
```
czgdp1807, line 4399: What if its already imported and added? We will add it to the current scope even though its already there?

Smit-create: I think that here we should change the name of the symbol too. If we are already using some variable with the same name then it might be a conflict, for example:
```py
import math

def f():
    pi: f64 = 4.5030
    print(math.pi, math.e)

f()
```
This gives an error:
```
semantic error: pi already defined
```

To tackle this, we need to change the symbol name to `module_name@attr_name`.

> We will add it to the current scope even though its already there?

`import_from_module` ensures that the symbol is not present in the `current_scope`.

czgdp1807: > To tackle this, we need to change the symbol name to `module_name@attr_name`.

Correct, name mangling is needed here I think. @certik Thoughts?

> `import_from_module` ensures that the symbol is not present in the `current_scope`.

Please test this. Try using `math.pi` twice.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -395,13 +395,14 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
...
-                const Location &loc) {
+                const Location &loc, bool skip_current_scope_check=false) {
     ASR::symbol_t *t = m->m_symtab->resolve_symbol(cur_sym_name);
     if (!t) {
         throw SemanticError("The symbol '" + cur_sym_name + "' not found in the module '" + mname + "'",
                 loc);
     }
-    if (current_scope->get_scope().find(cur_sym_name) != current_scope->get_scope().end()) {
+    if (!skip_current_scope_check &&
+        current_scope->get_scope().find(cur_sym_name) != current_scope->get_scope().end()) {
```
czgdp1807, line 405: I think the check should be on `new_sym_name` as that's the mangled thing, correct?

Smit-create: Yes, I think that's a typo and needs to be fixed. Thanks!

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: Done!

czgdp1807 (approved):

PR #1248, opened 2022-11-01
Fix `Pointer[CPtr]` passing to ``BindC`` functions

czgdp1807:
Taken from https://github.com/lcompilers/lpython/pull/1247/files


certik (approved):
Looks good to me!

Issue #1249, opened 2022-11-01
ASR: Support array storage in memory order - row major or col major


Shaikh-Ubaid:
First encountered here https://github.com/lcompilers/lpython/pull/1118#discussion_r973492701

https://github.com/lcompilers/lpython/blob/8ec42d225a08f5b3d6bce9dd68680ddabdfb8a6c/src/libasr/codegen/asr_to_wasm.cpp#L1432-L1442

Currently, when syncing `lpython` and `lfortran` libasr's, we are needing to keep track of this. Both the `libasr's` slightly diverge due to it.

certik:
@czgdp1807 another argument to possibly use `Array` ASR node.

czgdp1807:
The refactoring makes sense. The question is when do you want to do it?

certik:
After we deliver on Minpack. We are close, don't want to delay it.

certik:
Solution how to do this: https://github.com/lfortran/lfortran/issues/1601, the row/column major becomes part of the physical type.

PR #1252, opened 2022-11-02
ASR: Support Array storage order

Shaikh-Ubaid:
towards #1249


src/libasr/ASR.asdl
```diff
@@ -273,8 +273,8 @@ expr
 
     | Var(symbol v)
 
-    | ArrayConstant(expr* args, ttype type)
-    | ArrayItem(expr v, array_index* args, ttype type, expr? value)
+    | ArrayConstant(expr* args, ttype type, arraystorage storage_format)
+    | ArrayItem(expr v, array_index* args, ttype type, arraystorage storage_format, expr? value)
     | ArraySection(expr v, array_index* args, ttype type, expr? value)
     | ArraySize(expr v, expr? dim, ttype type, expr? value)
     | ArrayBound(expr v, expr? dim, ttype type, arraybound bound,
```
Shaikh-Ubaid, line 280: Currently, the `arraystorage` was being needed in the `ArrayItem` in the `wasm` backend. Hence, I added the attribute `arraystorage` only to `ArrayItem` and also `ArrayConstant`.

Shaikh-Ubaid: It seems it is possible that the `arraystorage` might be needed in others like `ArraySection`. I guess, we can add it as the need arises.

Shaikh-Ubaid:

Shaikh-Ubaid:
I will send a `PR` to `LFortran` with these changes once this is complete/merged.

certik:
It seems it should rather be part of the type. Here is one way to do it cleanly:
```
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -327,22 +327,23 @@ expr
 --     integer, so we also use kind=4 for the default logical.)
 
 ttype
-    = Integer(int kind, dimension* dims)
-    | Real(int kind, dimension* dims)
-    | Complex(int kind, dimension* dims)
-    | Character(int kind, int len, expr? len_expr, dimension* dims)
-    | Logical(int kind, dimension* dims)
+    = Integer(int kind)
+    | Real(int kind)
+    | Complex(int kind)
+    | Character(int kind, int len, expr? len_expr)
+    | Logical(int kind)
     | Set(ttype type)
     | List(ttype type)
     | Tuple(ttype* type)
-    | Struct(symbol derived_type, dimension* dims)
-    | Enum(symbol enum_type, dimension *dims)
-    | Union(symbol union_type, dimension *dims)
-    | Class(symbol class_type, dimension* dims)
+    | Struct(symbol derived_type)
+    | Enum(symbol enum_type)
+    | Union(symbol union_type)
+    | Class(symbol class_type)
     | Dict(ttype key_type, ttype value_type)
     | Pointer(ttype type)
+    | Array(ttype type, dimension* dims, arraystorage storage_format)
     | CPtr()
-    | TypeParameter(identifier param, dimension* dims)
+    | TypeParameter(identifier param)
 
 restriction_arg = RestrictionArg(identifier restriction_name, symbol restriction_func)
 
```
But that requires more refactoring, so we can do it in a separate PR.

certik (approved):
I think this is good for now, but we'll have to later refactor how we handle arrays and move this into the type itself I think.

PR #1255, opened 2022-11-02
C: Codegen to Compare data types

Smit-create:


src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -78,8 +78,7 @@ class BaseCCPPVisitor : public ASR::BaseVisitor<Struct>
     size_t template_number;
     std::string from_std_vector_helper;
 
-    std::unique_ptr<CCPPList> list_api;
-    std::unique_ptr<CCPPTuple> tuple_api;
+    std::unique_ptr<CCPPDSUtils> c_ds_api;
```
czgdp1807, line 81: Is this for wrapping up `list_api` and `tuple_api`?

Smit-create: I removed the individual classes of List/Tuple as there was a lot of repetitive code logic and also it needed frequent interaction of both the classes, I suppose that will be also needed for upcoming data structures like `dict`, `set`, etc.

czgdp1807 (approved):

PR #1256, opened 2022-11-02
Add an option: `--debug-with-line-column` to convert linear location to line+column in the debug info

Thirumalai-Shaktivel:
Related: #772


certik (approved):
Very good!

Thanks for this. Yes, makes sense to test `--debug-with-line-column` in `run_test.py`, since if that works, the linear location also works.

Thirumalai-Shaktivel:
Thanks for the review!

Issue #1222, opened 2022-10-22
WASM to X86 Backend


Shaikh-Ubaid:
There is an idea to implement a `wasm` to `x86` backend. We have to benchmark `ASR->x86` and `ASR->WASM->x86`. If the performance difference is not much, then the big advantage of `WASM->x86` would (hopefully) be that it would be a easier and quicker for us to deliver a working backend.

- We can start with 32bit, since it is already figured out, unless it is more complicated,then we can just do 64 bit.

***Notes:**
- This was first discussed at https://github.com/certik/test_wasm/pull/4#issuecomment-1135992438 (last line) and https://github.com/certik/test_wasm/pull/4#issuecomment-1136375282 (last paragraph).

Shaikh-Ubaid:
For `x86` do we need the binary format or text based format or both?

certik:
I think you can use the X86 assembler, which just generates a binary, but in Debug mode it can also do a text format.

Alternatively, we just do a binary, and then do x86 -> text assembly.

Shaikh-Ubaid:
https://github.com/lcompilers/lpython/blob/fc08c06fbcbc60d1000b7da0f2da4f8a72f3c622/src/libasr/codegen/asr_to_wasm.cpp#L1935-L1949

In `wasm`, when printing strings, we have the length of the string and its location in memory onto the stack. In `x86` as well, when there is a call to print string, we would be having the length of the string and its memory location onto the `x86` stack. The concern here is that the memory location of the string is not available with the `wasm->x86` backend (during compile-time), but instead is available with `x86` (during runtime, since the location is already onto the stack). Thus, if we (somehow) create a label for each string, we are unable to reference the label for a particular string (the label and the string's memory might not be same or we might not be able to create a relation between them). It seems we might need to dynamically allocate strings in `x86`.

Shaikh-Ubaid:
>In wasm, when printing strings, we have the length of the string and its location in memory onto the stack.

This idea of printing strings using its length and memory location is more of our own concept/choice. It is possible that `WASI` might have different mechanism of printing string. At the moment, we do not have support for `WASI`.

Shaikh-Ubaid:
Prospective Roadmap:
- [ ] support printing `strings`
- [ ] support `if else` (thus support `assert`)
- [ ] support `stop` and `error stop` (we need `error stop` for `CI` tests)
- [ ] support `wasm_x86` at the `CI`
- [ ] support printing negative integers
[...]

PR #1262, opened 2022-11-03
CI: Support integration_tests for X86 and WASM_X86 backends

Shaikh-Ubaid:
This `PR` adds support for `integration_tests` for the `x86` and `wasm_x86` backends. There is a basic test case `integration_tests/print_03.py` (which just prints an integer `5`) (temporarily) added to test if the generated `x86` binary executes.

PR #1264, opened 2022-11-04
Implement if-else in wasm_to_x86 backend

Thirumalai-Shaktivel:
Related #1222

### DIFF
--- a/src/libasr/codegen/asr_to_wasm.cpp
+++ b/src/libasr/codegen/asr_to_wasm.cpp
@@ -2048,11 +2048,9 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
         for (size_t i = 0; i < x.n_body; i++) {
             this->visit_stmt(*x.m_body[i]);
         }
-        if (x.n_orelse) {
-            wasm::emit_b8(m_code_section, m_al, 0x05);  // starting of else
-            for (size_t i = 0; i < x.n_orelse; i++) {
-                this->visit_stmt(*x.m_orelse[i]);
-            }
+        wasm::emit_b8(m_code_section, m_al, 0x05);  // starting of else
+        for (size_t i = 0; i < x.n_orelse; i++) {
+            this->visit_stmt(*x.m_orelse[i]);
         }
         nesting_level--;
         wasm::emit_expr_end(m_code_section, m_al);  // emit if end
@@ -2119,6 +2117,7 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
         }
         wasm::emit_i32_const(m_code_section, m_al, 1);  // non-zero exit code
         exit();
+        wasm::emit_b8(m_code_section, m_al, 0x05);  // starting of else
         wasm::emit_expr_end(m_code_section, m_al);  // emit if end
     }
 };
--- a/src/libasr/codegen/wasm_to_x86.cpp
+++ b/src/libasr/codegen/wasm_to_x86.cpp
@@ -252,6 +252,9 @@ Result<int> wasm_to_x86(Vec<uint8_t> &wasm_bytes, Allocator &al,
                 .count();
     }
 
+    //! Helpful for debugging
+    // std::cout << x86_visitor.m_a.get_asm() << std::endl;
+
     if (time_report) {
         std::cout << "Codegen Time report:" << std::endl;
         std::cout << "Decode wasm: " << std::setw(5) << time_decode_wasm
--- a/src/libasr/codegen/wasm_to_x86.cpp
+++ b/src/libasr/codegen/wasm_to_x86.cpp
@@ -16,6 +16,7 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
    public:
     X86Assembler &m_a;
     uint32_t cur_func_idx;
+    std::vector<uint32_t> unique_id;
 
     X86Visitor(X86Assembler &m_a, Allocator &al,
                diag::Diagnostics &diagonostics, Vec<uint8_t> &code)
@@ -66,6 +67,27 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
             // push eax value onto stack
             m_a.asm_push_r32(X86Reg::eax);
         }
+    }
+
+    void visit_EmtpyBlockType() {}
+
+    void visit_If() {
+        unique_id.push_back(offset);
+        m_a.asm_pop_r32(X86Reg::eax);
+        m_a.asm_cmp_r32_imm8(LFortran::X86Reg::eax, 1);
+        m_a.asm_je_label(".then_" + std::to_string(unique_id.back()));
+        m_a.asm_jmp_label(".else_" + std::to_string(unique_id.back()));
+        m_a.add_label(".then_" + std::to_string(unique_id.back()));
+        {
+            decode_instructions();
+        }
+        m_a.add_label(".endif_" + std::to_string(unique_id.back()));
+        unique_id.pop_back();
+    }
+
+    void visit_Else() {
+        m_a.asm_jmp_label(".endif_" + std::to_string(unique_id.back()));
+        m_a.add_label(".else_" + std::to_string(unique_id.back()));
     }
 
     void visit_LocalGet(uint32_t localidx) {
--- a/integration_tests/if_01.py
+++ b/integration_tests/if_01.py
@@ -1,0 +1,28 @@
+def Test_if_01():
+    if True:
+        print(1)
+
+    if False:
+        print(0)
+
+    if 1 < 0:
+        print(0)
+    else:
+        print(1)
+
+    if 1 > 0:
+        print(1)
+    else:
+        print(0)
+
+    if 1 < 0:
+        print(1)
+    elif 1 > 0:
+        print(1)
+    else:
+        print(0)
+
+def Verify():
+    Test_if_01()
+
+Verify()
--- a/integration_tests/if_01.py
+++ b/integration_tests/if_01.py
@@ -22,7 +22,38 @@ def Test_if_01():
     else:
         print(0)
 
+def Test_if_02():
+    if True:
+        print(1)
+        if True:
+            print(2)
+            if False:
+                print(3)
+            elif True:
+                print(4)
+            else:
+                print(5)
+        else:
+            print(6)
+
+    if True:
+        print(7)
+        if False:
+            print(8)
+            if False:
+                print(9)
+            else:
+                print(10)
+        else:
+            if True:
+                print(11)
+            else:
+                print(12)
+            print(13)
+    print(14)
+
 def Verify():
     Test_if_01()
+    Test_if_02()
 
 Verify()
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -775,7 +775,7 @@ int compile_to_binary_x86(
 
     // ASR -> X86
     auto asr_to_x86_start = std::chrono::high_resolution_clock::now();
-    LFortran::Result<int> r3 = LFortran::asr_to_x86(*asr, al, outfile, time_report);
+    LFortran::Result<int> r3 = LFortran::asr_to_x86(*asr, al, outfile, time_report, diagnostics);
     auto asr_to_x86_end = std::chrono::high_resolution_clock::now();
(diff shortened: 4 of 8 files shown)


src/libasr/codegen/wasm_to_x86.cpp
```diff
@@ -68,6 +69,51 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
...
+
+    void visit_EmtpyBlockType() {}
+
+    bool is_else_stmt_available() {
+        auto offset_ = offset;
+        uint8_t cur_byte = wasm::read_b8(code, offset_);
+        while (cur_byte != 0x0B) {
+            if (cur_byte == 0x05) {
+                cur_byte = wasm::read_b8(code, offset_);
+                if (cur_byte == 0x0B || cur_byte == 0x41) {
```
Thirumalai-Shaktivel, line 84: I have a doubt:
Does `wasm` always have an else block (even though else is empty) in the if statement?
image
@Shaikh-Ubaid 
If so how Assert is handled in asr_to_wasm?

Thirumalai-Shaktivel: I asked this because, the else block is not empty in the `visit_assert()`, it has `0x41`

Shaikh-Ubaid: >Does wasm always have an else block (even though else is empty) in the if statement?

Currently, if there is no `else` block, then in the `wat` generated, we would just have `if` `<some instructions>` `end`.

Shaikh-Ubaid: >I asked this because, the else block is not empty in the visit_assert(), it has 0x41

Do you mean `visit_Assert()` in `ASR->WASM` backend?

Thirumalai-Shaktivel: yes

src/libasr/codegen/wasm_to_x86.cpp
```diff
@@ -68,6 +69,51 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
...
+                    return false;
+                } else {
+                    return true;
+                }
+            }
+            cur_byte = wasm::read_b8(code, offset_);
+        }
```
Thirumalai-Shaktivel, line 91: But this logic, detects the else block, even for the following syntax!!
```
if True:
	pass
```

Thirumalai-Shaktivel: That is the reason, I check the next byte to be 0x0B (which checks the end of the if stmt).

Shaikh-Ubaid: At present, if there is no `else` block, then `visit_Else()` should not be visited. I guess, we might be missing something.

Shaikh-Ubaid:
I looked into the `PR`. It seems, it would be better if `wasm` backend adds an `else` block always (it would be empty when it is not needed).

If an `else` is always present, then there would not be any need of checking if any `else` is present (since, it would be always present (although empty when not needed)).

Shaikh-Ubaid:
I will rebase this on top of https://github.com/lcompilers/lpython/pull/1262 (so that we can print some value inside the `if-else` ladder), make the necessary updates to the `wasm` backend and push to this `PR`.

Thirumalai-Shaktivel:
This looks great! Thanks for updating.

Shaikh-Ubaid:
>I will rebase this on top of https://github.com/lcompilers/lpython/pull/1262

I rebased on top of latest `main`. Rebasing on top of https://github.com/lcompilers/lpython/pull/1262 is including many commits of that `PR` (it also makes this PR dependent on it).

Shaikh-Ubaid (approved):

Thirumalai-Shaktivel:
Let's implement visit_Assert() in this PR itself.
Assertion in wasm_x86, behaves differently!
```py
def Main0():
    assert 1 == 1

Main0()
```
Output:
```console
$ lpython examples/expr2.py --backend wasm_x86 && ./a.out
AssertionError
```
```py
def Main0():
    assert 0 == 1

Main0()
```
```console
$ lpython examples/expr2.py --backend wasm_x86 && ./a.out
```

Thirumalai-Shaktivel:
@Shaikh-Ubaid What is EmptyBlockType? How is Assert handling in `WASM` (it's seems wasm generates correct output)?

src/libasr/codegen/asr_to_wasm.cpp
```diff
@@ -2105,9 +2105,9 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
 
     void visit_Assert(const ASR::Assert_t &x) {
         this->visit_expr(*x.m_test);
-        wasm::emit_i32_eqz(m_code_section, m_al);
         wasm::emit_b8(m_code_section, m_al, 0x04);  // emit if start
         wasm::emit_b8(m_code_section, m_al, 0x40);  // empty block type
+        wasm::emit_b8(m_code_section, m_al, 0x05);  // starting of else
         if (x.m_msg) {
             std::string msg =
                 ASR::down_cast<ASR::StringConstant_t>(x.m_msg)->m_s;
```
Shaikh-Ubaid, line 2113: The previous version was more correct, right? The previous version would say -> If `Assert` condition failed, raise an error (which is what `assert` does, I think).

src/libasr/codegen/asr_to_wasm.cpp
```diff
@@ -2117,7 +2117,6 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
         }
         wasm::emit_i32_const(m_code_section, m_al, 1);  // non-zero exit code
         exit();
-        wasm::emit_b8(m_code_section, m_al, 0x05);  // starting of else
         wasm::emit_expr_end(m_code_section, m_al);  // emit if end
     }
 };
```
Shaikh-Ubaid, line 2122: The `else` portion was empty in the previous version.

Shaikh-Ubaid (changes requested):
The right way to support `Assert` would instead be to implement the `visit_I32Eqz` in the `WASM->X86` backend.

```cpp
void visit_I32Eqz() {
        m_a.asm_pop_r32(X86Reg::eax);
        m_a.asm_mov_r32_imm32(X86Reg::ebx, 0U);
        m_a.asm_cmp_r32_r32(X86Reg::eax, X86Reg::ebx);
        m_a.asm_push_r32(X86Reg::eax);
    }
```

certik:
Thanks @Thirumalai-Shaktivel for working on this, very helpful. Thanks @Shaikh-Ubaid for the review.

src/libasr/codegen/asr_to_wasm.cpp
Thirumalai-Shaktivel: The main motive for these changes was that `visit_I32Eqz()` in the `wasm_to_x86` is not working correctly.
Here, the changes implies:
```py
If True:
	// Do nothing
else:
	print("AssertionError")
```
which is similar to `assert False`

Thirumalai-Shaktivel:
> The right way to support `Assert` would instead be to implement the `visit_I32Eqz` in the `WASM->X86` backend.
> 
> ```c++
> void visit_I32Eqz() {
>         m_a.asm_pop_r32(X86Reg::eax);
>         m_a.asm_mov_r32_imm32(X86Reg::ebx, 0U);
>         m_a.asm_cmp_r32_r32(X86Reg::eax, X86Reg::ebx);
>         m_a.asm_push_r32(X86Reg::eax);
>     }
> ```

This seems not working in wasm_x86 backend.

src/libasr/codegen/asr_to_wasm.cpp
Thirumalai-Shaktivel: The same approach as we do in the LLVM backend: https://github.com/lcompilers/lpython/blob/d4a6d5360ca520d9a25d658e6660cb39a2eda4e2/src/libasr/codegen/asr_to_llvm.cpp#L4845-L4862

Thirumalai-Shaktivel:
What does I32Eqz do? isn't it a NOT operator?

Shaikh-Ubaid:
>What does I32Eqz do? isn't it a [NOT](https://github.com/sunfishcode/wasm-reference-manual/blob/master/WebAssembly.md#integer-equal-to-zero) operator?

It checks if the stack top is equal to zero. Yes, it works similar to a not operator.

>The same approach as we do in the LLVM backend: https://github.com/lcompilers/lpython/blob/d4a6d5360ca520d9a25d658e6660cb39a2eda4e2/src/libasr/codegen/asr_to_llvm.cpp#L4845-L4862

If it is similar to the `LLVM` backend, we can follow it.



```
void visit_I32Eqz() {
        m_a.asm_pop_r32(X86Reg::eax);
        m_a.asm_mov_r32_imm32(X86Reg::ebx, 0U);
        m_a.asm_cmp_r32_r32(X86Reg::eax, X86Reg::ebx);
        m_a.asm_push_r32(X86Reg::eax);
}.
```
>This seems not working in the wasm_x86 backend

It should (hopefully) work. I tested it locally on my system before sharing here. If we are following the `LLVM` approach, then we might not need it (I guess) and we can add it later when needed.

Thirumalai-Shaktivel:
> It should (hopefully) work. I tested it locally on my system before sharing here.

Were you able to get "AssertionError" for `assert False`?
This is weird, it didn't work for me!

Thirumalai-Shaktivel:
```py
def Test_assert_01():
    if True:
        assert True

    assert 1 != 1, "One is equal to one."

def Verify():
    Test_assert_01()

Verify()
```
```console
$ lpython examples/expr2.py --backend wasm_x86 && ./a.out
Call to imported function with index 4 not yet supported
Call to imported function with index 4 not yet supported
```

Thirumalai-Shaktivel:
Sorry my bad, print_str doesn't work yet in this PR! I will rebase that PR and report back

Shaikh-Ubaid:
>Sorry my bad, print_str doesn't work yet in this PR! I will rebase that PR and report back.

Let's wait for the `PRs` to be merged. It seems many `PRs` are interdependent. Let's work on `Assertion` in another `PR` once all the necessary `PRs` are merged.

Thirumalai-Shaktivel:
Yup! Now, I feel the same.
I will another PR for `Assert` once the `print_str` is merged

Shaikh-Ubaid (approved):

Thirumalai-Shaktivel:
Thank you for your approval and guidance!

src/libasr/codegen/wasm_to_x86.cpp
```diff
@@ -16,6 +16,7 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
    public:
     X86Assembler &m_a;
     uint32_t cur_func_idx;
+    std::vector<uint32_t> unique_id;
```
certik, line 19: Later:
```suggestion
    uint32_t unique_id;
```

src/libasr/codegen/wasm_to_x86.cpp
```diff
@@ -68,6 +69,27 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
         }
     }
 
+    void visit_EmtpyBlockType() {}
+
+    void visit_If() {
+        unique_id.push_back(offset);
```
certik, line 75:
```suggestion
        unique_id++;
        uint32_t label = unique_id;
```

src/libasr/codegen/wasm_to_x86.cpp
```diff
@@ -68,6 +69,27 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
...
+        m_a.asm_pop_r32(X86Reg::eax);
+        m_a.asm_cmp_r32_imm8(LFortran::X86Reg::eax, 1);
+        m_a.asm_je_label(".then_" + std::to_string(unique_id.back()));
+        m_a.asm_jmp_label(".else_" + std::to_string(unique_id.back()));
+        m_a.add_label(".then_" + std::to_string(unique_id.back()));
```
certik, line 80: For later (but probably quite soon): we should not be using strings for labels, but just unique integers. The unique integer can just be a counter in this visitor, just an integer, see above.

certik: (we might need to have a stack of unique ids...)

certik (approved):
Looks good to me, thanks!

PR #1265, opened 2022-11-04
WASM_X86: Initial support for print_str

Shaikh-Ubaid:
It seems to work, but the approach seems hackish.

---

***Example:***
```python
def Main0():
    x: i32
    x = (2+3)*5
    print(x)
    print("Hi")
    print("Hello")
    print((5-2) * 7)
    print("Bye")

Main0()
```
***Output:***
```bash
(lp) ubaid@ubaid-Lenovo-ideapad-330-15ARR:~/OpenSource/lpython$ lpython examples/expr2.py --backend x86
(lp) ubaid@ubaid-Lenovo-ideapad-330-15ARR:~/OpenSource/lpython$ ./a.out 
25
Hi
Hello
21
Bye
(lp) ubaid@ubaid-Lenovo-ideapad-330-15ARR:~/OpenSource/lpython$ lpython examples/expr2.py --backend wasm_x86
(lp) ubaid@ubaid-Lenovo-ideapad-330-15ARR:~/OpenSource/lpython$ ./a.out 
25
Hi
Hello
21
Bye
```

***Note:*** Currently only single parameter is supported for `print` by both `x86` and `wasm_x86` backends.

### DIFF
--- a/src/libasr/codegen/wasm_to_x86.cpp
+++ b/src/libasr/codegen/wasm_to_x86.cpp
@@ -17,6 +17,8 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
     X86Assembler &m_a;
     uint32_t cur_func_idx;
     std::vector<std::string> unique_id;
+    int32_t last_vis_i32_const, last_last_vis_i32_const;
+    std::unordered_map<int32_t, std::string> loc_to_str;
 
     X86Visitor(X86Assembler &m_a, Allocator &al,
                diag::Diagnostics &diagonostics, Vec<uint8_t> &code)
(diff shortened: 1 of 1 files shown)


czgdp1807:
Why is it `Main0` and not, `main0`?

Thirumalai-Shaktivel:
In the `x86` backend, we get the following:
```console
$ lpython examples/expr2.py  --backend x86 && ./a.out 
code generation error: Subroutine code not generated for 'main0'


```

czgdp1807:
Please generate functions after performing topological sort on them. Here's how it is done in C backend,

https://github.com/lcompilers/lpython/blob/d4a6d5360ca520d9a25d658e6660cb39a2eda4e2/src/libasr/codegen/asr_to_c.cpp#L690-L692

https://github.com/lcompilers/lpython/blob/d4a6d5360ca520d9a25d658e6660cb39a2eda4e2/src/libasr/codegen/asr_to_c.cpp#L727-L737

Shaikh-Ubaid:
>Please generate functions after performing topological sort on them. Here's how it is done in C backend,

Thank you for sharing. This would be very helpful. I think we can also use it in the `wasm` backend. I will submit a `PR` with these changes. Thank you again.

Shaikh-Ubaid:
>Why is it Main0 and not, main0?

There were functions like `_lpython_main_program()` and `_lcompilers_main_program()` which get defined before the `main0()` (I think that happens because the ascii value of `m` (109) is greater the `_` (95)) If we rename `main0()` to `Main0()` then it gets defined before the `_lpython_main_program()` and `_lcompilers_main_program()`, since I guess/think ascii of `M` (77) is less than `_` (95).

certik (approved):

certik (approved):
I think this is fine for now.

Shaikh-Ubaid:
I rebased this on top of latest main. I am adding it to auto-merge as it seems to be approved.

PR #1271, opened 2022-11-05
Disable implicit casting by default and implement intrinsic functions for casting

czgdp1807:
Overrides https://github.com/lcompilers/lpython/pull/1194

PR #1275, opened 2022-11-06
[ASR]: Fix character types

### LINKED ISSUE #1267: Cast IntegerToCharacter seems to use incorrect type
Here is an example:
```python
def main():
    x: i32
    x = 53
    s: str
    s = str(x)
    print(s)

main()
```
Which produces the following ASR code:
```
...
                            s:
                                (Variable 
                                    2 
                                    s 
                                    Local 
                                    () 
                                    () 
                                    Default 
                                    (Character 1 -2 () []) 
                                    Source 
                                    Public 
                                    Required 
                                    .false.
                                ), 
...

                    (= 
                        (Var 2 s) 
                        (Cast 
                            (Var 2 x) 
                            IntegerToCharacter 
                            (Character 1 1 () []) 
                            ()
                        ) 
                        ()
                    )
```
I think the `(Character 1 1 () [])` type in `Cast` should rather be `(Character 1 -2 () [])` in this case.

Smit-create:
Fixes #1267


Smit-create:
@certik @czgdp1807 This is ready for review!

czgdp1807 (approved):
This is okay but I would suggest to wait for https://github.com/lcompilers/lpython/pull/1271 to be merged.

PR #1276, opened 2022-11-08
Implement Integer compare in wasm to x86 backend

Thirumalai-Shaktivel:
Related: #1222


integration_tests/if_02.py
```diff
@@ -0,0 +1,40 @@
...
+    #     print(1)
+
+    # y = 5
+    # if y <= 10:
+    #     print(1)
+    # if y >= 5:
+    #     print(1)
+    # if y >= 1:
+    #     print(1)
+    print(y)
```
Thirumalai-Shaktivel, line 35: This prints as:
```console
$ lpython integration_tests/if_02.py --backend wasm_x86 && ./a.out
00
```
I don't know, what's causing this issue!

Thirumalai-Shaktivel: I think every other Compare operator works correctly, except `NotEq`

Thirumalai-Shaktivel:
This PR is dependable on #1264

src/libasr/codegen/wasm_to_x86.cpp
```diff
@@ -154,6 +154,41 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
         m_a.asm_push_r32(X86Reg::eax);
     }
 
+    void handle_I32Compare(const std::string &compare_op) {
+        unique_id.push_back(std::to_string(offset));
```
Shaikh-Ubaid, line 158: @Thirumalai-Shaktivel, please, could you possibly share why we need to store the `offset` in `unique_id`? Are we using it later outside the `handleI32Compare()` function?

If it is not used later, we need/should not store it, I think.

Thirumalai-Shaktivel: Yup, I forgot to pop the uniqu_id!
Now, If I think of ti, we have options:
1. 
```diff
diff --git a/src/libasr/codegen/wasm_to_x86.cpp b/src/libasr/codegen/wasm_to_x86.cpp
index cf968d5c4..12ea3b156 100644
--- a/src/libasr/codegen/wasm_to_x86.cpp
+++ b/src/libasr/codegen/wasm_to_x86.cpp
@@ -199,30 +199,29 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
     }
 
     void handle_I32Compare(const std::string &compare_op) {
-        unique_id.push_back(std::to_string(offset));
         m_a.asm_pop_r32(X86Reg::ebx);
         m_a.asm_pop_r32(X86Reg::eax);
         m_a.asm_cmp_r32_r32(X86Reg::eax, X86Reg::ebx);
         if (compare_op == "Eq") {
-            m_a.asm_je_label(".compare_1" + unique_id.back());
+            m_a.asm_je_label(".compare_1" + std::to_string(offset));
...
```
2.
```diff
diff --git a/src/libasr/codegen/wasm_to_x86.cpp b/src/libasr/codegen/wasm_to_x86.cpp
index cf968d5c4..2e23d7230 100644
--- a/src/libasr/codegen/wasm_to_x86.cpp
+++ b/src/libasr/codegen/wasm_to_x86.cpp
@@ -224,6 +224,7 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
         m_a.asm_mov_r32_imm32(X86Reg::eax, 1);
         m_a.add_label(".compare.end_" + unique_id.back());
         m_a.asm_push_r32(X86Reg::eax);
+        unique_id.pop_back();
     }
 
     void visit_I32Eq() { handle_I32Compare("Eq"); }
```
which do you prefer the most?

Thirumalai-Shaktivel: As you said, yup, it is not used later on. So, Instead of storing the unique_id, we just use `std::to_string(offset)`. I think option 1 is good to go.

src/libasr/codegen/wasm_to_x86.cpp
```diff
@@ -154,6 +154,41 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
...
+            m_a.asm_jne_label(".compare_1" + unique_id.back());
+        } else {
+            throw CodeGenError("Comparison operator not implemented");
+        }
+        m_a.asm_mov_r32_imm32(X86Reg::eax, 0);
+        m_a.asm_jmp_label(".compare.end_" + unique_id.back());
+        m_a.add_label(".compare_1" + unique_id.back());
+        m_a.asm_mov_r32_imm32(X86Reg::eax, 1);
+        m_a.add_label(".compare.end_" + unique_id.back());
+        m_a.asm_push_r32(X86Reg::eax);
```
Thirumalai-Shaktivel, line 182: In the last discussion, you said we can push value directly into the stack.
Here why not we just push the result value into the stack instead of moving it into the eax?
Also, I have another doubt:
How do we access the pushed value in the stack?

Thirumalai-Shaktivel: > https://github.com/lcompilers/lpython/issues/1222#issuecomment-1305148759
How to perform that?

src/libasr/codegen/wasm_to_x86.cpp
Thirumalai-Shaktivel: Also, the reason why I used the unique_id is because: I thought `offset` value would be incremented inside `handle_compareI32()` function. I checked and it seems the offset is not modified. So, we can use `std::to_string(offset)`.
I will make the able change with the loop stmt.
Thanks for noticing it.

src/libasr/codegen/wasm_to_x86.cpp
Thirumalai-Shaktivel: Pushing the value directly into the stack works:
```c++
m_a.asm_push_imm8(0);
m_a.asm_jmp_label(".compare.end_" + std::to_string(offset));
m_a.add_label(".compare_1" + std::to_string(offset));
m_a.asm_push_imm8(1);
m_a.add_label(".compare.end_" + std::to_string(offset));
```

Shaikh-Ubaid: >How do we access the pushed value in the stack?

It can be accessed during run-time only.

>How would we use the value on top of stack?

Other instructions can use it during run-time time. They perform some operation(s) on it and store the result back onto stack.

src/libasr/codegen/wasm_to_x86.cpp
Shaikh-Ubaid: >So, Instead of storing the unique_id, we can just use std::to_string(offset)

Yup, we can just use `offset` as a label.

src/libasr/codegen/wasm_to_x86.cpp
Shaikh-Ubaid: >Pushing the value directly into the stack works:

If it works, we can use this way. (I think it could be better as it could use (slightly) lesser instructions)

src/libasr/codegen/wasm_to_x86.cpp
Shaikh-Ubaid: I incorporated this change into this commit [WASM_X86: Use offset directly as label](https://github.com/lcompilers/lpython/pull/1285/commits/068563a24a95311c321a99b10fbf5f95f2a76a31)

src/libasr/codegen/wasm_to_x86.cpp
Shaikh-Ubaid: I added this change/improvement into this commit
[Refactor: WASM_X86: Directly push value onto stack](https://github.com/lcompilers/lpython/pull/1285/commits/15779646094f2c40110bfe06cd1dfa0041985e30).

Thirumalai-Shaktivel:
Thank you for updating the changes.

Shaikh-Ubaid:
Thank you for your suggestions and guidance on them. They were really helpful.

PR #1279, opened 2022-11-09
ASR: Constant string attributes refactoring

### LINKED ISSUE #1278: String attributes need refactoring
https://github.com/lcompilers/lpython/blob/d4a6d5360ca520d9a25d658e6660cb39a2eda4e2/src/lpython/semantics/python_comptime_eval.h#L69-L76

These string methods are never used during compile-time evaluation as they are represented by ConstantStr in ASR and handled differently in `python_ast_to_asr.cpp` and so this makes the file unnecessarily lengthy as well as the `python_ast_to_asr`. Moreover, the following eval functions are also implemented incorrectly:
https://github.com/lcompilers/lpython/blob/d4a6d5360ca520d9a25d658e6660cb39a2eda4e2/src/lpython/semantics/python_comptime_eval.h#L800-L810

We should remove all the value functions from this map and replace them with `not_implemented`.

Smit-create:
Fixes #1278

### DIFF
--- a/integration_tests/test_str_attributes.py
+++ b/integration_tests/test_str_attributes.py
@@ -9,11 +9,14 @@ def capitalize():
     assert "empty string" .capitalize() == "Empty string"
     assert "".capitalize() == ""
 
+
 def lower():
     s: str
     s = "AaaaAABBbbbbBB!@12223BN"
     assert s.lower() == "aaaaaabbbbbbbb!@12223bn"
     assert "DDd12Vv" .lower() == "ddd12vv"
+    assert "".lower() == ""
+
 
 def strip():
     s: str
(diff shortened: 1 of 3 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -5258,6 +5258,194 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                            if (j != 0) {
+                                j = lps[j - 1];
+                            } else {
+                                i = i + 1;
+                            }
+                        }
+                    }
+                }
+                tmp = ASR::make_IntegerConstant_t(al, loc, res,
+                    ASRUtils::TYPE(ASR::make_Integer_t(al,loc, 4, nullptr, 0)));
```
czgdp1807, line 5333: Please put this into a function and call it here.

czgdp1807 (changes requested):

integration_tests/test_str_attributes.py
```diff
@@ -67,10 +76,13 @@ def startswith():
...
-swapcase()
-find()
-startswith()
+def check():
+    capitalize()
+    lower()
+    strip()
+    swapcase()
+    find()
+    startswith()
```
namannimmo10, line 88: 🙂
```suggestion
    startswith()

check()
```

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: Done

integration_tests/test_str_attributes.py
Smit-create: Thanks

czgdp1807 (approved):

Issue #1280, opened 2022-11-09
Import nested modules


certik:
It is a high priority for us to be able to get the following imports working:
```python
from A.B.C.D import x
from A.B.C.E import y
```

certik:
Now when #1281 is in, what remains to be done?

I'll test it soon with our production code and see if anything remains. One thing that we need is finding these nested modules using the `-I` flag. I don't know if this works already or not yet.

Thirumalai-Shaktivel:
The following doesn't work
```console
$ pwd
/home/thirumalai
$ lpython -I /home/thirumalai/Open_Source/lpython integration_tests/test_import_01.py
$
```
I will try to debug and report back.

certik:
The use case we have is `lpython a.py` and in `a.py` we do `from A.B.C.D import x`, where the package `A` lives somewhere completely else and have to specify the path using `-I`.

Thirumalai-Shaktivel:
The following work:
```console
(lf) ➜  $ pwd
/home/thirumalai
(lf) ➜  $ cat a.py                                
import test_import
from test_import import print_a

print(print_a())
print(test_import.print_b())
print(test_import.print_c())
(lf) ➜  $ lpython -I /home/thirumalai/Open_Source/lpython/integration_tests a.py
A
B
C
```
The nested modules has the following error:
```console
$ lpython -I /home/thirumalai/Open_Source/lpython/integration_tests a.py             
semantic error: Could not find the module '__init__.py'. If an import path is available, please use the `-I` option to specify it
 --> a.py:1:1
  |
1 | from test_import.import_module_01 import multiply, e
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).
```
I will debug and report back

certik:
@Thirumalai-Shaktivel this now works, doesn't it?

Thirumalai-Shaktivel:
Yup, It works. I just added the comment to keep track of all the import issues.

certik:
I am going to close this one then.

PR #1281, opened 2022-11-09
[ASR] Implement nested modules import

Thirumalai-Shaktivel:
Related: #1280


integration_tests/test_import_02.py
```diff
@@ -0,0 +1,5 @@
+from test_import.import_module_01 import multiply
+from test_import.import_module_02 import add
+
+print(add(10, 20))
+print(multiply(10, 20))
```
Thirumalai-Shaktivel, line 5: This PR is a good start for nested modules!
Here is an example that works

certik (approved):
This looks good to me. Is it really this simple? I guess it is since we are importing symbols directly so far.

certik:
Thank you so much for fixing this! I really appreciate it.

Thirumalai-Shaktivel:
There might be a lot of bugs, like not using the correct to import, importing all the symbol (even though we specfied only some symbols to import). 
But, this is good start. We can always fix it on the go. Please report any bug you face or provide any suggestion to test.
Thanks for the approval

Issue #1282, opened 2022-11-09
Change the resulting type of `/`


certik:
This is the current behavior (in Python source code):
```
i32 / i32 -> f64
i64 / i64 -> f64
f32 / f32 -> f64
f64 / f64 -> f64
```
I propose we change it to the following:
```
i32 / i32 -> f64
i64 / i64 -> f64
f32 / f32 -> f32
f64 / f64 -> f64
```
So if an integer is divided, it always becomes f64. But for floating point it would keep the kind. This is in line with other operators like `*` (the only difference is that for integer arguments `/` must return a float):
```
i32 * i32 -> i32
i64 * i64 -> i64
f32 * f32 -> f32
f64 * f64 -> f64
```

I guess one could also argue that `i32 / i32 -> f64` must return `i32 / i32 -> f32`, but the way I think about it is that integer division should return the default floating point type, which in Python is `f64`.

Summary: I am proposing we change `f32 / f32 -> f64` to `f32 / f32 -> f32`, but keep all other behavior exactly the same.

certik:
@czgdp1807 let me know what you think. If you agree, then you can go ahead and do it, I think you know how to do it quickly.

PR #1284, opened 2022-11-10
Fixing return type of `/` operator

czgdp1807:
Now we follow the rules below for `/` operator,

```
i32 / i32 -> f64
i64 / i64 -> f64
f32 / f32 -> f32
f64 / f64 -> f64
c32 / c32 -> c32
c64 / c64 -> c64
```

A bit surprising that `c32` and `c64` were never tried before with `/` operator.

Closes https://github.com/lcompilers/lpython/issues/1282


certik (approved):
Awesome, thanks for this.

PR #1286, opened 2022-11-10
Support `-I` option to specify import path for the nested modules.

Thirumalai-Shaktivel:


certik (approved):
We'll have to later figure out how to test it, but this is good for now. Thanks!

PR #1290, opened 2022-11-12
Implement Loop statement

Thirumalai-Shaktivel:


src/libasr/codegen/wasm_to_x86.cpp
```diff
@@ -129,8 +129,38 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
 
     void visit_EmtpyBlockType() {}
 
+    void visit_Br(uint32_t /*label_index*/) {
+        // Branch is used to jump to the `loop.head`.
+        m_a.asm_jmp_label(".loop.head_" + unique_id.rbegin()[1]);
+    }
+
```
Thirumalai-Shaktivel, line 136: What is the use of `branch`?
My doubt is:
1. the label_index is 1 if the visitor is either `break` or `end` of the `if.block`
2. the label_index is 2 if the visitor is `continue`
if.block and continue must have the same behavior, right?

Thirumalai-Shaktivel: I see, the label_index is not constant, it changes based on the statements inside the loop.

certik (approved):
This is great. @Shaikh-Ubaid you should still review this.

Thirumalai-Shaktivel:
Thanks for reviewing and merging this PR

Shaikh-Ubaid:
>@Shaikh-Ubaid you should still review this.

Yes, I will review this in the evening today.

src/libasr/codegen/wasm_to_x86.cpp
Shaikh-Ubaid: >What is the use of branch?

It is like a `jump` instruction. Everytime we wish to continue a `loop` in `wasm`, we need to `branch` (only if the `loop` condition is `true`) to the `loop`.

https://samrat.me/posts/2020-03-29-webassembly-control-instr-examples/ offers a good explanation.

Shaikh-Ubaid: >I see, the label_index is not constant, it changes based on the statements inside the loop.

Yes, the `label_index` is used as the target of the `branch`.

https://github.com/lcompilers/lpython/blob/ce4612d5d2a7207ab8147bb102e39bbbe5e67e2c/src/libasr/codegen/asr_to_wasm.cpp#L2025-L2029

As shared in the above comment, the label index counts from the inner most `block` (A `block` could be a `loop` or an `if-else` construct). In our case, a `label_index` of `0` points to the `if-else` condition (which is the loop condition) and a `label_index` of `1` points to the actual `loop`.

Also, on branching to a `loop`, the control passes to the beginning of the `loop`, where as on branching to an `if` block, the control passes to the end of it.

src/libasr/codegen/wasm_to_x86.cpp
```diff
@@ -129,23 +130,59 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
...
+        `loop.branch` are enclosed within the `if.block`. If the condition
+        fails, the loop is exited through `else.block`.
+        .head
+            .If
+                # Statements
+                .Br
+            .Else
+            .endIf
+        .end
+        */
```
Shaikh-Ubaid, line 157: Great note, Thirumalai !. It explains the `loop` structure very beautifully. Thank you so much for it.

src/libasr/codegen/wasm_to_x86.cpp
Shaikh-Ubaid: >// Unlike with other index spaces, indexing of labels is relative by 
 // nesting depth, that is, label 0 refers to the innermost structured 
 // control instruction enclosing the referring branch instruction, while 
 // increasing indices refer to those farther out. 

The following example should (hopefully) explain the above more clearly.


```
// Code in some dummy language, (might) not be exactly in `WebAssembly Text Format`

(loop // label index = 3
    (if // label index = 2
        (if // label index = 1
            (if // label index = 0
                br some_label_index
            )
        )
    )
) 
```
Depending on where we would like to branch, we can have appropriate value for `some_label_index`.

src/libasr/codegen/wasm_to_x86.cpp
```diff
@@ -129,23 +130,59 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
...
 
+    void visit_Br(uint32_t label_index) {
+        // Branch is used to jump to the `loop.head` or `loop.end`.
+        if (if_unique_id.size() - loop_unique_id.size() == label_index - 1) {
+            // cycle/continue or loop.end
+            m_a.asm_jmp_label(".loop.head_" + loop_unique_id.back());
+        } else {
+            // exit/break
+            m_a.asm_jmp_label(".loop.end_" + loop_unique_id.back());
+        }
```
Shaikh-Ubaid, line 141: This `if-else` condition inside the `visit_Br()` is not clear to me. @Thirumalai-Shaktivel, please, could you possibly share how this `if-else` works and chooses whether to iterate the loop or end it?

Shaikh-Ubaid:
I wonder if we need the changes (replacing `print` statements with `assert`) in the file `integration_tests/if_01.py`. Previously, this `if` test case just had `print` statements. This was useful to us in the way that if we add a new backend (in the future) and if we add support for `if-else` in this new backend (with no support for `exit code` yet), we could easily enable this `integration_tests/if_01.py`. Now, with `assert` being used in this test case we would either need to write a new `simple` `if-else` test case or also add support for `assert` before enabling `if-else` test case.

It seems we write new backends less frequently, hence the above seems to be more of a discussion topic/point. I would like to request your views/suggestions on it.

Shaikh-Ubaid (commented):
This is an Amazing PR, Thirumalai! Great Work! Thank you so much for this.

src/libasr/codegen/wasm_to_x86.cpp
Thirumalai-Shaktivel: I got it, great explanation. Thanks

src/libasr/codegen/wasm_to_x86.cpp
Thirumalai-Shaktivel: Consider this example:
```py
while i < 10:
    j = 0
    while j < 10:
        k += 1
        if i == 0:
            if j == 3:
                continue
            else:
                j += 1
        j += 1
    i += 1
assert k == 95
```
Output printed using:
```diff
diff --git a/src/libasr/codegen/wasm_to_x86.cpp b/src/libasr/codegen/wasm_to_x86.cpp
index 9db9d2b4b..48f947076 100644
--- a/src/libasr/codegen/wasm_to_x86.cpp
+++ b/src/libasr/codegen/wasm_to_x86.cpp
@@ -131,6 +131,9 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
     void visit_EmtpyBlockType() {}
 
     void visit_Br(uint32_t label_index) {
+std::cout << "label_index: " << label_index << "\n";
+std::cout << "Loop_size: " << loop_unique_id.size() << "\n";
+std::cout << "If_size: " << if_unique_id.size() << "\n";
         // Branch is used to jump to the `loop.head` or `loop.end`.
```
```py
$ lpython examples/expr2.py --backend wasm_x86 && ./expr2.out
label_index: 3
Loop_size: 2
If_size: 4
label_index: 1
Loop_size: 2
If_size: 2
label_index: 1
Loop_size: 1
If_size: 1
```
We have 2 loop's and 2 if's. But in the backend, we get 2 loop's and 4 if's, right?
The first br comes from the continue, if we substract 4 if's with 2 loop's: we get the actual 2 if's. so the br label is 2+1. and So on...
Any doubts, please ask

Thirumalai-Shaktivel:
Exactly, I experiened the same issue, while implementing this PR. The problem was that in the integration_tests there is a file named `loop_01.py`. It had both integer and string to iterate in the loop. As wasm_x86 doesn't support StringLen, I was not able to use it. I created a new file and added all the tests there.
Also, why don't we have more simpler tests? All the if, loop, ... are being testing more robustly in LFortran.
Now, there are two options: either we add more tests in LPython or submit the PR for new backend's to LFortran.

src/libasr/codegen/wasm_to_x86.cpp
Thirumalai-Shaktivel: > how this if-else works and chooses whether to iterate the loop or end it?

I didn't notice, how the loop is being exited. I will look into it soon.

Thirumalai-Shaktivel: I see, the Loop is exited through the else block:
```asm
.loop.head_297:
    mov eax, [ebp-4]
    push eax
    push 0x0000000a
    pop ebx
    pop eax
    cmp eax, ebx
    jl .compare_1303
    push 0x00
    jmp .compare.end_303
.compare_1303:
    push 0x01
.compare.end_303:
    pop eax
    cmp eax, 0x01
    je .then_304
    jmp .else_304
.then_304:
    mov eax, [ebp-4]
    push eax
    push 0x00000001
    pop ebx
    pop eax
    add eax, ebx
    push eax
    pop eax
    mov [ebp-4], eax
    jmp .loop.head_297
    jmp .endif_304
.else_304:
.endif_304:
.loop.end_297:
    mov esp, ebp
    pop ebp
    ret
```

Thirumalai-Shaktivel: I have a doubt, here the jump and label `.endif_304` is useless? is it ok? @Shaikh-Ubaid

Thirumalai-Shaktivel: @Shaikh-Ubaid, where you able to understand or get the point?

Shaikh-Ubaid: >I have a doubt, here the jump and label .endif_304 is useless? is it ok? @Shaikh-Ubaid

I think for the moment that should be fine.

>@Shaikh-Ubaid, were you able to understand or get the point?

Yup, thank you so much for the explanation, Thirumalai.

Thirumalai-Shaktivel: > It seems this condition (if_unique_id.size() - loop_unique_id.size() == label_index - 1) might not be perfect to determine to continue or end the loop. 

+1, I agree with this.
Also, I didn't test loop within if statement.
There you go, you found the actual bug in the loop statement.
Thank you for testing this. I will create this a new issue. Let's work on this and fix it.

PR #1293, opened 2022-11-13
[WIP] ASR: Add `deterministic` and `side-effect-free` flags to Function

### LINKED ISSUE #1266: ASR: Add "deterministic" and "side-effect-free" flags to Function
Roughly like this:
```diff
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -87,7 +87,8 @@ symbol
     | Function(symbol_table symtab, identifier name, expr* args, stmt* body,
         expr? return_var, abi abi, access access, deftype deftype,
         string? bindc_name, bool elemental, bool pure, bool module, bool inline,
-        bool static, ttype* type_params, symbol* restrictions, bool is_restriction)
+        bool static, ttype* type_params, symbol* restrictions, bool is_restriction,
+        bool deterministic, bool side_effect_free)
     | GenericProcedure(symbol_table parent_symtab, identifier name,
         symbol* procs, access access)
     | CustomOperator(symbol_table parent_symtab, identifier name,
```
It's probably time to lump all these attributes into some "tuple" in ASR.asdl, but that's a separate issue.

Using the properties I and II listed at Wikipedia: https://en.wikipedia.org/wiki/Pure_function:

* *deterministic*: arguments fully determine the result, no dependence on external state (no reading of global variables)
* *side-effect-free*: no writing to global variables / changing some external state (no writing of global variables)

Note: a Fortran "pure" keyword means side-effect-free, but doesn't have to be deterministic. Fortran "simple" means both deterministic and side-effect-free.

A few examples:

### deterministic=true, side_effect_free=true

```python
def f(x: i32) -> i32:
    return x**2
```

The optimizer can replace the expression `f(1) + f(2) + f(3)` with just `14` (1+4+9 = 14).

### deterministic=true, side_effect_free=false

```python
def f(x: i32) -> i32:
    print(x)
    return x**2
```

The optimizer can replace the expression `f(1) + f(2) + f(3)` with just `14` (1+4+9 = 14), but it must still call `f(1); f(2); f(3);` (can drop the results) in this order and arguments and in the same position with regards to other functions with this exact side-effect (printing), as well as functions that are not deterministic (that read the console / screen output).

### deterministic=false, side_effect_free=true

```python
y = 5
def f(x: i32) -> i32:
    return x**2 + y
```

If the optimizer knows the global state (`y` in this case) does not change during the execution of `f(1) + f(2) + f(3)`, then it can compute this sum in any order.

### deterministic=false, side_effect_free=false

```python
y = 5
def f(x: i32) -> i32:
    print(x)
    return x**2 + y
```

In general the order of execution of `f(1) + f(2) + f(3)` must be preserved.

akshanshbhatt:
Closes #1266


certik:
Let's first send a PR that only introduces the two arguments into Function and sets them to false, and get everything updated and CI passing.

Then let's work on actually setting them correctly in a second PR.

certik:
Now when #1427 is merged, we can now finish this PR.

Smit-create:
I'll complete this PR as a part of https://github.com/lfortran/lfortran/issues/1303

Smit-create:
1. For checking determinisitic functions, we need to visit each stmt inside the body and check if any variables is accessed that comes from the parent symtabs of the current function symtab -- I suppose this can be easily check with variable dependencies the function is having.
2. For side-effect-true, I'm not sure what's the best way to detect as there can be many ways of changing the external state.

Smit-create:
In the following example:
```py
y: i32 = 5
def f(x: i32) -> i32:
    return x**2 + y
```

`f` depends on `y`, but the ASR does not include it in the dependency list. Is this a bug @czgdp1807?

<details>
<summary> ASR </summary>
<p>

```console
(TranslationUnit
    (SymbolTable
        1
        {
            f:
                (Function
                    (SymbolTable
                        2
                        {
                            _lpython_return_variable:
                                (Variable
                                    2
                                    _lpython_return_variable
                                    []
                                    ReturnVar
                                    ()
                                    ()
                                    Default
                                    (Integer 4 [])
                                    Source
                                    Public
                                    Required
                                    .false.
                                ),
                            x:
                                (Variable
                                    2
                                    x
                                    []
                                    In
                                    ()
                                    ()
                                    Default
                                    (Integer 4 [])
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
                        })
                    f
                    (FunctionType
                        [(Integer 4 [])]
                        (Integer 4 [])
                        Source
                        Implementation
                        ()
                        .false.
                        .false.
                        .false.
                        .false.
                        .false.
                        []
                        []
                        .false.
                    )
                    []
                    [(Var 2 x)]
                    [(=
                        (Var 2 _lpython_return_variable)
                        (IntegerBinOp
                            (IntegerBinOp
                                (Var 2 x)
                                Pow
                                (Integer 4 [])
                                ()
                            )
                            Add
                            (Var 1 y)
                            (Integer 4 [])
                            ()
                        )
                        ()
                    )
                    (Return)]
                    (Var 2 _lpython_return_variable)
                    Public
                    .true.
                    .false.
                ),
            main_program:
                (Program
                    (SymbolTable
                        3
                        {
                            
                        })
                    main_program
                    []
                    []
                ),
            y:
                (Variable
                    1
                    y
                    []
                    Local
                    (IntegerConstant 5 (Integer 4 []))
                    ()
                    Default
                    (Integer 4 [])
                    Source
                    Public
                    Required
                    .false.
                )
        })
    []
)
```
</p>
</details>

Issue #1298, opened 2022-11-14
[WASM_x86] Bug in the loop statement


Thirumalai-Shaktivel:
It seems this condition `(if_unique_id.size() - loop_unique_id.size() == label_index - 1)` might not be perfect to determine to `continue` or `end` the `loop`. Consider the following example where we have a `loop` inside an `if` construct:

```python
def main0():
    x: i32
    x = (2+3)*5
    j: i32 = 0
    if True:
        while x > 0:
            j += 2
            x -= 1
    print(x)
    print(j)

main0()
```

Output:
```bash
(lp) ubaid@ubaid-Lenovo-ideapad-330-15ARR:~/OpenSource/lpython$ lpython examples/expr2.py
0
50
(lp) ubaid@ubaid-Lenovo-ideapad-330-15ARR:~/OpenSource/lpython$ lpython examples/expr2.py --backend wasm_x86 -o loop
(lp) ubaid@ubaid-Lenovo-ideapad-330-15ARR:~/OpenSource/lpython$ ./loop 
24
2
(lp) ubaid@ubaid-Lenovo-ideapad-330-15ARR:~/OpenSource/lpython$ 
```

The `loop` in this case just executes once and then exits. This is because in this case (when `visit_Br()` is visited) the `if` count would be `2` and `loop` count would be `1`. The `label_index` would be `1`. The condition would be `2 - 1 == 1 - 1`, that is `1 == 0`, which is not `true` and hence the `loop` exits.

I think we might need to simulate how `wasm` works in `x86`. We might need to use same label indexes (or same label index space) for `if`'s and `loop`'s (just like `wasm` does).

_Originally posted by @Shaikh-Ubaid in https://github.com/lcompilers/lpython/pull/1290#discussion_r1021693996_

PR #1299, opened 2022-11-15
[WASM_x86] Fix a bug that calculates the label_index in the Br visitor

Thirumalai-Shaktivel:
Fixes: #1298

### DIFF
--- a/integration_tests/loop_02.py
+++ b/integration_tests/loop_02.py
@@ -75,6 +75,18 @@ def test_loop_03():
         i += 1
     assert k == 826
 
+    i = 0
+    if i == 0:
+        while i < 10:
+            j = 0
+            if j == 0:
+                while j < 10:
+                    k += 1
+                    if i == 9:
+                        break
+                    j += 1
+                i += 1
+    assert k == 917
 
 def verify():
     test_loop_01()
--- a/src/libasr/codegen/wasm_to_x86.cpp
+++ b/src/libasr/codegen/wasm_to_x86.cpp
@@ -34,6 +34,7 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
     uint32_t cur_func_idx;
     std::vector<std::string> if_unique_id;
     std::vector<std::string> loop_unique_id;
+    uint32_t cur_nesting_length;
     int32_t last_vis_i32_const, last_last_vis_i32_const;
     std::unordered_map<int32_t, std::string> loc_to_str;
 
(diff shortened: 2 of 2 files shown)


Shaikh-Ubaid:
Thirumailai, it seems the condition `(loop_unique_id.size() + if_unique_id.size() - cur_nesting_length == label_index  + 1)` might not be simple to understand. I think there could be a simpler way if we try to simulate similar to how `wasm` does. `wasm` uses the same `label index space` for `if`'s and `loop`'s. Similary, if we use the same `label index space` for `if`'s and `loop`'s in `wasm->x86` backend, I think we can find a simpler approach. Ideally, we should attempt to follow the label indexing as described here https://github.com/lcompilers/lpython/pull/1290#discussion_r1020954737.

>`wasm` uses the same `label index space` for `if`'s and `loop`'s

By same `label index space`, I mean that `if` ids and `loop` ids would be counted from the same integer domain. That is, if there is an `if` with an `id = 4 (let's say)`, there could not be a `loop` with that same `id = 4`.

Shaikh-Ubaid:
>(loop_unique_id.size() + if_unique_id.size() - cur_nesting_length) is same as nesting_level - cur_loop_nesting_level, right?

Yup, got it. I think the condition fine for the moment.

>Can you please share where the label index space is used in the wasm? or how it is implemented? (P.S. I didn't get it properly)

It is more abstract/conceptual. In simple words, what I mean by `wasm uses the same label index space for if's and loop's` is that `if`'s and `loop`'s use the same Id sequence (i.e. they take ids from the same set).

As the tests pass, let's merge this for the moment and we can improve upon it incrementally.

Shaikh-Ubaid (approved):
I tested it locally. It seems to work.

Thirumalai-Shaktivel:
Before spilting into two id's (if_id, loop_id), I thought to use the same approach as wasm  does i.e., using nesting_level. Like push_back both loop_id and if_id into one stack. Later I realized it is difficult to retrieve the correct id's in the nested if_loop statement. 
Yup, this (`loop_unique_id.size() + if_unique_id.size() - cur_nesting_length`) seems to work for many cases. But, if you find any bug, do let me know or create an issue.

Shaikh-Ubaid:
> I thought to use the same approach as wasm does i.e., using nesting_level

Yes, I guessed. Since `wasm` already uses the nesting level and as we are working on `wasm` to `x86`, ideally, we should not be needing to use nesting level again (since it is something already figured out at the wasm level).

>ideally, we should not be needing to use nesting level again

I am not sure how much this ideal case is possible, but we need to try/experiment and see.

PR #1301, opened 2022-11-15
Handle struct return type in C backend

czgdp1807:


certik (approved):
Seems ok, didn't review all the details.

PR #1302, opened 2022-11-15
Run the executable if no error

Thirumalai-Shaktivel:


src/bin/lpython.cpp
```diff
@@ -1526,32 +1506,48 @@ int main(int argc, char *argv[])
...
                 throw LFortran::LCompilersException("Unsupported backend.");
             }
-            if (err) return err;
-            return link_executable({tmp_o}, outfile, runtime_library_dir,
-                    backend, static_link, true, compiler_options);
+
+            if (compiler_options.arg_o == "") {
+                if (backend == Backend::wasm) {
+                    err = system(("js " + outfile +".js").c_str());
+                } else {
```
Thirumalai-Shaktivel, line 1539: @Shaikh-Ubaid, the Wasm Backend executable can be executed using `js filename.out.js`, right?

Shaikh-Ubaid: It can be executed using `node a.out.js` (where `a.out` and `a.out.js` are the output files generated).

I am unsure if we need auto-execution for `wasm` backend (or other backends excluding `llvm`). It seems that when a user specifies a particular backend (apart from `llvm`, for example, let's say `--backend=wasm`), the (primary) intention of the user is to generate target backend code from the source code. Thus, if a user executes `lpython examples/expr2.py --backend wasm`, he's trying to generate `wasm` code from the source code. He could later choose to use the generated `wasm` code for his own purposes or choose to execute it.

I think it would be better if we leave the choice of what should be done with the output code (`wasm` code in our example) onto the user. One added advantage of leaving the choice with the user is that the user would get to know what all files are generated using a particular backend (`wasm` backend generates two files `a.out` and `a.out.js`). The user also gets the opportunity to explore how the generated target (`wasm`) code could be executed.

I am just sharing my thoughts here. Anything is-fine/works for me.

Thirumalai-Shaktivel: @czgdp1807 @certik Can you please share your thoughts as well?

czgdp1807: > It seems that when a user specifies a particular backend (apart from llvm, for example, let's say --backend=wasm), the (primary) intention of the user is to generate target backend code from the source code.

Not sure. I guess `--show-wasm` should generate the code?

Overall I guess at least the executable shouldn't be deleted after automatic execution. For now for LLVM backend it gets deleted.

Thirumalai-Shaktivel: After [this PR](https://github.com/lcompilers/lpython/pull/1093/files#diff-34a6ddf1f5bdf792153446e9e32bf70fcd561630f4acf4543effa1a84922e7dbL965-R966) got merged. LPython doesn't not generate `a.out` anymore. Instead it generates `filename.out`. These changes were made because of some problems in testing the runtime outputs, see [here](https://github.com/lcompilers/lpython/pull/1093#discussion_r1001948244) for more details. Also, to stop the automatic execution one can use `-o` option to specify the executable to be written into.
My appologize for sharing the updated information.

certik: I don't have a strong opinion either way on this. I feel calling "node" from LPython seems fragile.

I think for now, why don't we do the following:

* `lpython a.py` compiles `a.py` with the default backend (currently LLVM, perhaps in the future it will be WASM + x86/arm) and runs it
* If any other option is passed to `lpython`, then it will not run it by default.

Although I can see how this is not super consistent. But it gets us started. I think the most common use case is `python a.py`, so we now already have `lpython a.py`. Everything else seems lower priority.

We could also consider `lpython run a.py`.

Thirumalai-Shaktivel: Do you mean `lpython --run a.py --backend x86` to run by default?

Thirumalai-Shaktivel: I think `lpython a.py --backend x86 && ./a.out` is much familiar than using `--run`. So, Let's only support `lpython a.py` to run by default.
Also, I have a doubt: Currently, we write the executable to `filename.out`. Is this fine for other backend as well?

src/bin/lpython.cpp
```diff
@@ -1526,32 +1506,52 @@ int main(int argc, char *argv[])
...
+                        err = system(("./" + outfile).c_str());
+                    }
+                }
+                if (err != 0) {
+                    if (0 < err && err < 256) {
+                        return err;
+                    } else {
+                        return 1;
+                    }
+                }
```
Thirumalai-Shaktivel, line 1552: This is separate topic:
I recently came across something like:
```py
def test():
	quit(10)

test()
```
```console
$ lpython a.py
err = 2560
STOP
$ echo $?
1
```
So, I thought to replace this with the following change:
```suggestion
                if (err != 0) {
                    if (0 < err && err < 256) {
                        return err;
                    } if else (err % 256 == 0) {
                    	return err / 256;
                    } else {
                        return 1;
                    }
                }
```
@certik

src/bin/lpython.cpp
czgdp1807: > Do you mean `lpython --run a.py --backend x86` to run by default?

Makes sense to me.

czgdp1807 (approved):
LGTM. Will ask @certik as well before merging. Conflicts should be resolved as well.

certik (approved):
Looks good to me.

Open up a new issue for the error return logic.

Thirumalai-Shaktivel:
Thanks for the approvals. Let's merge this.

PR #1304, opened 2022-11-16
Remove dead code from ``convert_call_args``

czgdp1807:
See https://github.com/lfortran/lfortran/pull/1017


certik (approved):
Didn't review the details, but it looks ok if everything still works.

Issue #1305, opened 2022-11-16
Importing modules does not work


certik:
```
$ tree      
.
├── a
│   ├── __init__.py
│   └── b.py
└── r.py

1 directory, 3 files
$ cat a/b.py 
def f():
    print("b.f() called")
$ cat r.py 
from a.b import f
f()
```
This gives:
```console
$ lpython r.py 
semantic error: Could not find the module '__init__.py'. If an import path is available, please use the `-I` option to specify it
 --> r.py:1:1
  |
1 | from a.b import f
  | ^^^^^^^^^^^^^^^^^ 


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).
$ lpython -I . r.py
semantic error: Could not find the module '__init__.py'. If an import path is available, please use the `-I` option to specify it
 --> r.py:1:1
  |
1 | from a.b import f
  | ^^^^^^^^^^^^^^^^^ 


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).
```

certik:
@Thirumalai-Shaktivel any idea why this does not work?

Thirumalai-Shaktivel:
Currently, we don't support
```py
from a.b import f
```
Instead this works:
```py
from a import f
```
```console
$ lpython r.py
b.f() called
```
I will try to work on this soon.

Thirumalai-Shaktivel:
Fixed here: https://github.com/lcompilers/lpython/pull/1306

certik:
Ok, the above now works with #1306, but more deeply nested doesn't:
```console
$ tree a
a
├── __init__.py
├── b.py
└── c
    ├── __init__.py
    └── d.py
(lf) repos/lpython(nested_modules_02) $ cat a/c/d.py 
def g():
    print("a.c.g() called")
(lf) repos/lpython(nested_modules_02) $ cat r.py 
from a.b import f
f()
from a.c.d import g
g()
(lf) repos/lpython(nested_modules_02) $ python r.py 
b.f() called
a.c.g() called
(lf) repos/lpython(nested_modules_02) $ lpython r.py
semantic error: Could not find the module 'a/c.d.py'. If an import path is available, please use the `-I` option to specify it
 --> r.py:3:1
  |
3 | from a.c.d import g
  | ^^^^^^^^^^^^^^^^^^^ 


Note: if any of the above error or warning messages are not clear or are lacking
[...]

PR #1306, opened 2022-11-17
Move nested modules handling into a separate function

Thirumalai-Shaktivel:
Related: #992, #1305


certik (approved):
I think this looks good, thanks for fixing it!

Let's merge this after the sync.

certik:
I rebased on top of the sync, let's see if everything passes. If so, we can merge.

PR #1310, opened 2022-11-17
Implement CPtr deepcopy and handle CPtr type passing into functions

czgdp1807:


integration_tests/bindc_03.py
```diff
@@ -15,7 +15,6 @@ def gpy(a: CPtr, value: i32, offset_value: bool) -> None:
 def get_array(size: i32) -> CPtr:
     pass
 
-@ccallable
```
certik, line 18: Doesn't this change break some functionality that we need?

certik (approved):
Otherwise it looks good.

PR #1315, opened 2022-11-19
Use replace_regex to replace all occurrences of "." with "/"

Thirumalai-Shaktivel:
Related: #1305

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -3408,7 +3408,7 @@ public:
                     // Check for the nested modules with "."
                     // Example: from x.y import z
                     if (mod_sym.find(".") != std::string::npos) {
-                        mod_sym.replace(mod_sym.find("."), 1, "/");
+                        mod_sym = replace(mod_sym, "[.]", "/");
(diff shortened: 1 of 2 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3408,7 +3408,7 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                     // Check for the nested modules with "."
                     // Example: from x.y import z
                     if (mod_sym.find(".") != std::string::npos) {
-                        mod_sym.replace(mod_sym.find("."), 1, "/");
+                        mod_sym = replace(mod_sym, "[.]", "/");
```
czgdp1807, line 3411: Seems like this was hard coded before.

czgdp1807 (approved):
LGTM.

src/lpython/semantics/python_ast_to_asr.cpp
Thirumalai-Shaktivel: Previously, I thought `replace()` would replace every occrurances in the string. Now, I realized that it replaces only the 1st occurance or simple the index of the `fine()`.

Thirumalai-Shaktivel:

PR #1320, opened 2022-11-22
Fixes for ASR and LLVM backend

czgdp1807:


src/libasr/asr_verify.cpp
```diff
@@ -63,7 +63,7 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
             diagnostics.message_label("ASR verify: " + error_msg,
                 {loc}, "failed here",
                 diag::Level::Error, diag::Stage::ASRVerify);
-            throw VerifyAbort();
+            LFORTRAN_ASSERT(false);
```
certik, line 66: I think you have to abort here, otherwise you won't get the error message?

certik:
Should some tests be enabled that now work?

czgdp1807:
I will spawn of better PRs from this one which will have tests for each bug and feature.

src/libasr/asr_verify.cpp
czgdp1807: Well, I added this line for my own conviction to make the stacktrace include the code path which causes the ASR verify to fail. Otherwise it always stops at call to ASR verify and doesn’t include what caused the failure.

certik: It should work with `--show-stacktraces`, as it records the stacktrace to the failure when the exception was thrown. Does that not work?

czgdp1807: No it doesn't. See example below,

```bash
  File "/Users/czgdp1807/lpython_project/lpython/src/bin/lpython.cpp", line 1492
    return emit_llvm(arg_file, runtime_library_dir, lpython_pass_manager, compiler_options);
  File "/Users/czgdp1807/lpython_project/lpython/src/bin/lpython.cpp", line 547
    res = fe.get_llvm3(*asr, pass_manager, diagnostics, infile);
  File "/Users/czgdp1807/lpython_project/lpython/src/lpython/python_evaluator.cpp", line 58
    run_fn, infile);
  File "/Users/czgdp1807/lpython_project/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 6448
    pass_manager.apply_passes(al, &asr, pass_options, diagnostics);
  File "/Users/czgdp1807/lpython_project/lpython/src/libasr/pass/pass_manager.h", line 196
    _apply_passes(al, asr, _passes, pass_options, diagnostics);
  File "/Users/czgdp1807/lpython_project/lpython/src/libasr/pass/pass_manager.h", line 99
    LFORTRAN_ASSERT(LFortran::asr_verify(*asr, true, diagnostics));
AssertFailed: LFortran::asr_verify(*asr, true, diagnostics)
```
Ends there, we don't have the paths of `asr_verify.cpp` in the stacktrace. I think this is because `VerifyAbort` is getting caught at the following line,

https://github.com/lcompilers/lpython/blob/93cc7c2058da1c902a032a2a8338697625dada7d/src/libasr/asr_verify.cpp#L709-L714

czgdp1807:
@certik This is ready for review and is a merge-able PR. Please see my above comment.

src/libasr/asr_verify.cpp
certik: I think the issue is that the `verify()` call itself is wrapped in an assert:
```
LFORTRAN_ASSERT(LFortran::asr_verify(*asr, true, diagnostics));
```
I have reworked this at a few places, but we need to fix it everywhere. See this in `asr_utils`:
```
#if defined(WITH_LFORTRAN_ASSERT)
        diag::Diagnostics diagnostics;
        if (!asr_verify(*tu, true, diagnostics)) {
            std::cerr << diagnostics.render2();
            throw LCompilersException("Verify failed");
        };
#endif
```
When you do this, it will print the stacktrace for verify failures.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3249,6 +3253,24 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                Vec<ASR::symbol_t *> v;
+                v.reserve(al, 1);
+                overload_defs[sym_name] = v;
+            } else {
+                overload_number = std::to_string(overload_defs[sym_name].size());
+            }
+            sym_name = "__lpython_overloaded_" + overload_number + "__" + sym_name;
+        }
+        if (parent_scope->get_scope().find(sym_name) != parent_scope->get_scope().end()) {
+            throw SemanticError("Procedure, " + std::string(x.m_name) +  " is already defined", x.base.base.loc);
```
certik, line 3271: I would remove the comma, and also in Python they are called "functions" I think:
```suggestion
            throw SemanticError("Function " + std::string(x.m_name) +  " is already defined", x.base.base.loc);
```

certik (approved):
This looks good, I left a comment for a better error message, which you can fix in a subsequent PR, whatever is easier.

src/lpython/semantics/python_ast_to_asr.cpp
czgdp1807: Done.

Issue #1319, opened 2022-11-21
-I doesn't work


certik:
Here is an example:

https://github.com/certik/mathfn/issues/1

certik:
@Thirumalai-Shaktivel you might now what the issue is.

Thirumalai-Shaktivel:
> Refer, [certik/mathfn#1 (comment)](https://github.com/certik/mathfn/issues/1#issuecomment-1324717535) for self-assignment.

@czgdp1807 are you working on this issue?

Thirumalai-Shaktivel:
Following fix this issue:
```diff
diff --git a/src/lpython/semantics/python_ast_to_asr.cpp b/src/lpython/semantics/python_ast_to_asr.cpp
index 2f55a8849..ee0afd8c6 100644
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -3418,7 +3418,6 @@ public:
             if (import_path != "" &&
                     !path_exists(paths[0] + '/' + mod_sym + ".py")) {
                 paths = {import_path};
-                if (parent_dir != "") paths[0] += '/' + parent_dir;
                 if(path_exists(paths[0])) {
                     // Check for the nested modules with "."
                     // Example: from x.y import z
```
Now, we get this error:
```console
$ lpython -I src examples/example_sin.py
semantic error: sin already defined
 --> examples/example_sin.py:1:1
  |
1 | from mathfn.sin import sin
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).

```

czgdp1807:
Yes I am. Will make a PR shortly.

> So, here we should not use the `parent_dir`.

Yes. I figured out earlier today.

> Now, we get this error:

I am fixing this as well. Will send the PR. The logic of `set_module_symbol` is on a case by case basis and is hard coded. So it will fail on such cases.

Thirumalai-Shaktivel:
Yup, Thanks for fixing it.

PR #1322, opened 2022-11-23
Fix module imports

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1319

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -46,22 +46,31 @@ message("LPYTHON_RTLIB_LIBRARY: ${LPYTHON_RTLIB_LIBRARY}")
 
 macro(RUN)
     set(options FAIL)
-    set(oneValueArgs NAME)
+    set(oneValueArgs NAME IMPORT_PATH)
     set(multiValueArgs LABELS EXTRAFILES)
     cmake_parse_arguments(RUN "${options}" "${oneValueArgs}"
                           "${multiValueArgs}" ${ARGN} )
     set(name ${RUN_NAME})
+    set(import_path ${RUN_IMPORT_PATH})
     if (NOT name)
         message(FATAL_ERROR "Must specify the NAME argument")
     endif()
 
     if (${KIND} IN_LIST RUN_LABELS)
         if (KIND STREQUAL "llvm")
-            add_custom_command(
-                OUTPUT ${name}.o
-                COMMAND lpython -c ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py -o ${name}.o
-                DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
-                VERBATIM)
+            if (import_path)
+                add_custom_command(
+                    OUTPUT ${name}.o
+                    COMMAND lpython -c -I ${import_path} ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py -o ${name}.o
+                    DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
+                    VERBATIM)
+            else ()
+                add_custom_command(
+                    OUTPUT ${name}.o
+                    COMMAND lpython -c ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py -o ${name}.o
+                    DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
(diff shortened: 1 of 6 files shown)


czgdp1807:
@certik This is ready for review. Please let me know if anything's to be done logic wise. Otherwise I will clean up the git history and remove debug prints and merge. Thanks.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3484,6 +3522,7 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
             }
             bool ltypes, enum_py;
             set_module_symbol(msym, paths);
+            // std::cout<<"msym: "<<msym<<std::endl;
```
certik, line 3525: Some debug print statements should be removed.

integration_tests/CMakeLists.txt
```diff
@@ -239,7 +248,8 @@ RUN(NAME test_for_loop       LABELS cpython llvm c)
 RUN(NAME modules_01          LABELS cpython llvm wasm wasm_x86)
 RUN(NAME modules_02          LABELS cpython llvm wasm wasm_x86)
 RUN(NAME test_import_01      LABELS cpython llvm)
-RUN(NAME test_import_02      LABELS cpython llvm)
+RUN(NAME test_import_02      IMPORT_PATH integration_tests
+        LABELS cpython llvm)
```
certik, line 252: Isn't this change equivalent to what there was before? I think it is relative to the current directory, which I am not sure if it is the directory with the sources, and the `integration_tests` is a non-existent directory then. Unless the working directory is the root LPython git repository's directory, which it is not if I do it by hand. So the directory doesn't exist, and LPython just finds the same modules as before.

tests/reference/asr-test_import_02-55b47fa.stderr
```diff
@@ -1,4 +1,4 @@
-warning: The module 'test_import/test_import_2' cannot be loaded
+warning: The module 'test_import_2' located in tests/errors/test_import/test_import_2.py cannot be loaded
```
certik, line 1: This is a good change / test.

certik (approved):
I think this change is OK. I am a little confused exactly what is being tested, but the `tests/reference/asr-test_import_02` test is a good one.

integration_tests/CMakeLists.txt
czgdp1807: Correct. I just wanted to make sure that things work if we pass import paths separately. For example `main` fails if I pass `integration_tests` as import path (see example below). You can see even though behaviour of both commands should have been the same but it isn't. So I thought to test `IMPORT_PATH` as well.

```bash
(lp) 9:09:50:~/lpython_project/lpython % lpython -I integration_tests integration_tests/test_import_02.py
semantic error: Could not find the module 'test_import.import_module_01.py'. If an import path is available, please use the `-I` option to specify it
 --> integration_tests/test_import_02.py:1:1
  |
1 | from test_import.import_module_01 import multiply, e
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


(lp) 9:10:30:~/lpython_project/lpython % lpython integration_tests/test_import_02.py 
2.71828182845904509e+00
1.45136923488338110e+00
30
200
```

tests/reference/asr-test_import_02-55b47fa.stderr
czgdp1807: Yes. I provided the full path to the file which is not being loaded so that its easy for us and the users to actually see what went wrong and where. Before this it was a bit cryptic.

src/lpython/semantics/python_ast_to_asr.cpp
czgdp1807: Yes. Going to do that and clean up the git history.

integration_tests/CMakeLists.txt
certik: ok

czgdp1807: I guess we should make a directory outside integration_tests something called, `test_modules` and define our modules to be tested inside it. We pass the import path of `test_modules` in the command to check if it works.

certik: Yes.

PR #1328, opened 2022-11-28
Added dependencies attribute in ``Variable`` symbol

czgdp1807:


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -5909,7 +5910,14 @@ Result<ASR::TranslationUnit_t*> python_ast_to_asr(Allocator &al, LocationManager
         return res.error;
     }
     ASR::TranslationUnit_t *tu = ASR::down_cast2<ASR::TranslationUnit_t>(unit);
-    LFORTRAN_ASSERT(asr_verify(*tu, true, diagnostics));
+    PassUtils::UpdateDependenciesVisitor symbol_table_update(al);
+    symbol_table_update.visit_TranslationUnit(*tu);
```
certik, line 5914: I think that this is a problematic design for two reasons:

* The ASR is not correct, but we are running an ASR pass. The contract is that if we are running an ASR pass, the ASR is valid both before and after.
* For performance reasons, the AST -> ASR must be just two passes at most. We need two passes due to first constructing a symbol table, and then using it, it's hard to do everything in single pass. But this change seems to add additional two passes. I think that is too much.

We should figure out the dependencies as we are constructing it, and the dependencies in declarations should be such that you only depend on what is specified previously in the code. If it is specified below the current line, then you can't use it in a declaration. That should make it possible to figure out the dependencies in the same pass that constructs the ASR.

certik (changes requested):
Thanks for working on this, it looks good overall, but there is one important issue that we need to fix, see my comment.

src/lpython/semantics/python_ast_to_asr.cpp
czgdp1807: I see. 

> We should figure out the dependencies as we are constructing it, and the dependencies in declarations should be such that you only depend on what is specified previously in the code. 

I have thought about this and I can do this. However, it will be very bug prone and a lot tracking will be required to achieve this.

 What do you think of the following proposal?

"Merge this PR for now but as a follow up remove `PassUtils::UpdateDependenciesVisitor` from AST to ASR construction phase".

Let me know your thoughts. My main concern is that it would be very tricky to correctly track dependencies of several symbols in the AST to ASR phase. So it might take more time and efforts than usual. So I would suggest going for a slower but robust approach initially and then later on I will remove these `PassUtils::UpdateDependenciesVisitor` and make every thing work resulting in a faster solution. 

Also it depends on how urgent this feature is needed. Let me know about that as well.

certik: Well, this will be needed for LFortran as well then, and it will slow the compilation down for no good reason (I think).

So to go forward, we can extend ASR and implement whatever can be merged, but not to call this pass, that is, make it optional and off by default. Possibly exposed via a command line option.

I think we should not compromise on speed at all, as it is very hard to later recover, even if it seems simple.

Let's discuss a possible path forward at our next video meeting. If designed well, it seems to me it should be equivalent to the pass, so we maybe create some class or something to do this.

src/libasr/pass/array_op.cpp
```diff
@@ -978,6 +978,8 @@ void pass_replace_array_op(Allocator &al, ASR::TranslationUnit_t &unit,
     std::string rl_path = pass_options.runtime_library_dir;
     ArrayOpVisitor v(al, rl_path);
     v.visit_TranslationUnit(unit);
+    PassUtils::UpdateDependenciesVisitor u(al);
+    u.visit_TranslationUnit(unit);
```
certik, line 982: Looks like we now have these extra passes that need to be run.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -13,6 +13,7 @@
 
 #include <libasr/asr.h>
 #include <libasr/asr_utils.h>
+#include <libasr/pass/pass_utils.h>
```
certik, line 16: Why do we need this include?

czgdp1807: This will be removed.

integration_tests/variable_decl_02.py
```diff
@@ -0,0 +1,14 @@
...
+    c: i64 = i64(d + 1)
+    b: i64 = i64(d) + c + i64(2)
+    a: i32 = i32(b) + 20
+    print(a, b, c, d)
+    assert d == 0
+    assert c == i64(1)
+    assert b == i64(3)
+    assert a == 23
+
+f()
```
certik, line 14: This test already works with the current master:
```
$ lpython a.py 
23 3 1 0
```
So that's not a good test for this PR?

src/libasr/pass/array_op.cpp
czgdp1807: There are two choices with us,

1. Either track dependencies while writing a pass. So a pass involves first doing the job of writing the visitor like in the above case `ArrayOpVisitor` and then track dependencies in the pass itself i.e., inside `ArrayOpVisitor`.
2. Write the pass visitor `ArrayOpVisitor` and then call `PassUtils::UpdateDependenciesVisitor`.

The second approach decouples writing a pass and tracking dependencies. Calling `PassUtils::UpdateDependenciesVisitor` is not necessary after every calling every pass visitor, just only those which add functions, structures or modify variables. Passes of such kind are very few in number AFAIK.

integration_tests/variable_decl_01.py
```diff
@@ -0,0 +1,19 @@
+from ltypes import i32, i64
+from numpy import empty, int64
+
+def f(n: i32, m: i32):
+    l: i32 = 2
+    a: i64[n, m, l] = empty((n, m, l), dtype=int64)
```
certik, line 6: `a` depends on `n` and `m`, which is fine, every time the function gets called with some arguments, it will allocate a "stack" (or equivalent) array of length that is an expression of `n` and `m`.

What is `l`? I thought we generate an assignment for it like `l = 2` , but by looking at the ASR using this PR:
```
                            a:
                                (Variable 
                                    171 
                                    a 
                                    [n
                                    m
                                    l] 
                                    Local 
                                    () 
                                    () 
                                    Default 
                                    (Integer 8 [((IntegerConstant 0 (Integer 4 []))                                 (Var 171 n))
                                    ((IntegerConstant 0 (Integer 4 [])) 
                                    (Var 171 m))
                                    ((IntegerConstant 0 (Integer 4 [])) 
                                    (Var 171 l))]) 
                                    Source 
                                    Public 
                                    Required 
                                    .false.
                                ), 
                            l:
                                (Variable 
                                    171 
                                    l 
                                    [] 
                                    Local 
                                    () 
                                    Default 
                                    (Integer 4 []) 
                                    Source 
                                    Public 
                                    Required 
                                    .false.
                                ), 
```
It shows that `l` is actually a constant. In that case, we can use it in declarations and yes, doing the dependency is fine.

So I guess this is fine.

src/libasr/ASR.asdl
```diff
@@ -103,12 +103,12 @@ symbol
         ttype type, symbol? parent)
     | UnionType(symbol_table symtab, identifier name, identifier* dependencies,
         identifier* members, abi abi, access access, symbol? parent)
-    | Variable(symbol_table parent_symtab, identifier name, intent intent,
-        expr? symbolic_value, expr? value, storage_type storage, ttype type,
+    | Variable(symbol_table parent_symtab, identifier name, identifier* dependencies,
+        intent intent, expr? symbolic_value, expr? value, storage_type storage, ttype type,
```
certik, line 107: I see that we always use `identifier* dependencies` for all symbols, such as Functions. What is the convention --- does the dependency have to be in the parent scope, i.e. the symbol table that contains the given symbol?

What if a function calls another function from another module? What if a variable declaration depends on another variable from another module? I guess then it would be ExternalSymbol in the same scope.

What if it depends on something that is further up in the scope (parent's parent)?

Is the logic that you can find it by "resolving" it recursively?

Why cannot we just point to the symbols directly?

integration_tests/variable_decl_02.py
czgdp1807: I didn't try it on `main`. However I would suggest to keep this test as it verifies the changes of this PR as well and its a very small one so shouldn't effect the test timings by much. Originally I wasn't planning to add new tests as existing ones felt sufficient but then I added two new tests anyways to make sure whatever I changes I did work correctly.

certik:
I think that this is probably fine, but I still have some questions.

src/libasr/pass/array_op.cpp
certik: Ok!

src/libasr/ASR.asdl
certik: Let's document this in here: https://github.com/lfortran/lfortran/blob/bb49552e0fc51706ba4f0772636451c5b0aab8fc/doc/src/asr/asr_nodes/symbol_nodes/symbol.md, we should probably split it into one file per symbol node, and document what the intention and restrictions are of each argument.

czgdp1807: I am following the pattern of storing dependencies in `Module`. 

> Why cannot we just point to the symbols directly?

We can. There's nothing preventing us from doing so. Its a judgement call. Store the dependencies as strings and resolve them using `current_scope->resolve_symbol(string_name)` or just use the symbols directly. No strong opinions, both should work. What do you suggest to do?

czgdp1807: Strings feel easier to manage though. 😅.

czgdp1807:
> I think that this is probably fine, but I still have some questions.

Answered. Let me know if you have more. Will answer them as well.

certik (approved):
I think it's probably good enough. We can improve upon it as needed.  Thanks for your work!

PR #1331, opened 2022-11-29
Added option to pass custom C postprocessor

czgdp1807:
cc: @certik


integration_tests/CMakeLists.txt
```diff
@@ -82,11 +83,20 @@ macro(RUN)
...
-            add_custom_command(
-                OUTPUT ${name}.c
-                COMMAND lpython --show-c ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py > ${name}.c
-                DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
-                VERBATIM)
+            if (custom_c_postprocessor)
+                add_custom_command(
+                    OUTPUT ${name}.c
+                    COMMAND lpython --show-c ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
+                            --custom-c-postprocessor ${CMAKE_CURRENT_SOURCE_DIR}/${custom_c_postprocessor} > ${name}.c
```
certik, line 90: I think as part of this change, let's not do `--show-c`, but rather save the result to a file ourselves, so no need for ">" which is quite fragile in build systems.

czgdp1807: I agree. Will do this on Thursday.

certik:
Thanks for this, the other thing we need is to allow passing arguments to the script. I have just refactored such a script so that it accepts an argument of a `some_file.inc` that gets included at the top of the generated file as a header, instead of heaving the header hardwired in the script. I also need to add another argument with a file with a list of functions to modify.

czgdp1807:
Well what do you think about this - Instead of passing a bunch of arguments to the post processor via LPython, can we just create a `json` file specifying all the configuration and the post processor can read this `json` file and do the necessary things. We can pass the path to this `json` to the post processor via command line argument. So LPython will only call, `python post_processor.py path_to_json_config_file` and not `python post_processor.py some_headers some_other_information some_more_information`.

certik:
I will test this PR manually and see if it satisfyies our use case.

PR #1335, opened 2022-11-30
WASM_X64: Initial backend

Shaikh-Ubaid:
This `PR` adds initial support for `wasm` to `x64` backend.

### DIFF
--- a/src/libasr/codegen/wasm_to_x64.cpp
+++ b/src/libasr/codegen/wasm_to_x64.cpp
@@ -1,0 +1,136 @@
+#include <fstream>
+#include <chrono>
+#include <iomanip>
+
+#include <libasr/assert.h>
+#include <libasr/codegen/wasm_decoder.h>
+#include <libasr/codegen/wasm_to_x64.h>
+#include <libasr/codegen/x86_assembler.h>
+
+namespace LFortran {
+
+namespace wasm {
+
+/*
+
+This X64Visitor uses stack to pass arguments and return values from functions.
+Since in X64, instructions operate on registers (and not on stack),
+for every instruction we pop elements from top of stack and store them into
+registers. After operating on the registers, the result value is then
+pushed back onto the stack.
+
+One of the reasons to use stack to pass function arguments is that,
+it allows us to define and call functions with any number of parameters.
+As registers are limited in number, if we use them to pass function arugments,
+the number of arguments we could pass to a function would get limited by
+the number of registers available with the CPU.
+
+*/
+
+class X64Visitor : public WASMDecoder<X64Visitor>,
+                   public WASM_INSTS_VISITOR::BaseWASMVisitor<X64Visitor> {
+   public:
+    X86Assembler &m_a;
+
+    X64Visitor(X86Assembler &m_a, Allocator &al,
+               diag::Diagnostics &diagonostics, Vec<uint8_t> &code)
+        : WASMDecoder(al, diagonostics),
+          BaseWASMVisitor(code, 0U /* temporary offset */),
+          m_a(m_a) {
+        wasm_bytes.from_pointer_n(code.data(), code.size());
+    }
+
+    void gen_x64_bytes() {
+        // update the initial value of asm text as per X64 text format
+        m_a.update_asm("BITS 64\n\n");
+
+        emit_elf64_header(m_a);
+
+        {
+            m_a.add_label("_start");
+
+            // exit with a fixed non-zero exit code
+            m_a.asm_mov_r64_imm64(LFortran::X64Reg::rax, 60); // sys_exit
+            m_a.asm_mov_r64_imm64(LFortran::X64Reg::rdi, 24 /* exit_code */); // exit code
+            m_a.asm_syscall(); // syscall
+        }
+
+        emit_elf64_footer(m_a);
+    }
+};
+
+}  // namespace wasm
+
+Result<int> wasm_to_x64(Vec<uint8_t> &wasm_bytes, Allocator &al,
+                        const std::string &filename, bool time_report,
+                        diag::Diagnostics &diagnostics) {
+    int time_decode_wasm = 0;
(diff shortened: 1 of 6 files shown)


src/libasr/codegen/wasm_to_x64.cpp
```diff
@@ -0,0 +1,136 @@
...
+            m_a.add_label("_start");
+
+            // exit with a fixed non-zero exit code
+            m_a.asm_mov_r64_imm64(LFortran::X64Reg::rax, 60); // sys_exit
+            m_a.asm_mov_r64_imm64(LFortran::X64Reg::rdi, 24 /* exit_code */); // exit code
+            m_a.asm_syscall(); // syscall
+        }
+
+        emit_elf64_footer(m_a);
+    }
```
Shaikh-Ubaid, line 59: Currently, we are generating fixed/pre-determined instructions.

The fixed instructions we generate in assembly textual format are
```asm
; setting exit code
mov rax, 60 ; sys_exit() system call
mov rdi, 24 ; exit code
syscall
```

Output:
```bash
(lp) lpython$ lpython examples/expr2.py --backend wasm_x64 -o tmp
(lp) lpython$ ./tmp
(lp) lpython$ echo $?
24
```

I tested with some other values  (>= `0` and <= `255` (`255` is probably max limit for exit code)) of exit codes. It seems to work.

src/libasr/codegen/wasm_to_x64.cpp
```diff
@@ -0,0 +1,136 @@
...
+        std::cout << "Generate asm: " << std::setw(5) << time_gen_x64_bytes
+                  << std::endl;
+        std::cout << "Verify:       " << std::setw(5) << time_verify
+                  << std::endl;
+        std::cout << "Save:       " << std::setw(5) << time_save << std::endl;
+        int total =
+            time_decode_wasm + time_gen_x64_bytes + time_verify + time_save;
+        std::cout << "Total:      " << std::setw(5) << total << std::endl;
+    }
+    return 0;
```
Shaikh-Ubaid, line 133: This code seems similar to `wasm_to_x86()` (and thus seems repetitive).

src/libasr/codegen/x86_assembler.h
```diff
@@ -567,6 +641,21 @@ class X86Assembler {
         EMIT("mov " + r2s(r32) + ", " + i2s(imm32));
     }
 
+    void rex(int W, int R, int X, int B) {
+        LFORTRAN_ASSERT(W <= 1);
+        LFORTRAN_ASSERT(R <= 1);
+        LFORTRAN_ASSERT(X <= 1);
+        LFORTRAN_ASSERT(B <= 1);
+        m_code.push_back(m_al, (0b01000000 | (W << 3) | (R << 2) | (X << 1) | B));
+    }
```
Shaikh-Ubaid, line 650: `W`, `R`, `X`, `B` are single bit values (they take value `0` or `1`), and I am unsure what `data type` (`int`, `uint8_t`, etc.) they should have. I currently used `int` data type for them (as `int` seems general).

Shaikh-Ubaid:
This seems ready. Please possibly review and please share feedback.

src/libasr/codegen/x86_assembler.h
certik: I would use the same data type that m_code.push_back() is using, which is `uint8_t`, so I would use that. That way there are no issues with automatic casting.

src/libasr/codegen/x86_assembler.h
```diff
@@ -567,6 +641,21 @@ class X86Assembler {
...
+
+    void asm_mov_r64_imm64(X64Reg r64, uint64_t imm64) {
+        rex(1, 0, 0, 0);
```
certik, line 653: I would write this as:
```suggestion
        m_code.push_back(m_al, rex(1, 0, 0, 0));
```

certik (approved):
Very good, great job on this!

I think you got it. I don't see any major issues, I just left some minor comments. Feel free to merge when you are ready and we can iterate upon this as we go. 

Very exciting, this is the real deal, this will take us all the way to a production Debug x86 backend.

src/libasr/codegen/x86_assembler.h
Shaikh-Ubaid: Got it, Thank you.

src/libasr/codegen/x86_assembler.h
Shaikh-Ubaid: Got it. It looks great. Yes, it makes it explicit and much simpler. Thank you so much for sharing.

Shaikh-Ubaid:
Thank you so much for the guidance and feedback. I incorporated the suggested changes. I tested the changes locally and they seem to work. I am adding this to auto-merge as it seems to be approved. Thank you again.

certik:
Perfect! Thanks.

So now the next step is to start generating these instructions from WASM.

src/libasr/codegen/x86_assembler.cpp
```diff
@@ -202,4 +202,74 @@ void emit_print_int(X86Assembler &a, const std::string &name)
...
+    a.add_var("filesize", a.pos()-a.origin());
+}
+
+void emit_exit_64(X86Assembler &a, std::string name, int exit_code) {
+    a.add_label(name);
+    // void exit(int status);
+    a.asm_mov_r64_imm64(LFortran::X64Reg::rax, 60); // sys_exit
+    a.asm_mov_r64_imm64(LFortran::X64Reg::rdi, exit_code); // exit code
+    a.asm_syscall(); // syscall
+}
```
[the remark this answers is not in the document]
Shaikh-Ubaid: >So now the next step is to start generating these instructions from WASM.

Yes, now I will attempt to support this `exit` function (with compile-time exit code). I guess, we will need support for `jmp` and `return` instructions. After their support, I think/guess basic function calls could/should work (hopefully).

PR #1336, opened 2022-12-01
Pointers in emulation mode

### LINKED ISSUE #1334: Feature request: pointers in emulation mode
This works with LPython:
```python
def mystruct_set(mystruct: CPtr, k: i16) -> None:
    b: Pointer[mystruct]
    c_p_pointer(mystruct, b)
    b.k = k
```
But not in CPython mode, because "b" is not declared in `c_p_pointer(mystruct, b)`:
```
UnboundLocalError: local variable 'b' referenced before assignment
```
Also the implementation of `c_p_pointer` makes no sense:
```python
def c_p_pointer(cptr, targettype):
    return pointer(targettype)
```
As it needs to use the `cptr` pointer. Rather, the implementation must look something like this:
```python
def mystruct_set(mystruct: CPtr, k: i16) -> None:
    struct_type = ctypes.POINTER(mystruct_type)
    newa = ctypes.cast(mystruct, struct_type)
    newa.contents.k = k
```

czgdp1807:
Closes #1334


certik:
This seems to work well:
```console
$ lpython structs_15.py 
4
514 2
1542 6
20 238
2
$ PYTHONPATH=../src/runtime/ltypes python structs_15.py
4
514 2
1542 6
20 -18
2
```
The only difference is due to https://github.com/lcompilers/lpython/issues/1337, which is a separate issue.

certik (approved):
These changes are fine with me. 

@czgdp1807 I assume you have some stuff that you still want to polish here, so go ahead and then merge. Otherwise we can squash & merge.

certik:
@czgdp1807 I need this, so I merged it. If there are some things you want to improve, then go ahead and send a new PR.

PR #1338, opened 2022-12-02
WASM_X64: Basic support for function calls

Shaikh-Ubaid:
This `PR` adds initial support for function calls in the `WASM_X64` backend. This `PR` also adds support of generating `X64` instructions from `wasm` instructions.

### DIFF
--- a/src/libasr/codegen/wasm_to_x64.cpp
+++ b/src/libasr/codegen/wasm_to_x64.cpp
@@ -31,6 +31,7 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
                    public WASM_INSTS_VISITOR::BaseWASMVisitor<X64Visitor> {
    public:
     X86Assembler &m_a;
+    uint32_t cur_func_idx;
 
     X64Visitor(X86Assembler &m_a, Allocator &al,
                diag::Diagnostics &diagonostics, Vec<uint8_t> &code)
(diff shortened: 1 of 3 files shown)


src/libasr/codegen/wasm_to_x64.cpp
```diff
@@ -40,19 +41,105 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
...
+            case 0: {  // print_i32
+                std::cerr << "Call to print_i32() will be printed as exit code\n";
+
+                // Currently, for print we are setting the value to be printed
+                // as the exit code. Later we can access/view this value from console
+                // using: echo $?
+                m_a.asm_pop_r64(X64Reg::rdi); // get exit code from stack top
+                m_a.asm_mov_r64_imm64(X64Reg::rax, 60); // sys_exit
+                m_a.asm_syscall(); // syscall
+                break;
```
Shaikh-Ubaid, line 57: Currently, instead of printing a value, we set the value as `exit code` and can then see/view the exit code value using `echo $?`.

src/libasr/codegen/wasm_to_x64.cpp
```diff
@@ -40,19 +41,105 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
...
+        m_a.asm_call_label(exports[func_idx].name);
+    }
+
+    void visit_I32Const(int32_t value) {
+        // direct addition of imm64 to stack is not available with us yet
+        // so temporarily using a combination of instructions
+        // TODO: Update this once we have support for push_imm64()
+        m_a.asm_mov_r64_imm64(X64Reg::rax, value);
+        m_a.asm_push_r64(X64Reg::rax);
+    }
```
Shaikh-Ubaid, line 107: **doubt:** We are building a `64-bit` assembly backend. Therefore, for `visit_I32Const()`, should we be adding a `32-bit` constant or a `64-bit` constant to the stack? At the moment, I used a `64-bit` constant as in the above code.

Shaikh-Ubaid:
The `textual` `64-bit` assembly format that is currently generated seems to `segment fault`. I think we can fix in some other `PR`.

Shaikh-Ubaid:

src/libasr/codegen/wasm_to_x64.cpp
certik: As long as we correctly implement the WASM semantics for all types (in this case I32), then we can do whatever we want. 

I think the 64bit x86 supports using 32-bit values, so we should use it I think.

src/libasr/codegen/wasm_to_x64.cpp
certik: Beautiful!

src/libasr/codegen/wasm_to_x64.cpp
Shaikh-Ubaid: Got it. Thank you.

Shaikh-Ubaid:
@certik, please, could you possibly share if this is fine to merge?

certik (approved):

Shaikh-Ubaid:

PR #1340, opened 2022-12-02
Update python tests for pointers in emulation mode

czgdp1807:
Continues https://github.com/lcompilers/lpython/pull/1336


czgdp1807:
@certik This is ready. I enable CPython tests for a bunch of tests. The only thing left is `structs_13` which I will do only when a use case arises for emulation mode.

certik (approved):
Looks great, thanks!

PR #1345, opened 2022-12-02
[ASR Pass]: Fix List Concat and List Section

### LINKED ISSUE #1343: BUG: [ASR pass] List Concat
```py
from ltypes import i32

def f():
    a: list[i32]
    b: list[i32]
    a = [1, 2]
    b = [3, 4]
    c: list[str]
    d: list[str]
    c = ['a', 'b']
    d = ['c', 'd', 'e']
    c += d
    a += b
    i: i32
    for i in range(len(c)):
        print(c[i])
    for i in range(len(a)):
        print(a[i])
f()
```
The current implementation of the `ListConcat` node in ASR is handled using `pass_list_expr.cpp`. The bug here is that it only handles `ListConcat` of a single list type. As seen in the above example, once the `_lcompilers_list_concat` function is inserted in the symbol table, all other `ListConcat` nodes will also use the same function. In the above example, `_lcompilers_list_concat` was created for `list[str]` and it was used for `list[i32]` which will fail in LLVM.

The bug can be verified using: `% lpython --pass=pass_list_expr a.py --show-asr --indent`

For handling this we should create a map of list types and their respective function names.

### LINKED ISSUE #1347: [BUG]: List Section doesn't work with `list[str]`
```py
def f():
    c: list[str]
    c = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h']
    d: list[str]
    d = c[3:6]
    assert d[0] == 'd'
    assert d[1] == 'e'
    assert d[2] == 'f'
    assert len(d) == 3
f()
```
Error:
```
Both operands to ICmp instruction are not of the same type!
  %27 = icmp sgt i8* %24, i32 %26
  store i32 %29, i8** %end, align 4
 i8*Stored value type does not match pointer operand type!
  store i8* %30, i32* %__1_k, align 8
 i32Both operands to ICmp instruction are not of the same type!
  %33 = icmp sgt i8* %31, i32 %32
  %67 = add i32 %65, i8* %66
  %call_arg_value = alloca i32, align 4
 i8**  %53 = call %list @_lcompilers_list_section(%list* %c, i32* %call_arg_value, i32* %call_arg_value1, i32* %call_arg_value2)
```

Smit-create:
Fixes #1343
Fixes #1347

### DIFF
--- a/integration_tests/test_list_09.py
+++ b/integration_tests/test_list_09.py
@@ -34,5 +34,15 @@ def test_list_concat():
     y.clear()
     for i in range(1, 100):
         assert z[i-1] == i
+    
+    c: list[str]
+    d: list[str]
+    c = ['a', 'b']
+    d = ['c', 'd', 'e']
+    c += d
+    assert len(c) == 5
+    for i in range(5):
+        assert ord(c[i]) - ord('a') == i
+
 
 test_list_concat()
--- a/integration_tests/test_list_10.py
+++ b/integration_tests/test_list_10.py
@@ -38,5 +38,15 @@ def test_list_section():
     for i in range(0, len(x), 2):
         assert x[i] == y[j]
         j += 1
+    
+    c: list[str]
+    c = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h']
+    d: list[str]
+    d = c[3:6]
+    assert d[0] == 'd'
+    assert d[1] == 'e'
+    assert d[2] == 'f'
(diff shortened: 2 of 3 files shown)


Thirumalai-Shaktivel:
LGTM, thanks for fixing this!

certik (approved):
I think this is fine. I don't like generally using strings inside the compiler as they are slow, unless we have to. But I think we can iterate on this, so it's fine for now.

Issue #1333, opened 2022-11-30
Feature request: an equivalent of "char        buffer[64];" in a struct


certik:
Right now if you do:
```
@dataclass
class some_struct:
    buffer: i8[64]
```
It will produce an array with a descriptor. But rather, we need it to produce `char buffer[64]`, that is, 64 bytes of memory. It could be that we can use a descriptor by default, but if we do:
```
@ccallable
@dataclass
class some_struct:
    buffer: i8[64]
```
Then it would use the array directly.

czgdp1807:
For both C and LLVM backends, correct? And specifically for `i8` type?

certik:
Yes, I would always do it for the LLVM backend to ensure the design is good. This we also need for the C backend. Yes, we need it for i8, and if we needed larger ones later, I think we could "emulate" them using i8.

certik:
This is now a pretty high priority, we need to figure out a fix.

czgdp1807:
The underlying issue is https://github.com/lcompilers/lpython/issues/1350. If we figure out a decent protocol then solving this issue is a by-product.

certik:
The issue #1350 is general, and we need to figure out a solution.

In the meantime to unblock us, the issue is actually quite simple. Regular struct I think should not be BindC and we should have the freedom to rearrange or represent members (arrays) any way we like. In this particular case, when interoperability with C is needed, we should declare it as:
```
@ccallable
@dataclass
class some_struct:
    buffer: i8[64]
```
And then declare all members, such as `buffer`, as BindC. This then in return prescribes that it must be represented by a pointer, and in this case the memory is in the struct itself, 64 bytes, just like in C.

czgdp1807:
Makes sense. Will do so, once I come back from vacation.

PR #1346, opened 2022-12-02
Create "char buffer[...]" for "i8[...]" typed arrays

czgdp1807:
@certik Note that avoiding descriptors is impossible (as far as I can think) for "i8[:]" annotation, because we need to store lower and upper bounds. Hence I implemented this feature only for "i8[some_compile_time_dimension]".

### DIFF
--- a/src/libasr/codegen/asr_to_c.cpp
+++ b/src/libasr/codegen/asr_to_c.cpp
@@ -147,7 +147,13 @@ public:
                 std::string mem_var_name = current_scope->get_unique_name(itr.first + std::to_string(counter));
                 counter += 1;
                 sub += indent + convert_variable_decl(*mem_var, true, true, true, true, mem_var_name) + ";\n";
-                sub += indent + name + "->" + itr.first + " = " + mem_var_name + ";\n";
+                if( mem_var->m_abi == ASR::abiType::BindC &&
+                    ASR::is_a<ASR::Integer_t>(*mem_var->m_type) &&
+                    ASRUtils::extract_kind_from_ttype_t(mem_var->m_type) == 1 ) {
+                    sub += indent + "strcpy(" + name + "->" + itr.first + ", " + mem_var_name + ");\n";
+                } else {
+                    sub += indent + name + "->" + itr.first + " = " + mem_var_name + ";\n";
+                }
             } else if( ASR::is_a<ASR::Struct_t>(*mem_type) ) {
                 ASR::Struct_t* struct_t = ASR::down_cast<ASR::Struct_t>(mem_type);
                 ASR::StructType_t* struct_type_t = ASR::down_cast<ASR::StructType_t>(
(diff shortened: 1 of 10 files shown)


certik (approved):
I only reviewed quickly, it seems good to me. I have not reviewed every detail, but if it looks good to you as well, then go ahead and merge. We can fix it later if something doesn't work.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1747,10 +1759,15 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-            if (ASRUtils::expr_type(x.m_v)->type == ASR::ttypeType::Pointer) {
+            int a_kind = ASRUtils::extract_kind_from_ttype_t(x_mv_type);
+            bool is_i8_array = (ASR::is_a<ASR::Integer_t>(*x_mv_type) && a_kind == 1 &&
+                                ASRUtils::expr_abi(x.m_v) == ASR::abiType::BindC);
+            if (ASR::is_a<ASR::Pointer_t>(*x_mv_type) ||
+               (is_i8_array && ASR::is_a<ASR::StructInstanceMember_t>(*x.m_v))) {
                 array = CreateLoad(array);
             }
             bool is_data_only = is_argument && !ASRUtils::is_dimension_empty(m_dims, n_dims);
+            is_data_only = is_data_only || is_i8_array;
```
certik, line 1770: In here I am not sure we should special case i8. I think there is nothing special about i8, we should have exactly the same handling for all integer (and real and complex) arrays.

src/libasr/codegen/asr_to_c.cpp
```diff
@@ -238,16 +244,26 @@ class ASRToCVisitor : public BaseCCPPVisitor<ASRToCVisitor>
...
-                                        use_ref, dummy,
-                                        (v.m_intent != ASRUtils::intent_in &&
-                                        v.m_intent != ASRUtils::intent_inout &&
-                                        v.m_intent != ASRUtils::intent_out &&
-                                        !is_struct_type_member) || force_declare, is_fixed_size);
+                    if( t->m_kind == 1 && v.m_abi == ASR::abiType::BindC && is_fixed_size ) {
+                        if( !force_declare ) {
+                            force_declare_name = std::string(v.m_name);
+                        }
+                        sub = "char " + force_declare_name + dims;
```
certik, line 251: Why not using `int8_t` ? It seems that's cleaner and consistent with i16, i32, i64, etc.

czgdp1807: I think in the issue it was mentioned to create `char buffer[64]` for `buffer: i8[64]`.

src/libasr/codegen/asr_to_llvm.cpp
czgdp1807: Well, I guess the requirement was to create data only array for i8 if its a BindC. See https://github.com/lcompilers/lpython/issues/1333

czgdp1807:
> I reviewed it commit by commit, it's clean, but I don't like the special casing for i8.

+1. I also didn't like it while implementing this. However from what I understood by reading https://github.com/lcompilers/lpython/issues/1333, and in https://github.com/lcompilers/lpython/issues/1333#issuecomment-1332779425, I thought it was needed only for `i8` typed arrays.

My suggestion would be to always create data only arrays for fixed sizes and use descriptors otherwise.

certik:
See https://github.com/lcompilers/lpython/issues/1350 to discuss this more.

(There is nothing special about i8.)

PR #1351, opened 2022-12-07
Test for tuple return and fixes in ``CPtrToPointer`` node in C backend

czgdp1807:


certik (approved):
Awesome, looks great! Thanks for fixing it so quickly!

Issue #1316, opened 2022-11-21
Modify the handling of exit to throw the correct code


Thirumalai-Shaktivel:
I recently came across something like this:
```py
def test():
	quit(10)

test()
```
```console
$ lpython a.py
err = 2560
STOP
$ echo $?
1
```
So, I thought to replace this with the following change:
```suggestion
                if (err != 0) {
                    if (0 < err && err < 256) {
                        return err;
                    } else if (err % 256 == 0) {
                    	return err / 256;
                    } else {
                        return 1;
                    }
                }
```
@certik

_Originally posted by @Thirumalai-Shaktivel in https://github.com/lcompilers/lpython/pull/1302#discussion_r1027038771_

PR #1352, opened 2022-12-10
Modify the handling of exit code in lpython.cpp

Thirumalai-Shaktivel:
Closes: #1316

### DIFF
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -1649,7 +1649,7 @@ int main(int argc, char *argv[])
                     err = system(("js " + outfile +".js").c_str());
                 } else {
                     if (compiler_options.platform == LFortran::Platform::Windows) {
-                        err = system(outfile.c_str());
+                        return system(outfile.c_str());
                     } else {
                         err = system(("./" + outfile).c_str());
                     }
@@ -1658,7 +1658,9 @@ int main(int argc, char *argv[])
                     if (0 < err && err < 256) {
                         return err;
                     } else {
-                        return 1;
+                        // https://stackoverflow.com/a/27117435/15913193
+                        // https://linux.die.net/man/3/system
+                        return LFortran::get_exit_status(err);
                     }
                 }
             }
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -1658,8 +1658,6 @@ int main(int argc, char *argv[])
                     if (0 < err && err < 256) {
                         return err;
                     } else {
-                        // https://stackoverflow.com/a/27117435/15913193
-                        // https://linux.die.net/man/3/system
                         return LFortran::get_exit_status(err);
                     }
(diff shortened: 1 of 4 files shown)


src/bin/lpython.cpp
```diff
@@ -1657,6 +1657,8 @@ int main(int argc, char *argv[])
                 if (err != 0) {
                     if (0 < err && err < 256) {
                         return err;
+                    } else if (err % 256 == 0) {
+                        return err / 256;
```
certik, line 1661: What does this do exactly? Let's document it in a comment.

Thirumalai-Shaktivel: In Linux, the exit values are multiplied by 256
https://stackoverflow.com/a/3736358
unix.stackexchange.com/questions/394639/why-do-high-exit-codes-on-linux-shells-256-not-work-as-expected
So, when we do:
```py
quit(10)
```
```
>> echo $?
2560
```

src/bin/lpython.cpp
```diff
@@ -1658,7 +1658,9 @@ int main(int argc, char *argv[])
                     if (0 < err && err < 256) {
                         return err;
                     } else {
-                        return 1;
+                        // https://stackoverflow.com/a/27117435/15913193
+                        // https://linux.die.net/man/3/system
+                        return WEXITSTATUS(err);
```
certik, line 1663: This does not compile on Windows:
```
D:\a\lpython\lpython\lpython-0.8.0-14-gb2d00ab00\src\bin\lpython.cpp(1663): error C3861: 'WEXITSTATUS': identifier not found
```

Thirumalai-Shaktivel: I'm Sorry, @certik. I forgot the suggestions you gave on this.
Would this: 0e9995174cc1e3cdf8b4363f630f27414dc22fe5 be a valid fix

src/bin/lpython.cpp
```diff
@@ -1658,7 +1658,9 @@ int main(int argc, char *argv[])
                     if (0 < err && err < 256) {
                         return err;
                     } else {
-                        return 1;
+                        // https://stackoverflow.com/a/27117435/15913193
+                        // https://linux.die.net/man/3/system
```
certik, line 1662: Move these two links into the `utils.cpp` file to document `get_exit_status`?

certik (approved):
I think this looks good, I just left some minor comments.

src/lpython/utils.h
```diff
@@ -12,6 +12,8 @@ std::string get_runtime_library_header_dir();
 bool is_directory(std::string path);
 bool path_exists(std::string path);
 
+int32_t get_exit_status(int32_t err);
```
certik, line 15: Document here what the function does, you can put the links above here as well.

Thirumalai-Shaktivel:

PR #1354, opened 2022-12-10
WASM_X64: Fix asm text format

### LINKED ISSUE #1353: WASM_X64: `textual` `64-bit` assembly generated segment faults
The `textual` `64-bit` assembly format that is currently generated seems to `segment fault`. I think we can fix in some other `PR`.

_Originally posted by @Shaikh-Ubaid in https://github.com/lcompilers/lpython/issues/1338#issuecomment-1334710796_

Shaikh-Ubaid:
closes #1353

### DIFF
--- a/src/libasr/codegen/wasm_to_x64.cpp
+++ b/src/libasr/codegen/wasm_to_x64.cpp
@@ -208,7 +208,9 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
     void gen_x64_bytes() {
         {   // Initialize/Modify values of entities for code simplicity later
 
-            m_a.update_asm("BITS 64\n\n"); // Update initial value of asm text as per X64 text format
+            m_a.update_asm(""); // delete previous asm code
+            m_a.append_asm("BITS 64\n");
+            m_a.append_asm("\torg " + i2s((uint64_t)m_a.origin()) + "\n\n"); // specify origin info
             exports.back().name = "_start"; // Update _lcompilers_main() to _start
         }
 
--- a/src/libasr/codegen/wasm_to_x64.cpp
+++ b/src/libasr/codegen/wasm_to_x64.cpp
@@ -134,7 +134,7 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
             localidx -= no_of_params;
             std::string var_type = var_type_to_string[codes[cur_func_idx].locals[localidx].type];
             if (var_type == "i32") {
-                m_a.asm_mov_r64_m64(X64Reg::rax, &base, nullptr, 1, -8 * (1 + localidx));
+                m_a.asm_mov_r64_m64(X64Reg::rax, &base, nullptr, 1, -8 * (1 + (int)localidx));
                 m_a.asm_push_r64(X64Reg::rax);
             } else if (var_type == "f64") {
                 std::cerr << "Floats are not yet supported" << std::endl;
@@ -163,7 +163,7 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
             std::string var_type = var_type_to_string[codes[cur_func_idx].locals[localidx].type];
             if (var_type == "i32") {
                 m_a.asm_pop_r64(X64Reg::rax);
-                m_a.asm_mov_m64_r64(&base, nullptr, 1, -8 * (1 + localidx), X64Reg::rax);
+                m_a.asm_mov_m64_r64(&base, nullptr, 1, -8 * (1 + (int)localidx), X64Reg::rax);
             } else if (var_type == "f64") {
                 std::cerr << "Floats are not yet supported" << std::endl;
             } else {
--- a/src/libasr/codegen/wasm_to_x64.cpp
+++ b/src/libasr/codegen/wasm_to_x64.cpp
@@ -206,11 +206,7 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
     }
 
     void gen_x64_bytes() {
-        {   // Initialize/Modify values of entities for code simplicity later
-
-            m_a.update_asm(""); // delete previous asm code
-            m_a.append_asm("BITS 64\n");
-            m_a.append_asm("\torg " + i2s((uint64_t)m_a.origin()) + "\n\n"); // specify origin info
+        {   // Initialize/Modify values of entities
             exports.back().name = "_start"; // Update _lcompilers_main() to _start
         }
 
@@ -257,7 +253,7 @@ Result<int> wasm_to_x64(Vec<uint8_t> &wasm_bytes, Allocator &al,
     int time_save = 0;
     int time_verify = 0;
 
-    X86Assembler m_a(al);
+    X86Assembler m_a(al, true /* bits 64 */);
 
     wasm::X64Visitor x64_visitor(m_a, al, diagnostics, wasm_bytes);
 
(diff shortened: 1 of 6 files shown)


Shaikh-Ubaid:
**Example:**
```python
$ cat examples/expr2.py 
def main0():
    print(25)

main0()
```
**Output:**
```assembly
BITS 64
        org 0x0000000008048000

ehdr:
    db 0x7f
    db 0x45
    db 0x4c
    db 0x46
    db 0x02
    db 0x01
    db 0x01
    db 0x00
    db 0x00
    db 0x00
    db 0x00
    db 0x00
    db 0x00
    db 0x00
    db 0x00
    db 0x00
    dw 0x0002
    dw 0x003e
    dd 0x00000001
    dq _start
    dq e_phoff
    dq 0x0000000000000000
    dd 0x00000000
    dw ehdrsize
    dw phdrsize
    dw 0x0001
    dw 0x0000
    dw 0x0000
    dw 0x0000

ehdrsize equ 0x00000040

phdr:
    dd 0x00000001
    dd 0x00000005
    dq 0x0000000000000000
    dq 0x0000000008048000
    dq 0x0000000008048000
    dq filesize
    dq filesize
    dq 0x0000000000001000

phdrsize equ 0x00000038


e_phoff equ 0x0000000000000040

_lpython_main_program:
    push rbp
    mov rbp, rsp
    mov rax, 0x0000000000000000
    call main0
    mov rsp, rbp
    pop rbp
    ret
main0:
    push rbp
    mov rbp, rsp
    mov rax, 0x0000000000000000
    mov rax, 0x0000000000000019
    push rax
    pop rdi
    mov rax, 0x000000000000003c
    syscall
    mov rsp, rbp
    pop rbp
    ret
_start:
    push rbp
    mov rbp, rsp
    mov rax, 0x0000000000000000
    call _lpython_main_program
    mov rax, 0x0000000000000000
    push rax
    pop rdi
    mov rax, 0x000000000000003c
    syscall
    mov rsp, rbp
    pop rbp
    ret


filesize equ $ - $$
```

**On Execution:**
```bash
$ nasm -f bin main_x64.asm
$ chmod +x main_x64
$ ./main_x64 
$ echo $?
25
$ 
```

Shaikh-Ubaid:
`filesize` and `e_entry` are two variables that seems to be affected when the instruction(s) get encoded in size other than `64-bit`. Therefore, in this `PR` we assign value to these variables appropriately. For `e_entry`, we hard-code the `_start` label in the `asm` text format. For `filesize` we let it to be computed by `nasm`.

src/libasr/codegen/x86_assembler.h
```diff
@@ -345,6 +345,10 @@ class X86Assembler {
         return m_asm_code;
     }
 
+    void append_asm(std::string asm_code) {
+        m_asm_code += asm_code;
+    }
+
```
Shaikh-Ubaid, line 351: The `X86Assembler` already has an `emit()` function which adds text to `m_asm_code`, but it is defined as private so we are unable to call it using the assembler object `m_a`. I am unsure if we should make it public. I currently defined this `append_asm()` to allow adding text to `m_asm_code`.

Shaikh-Ubaid:

src/libasr/codegen/wasm_to_x64.cpp
```diff
@@ -109,7 +109,9 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
     void gen_x64_bytes() {
         {   // Initialize/Modify values of entities for code simplicity later
 
-            m_a.update_asm("BITS 64\n\n"); // Update initial value of asm text as per X64 text format
+            m_a.update_asm(""); // delete previous asm code
+            m_a.append_asm("BITS 64\n");
+            m_a.append_asm("\torg " + i2s((uint64_t)m_a.origin()) + "\n\n"); // specify origin info
```
certik, line 114: Instead, let's change this:
```
    X86Assembler(Allocator &al) : m_al{al} {
```
To:
```
    X86Assembler(Allocator &al, bool bits64) : m_al{al} {
```
where `bits64` is true for 64 bit and false for 32 bit.

src/libasr/codegen/x86_assembler.cpp
```diff
@@ -256,12 +256,14 @@ void emit_elf64_header(X86Assembler &a, uint32_t p_flags) {
...
+    a.add_var64("e_phoff", a.get_defined_symbol("phdr").value-a.origin());
 }
 
 void emit_elf64_footer(X86Assembler &a) {
-    a.add_var("e_entry", a.get_defined_symbol("_start").value);
-    a.add_var("filesize", a.pos()-a.origin());
+    a.define_symbol("filesize", a.pos()-a.origin());
+    a.append_asm("\n\nfilesize equ $ - $$"); // $ refers current address, $$ refers starting address
+    // a.add_var64("e_entry", a.get_defined_symbol("_start").value);
+    // a.add_var64("filesize", a.pos()-a.origin());
```
certik, line 266: The last two lines can be just deleted.

src/libasr/codegen/x86_assembler.h
certik: I would remove append_asm. I would put all the header right underneath of:
```
        m_asm_code = "BITS 32\n\n";
```

src/libasr/codegen/x86_assembler.cpp
```diff
@@ -256,12 +256,14 @@ void emit_elf64_header(X86Assembler &a, uint32_t p_flags) {
...
     a.add_var("phdrsize", a.pos()-a.get_defined_symbol("phdr").value);
-    a.add_var("e_phoff", a.get_defined_symbol("phdr").value-a.origin());
+    a.add_var64("e_phoff", a.get_defined_symbol("phdr").value-a.origin());
 }
 
 void emit_elf64_footer(X86Assembler &a) {
-    a.add_var("e_entry", a.get_defined_symbol("_start").value);
-    a.add_var("filesize", a.pos()-a.origin());
+    a.define_symbol("filesize", a.pos()-a.origin());
+    a.append_asm("\n\nfilesize equ $ - $$"); // $ refers current address, $$ refers starting address
```
certik, line 264: We can create a function:
```
a.add_var_size("filesize")
```
Which would emit `filesize equ $ - $$` in string assembly, but an equivalent of `a.add_var("filesize", a.pos()-a.origin());`.

certik: Here is how it can be implemented:
```c++
#    define EMIT_VAR_SIZE(a) emit("\n", a + " equ $ - $$\n")

    void add_var_size(const std::string &var) {
        uint64_t val = pos() - origin();
        define_symbol(var, val);
        EMIT_VAR_SIZE(var);
    }
```

src/libasr/codegen/x86_assembler.h
Shaikh-Ubaid: Yes, got it. I removed the `append_asm()` function.

src/libasr/codegen/wasm_to_x64.cpp
Shaikh-Ubaid: Got it. I added the `64bits` parameter to `X86Assembler()` constructor.

src/libasr/codegen/x86_assembler.cpp
Shaikh-Ubaid: Yes, removed.

src/libasr/codegen/x86_assembler.cpp
Shaikh-Ubaid: Got it. Thank you for the guidance. I added this `add_var_size()` function.

Shaikh-Ubaid:

certik (approved):
Looks good to me now.

Issue #1337, opened 2022-12-01
i16 and i8 do not print negative values correctly


certik:
```python
from ltypes import i8, i16, i32, i64
u: i64 = i64(-18)
print(u)
x: i32 = i32(-18)
print(x)
y: i16 = i16(-18)
print(y)
z: i8 = i8(-18)
print(z)
```
This prints:
```console
$ PYTHONPATH=../src/runtime/ltypes python a.py
-18
-18
-18
-18
$ lpython a.py
-18
-18
65518
238
```

PR #1355, opened 2022-12-12
Fix printing of ``i8`` and ``i16`` integers in LLVM

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1337

In this PR,

```python
from ltypes import i8, i16, i32, i64
u: i64 = i64(-18)
print(u)
x: i32 = i32(-18)
print(x)
y: i16 = i16(-18)
print(y)
z: i8 = i8(-18)
print(z)
```

**Output**

```bash
-18
-18
-18
-18
```

Also note https://github.com/lcompilers/lpython/issues/1356 is opened separately as its a parser issue.


certik (approved):
Awesome, thanks!

PR #1357, opened 2022-12-12
Enable C test for ``test_math.py``

### LINKED ISSUE #1246: C: Segmentation fault on generated test_math.c
I am trying to compile and run the generated C code from `integrations/test_math.py`. LPython generates the code successfully but running the generated code after compilation gives a Segmentation fault. Debugging more into this found the following issue:
The generated code contains:
```cpp
double log(double x)
{
    double _lpython_return_variable;
    _lpython_return_variable = _lfortran_dlog(x);
}
```

And the `_lfortran_dlog` in `lfortran_intrinsincs.c` also uses a function `log` from `math.h`
``` cpp
LFORTRAN_API double _lfortran_dlog(double x)
{
    return log(x);
}
```

I suppose there is an infinite loop of function calls between these functions and hence may lead to a segmentation fault.

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1246

Related discussion - https://github.com/lcompilers/lpython/pull/1243#discussion_r1008810444


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4742,13 +4742,30 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                         ASRUtils::expr_value(enum_member_variable->m_symbolic_value));
             } else if (ASR::is_a<ASR::Module_t>(*t)) {
                 ASR::Module_t *m = ASR::down_cast<ASR::Module_t>(t);
-                std::string sym_name = value + "@" + x.m_attr;
+                std::string sym_name = value + "_" + x.m_attr + "_external__";
                 ASR::symbol_t *sym = current_scope->resolve_symbol(sym_name);
```
Smit-create, line 4746: Should we insert this in the global scope itself?

src/runtime/math.py
```diff
@@ -687,6 +688,7 @@ def log1p(x: f64) -> f64:
     return log(1.0 + x)
 
 
+@overload
```
Smit-create, line 691: This is kind of hacky and shouldn't we use some name tweaks at the ASR level?

czgdp1807: I guess overload just does that. We can do it by hand but that will be needed only for `fmod` and `log` and nothing else. So ASR name tweaks will be hackish as well. The problem is the `log` which is called in `lfortran_intrinsics.h` is defined in our generated C code as well and hence gets linked to our definition instead of the one in `cmath`. That’s what causes infinite call loop leading to segmentation fault. Now `overload` automatically does name mangling and hence linker correctly links because now `log` is not present instead its as present as a mangled name. Doing the same mangling by hand in ASR is more hackish. Also overloaded  `log`/`fmod` are harmless as far as I can see.

czgdp1807: The other way is to do it for every imported symbol. For example, `from math import log, fabs` should be stored as `math_log__external_`, `math_fabs__external_`. Then it will be uniform for every imported symbol, so no hacks then. However, I am not sure that's worth the effort. @certik Do you have any other approach in mind?

src/lpython/semantics/python_ast_to_asr.cpp
czgdp1807: Not needed IMO. Otherwise it would get declared scope in the C code as well and will mix things up with other scopes as well.

src/runtime/math.py
Smit-create: > The other way is to do it for every imported symbol. For example, `from math import log, fabs` should be stored as `math_log__external_`, `math_fabs__external_`. Then it will be uniform for every imported symbol, so no hacks then

This sounds better to me. The reason is that the users might not be aware of this and may end up writing a function without an `overload` decorator (and it's quite intuitive not to use `overload` for a single function definition.) We need to figure out a generalized approach to handle this and the above might be the one.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -451,6 +451,22 @@ R"(#include <stdio.h>
...
+                ASR::symbol_t* var_sym = item.second;
+                if( ASR::is_a<ASR::ExternalSymbol_t>(*var_sym) ) {
+                    ASR::ExternalSymbol_t* v_ext = ASR::down_cast<ASR::ExternalSymbol_t>(var_sym);
+                    ASR::symbol_t* v_sym = v_ext->m_external;
+                    if (ASR::is_a<ASR::Variable_t>(*v_sym)) {
+                        ASR::Variable_t* v = ASR::down_cast<ASR::Variable_t>(v_sym);
+                        if( v->m_symbolic_value ) {
+                            this->visit_expr(*v->m_symbolic_value);
+                            decl += indent + "#define " + std::string(v_ext->m_name) + " " + src + "\n";
+                            src.clear();
```
certik, line 464: I think we have to be careful here.

What happens if `v_ext->m_name` is equal to "x" and then we have another variable like "expr`, will the `x` in it be substituted? I think it will be, so we do not want to do that. Also we do not want to have any such macro defines in header files. We currently do not generate them, but eventually we will.

I would not use C macros.

czgdp1807: > I would not use C macros.

So constant C variables should work for this case?

certik: Either that, or we can initialize the global variables when the program starts. I think both seems better than macros.

src/runtime/math.py
czgdp1807: I did a more sensible change (by implementing log for f32, fmod for (f32, f32) and log1p for f32). So the overloads make sense now. 



> The other way is to do it for every imported symbol. For example, `from math import log, fabs` should be stored as `math_log__external_`, `math_fabs__external_`.

This is not possible actually because call names won't resolve correctly. This is a problem on our compiler's end and I fixed it by implementing, log and fmod correctly.

certik:
This needs some work. There might be good ideas here.

PR #1359, opened 2022-12-13
[LLVM]: Implement Tuple/List Compare

### LINKED ISSUE #1245: ASR: [BUG] Tuple/List Compare
```py
% cat a.py
from ltypes import i32

def f():
    a: tuple[i32, i32]
    b: tuple[i32, i32]
    a = (1, 2)
    b = (1, 2)
    assert a == b
f()
```

LPython generates the ASR but looks incorrect:
```
(Assert 
                        (Var 2 b) 
                        ()
                    )
```

And, therefore it fails in the backend:
```
Branch condition is not 'i1' type!
  br %tuple %16, label %then, label %else
  %16 = load %tuple, %tuple* %b, align 4
```

The same code works in CPython.

Smit-create:
Fixes #1245 
Completes https://github.com/lcompilers/lpython/pull/1342#issuecomment-1345525084

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1671,6 +1671,21 @@ public:
         list_api->list_clear(plist);
     }
 
+    void visit_TupleCompare(const ASR::TupleCompare_t& x) {
+        int64_t ptr_loads_copy = ptr_loads;
+        ptr_loads = 0;
+        this->visit_expr(*x.m_left);
+        llvm::Value* left = tmp;
+        this->visit_expr(*x.m_right);
+        llvm::Value* right = tmp;
+        ptr_loads = ptr_loads_copy;
+        tmp = llvm_utils->is_equal_by_value(left, right, *module,
+                ASRUtils::expr_type(x.m_left));
+        if (x.m_op == ASR::cmpopType::NotEq) {
+            tmp = builder->CreateNot(tmp);
+        }
+    }
+
     void visit_TupleLen(const ASR::TupleLen_t& x) {
         LFORTRAN_ASSERT(x.m_value);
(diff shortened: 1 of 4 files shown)


src/libasr/pass/pass_manager.h
```diff
@@ -129,7 +129,6 @@ namespace LCompilers {
                 "select_case",
                 "inline_function_calls",
                 "unused_functions",
-                "pass_compare"
```
czgdp1807, line 132: Probably this means we need backend specific list of passes. Because C/C++/Julia will need this pass but LLVM won't.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1671,6 +1671,21 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+        this->visit_expr(*x.m_left);
+        llvm::Value* left = tmp;
+        this->visit_expr(*x.m_right);
+        llvm::Value* right = tmp;
+        ptr_loads = ptr_loads_copy;
+        tmp = llvm_utils->is_equal_by_value(left, right, *module,
+                ASRUtils::expr_type(x.m_left));
+        if (x.m_op == ASR::cmpopType::NotEq) {
+            tmp = builder->CreateNot(tmp);
+        }
```
czgdp1807, line 1686: This seems good to me. Is this working?

Smit-create: Yes, `TupleCompare` works fine. I will need to add the `ListCompare` to make the CI pass.

Smit-create:
The tests pass locally for me as well as on ubuntu CI. Not sure why does that segfault on macos

Smit-create:
@czgdp1807 @certik This is ready for review.

PR #1362, opened 2022-12-15
Support defining Union inside a ``dataclass``

czgdp1807:


integration_tests/structs_16.py
```diff
@@ -0,0 +1,19 @@
+from ltypes import i32, i64, dataclass, union, Union
+
+@dataclass
+class A:
+    @union
+    class B(Union):
+        x: i32
+        y: i64
+    b: B
+    c: i32
```
certik, line 10: Perfect! While at this, given the new `StructStaticMember`, can you also please add a test for:
```python
@dataclass
class A:
    @dataclass
    class B(Union):
        x: i32
        y: i64
    b: B
    c: i32
```
Essentially a "struct in struct"? We will need it too sooner or later, so now is the time to make sure it works.

czgdp1807: Done. Please let me know if there is anything else to do here.

certik (approved):
I think this is fine. I haven't checked the verify() changes in detail, hopefully they are correct.

PR #1378, opened 2022-12-23
Order of declaration: Only include symbols if they are in scope

### LINKED ISSUE #1368: Segfault when instantiating an array that references a global for its size
Something changed between commit fc9d8996f415ecbddc2f7f314380f7ff132b09f2 and ca190303cff9bc7e90b5a2c013eaee5e18377cac. The following worked with the former but segfaults with the latter:

```python
from ltypes import i32

SIZE: i32 = i32(1)

def main() -> None:
    xs: i32[SIZE]

main()
```

```shell
$ lpython --show-c main.py
  Binary file "/home/dylon/Workspace/lpython/src/bin/lpython", in _start()
  File "/build/glibc-CVJwZb/glibc-2.27/csu/../csu/libc-start.c", line 310, in __libc_start_main()
  File "/home/dylon/Workspace/lpython/src/bin/lpython.cpp", line 1557, in main()
    return emit_c(arg_file, runtime_library_dir, compiler_options);
  File "/home/dylon/Workspace/lpython/src/bin/lpython.cpp", line 285, in (anonymous namespace)::emit_c(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, LFortran::CompilerOptions&)
    auto res = LFortran::asr_to_c(al, *asr, diagnostics,
  File "/home/dylon/Workspace/lpython/src/libasr/codegen/asr_to_c.cpp", line 1163, in LFortran::asr_to_c[abi:cxx11](Allocator&, LFortran::ASR::TranslationUnit_t&, LFortran::diag::Diagnostics&, LFortran::Platform&, long)
    v.visit_asr((ASR::asr_t &)asr);
  File "/home/dylon/Workspace/lpython/src/libasr/../libasr/asr.h", line 4199, in LFortran::ASR::BaseVisitor<LFortran::ASRToCVisitor>::visit_asr(LFortran::ASR::asr_t const&)
    void visit_asr(const asr_t &b) { visit_asr_t(b, self()); }
  File "/home/dylon/Workspace/lpython/src/libasr/../libasr/asr.h", line 4176, in void LFortran::ASR::visit_asr_t<LFortran::ASRToCVisitor>(LFortran::ASR::asr_t const&, LFortran::ASRToCVisitor&)
    case asrType::unit: { v.visit_unit((const unit_t &)x); return; }
  File "/home/dylon/Workspace/lpython/src/libasr/../libasr/asr.h", line 4200, in LFortran::ASR::BaseVisitor<LFortran::ASRToCVisitor>::visit_unit(LFortran::ASR::unit_t const&)
    void visit_unit(const unit_t &b) { visit_unit_t(b, self()); }
  File "/home/dylon/Workspace/lpython/src/libasr/../libasr/asr.h", line 3952, in void LFortran::ASR::visit_unit_t<LFortran::ASRToCVisitor>(LFortran::ASR::unit_t const&, LFortran::ASRToCVisitor&)
    case unitType::TranslationUnit: { v.visit_TranslationUnit((const TranslationUnit_t &)x); return; }
  File "/home/dylon/Workspace/lpython/src/libasr/codegen/asr_to_c.cpp", line 602, in LFortran::ASRToCVisitor::visit_TranslationUnit(LFortran::ASR::TranslationUnit_t const&)
    visit_symbol(*sym);
  File "/home/dylon/Workspace/lpython/src/libasr/../libasr/asr.h", line 4202, in LFortran::ASR::BaseVisitor<LFortran::ASRToCVisitor>::visit_symbol(LFortran::ASR::symbol_t const&)
    void visit_symbol(const symbol_t &b) { visit_symbol_t(b, self()); }
  File "/home/dylon/Workspace/lpython/src/libasr/../libasr/asr.h", line 3962, in void LFortran::ASR::visit_symbol_t<LFortran::ASRToCVisitor>(LFortran::ASR::symbol_t const&, LFortran::ASRToCVisitor&)
    case symbolType::Function: { v.visit_Function((const Function_t &)x); return; }
  File "/home/dylon/Workspace/lpython/src/libasr/../libasr/codegen/asr_to_c_cpp.h", line 441, in LFortran::BaseCCPPVisitor<LFortran::ASRToCVisitor>::visit_Function(LFortran::ASR::Function_t const&)
    if (ASR::is_a<ASR::Variable_t>(*var_sym)) {
  File "/home/dylon/Workspace/lpython/src/libasr/asr.h", line 31, in bool LFortran::ASR::is_a<LFortran::ASR::Variable_t, LFortran::ASR::symbol_t>(LFortran::ASR::symbol_t const&)
    return T::class_type == x.type;
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", in killpg()
```

If I make `SIZE` a local or replace it altogether with an int literal, it succeeds as before:

```python
from ltypes import i32

def main() -> None:
    SIZE: i32 = i32(1)
    xs: i32[SIZE]

main()
```

```shell
$ lpython --show-c main.py
#include <inttypes.h>

#include <stdlib.h>
#include <stdbool.h>
#include <stdio.h>
#include <string.h>

#define ASSERT(cond)                                                           \
    {                                                                          \
        if (!(cond)) {                                                         \
            printf("%s%s", "ASSERT failed: ", __FILE__);                       \
            printf("%s%s", "\nfunction ", __func__);                           \
            printf("%s%d%s", "(), line number ", __LINE__, " at \n");          \
            printf("%s%s", #cond, "\n");                                       \
    }
#define ASSERT_MSG(cond, msg)                                                  \
    {                                                                          \
        if (!(cond)) {                                                         \
            printf("%s%s", "ASSERT failed: ", __FILE__);                       \
            printf("%s%s", "\nfunction ", __func__);                           \
            printf("%s%d%s", "(), line number ", __LINE__, " at \n");          \
            printf("%s%s", #cond, "\n");                                       \
            printf("%s", "ERROR MESSAGE:\n");                                  \
            printf("%s%s", msg, "\n");                                         \
    }


struct dimension_descriptor
{
    int32_t lower_bound, length;
};

struct i32
{
    int32_t *data;
    struct dimension_descriptor dims[32];
    int32_t n_dims;
    bool is_allocated;
};


// Implementations
void _xx_lcompilers_changed_main_xx()
{
    int32_t SIZE = 1;
    struct i32 xs_value;
    struct i32* xs = &xs_value;
    int32_t *xs_data = (int32_t*) malloc(sizeof(int32_t)*SIZE);
    xs->data = xs_data;
    xs->n_dims = 1;
    xs->dims[0].lower_bound = 0;
    xs->dims[0].length = SIZE;
    SIZE = 1;
}

void _lpython_main_program()
{
    _xx_lcompilers_changed_main_xx();
}

{
    _lpython_main_program();
    return 0;
}
```

Smit-create:
Fixes #1368 

cc @czgdp1807


src/libasr/asr_utils.cpp
```diff
@@ -74,7 +74,8 @@ std::vector<std::string> determine_function_definition_order(
             ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(itr.second);
             for( size_t i = 0; i < func->n_dependencies; i++ ) {
                 std::string dep = func->m_dependencies[i];
-                deps.push_back(dep);
+                if (symtab->get_symbol(dep) != nullptr)
+                    deps.push_back(dep);
```
certik, line 78: I think the "contract" (checked by "verify") is that all the identifiers in dependencies are from the local symbol table, so then `symtab->get_symbol(dep)` should never be null. Can you post an example where this fails? It seems this PR is a workaround for some bug, that we should rather fix directly.

Smit-create: In this example:
```py
from ltypes import i32

SIZE: i32 = i32(1)

def main() -> None:
    xs: i32[SIZE]

main()
```

The ASR has:
```console
(SymbolTable
                        2
                        {
                            xs:
                                (Variable 
                                    2 
                                    xs 
                                    [SIZE] 
                                    Local 
                                    () 
                                    () 
                                    Default 
                                    (Integer 4 [((IntegerConstant 0 (Integer 4 [])) 
                                    (Var 1 SIZE))]) 
                                    Source 
                                    Public 
                                    Required 
                                    .false.
                                )
                            
                        }) 
```

We see that `xs` has `SIZE` as the dependent variable which is present in the parent symbol.

czgdp1807: I guess, having `SIZE` in dependencies is correct, after all it depends on it. Yes, but including it in declaration order doesn't make sense, because parent scope must have been already visited. I guess a better way would be to include `SIZE` in `main`'s (the function) dependency list, because we want to make sure `SIZE` is declared before `main`. What do you say @certik ?

certik: I see, thanks @Smit-create for the example. Now it make sense what is going on.

What is the best design?

The motivation for these dependencies is to ensure we can create the symbol table in the right order. Yes, things can depend on parent scope, but as Gagandeep said, the parent scope is implicitly always created first anyway.

One option is to treat dependencies as strictly "local" dependencies, only within the given symbol table. That is how it is used in the backend.  We can rename the argument to "local_dependencies". Any dependency on parent scope is implicit.

Then in the parent scope, there is also a dependency: `main` depends on `SIZE`.

How to easily figure out these dependencies? Not sure.

Another approach is to always store all dependencies, but even then we still need to add SIZE to main in the parent scope, otherwise main() could have been generated before SIZE.

There is another issue. One can have two mutually recursive functions that call each other. That is the reason why we need to do a two pass approach in AST->ASR. So they will both depend on each other. I guess that could still be encoded as a cyclic dependency.

Smit-create: I think we can only include the local dependencies with the given symbol table. What do you think @czgdp1807 ?

certik: I would only use dependencies for local symbols.

certik:
I think this is fine, let's add a test to make sure it actually generates the C code for the global variable.

certik (approved):
Thanks!

PR #1379, opened 2022-12-23
WASM_X64: Support printing integers

Shaikh-Ubaid:
**Example:**
```python
def Main0():
    x: i32
    x = (2+3)* (-5)
    y: i32 = 100
    z: i32 = 2147483647
    w: i32 = -2147483648
    print(x)
    print(y)
    print(z)
    print(w)
    print("Hi")
    print("Hello")
    print((5-2) * 7)
    print("Bye")

Main0()
```
**Binary Output:**
```bash
(lp) lpython$ lpython integration_tests/print_03.py  --backend wasm_x64 -o tmp
(lp) lpython$ ./tmp
-25
100
2147483647
-2147483648
Hi
Hello
21
Bye
(lp) lpython$ 
```

**Text Format Output:**
```bash
(wasm_asm) lpython$ nasm -fbin ./main.asm && chmod +x main && ./main
-25
100
2147483647
-2147483648
Hi
Hello
21
Bye
(wasm_asm) lpython$ 
```

### DIFF
--- a/src/libasr/codegen/wasm_to_x64.cpp
+++ b/src/libasr/codegen/wasm_to_x64.cpp
@@ -179,9 +179,6 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
     }
 
     void visit_I32Const(int32_t value) {
-        // direct addition of imm64 to stack is not available with us yet
-        // so temporarily using a combination of instructions
-        // TODO: Update this once we have support for push_imm64()
         m_a.asm_mov_r64_imm64(X64Reg::rax, abs(value));
         if (value < 0) m_a.asm_neg_r64(X64Reg::rax);
         m_a.asm_push_r64(X64Reg::rax);
--- a/src/libasr/codegen/wasm_to_x64.cpp
+++ b/src/libasr/codegen/wasm_to_x64.cpp
@@ -179,7 +179,7 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
     }
 
     void visit_I32Const(int32_t value) {
-        m_a.asm_mov_r64_imm64(X64Reg::rax, abs(value));
+        m_a.asm_mov_r64_imm64(X64Reg::rax, labs((int64_t)value));
         if (value < 0) m_a.asm_neg_r64(X64Reg::rax);
         m_a.asm_push_r64(X64Reg::rax);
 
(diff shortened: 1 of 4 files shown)


Shaikh-Ubaid:

src/libasr/codegen/wasm_to_x64.cpp
```diff
@@ -50,14 +50,8 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
...
-                std::cerr << "Call to print_i32() will be printed as exit code\n";
-
-                // Currently, for print we are setting the value to be printed
-                // as the exit code. Later we can access/view this value from console
-                // using: echo $?
-                m_a.asm_pop_r64(X64Reg::rdi); // get exit code from stack top
-                m_a.asm_mov_r64_imm64(X64Reg::rax, 60); // sys_exit
-                m_a.asm_syscall(); // syscall
+                m_a.asm_call_label("print_i64");
+                m_a.asm_pop_r64(X64Reg::r15); // pop the passed argument
```
certik, line 54: After this is merged, let's document what the "calling convention" is. It looks like the caller is responsible to put arguments on stack, as well as removing them from stack once the called function (`print_i64` in this case) finishes?

certik (approved):
I think this looks great! Thanks for this.

src/libasr/codegen/wasm_to_x64.cpp
Shaikh-Ubaid: Yes, the caller is responsible to put arguments on stack. 

In `wasm`, the arguments are already used up (that is there is not need to externally remove them) by the called function.

In assembly `x86` and/or `x64`, we are currently manually/explicitly removing the passed arguments in the `visit_Call()` function:
https://github.com/lcompilers/lpython/blob/7aaedcc9ff91994d8846d9810b35ec7fc8bdffda/src/libasr/codegen/wasm_to_x64.cpp#L97-L123

Shaikh-Ubaid: For the `print_i64`, as it is an imported function and as this is an initial support, we are removing the passed arguments just after the call:

https://github.com/lcompilers/lpython/blob/7aaedcc9ff91994d8846d9810b35ec7fc8bdffda/src/libasr/codegen/wasm_to_x64.cpp#L50-L56

Shaikh-Ubaid:

src/libasr/codegen/wasm_to_x64.cpp
certik: I think we should follow the WASM convention and the function should remove them, so in this case `print_i64` should remove the arguments from stack. What do you think?

PR #1382, opened 2022-12-24
Support data only array members for ``BindC`` structs

czgdp1807:
https://github.com/lcompilers/lpython/issues/1333#issuecomment-1359550823

Closes https://github.com/lcompilers/lpython/issues/1333

### DIFF
--- a/integration_tests/structs_14.py
+++ b/integration_tests/structs_14.py
@@ -1,5 +1,5 @@
-from ltypes import i8, dataclass, i32, f32, c32, f64, i16, i64, ccallable
-from numpy import empty, int8, int16, int32, int64, float32, complex64, float64
+from ltypes import i8, dataclass, i32, f32, c32, f64, i16, i64, c64, ccallable
+from numpy import empty, int8, int16, int32, int64, float32, complex64, complex128, float64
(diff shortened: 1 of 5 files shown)


certik (approved):
I think this looks god now. I would add some tests.

czgdp1807:
> I would add some tests.

`integration_tests/structs_14.py` already has some and I will add more there. Do you have any other use case which you want me to cover?

integration_tests/structs_14.py
```diff
@@ -1,32 +1,72 @@
...
+    buffer4: f64[32]
 
 @ccallable
 @dataclass
 class buffer_struct_clink:
     buffer: i8[32]
+    buffer1: i32[32]
+    buffer2: f32[32]
+    buffer3: c32[32]
+    buffer4: f64[32]
```
certik, line 20: Let's do also i16 and i64, those are commonly used. Let's do c64 too, might as well. Then everything is covered.

certik (approved):
I think that this is good. 

We can later see if there is a way to simplify some of these things, but for now this seems good enough and fixes the immediate problem.

Issue #1358, opened 2022-12-12
Implement basic command line handling (sys.argv)


certik:
We need to access the command line. In Python that is done via the `sys.argv` list. It seems it would be of type `List[str]`, we need to fill it with the command line arguments. Example:
```
from sys import argv
print(argv)
```

We should also make sure we can use such a list (e.g., `len`, indexing, string comparison, etc.), but those would be bugs in list support, so we would open separate issues if needed.

Thirumalai-Shaktivel:
I have a doubt. Consider an example
```py
x: list[i32] = [1]
print(x[0])
```
ASR
```py
(TranslationUnit 
    (SymbolTable
        1
        {
            _lpython_main_program:
                (Function 
                    (SymbolTable
                        3
                        {
                            
                        }) 
                    _lpython_main_program 
                    [] 
                    [] 
                    [(Print 
                        () 
                        [(ListItem 
                            (Var 1 x) 
                            (IntegerConstant 0 (Integer 4 [])) 
                            (Integer 4 []) 
                            ()
                        )] 
                        () 
                        ()
                    )] 
                    () 
                    Source 
                    Public 
                    Implementation 
                    () 
                    .false. 
                    .false. 
                    .false. 
                    .false. 
                    .false. 
                    [] 
                    [] 
                    .false.
                ), 
            main_program:
                (Program 
                    (SymbolTable
                        2
                        {
                            
                        }) 
                    main_program 
                    [] 
                    [(SubroutineCall 
                        1 _lpython_main_program 
                        () 
                        [] 
                        ()
                    )]
                ), 
            x:
                (Variable 
                    1 
                    x 
                    [] 
                    Local 
                    (ListConstant 
                        [(IntegerConstant 1 (Integer 4 []))] 
                        (List 
                            (Integer 4 [])
                        )
                    ) 
                    () 
                    Default 
                    (List 
                        (Integer 4 [])
                    ) 
                    Source 
                    Public 
                    Required 
                    .false.
                )
            
        }) 
    []
)
```
From the above ASR, the variable `x` must be included in the `_lpython_main_program` SymbolTable, right?
Is this a bug or we have not handled the global varible yet?

certik:
Yes global variables are not always handled correctly, that's a separate issue.

certik:
I would implement either:
```python
from sys import argv
def f():
    print(argv)
```
But if it doesn't work due to the global variable issue, then let's try this:
```python
def f():
    from sys import argv
[...]

PR #1383, opened 2022-12-25
Implement `argv` in `sys` to get command line arguments

Thirumalai-Shaktivel:
Related: https://github.com/lcompilers/lpython/issues/1358


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1165,3 +1165,24 @@ LFORTRAN_API void _lpython_close(int64_t fd)
...
+char **argv;
+
+LFORTRAN_API void _lpython_set_argv(int32_t argc_1, char **argv_1) {
+    argv = malloc(argc_1 * sizeof(char *));
+    for (size_t i = 0; i < argc_1; i++) {
+        argv[i] = strdup(argv_1[i]);
+    }
+    argc = argc_1;
+}
+__attribute__((section(".init_array"))) static void *ctr = &_lpython_set_argv;
```
Thirumalai-Shaktivel, line 1180: This doesn't seem to work in macOS and Windows.

certik: Just use regular C, not these `__attribute__` or stuff.

I think we either have to call some `libc` functions to access the arguments, or we need to emit this `_lpython_set_argv` function in the ASR->LLVM/C backends when the program starts.

certik: Looks like there is no easy way to do that: https://stackoverflow.com/questions/2471553/access-command-line-arguments-without-using-char-argv-in-main#24718544

So I recommend we simply get those arguments in `main` (both in C and in LLVM), and then call our function like `_lcompilers_set_argv`, to set these in our "internal structure", like you did above. That seems like a portable, clean approach that should work everywhere.

Thirumalai-Shaktivel:
The CI failure seems unrelated...

Thirumalai-Shaktivel:
It seems the C backend causes Segfault.

integration_tests/CMakeLists.txt
```diff
@@ -393,3 +393,5 @@ RUN(NAME func_static_01 LABELS cpython llvm c wasm)
 RUN(NAME func_static_02 LABELS cpython llvm c wasm)
 RUN(NAME func_dep_03    LABELS cpython llvm c)
 RUN(NAME func_dep_04    LABELS cpython llvm c)
+
+RUN(NAME test_argv_01    LABELS llvm)
```
certik, line 397: Also enable `cpython` test here.

integration_tests/test_argv_01.py
```diff
@@ -0,0 +1,12 @@
+from ltypes import i32
+from sys import argv
+
+def test():
+    exe: str = argv()[0]
```
certik, line 5: I think this has to be:
```suggestion
    exe: str = argv[0]
```

src/runtime/sys.py
```diff
@@ -6,3 +6,23 @@ def exit(error_code: i32):
     """
 
     quit(error_code)
+
+def argv() -> list[str]:
+    # TODO: use argv as a global `list[str]` variable
```
certik, line 11: I see, it's in the TODO list here.

certik:
This looks good. Now the issue is how to create the argv global variable and initialize it.

One option is to add a global statement like:
```
argv = ...
```

But I think this might not be working yet. In that case I recommend to finish this PR, rename `argv()` to `_lpython_argv()`, merge. Then we will work on the global statements. Then we can do `argv: list[str] = _lpython_argv()` or something like that, and that should work.

How does CPython's `sys.py` does this? Is it implemented in C?

integration_tests/test_argv_01.py
Thirumalai-Shaktivel: Exactly, but global `list` doesn't work yet.

Thirumalai-Shaktivel:
Yup, sys is implementes in C, See: https://stackoverflow.com/a/51580860/15913193

certik:
It's worth studying how they do it. They have this function here https://github.com/python/cpython/blob/91cdd75ba50eea4d0a608324e4a84afab8ae3fbd/Python/sysmodule.c#L2692 and functions above it, that set the global argv list. Are these called when the CPython interpreter/program starts every time?

certik (approved):
I think this is good to merge as is.

Then we'll improve upon it with subsequent PRs.

Thanks!

Thirumalai-Shaktivel:
Thanks for approving and merging this PR.

Issue #1376, opened 2022-12-23
List does not work in the global scope


Thirumalai-Shaktivel:
Consider an example
```py
x: list[i32] = [1]
print(x[0])
```
ASR
```py
(TranslationUnit 
    (SymbolTable
        1
        {
            _lpython_main_program:
                (Function 
                    (SymbolTable
                        3
                        {
                            
                        }) 
                    _lpython_main_program 
                    [] 
                    [] 
                    [(Print 
                        () 
                        [(ListItem 
                            (Var 1 x) 
                            (IntegerConstant 0 (Integer 4 [])) 
                            (Integer 4 []) 
                            ()
                        )] 
                        () 
                        ()
                    )] 
                    () 
                    Source 
                    Public 
                    Implementation 
                    () 
                    .false. 
                    .false. 
                    .false. 
                    .false. 
                    .false. 
                    [] 
                    [] 
                    .false.
                ), 
            main_program:
                (Program 
                    (SymbolTable
                        2
                        {
                            
                        }) 
                    main_program 
                    [] 
                    [(SubroutineCall 
                        1 _lpython_main_program 
                        () 
                        [] 
                        ()
                    )]
                ), 
            x:
                (Variable 
                    1 
                    x 
                    [] 
                    Local 
                    (ListConstant 
                        [(IntegerConstant 1 (Integer 4 []))] 
                        (List 
                            (Integer 4 [])
                        )
                    ) 
                    () 
                    Default 
                    (List 
                        (Integer 4 [])
                    ) 
                    Source 
                    Public 
                    Required 
                    .false.
                )
            
        }) 
    []
)
```
From the above ASR, the variable `x` must be included in the `_lpython_main_program` SymbolTable, right?
Is this a bug or we have not handled the global variable yet? or is this expected output?

```console
$ lpython examples/expr2.py
...
  File "/home/thirumalai/Open_Source/lpython/src/libasr/../libasr/asr.h", line 4150, in LFortran::ASR::BaseVisitor<LFortran::ASRToLLVMVisitor>::visit_symbol(LFortran::ASR::symbol_t const&)
    void visit_symbol(const symbol_t &b) { visit_symbol_t(b, self()); }
  File "/home/thirumalai/Open_Source/lpython/src/libasr/../libasr/asr.h", line 3919, in ??
    case symbolType::Variable: { v.visit_Variable((const Variable_t &)x); return; }
[...]

PR #1384, opened 2022-12-25
Initial support for global Lists

Thirumalai-Shaktivel:
Related: #1376


Thirumalai-Shaktivel:
Here I'm trying to move all the symbols in the global to the `_lpython_main_program` scope.
```py
x: list[i32] = [1,2,3,4]
print(x[0])
```
Here is the current ASR output:
```py
(TranslationUnit 
    (SymbolTable
        1
        {
            _lpython_main_program:
                (Function 
                    (SymbolTable
                        3
                        {
                            x:
                                (Variable 
                                    3 
                                    x 
                                    [] 
                                    Local 
                                    (ListConstant 
                                        [(IntegerConstant 1 (Integer 4 []))
                                        (IntegerConstant 3 (Integer 4 []))
                                        (IntegerConstant 4 (Integer 4 []))] 
                                        (List 
                                            (Integer 4 [])
                                        )
                                    ) 
                                    () 
                                    Default 
                                    (List 
                                        (Integer 4 [])
                                    ) 
                                    Source 
                                    Public 
                                    Required 
                                    .false.
                                )
                            
                        }) 
                    _lpython_main_program 
                    [] 
                    [] 
                    [(Print 
                        () 
                        [(ListItem 
                            (Var 3 x) 
                            (IntegerConstant 0 (Integer 4 [])) 
                            (Integer 4 []) 
                            ()
                        )] 
                        () 
                        ()
                    )] 
                    () 
                    Source 
                    Public 
                    Implementation 
                    () 
                    .false. 
                    .false. 
                    .false. 
                    .false. 
                    .false. 
                    [] 
                    [] 
                    .false.
                ), 
            main_program:
                (Program 
                    (SymbolTable
                        2
                        {
                            
                        }) 
                    main_program 
                    [] 
                    [(SubroutineCall 
                        1 _lpython_main_program 
                        () 
                        [] 
                        ()
                    )]
                )
            
        }) 
    []
)
```
Am I doing this correct? 
What are your thoughts? @certik @czgdp1807 @Smit-create

certik:
Something like this. I think for global statements we have a pass.

For global variables it gets tricky, if other functions access them. Not sure what the best solution there is.

Thirumalai-Shaktivel:
+1, This seems to be tricky to handle.
We already know that all the global statements will be included inside the `_lpython_main_program` (By this PR), so while looking into the global scope, we will look into the `_lpython_main_program` scope as well. 

@certik @czgdp1807 @Smit-create 
Can you please share your thoughts as well? Like Different possible ways to make LPython supports global scope or how to move forward with this PR.

certik:
We might need to discuss this over video. I think using TranslationUnit symbol table is fine. I think LLVM supports global variables. But we need to figure out all the details and dependencies.

certik:
Here is an example to consider:
```python
x: list[i32] = [1,2,3,4]
def f():
    print(x[1])
print(x[0])
f()
```

src/libasr/pass/global_stmts.cpp
```diff
@@ -100,6 +100,11 @@ void pass_wrap_global_stmts_into_function(Allocator &al,
                 throw LCompilersException("Unsupported type of global scope node");
             }
         }
+        std::vector<std::string> symbols;
+        unit.m_global_scope->move_symbols_from_global_scope(fn_scope, symbols);
+        for (auto &a: symbols) {
+            unit.m_global_scope->erase_symbol(a);
+        }
```
certik, line 107: I would not do this every time. Either as an option (via an argument), or even better, refactor this into a function and let the frontend call it. Add verify() in this refactored function.

certik: You can add this function into `global_stmts.h`, so there will be two separate functions: one moves the global statements/expressions, the other moves the global variables.

Thirumalai-Shaktivel:
The following example seems to work:
```
x: list[i32]
x = [1,2,3]
print(x, x[0])
```
But, I got some doubts:
We already support global declaration for int, float,... But the global declaration fails for the list (or maybe tuple, dict...). So, I was thought to leave int, float, ... declarations and move only those that fail in the global scope (like list).

@czgdp1807 @certik thoughts?

Thirumalai-Shaktivel:
We need to discuss more on this for further implementations.
Currently, we move the variables like list, dict, tuples into the function. After moving, we create an External symbol for that moved item and erase it.

I have a doubt, can we have an External symbol in TranslationUnit sym_table?
Does this violate the design?

I think the ASR seems correct, but the LLVM fails:
```py
x: list[i32]
x = [1,2,3]
print(x, x[0])

def main0():
    print(x[0])
```
<details>
<summary>ASR</summary>
```clojure
(TranslationUnit
    (SymbolTable
        1
        {
            _lpython_main_program:
                (Function
                    (SymbolTable
                        4
                        {
                            x:
                                (Variable
                                    4
                                    x
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (List
                                        (Integer 4 [])
                                    )
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
                        })
                    _lpython_main_program
                    [main0]
                    []
                    [(=
                        (Var 1 x)
                        (ListConstant
                            [(IntegerConstant 1 (Integer 4 []))
                            (IntegerConstant 2 (Integer 4 []))]
                            (List
                                (Integer 4 [])
                            )
                        )
                        ()
                    )
                    (Print
                        ()
                        [(ListItem
                            (Var 1 x)
                            (IntegerConstant 1 (Integer 4 []))
                            (Integer 4 [])
                            ()
                        )]
                        ()
                        ()
                    )
                    (SubroutineCall
                        1 main0
                        ()
                        []
                        ()
                    )]
                    ()
                    Source
                    Public
                    Implementation
                    ()
                    .false.
                    .false.
                    .false.
                    .false.
                    .false.
                    []
                    []
                    .false.
                ),
            main0:
                (Function
                    (SymbolTable
                        2
                        {
                            
                        })
                    main0
                    []
                    []
                    [(Print
                        ()
                        [(ListItem
                            (Var 1 x)
                            (IntegerConstant 1 (Integer 4 []))
                            (Integer 4 [])
                            ()
                        )]
                        ()
                        ()
                    )]
                    ()
                    Source
                    Public
                    Implementation
                    ()
                    .false.
                    .false.
                    .false.
                    .false.
                    .false.
                    []
                    []
                    .false.
                ),
            main_program:
                (Program
                    (SymbolTable
                        3
                        {
                            
                        })
                    main_program
                    []
                    [(SubroutineCall
                        1 _lpython_main_program
                        ()
                        []
                        ()
                    )]
                ),
            x:
                (ExternalSymbol
                    1
                    x
                    4 x
                    _lpython_main_program
                    []
                    x
                    Public
                )
        })
    _lpython_main_program
    []
)
```
</details>
```llvm
Instruction does not dominate all uses!
  %x = alloca %list, align 8
  %0 = getelementptr %list, %list* %x, i32 0, i32 0
Instruction does not dominate all uses!
  %x = alloca %list, align 8
  %4 = getelementptr %list, %list* %x, i32 0, i32 2
```

certik:
Related: https://github.com/lfortran/lfortran/issues/1059

certik:
Let's create two passes:

* One pass transforms the executable code into a function in the same translation unit, but leaves global variables in translation unit. After this pass, there are no global statements or expressions, only global variables (in translation unit) and functions that access those global variables

* Second pass transforms the global variables into a module. We have to support module variables anyway in all backends, so this might make it easy to support global variables. This pass requires to modify all functions that access the global variable to import the variable from the module.

Thirumalai-Shaktivel:
ASR output with current changes:
```Clojure
(TranslationUnit
    (SymbolTable
        1
        {
            _global_symbols:
                (Module
                    (SymbolTable
                        5
                        {
                            _lpython_main_program:
                                (Function
                                    (SymbolTable
                                        4
                                        {
                                            x:
                                                (ExternalSymbol
                                                    4
                                                    x
                                                    5 x
                                                    _global_symbols
                                                    []
                                                    x
                                                    Public
                                                )
                                        })
                                    _lpython_main_program
                                    [main0]
                                    []
                                    [(=
                                        (Var 4 x)
                                        (ListConstant
                                            [(IntegerConstant 1 (Integer 4 []))]
                                            (List
                                                (Integer 4 [])
                                            )
                                        )
                                        ()
                                    )
                                    (SubroutineCall
                                        5 main0
                                        ()
                                        []
                                        ()
                                    )]
                                    ()
                                    Source
                                    Public
                                    Implementation
                                    ()
                                    .false.
                                    .false.
                                    .false.
                                    .false.
                                    .false.
                                    []
                                    []
                                    .false.
                                ),
                            main0:
                                (Function
                                    (SymbolTable
                                        2
                                        {
                                            x:
                                                (Variable
                                                    2
                                                    x
                                                    []
                                                    Local
                                                    ()
                                                    ()
                                                    Default
                                                    (List
                                                        (Integer 4 [])
                                                    )
                                                    Source
                                                    Public
                                                    Required
                                                    .false.
                                                )
                                        })
                                    main0
                                    []
                                    []
                                    [(=
                                        (Var 2 x)
                                        (ListConstant
                                            [(IntegerConstant 1 (Integer 4 []))]
                                            (List
                                                (Integer 4 [])
                                            )
                                        )
                                        ()
                                    )
                                    (Print
                                        ()
                                        [(ListItem
                                            (Var 2 x)
                                            (IntegerConstant 0 (Integer 4 []))
                                            (Integer 4 [])
                                            ()
                                        )]
                                        ()
                                        ()
                                    )]
                                    ()
                                    Source
                                    Public
                                    Implementation
                                    ()
                                    .false.
                                    .false.
                                    .false.
                                    .false.
                                    .false.
                                    []
                                    []
                                    .false.
                                ),
                            x:
                                (Variable
                                    5
                                    x
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (List
                                        (Integer 4 [])
                                    )
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
                        })
                    _global_symbols
                    []
                    .false.
                    .false.
                ),
            main_program:
                (Program
                    (SymbolTable
                        3
                        {
                            _lpython_main_program:
                                (ExternalSymbol
                                    3
                                    _lpython_main_program
                                    5 _lpython_main_program
                                    _global_symbols
                                    []
                                    _lpython_main_program
                                    Public
                                )
                        })
                    main_program
                    []
                    [(SubroutineCall
                        3 _lpython_main_program
                        ()
                        []
                        ()
                    )]
                )
        })
    []
)
```

Thirumalai-Shaktivel:
After everything works (List in the LLVM), then I will create a separate pass for `pass_wrap_symbols_into_module`

Thirumalai-Shaktivel:
https://github.com/lcompilers/lpython/blob/2c5034a299c7ec3edcd6bb13daab1434b8e6d0cc/src/libasr/codegen/asr_to_llvm.cpp#L2855

This `AllocaInst` is in the `decl_vars` function to declare variables, I think. My doubt is, here ptr is declared but where does it get loaded or stored for Lists in LLVM?
@czgdp1807

examples/expr2.py
```diff
@@ -1,9 +1,15 @@
...
-    x: i32
-    x = (2+3)*5
-    print(x)
+from ltypes import i32
 
-main0()
+x: list[i32]
+x = [1, 2, 3]
+# i: i32 = x[0]
+print(x)
```
Thirumalai-Shaktivel, line 6: This works:
```
[1, 2, 3]
```

certik:
In here I would simply focus on creating a pass that transform the global statements and variables from TranslationUnit into a module. And you can create integration_test, and test it using ASR tests in `tests/`. Let's not worry if things do not work with LLVM.

Then in a separate PR let's test that module variables (of all types, including lists) work.

Thirumalai-Shaktivel:
I followed the first approach from here: https://github.com/lcompilers/lpython/pull/1491#discussion_r1118060606
The `examples/expr2.py` works.

@czgdp1807 What do you think about this?

certik:
Looks good so far. You can add a test, get everything passing. Then you can figure out how to change `void*` into "struct" as needed. Regarding LLVM pointers, use ASR to figure out all the types, not LLVM.

czgdp1807 (approved):
LGTM.

Thirumalai-Shaktivel:
> Then you can figure out how to change void* into "struct" as needed.

Should I try this in this PR? @certik 
After getting this done, I will work on importing the global variables.

Thirumalai-Shaktivel:
Actually, it was easy to figure out.

certik:
There are some test failures.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2296,6 +2297,14 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 }
             }
             llvm_symtab[h] = ptr;
+        } else if (x.m_type->type == ASR::ttypeType::List) {
+            llvm::StructType* list_type = (llvm::StructType *)
+                get_type_from_ttype_t_util(x.m_type);
```
czgdp1807, line 2302:
```suggestion
                get_type_from_ttype_t_util(x.m_type));
```

czgdp1807 (approved):

czgdp1807:
> Actually, it was easy to figure out.

`CreateBitCast` I guess?

Thirumalai-Shaktivel:
> `CreateBitCast` I guess?

The thing is I switched from the `void*` type to the `list` type, so there is no need to do BitCast, I think.
```llvm
; ModuleID = 'LFortran'
source_filename = "LFortran"

%list = type { i32, i32, i32* }

@i = global i32 0
@x = global %list zeroinitializer
...
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2296,6 +2296,14 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
             }
             llvm_symtab[h] = ptr;
+        } else if (x.m_type->type == ASR::ttypeType::List) {
+            llvm::StructType* list_type = static_cast<llvm::StructType*>(
+                get_type_from_ttype_t_util(x.m_type));
+            llvm::Constant *ptr = module->getOrInsertGlobal(x.m_name, list_type);
+            module->getNamedGlobal(x.m_name)->setInitializer(
+                llvm::ConstantStruct::get(list_type,
+                llvm::Constant::getNullValue(list_type)));
+            llvm_symtab[h] = ptr;
```
Thirumalai-Shaktivel, line 2306: Sorry, I think I understood this wrong!
@czgdp1807 is the approach used here correct?

czgdp1807: Using global variable is better. It stays in scope and is independent of all LLVM functions. I checked the generated LLVM code and it looks fine to me.

Thirumalai-Shaktivel: Perfect

tests/reference/llvm-print_04-443a8d8.stdout
```diff
@@ -1,10 +1,10 @@
...
 source_filename = "LFortran"
 
-@u = global i64 -922337203685477580
-@x = global i32 -2147483648
-@y = global i16 -32768
-@z = global i8 -128
+@u = global i64 0
+@x = global i32 0
+@y = global i16 0
+@z = global i8 0
```
certik, line 7: I think I would keep these initializations here. This is relatively minor, so we can do it in the next PR.

certik (approved):

tests/reference/llvm-print_04-443a8d8.stdout
certik: I created an issue for this at https://github.com/lcompilers/lpython/issues/1582.

PR #1386, opened 2022-12-27
Return path of runtime library header of development version correctly

czgdp1807:
With this PR,

`/Users/czgdp1807/lpython_project/lpython/src/bin/../libasr/runtime`

On `main`,

`/Users/czgdp1807/lpython_project/lpython/src/bin/../runtime/impure`

`lfortran_intrinsics.h` isn't present in `src/runtime/impure`. Its always present in `src/libasr/runtime` for development version of LPython.

### DIFF
--- a/src/lpython/utils.cpp
+++ b/src/lpython/utils.cpp
@@ -79,7 +79,23 @@ std::string get_runtime_library_header_dir()
     char *env_p = std::getenv("LFORTRAN_RUNTIME_LIBRARY_HEADER_DIR");
     if (env_p) return env_p;
 
-    return get_runtime_library_dir() + "/impure";
+    std::string path;
+    int dirname_length;
+    get_executable_path(path, dirname_length);
+    std::string dirname = path.substr(0, dirname_length);
+    if (   endswith(dirname, "src/bin")
+        || endswith(dirname, "src\\bin")
+        || endswith(dirname, "SRC\\BIN")) {
+        // Development version
+        return dirname + "/../libasr/runtime";
+    } else if (endswith(dirname, "src/lpython/tests") ||
+               endswith(to_lower(dirname), "src\\lpython\\tests")) {
+        // CTest Tests
+        return dirname + "/../../runtime/impure";
+    } else {
+        // Installed version
+        return dirname + "/../share/lpython/lib/impure";
+    }
 }
 
 bool is_directory(std::string path) {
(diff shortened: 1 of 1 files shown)


src/lpython/utils.cpp
```diff
@@ -79,7 +79,23 @@ std::string get_runtime_library_header_dir()
     char *env_p = std::getenv("LFORTRAN_RUNTIME_LIBRARY_HEADER_DIR");
     if (env_p) return env_p;
 
-    return get_runtime_library_dir() + "/impure";
+    std::string path;
+    int dirname_length;
+    get_executable_path(path, dirname_length);
```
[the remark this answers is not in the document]
czgdp1807: It returns absolute path always.

src/lpython/utils.cpp
```diff
@@ -79,7 +79,23 @@ std::string get_runtime_library_header_dir()
...
+        // Development version
+        return dirname + "/../libasr/runtime";
+    } else if (endswith(dirname, "src/lpython/tests") ||
+               endswith(to_lower(dirname), "src\\lpython\\tests")) {
+        // CTest Tests
+        return dirname + "/../../runtime/impure";
+    } else {
+        // Installed version
+        return dirname + "/../share/lpython/lib/impure";
+    }
```
certik, line 98: This code seems very messy and fragile. I know you copied it from `get_runtime_library_dir`, but I suggest we refactor it, so that the "mess" is concentrated at one place. I think determining the development vs installed version is fine. The CTest part is especially hacky, as it won't work in any subdirectories, etc.

One option is to create a function that determines if LFortran runs from the source directory or from installed directory. Also, does this work in a build directory, or only in in-tree builds?

Let's figure out how to fix this properly, instead of adding to this technical debt that will be harder to fix in the future.

czgdp1807: > This code seems very messy and fragile.

True.



> I know you copied it from `get_runtime_library_dir`, but I suggest we refactor it, so that the "mess" is concentrated at one place.

Yes. I just wanted a quick fix so copied from there. I agree we should encapsulate the logic to figure out whether LPython/LFortran is an installed version, source version or CTest part.

> One option is to create a function that determines if LFortran runs from the source directory or from installed directory.

Yes that sounds good. I think even the version should contain this information (a prefix like `dev`) i.e., whether the LFortran/LPython is installed or development version.
I don't know about the CTest part though. What should be a good logic to figure that out?

certik: I would just use the existing (messy) logic, but put it into one function and document it. Then use that function.

certik:
The runtime library handling should be unified with LFortran, it's still a bit hackish in both. I am making this a Draft.

Issue #1387, opened 2022-12-28
`buffer: i8[64];` in struct must generate `char buffer[64]`, not `struct i8* buffer`


certik:
This:
```python
@packed
@dataclass
class A:
    buffer: i8[64];
    k: i16
```
I think it must generate:
```c
struct A {
 int8_t buffer[64];
 int16_t k;
}
```
but it currently generates:
```c
struct A {
 struct i8* buffer;
 int16_t k;
}
```

PR #1388, opened 2022-12-28
Generate data only array members if they are of fixed size

czgdp1807:
cc: @certik 

Closes https://github.com/lcompilers/lpython/issues/1387

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -355,6 +355,8 @@ RUN(NAME structs_14          LABELS cpython llvm c)
 RUN(NAME structs_15          LABELS cpython llvm c)
 RUN(NAME structs_16          LABELS cpython llvm c)
 RUN(NAME structs_17          LABELS cpython llvm c)
+RUN(NAME structs_18          LABELS llvm c
+    EXTRAFILES structs_18b.c)
 RUN(NAME sizeof_01           LABELS llvm c
         EXTRAFILES sizeof_01b.c)
 RUN(NAME enum_01             LABELS cpython llvm c)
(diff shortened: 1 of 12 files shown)


integration_tests/CMakeLists.txt
```diff
@@ -351,7 +351,8 @@ RUN(NAME structs_11          LABELS cpython llvm c)
 RUN(NAME structs_12          LABELS cpython llvm c)
 RUN(NAME structs_13          LABELS llvm c
     EXTRAFILES structs_13b.c)
-RUN(NAME structs_14          LABELS cpython llvm c)
+RUN(NAME structs_14          LABELS llvm c
+    EXTRAFILES structs_14b.c)
```
certik, line 355: We might need to do a separate test, since we lost CPython test here. We want to ideally test everything via CPython, but at least what we were testing before.

integration_tests/structs_14b.c
```diff
@@ -0,0 +1,9 @@
+#include "structs_14b.h"
+
+int16_t sum_buffer_i16(int16_t data[], int32_t size) {
+    int16_t sum_buffer = 0;
+    for( int32_t i = 0; i < size; i++ ) {
+        sum_buffer += data[i];
+    }
+    return sum_buffer;
+}
```
certik, line 9: I think this is not enough. I think we need to pass the whole structure into C, then in C recreate the same structure using `char[64]` style, and then access the members and check that everything works. That would be robust.

The current test could still pass even with our pointer setup (yes, the memory would not be initialized properly, but one could get "lucky").

certik (approved):
I think this is good, although it's not immediately obvious to me from the test that it does the right thing. Let's add the C Struct test as I indicated in the comment, then I think it will be 100% clear that it works.

integration_tests/structs_14b.c
czgdp1807: Which one do you want to pass to C, `buffer_struct` (not decorated with `@ccall`) or `buffer_struct_clink` (decorated with `@ccall`)? I guess `buffer_struct_clink` should be passed to C for testing as its the one with `BindC` ABI, correct? If you want to pass both, then LLVM backend will also need to be updated to always (irrespective of the ABI) represent fixed sized array members as data only pointers

czgdp1807: I added test as per my understanding of the requirement. Please let me know if they look good or if you want any change in it. Thanks.

certik (approved):
I think this looks good to merge! Thanks for fixing this.

PR #1390, opened 2023-01-02
WASM_X64: Support if else

Shaikh-Ubaid:
This PR adds support for `if-else` statements in the `wasm_x64` backend.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -241,7 +241,7 @@ RUN(NAME expr_13             LABELS llvm c
 RUN(NAME expr_14             LABELS cpython llvm c)
 RUN(NAME loop_01             LABELS cpython llvm c)
 RUN(NAME loop_02             LABELS cpython llvm c wasm wasm_x86)
-RUN(NAME if_01               LABELS cpython llvm c wasm wasm_x86)
+RUN(NAME if_01               LABELS cpython llvm c wasm wasm_x86 wasm_x64)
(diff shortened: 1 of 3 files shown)


certik (approved):

Shaikh-Ubaid:
I initially attempted to support the idea of using unique integers shared at https://github.com/lcompilers/lpython/pull/1264#discussion_r1019468441. 

For example using: 
```cpp
void visit_If() {
        unique_idx++;
        std::string label = std::to_string(unique_idx);
        m_a.asm_pop_r64(X64Reg::rax); // now `rax` contains the logical value (true = 1, false = 0) of the if condition
        m_a.asm_cmp_r64_imm8(X64Reg::rax, 1);
        m_a.asm_je_label(".then_" + label);
        m_a.asm_jmp_label(".else_" + label);
        m_a.add_label(".then_" + label);
        {
            decode_instructions();
        }
        m_a.add_label(".endif_" + label);
        unique_idx--;
    }

    void visit_Else() {
        std::string label = std::to_string(unique_idx);
        m_a.asm_jmp_label(".endif_" + label);
        m_a.add_label(".else_" + label);
    }
```
But it happens that the `if-else` statements within the same scope have the same ids as follows:
```
if { ---- index 0
    if { --- index 1
        if { --- index 2

        } else { --- index 2

        }        
    } else { --- index 1

    }
   
    if { -- index 1
   
    } else { --- index 1
    
    }
} 
```

Thus, I used the stack approach similar to the `x86` backend.

Shaikh-Ubaid:
Thank you for the approval. I am adding it to auto-merge as it seems approved.

PR #1394, opened 2023-01-03
Improve punctuation and idiomatic usage of English.

rebcabin:


certik (approved):
Thanks for the fixes!

PR #1399, opened 2023-01-04
WASM_X64: Support loops

Shaikh-Ubaid:
This `PR` adds initial support for loops in the `wasm_x64` backend.

It seems that with merging/completion of this PR, both the backends `wasm_x86` and `wasm_x64` would be at a similar level.


Shaikh-Ubaid:

src/libasr/codegen/wasm_to_x64.cpp
```diff
@@ -236,6 +287,43 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
...
+        m_a.asm_push_imm8(1);
+        m_a.add_label(".compare.end_" + label);
+    }
+
+    void visit_I32Eq() { handle_I32Compare("Eq"); }
+    void visit_I32GtS() { handle_I32Compare("Gt"); }
+    void visit_I32GeS() { handle_I32Compare("GtE"); }
+    void visit_I32LtS() { handle_I32Compare("Lt"); }
+    void visit_I32LeS() { handle_I32Compare("LtE"); }
+    void visit_I32Ne() { handle_I32Compare("NotEq"); }
```
certik, line 325: Passing strings is slow. We should speed it up by using a template and do something like:

```suggestion
    void visit_I32Ne() { handle_I32Compare<m_a.asm_jne_label>(); }
```
And in `handle_I32Compare` just call `T(".compare_1" + label);` and skip the whole if statement style "switch" and string handling.

certik: This can be done after this is merged.

certik (approved):

src/libasr/codegen/wasm_to_x64.cpp
Shaikh-Ubaid: Got it. It is a great approach. Thank you for the guidance. 

>This can be done after this is merged.

Sure, I will submit it after this PR is merged.

certik: See https://github.com/lcompilers/lpython/issues/1409.

PR #1401, opened 2023-01-05
C: Fix strings allocation/declaration

Smit-create:


certik (approved):
I think that's fine for now. Although a better approach might be to always allocate the global variable, so that we don't have to have those if checks all the time.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -625,6 +626,9 @@ R"(#include <stdio.h>
                 // ASR assigns return variable only once at the end of function
                 alloc_return_var = true;
             }
+            if (ASRUtils::is_character(*m_target_type)) {
+                alloc_target = indent + "if (!" + target + ") " + target + " = (char *) malloc(40 * sizeof(char));\n";
```
certik, line 630: Why is this 40? That seems it will break.

Smit-create:
> Although a better approach might be to always allocate the global variable,

C doesn't support `char *x = (char *) malloc()` for global variables during declaration.

src/libasr/codegen/asr_to_c_cpp.h
Smit-create: We are currently fixing it to 40 (everywhere), but we need to define some strong support later that handles reallocation like lists.

czgdp1807: I think its just some pre-allocation we are doing at declaration. Yes, we need to have some flag which will help us figure out if a particular `char*` is allocated or not. If its not at a particular point of usage then allocate it and then use it.

tests/reference/c-print_01-4d44628.stdout
```diff
@@ -40,9 +40,12 @@ void f()
     char * x = (char*) malloc(40 * sizeof(char));
     char * y = (char*) malloc(40 * sizeof(char));
     printf("%s\n", "Hello World!");
+    if (!x) x = (char *) malloc(40 * sizeof(char));
     strcpy(x, ",");
+    if (!y) y = (char *) malloc(40 * sizeof(char));
```
czgdp1807, line 45: Actually here (i.e., `strcpy(x, ",")`), `x` needs to be pre-allocated. We do this right now by doing `malloc(40, sizeof(char))`. This means we are allocating a string size of 40 and assigning it to `x`. This is simple but sometimes means wastage of memory and sometimes means chances of segmentation fault. So a better way is to define our own `_lfortran_strcpy` in `lfortran_intrinsics.h` and then call C's `strcpy` in it. Then, our `_lfortran_strcpy` will be called everywhere. This is how it will look like,

```c
void _lfortran_strcpy(char* x, char* y) {
    x = (char*) malloc((strlen(x) + strlen(y) + 1) * sizeof(char));
    strcpy(x, y);
}
```

czgdp1807: @Smit-create Please do the above in this PR only. I think the above approach is better than what we are doing right now.

certik:
> C doesn't support `char *x = (char *) malloc()` for global variables during declaration.

We are already calling an "init" function to handle command line arguments, so we can create some mechanism to register this initializations and do it when the program starts.

integration_tests/CMakeLists.txt
```diff
@@ -383,7 +383,7 @@ RUN(NAME union_03            LABELS cpython llvm c)
 RUN(NAME test_str_to_int     LABELS cpython llvm c)
 RUN(NAME test_platform       LABELS cpython llvm c)
 RUN(NAME test_vars_01        LABELS cpython llvm)
-RUN(NAME test_version        LABELS cpython llvm c)
+RUN(NAME test_version        LABELS cpython llvm)
```
Smit-create, line 386: This isn't a valid test in C and was introduced in this PR #1326.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -623,6 +623,15 @@ LFORTRAN_API void _lfortran_strcat(char** s1, char** s2, char** dest)
     *dest = &(dest_char[0]);
 }
 
+// strcpy -----------------------------------------------------------
+
+LFORTRAN_API void _lfortran_strcpy(char** x, char *y)
+{
+    if (*x) free((void *)*x);
+    *x = (char*) malloc((strlen(y) + 1) * sizeof(char));
+    strcpy(*x, y);
```
czgdp1807, line 632: We are de-referencing `x` in all the three lines? So why not `_lfortran_strcpy(char* x, char *y)`? Any particular reasons?

certik (approved):

src/libasr/runtime/lfortran_intrinsics.c
Smit-create: We are actually changing the pointer values and so that won't work as the pointer will be a "copy" variable. Notice the difference here:
```c
#include <stdio.h>

void fun(int *a) {
    int x = 10;
    a = &x;
}

void fun2(int **a) {
    int x = 10;
    *a = &x;
}

int main() {
    int b = 5;
    int *a = &b;
    fun(a);
    printf("%d\n", *a); // Still prints 5
    fun2(&a);
    printf("%d\n", *a); // prints 10
    return 0;
}
```

czgdp1807: I see. Was expecting this. Thanks for the confirmation. Going to approve. Please rebase on top of latest `main` and enable auto merge.

czgdp1807 (approved):

PR #1402, opened 2023-01-05
WASM_X64: Initial support for floats

Shaikh-Ubaid:
This PR adds initial support for floats in the `wasm_x64` backend.

**Example:**
```bash
(lp) lpython$ cat examples/expr2.py 
def main0():
    print(52)
    print("hi")
    print(123.456)
    print("hi")
    print(52)
    print(23.414)
    print("hi")
    print(79)
main0()
(lp) lpython$ lpython examples/expr2.py --backend wasm_x64 -o tmp > main.asm
(lp) lpython$ ./tmp
52
hi
123.45600000
hi
52
23.41400000
hi
79
(lp) lpython$ 
```

### DIFF
--- a/src/libasr/codegen/wasm_to_x64.cpp
+++ b/src/libasr/codegen/wasm_to_x64.cpp
@@ -93,7 +93,7 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
                 break;
             }
             case 5: {  // flush_buf
-                emit_print_64(m_a, "string_newline", 2);
+                emit_print_64(m_a, "string_newline", 1);
                 break;
             }
             case 6: {  // set_exit_code
--- a/src/libasr/codegen/wasm_to_x64.cpp
+++ b/src/libasr/codegen/wasm_to_x64.cpp
@@ -41,6 +41,7 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
     uint32_t block_id;
     std::vector<std::pair<uint32_t, Block>> blocks;
 
+    bool decoding_data_segment;
 
     X64Visitor(X86Assembler &m_a, Allocator &al,
                diag::Diagnostics &diagonostics, Vec<uint8_t> &code)
@@ -49,6 +50,7 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
           m_a(m_a) {
         wasm_bytes.from_pointer_n(code.data(), code.size());
         block_id = 1;
+        decoding_data_segment = false;
     }
 
     void visit_Return() {}
@@ -253,9 +255,11 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
     }
 
     void visit_I32Const(int32_t value) {
-        m_a.asm_mov_r64_imm64(X64Reg::rax, labs((int64_t)value));
-        if (value < 0) m_a.asm_neg_r64(X64Reg::rax);
-        m_a.asm_push_r64(X64Reg::rax);
+        if (!decoding_data_segment) {
+            m_a.asm_mov_r64_imm64(X64Reg::rax, labs((int64_t)value));
+            if (value < 0) m_a.asm_neg_r64(X64Reg::rax);
+            m_a.asm_push_r64(X64Reg::rax);
+        }
 
         // TODO: Following seems/is hackish. Fix/Improve it.
         last_last_vis_i32_const = last_vis_i32_const;
@@ -334,6 +338,7 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
         emit_elf64_header(m_a);
         emit_print_int_64(m_a, "print_i64");
 
+        decoding_data_segment = true;
         // declare compile-time strings
         for (uint32_t i = 0; i < data_segments.size(); i++) {
             offset = data_segments[i].insts_start_index;
@@ -341,6 +346,7 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
             std::string label = "string" + std::to_string(last_vis_i32_const);
             label_to_str[label] = data_segments[i].text;
         }
+        decoding_data_segment = false;
 
         for (uint32_t idx = 0; idx < type_indices.size(); idx++) {
             m_a.add_label(exports[idx].name);
--- a/src/libasr/codegen/wasm_to_x64.cpp
+++ b/src/libasr/codegen/wasm_to_x64.cpp
@@ -42,6 +42,15 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
     std::vector<std::pair<uint32_t, Block>> blocks;
     std::map<std::string, double> double_consts;
 
+    /*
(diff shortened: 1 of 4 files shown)


Shaikh-Ubaid:
There seems some bug in the `x64` text format that is generated.

```bash
(wasm_asm) lpython$ nasm -fbin main.asm && chmod +x main && ./main
52
hi
123.100000000 // it should have been 123.45600000
hi
52
23.100000000  // it should have been 23.41400000
hi
79
```

I think/guess we can fix it in a separated PR.

Shaikh-Ubaid:

Shaikh-Ubaid:
It seems there are two types of floating-point instructions - vector and scalar instructions. I am unsure which one we need. To get started, I currently implemented the scalar version of instructions.

Shaikh-Ubaid:
This issue https://github.com/lcompilers/lpython/pull/1402#issuecomment-1372165794 is fixed now.


```bash
(lp) lpython$ cat examples/expr2.py 
def main0():
    print(52)
    print("hi")
    print(123.456)
    print("hi")
    print(52)
    print(23.414)
    print("hi")
    print(79)
main0()
(lp) lpython$ lpython examples/expr2.py --backend wasm_x64 -o tmp > main.asm
(lp) lpython$ ./tmp
52
hi
123.45600000
hi
52
23.41400000
hi
79
(wasm_asm) lpython$ nasm -fbin main.asm && chmod +x main && ./main
52
hi
123.45600000
hi
52
23.41400000
hi
79
(wasm_asm) lpython$ 
```

src/libasr/codegen/x86_assembler.cpp
```diff
@@ -396,4 +404,50 @@ void emit_print_int_64(X86Assembler &a, const std::string &name)
...
+
+    // print dot
+    emit_print_64(a, "string_dot", 1U);
+
+    // print fractional part
+    {
+        a.asm_cvttsd2si_r64_r64(X64Reg::rax, X64FReg::xmm0); // rax now contains value int(xmm0)
+        a.asm_cvtsi2sd_r64_r64(X64FReg::xmm1, X64Reg::rax);
+        a.asm_subsd_r64_r64(X64FReg::xmm0, X64FReg::xmm1);
+        a.asm_mov_r64_imm64(X64Reg::rax, 100000000); // to multiply by 10^8
```
Shaikh-Ubaid, line 435: Maybe here we should multiply by some larger number (probably `10^12`), so that we have more precision. I attempted multiplying by `10^12`, but it leads to random values. I guess it happens due to some integer overflow (since after multiplying we convert it to integer).

**doubt:** What is the precision we need? Would it be  fine if we implement the precision in later PRs?

Shaikh-Ubaid:

src/libasr/codegen/wasm_to_x64.cpp
```diff
@@ -35,14 +35,17 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
     int32_t last_vis_i32_const, last_last_vis_i32_const;
     std::map<std::string, std::string> label_to_str;
     std::vector<std::string> if_unique_id;
+    std::map<std::string, double> double_consts;
 
+    bool decoding_data_segment;
```
certik, line 40: Can you please add a comment here what this does (I assume it is true if we are decoding a data segment in WASM, and I can also see you are sometimes only executing some logic if not decoding a segment, but I do not understand why).

src/libasr/codegen/x86_assembler.cpp
certik: Ultimately we need to print all digits. But for now this is fine, and later I think we can simply implement this in the runtime library implemented in the frontend language, that we will just compile through ASR and WASM, so we won't need to maintain a specific assembly version of this.

src/libasr/codegen/x86_assembler.h
```diff
@@ -1195,6 +1237,110 @@ class X86Assembler {
...
+    }
+
+    // Move or Merge Scalar Double Precision Floating-Point Value
+    void asm_movsd_m64_r64(X64Reg *base, X64Reg *index,
+                uint8_t scale, int64_t disp, X64FReg r64) {
+        X86Reg r32 = X86Reg(r64 & 7);
+        X86Reg* base32 = nullptr, *index32 = nullptr;
+        if (base) {
+            base32 = new X86Reg;
+            *base32 = X86Reg(*base & 7);
```
certik, line 1270: This seems to be leaking memory. If you need to allocate, use the Allocator.

However, we should not need to allocate these simple structures, just use them by value. You can let the caller provide the allocation using a local variable.

certik:
Very good, overall this looks great. I left some minor comments to resolve.

src/libasr/codegen/wasm_to_x64.cpp
Shaikh-Ubaid: A data segment in `wasm` uses a set of instructions/expressions to specify the offset in memory where the string/data is to be stored.

For example, in `(data (;0;) (i32.const 4) "abc")`, the set of instructions/expressions contains one element which is `i32.const 4` and it evaluates to `4` which represents the memory offset where the string `abc` is to be stored. 

To obtain the string offset we need to decode these set of instructions. Since, we currently support compile-time strings and the string offset is stored in `last_vis_i32_const`, the instructions are not needed for us at the moment. Since, they are not needed, we should not be emitting them while decoding a data segment.

Shaikh-Ubaid: I added a comment about it in commit [ASM_X64: Add comment about decoding_data_segment](https://github.com/lcompilers/lpython/pull/1402/commits/16217d7a8e07f01ce687faef21c567b045402444).

src/libasr/codegen/x86_assembler.h
Shaikh-Ubaid: I fixed it in [X86Assembler: Use local varaibles for index32 and base32](https://github.com/lcompilers/lpython/pull/1402/commits/cf2c202aeb01c2a476885542260fb992798aa350).

Shaikh-Ubaid:

Shaikh-Ubaid:
**Example:**
```bash
(lp) lpython$ cat examples/expr2.py 
def main0():
    print(52)
    print("hi")
    print(123.456)
    print("abc")
    print(52)
    print(23.414)
    print("efgh")
    print(79)
main0()
(lp) lpython$ lpython examples/expr2.py --backend wasm_x64 -o tmp > main.asm
(lp) lpython$ ./tmp
52
hi
123.45600000
abc
52
23.41400000
efgh
79
(wasm_asm) lpython$ nasm -fbin main.asm && chmod +x main && ./main
52
hi
123.45600000
abc
52
23.41400000
efgh
79
```

czgdp1807 (approved):
This looks good to me overall. Though I am not an expert in WASM to x64/x86 backends. But the structure of this PR is good.

certik (approved):
I think that looks good.

Issue #1129, opened 2022-09-20
ASR: add string formatting nodes


certik:
One way to handle string formatting and printing in a language independent approach is to decouple string formatting form printing. The `Print` ASR node would just do printing and printing related flags. I think it's pretty much there.

Then we need an expr node like `StringFormat`, which will handle all string formatting. The string formatting is inherently runtime operation, so the `StringFormat` would accept the format string (in various styles) and the variables to format, and it would produce formatted string. It is not clear if a single `StringFormat` can handle all formatting styles. If not, we will introduce `StringFormatFortran`, `StringFormatC`, `StringFormatPython1` (`%s` style), `StringFormatPython2` (the `{}` style) etc. (there are also f-strings in Python, it would be nice to generate the original f-string from ASR if possible).

Then in Fortran `print "(i0, i0)", a, b` will become `Print(StringFormatFortran("(i0, i0)", a, b))`, and Python's `print("%d %d", a, b)` would become `Print(StringFormatPython1("%d %d", a, b))`.

certik:
There is also Python's `"%r" % some_object`, it seems it should be part of `StringFormatPython1`, to handle the `%r` correctly. Then `print(l)` becomes `Print(StringFormatPython1("%r", l))`.

czgdp1807:
We should test this approach with the following types (mixed types of container data structures).

1. `list[dict[i32, tuple[i32, f64]]]`
2. `tuple[list[i32], dict[i32, f64]]`
3. `dict[tuple[i32, str], list[dict[i32, i64]]]`
4. .
5. .
6. .
7. Any combination of aggregate data types like `list`, `tuple`, `dict` and so on.

Shaikh-Ubaid:
One of the concerns that I have with this approach is that
- As per the `ASR` Specification [here](https://github.com/lcompilers/lpython/wiki/Intrinsic-Functions-Design), it seems `ASR` is **self-contained** and **complete**
- If we have nodes like `StringFormatC` or `StringFormatPython3` in the `ASR`, it seems that the `ASR` in a way might not look complete, since there could always be a language for which we did not yet define an `ASR` node.
- Or whenever we add/define/develop a new language frontend, we would need to add a new node to the `ASR` for print formatting of this specific language, thus the `ASR` does not look self-contained (since it is being needed to be extended/modified for every new langauge frontend).

Shaikh-Ubaid:
[...]

PR #1405, opened 2023-01-06
C: Add printing support for Lists

Smit-create:


czgdp1807 (commented):
Can we just call the following pass in C backend instead of implementing printing of list in C backend again? Let me know if calling the pass will work.

https://github.com/lcompilers/lpython/blob/993fff2e4eb074ad2e1881fe3b08a73139c0d1c1/src/libasr/pass/print_list.cpp#L201

Smit-create:
The pass is not that robust, see with the current `main` branch the following fails in the LLVM backend:
```py
def test_print_list():
    a: list[str] = ["ab", "abc", "abcd"]

    print(a) # works
    print(a, a) # fails

test_print_list()
```
Also, that pass seems to be only working for single argument list type, not with nested tuple, see:
```py
def test_print_list():
    a: list[tuple[i32, i32]] = [(1,2), (3, 4), (5, 6)]

    print(a) # fails

test_print_list()
```

That will require much work on the pass too.

Moreover, `C`, at present, doesn't use any passes and already has a design that will need slight modifications to make all this work.

I will fix those issues with the pass in a new PR soon to make it work for LLVM.

czgdp1807:
Associated issue, https://github.com/lcompilers/lpython/issues/1129. But anyways we still need the ASR pass to work for lists.

czgdp1807 (approved):

PR #1421, opened 2023-01-12
Unify C Utils API to use pointers for tuple/list

Smit-create:


Smit-create:
I was thinking of unifying the list/tuple API in this PR by using pointers and malloc everywhere but realized that since we aren't deleting the memory using free, this might create memory problems and so it is good to use normal variables instead of pointers.

PR #1422, opened 2023-01-12
WASM_X64,X86: Manipulate stack pointer using sub/add instructions

Shaikh-Ubaid:
This `PR` solves the following issue:

Currently, we update the `stack` pointer (mostly) using `push` and `pop` operations. We can also update the `stack pointer` by adding/subtracting the desired `offset` from it. In case of `floating-point` operations, instead of pushing an integer register onto `stack` to create required space, it could/would be better if we update the stack pointer by adding/subtracting value from `esp` to create the required space. To support `add`/`sub` desired value, we need support for `add`/`sub` instruction in the `wasm_x64` backend.

### DIFF
--- a/src/libasr/codegen/wasm_to_x64.cpp
+++ b/src/libasr/codegen/wasm_to_x64.cpp
@@ -482,7 +482,7 @@ Result<int> wasm_to_x64(Vec<uint8_t> &wasm_bytes, Allocator &al,
     }
 
     //! Helpful for debugging
-    std::cout << x64_visitor.m_a.get_asm() << std::endl;
+    // std::cout << x64_visitor.m_a.get_asm() << std::endl;
 
     if (time_report) {
         std::cout << "Codegen Time report:" << std::endl;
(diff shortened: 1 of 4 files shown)


Shaikh-Ubaid:

src/libasr/codegen/wasm_to_x64.cpp
```diff
@@ -506,7 +500,7 @@ Result<int> wasm_to_x64(Vec<uint8_t> &wasm_bytes, Allocator &al,
     }
 
     //! Helpful for debugging
-    // std::cout << x64_visitor.m_a.get_asm() << std::endl;
+    std::cout << x64_visitor.m_a.get_asm() << std::endl;
```
certik, line 503: Probably want to remove.

certik (approved):
Besides that debugging print this looks good.

src/libasr/codegen/wasm_to_x64.cpp
Shaikh-Ubaid: Sorry, I will be careful next time. I fixed it in [WASM_X64: Comment out asm text print](https://github.com/lcompilers/lpython/pull/1422/commits/c039edfe2026b74a9bf260e37b5a6e2eb86bc152).

Shaikh-Ubaid:
Thank you for the approval. I rebased this. I am adding this to auto-merge as it seems approved.

certik:
Thanks!

PR #1426, opened 2023-01-14
Fix issues when WITH_RUNTIME_STACKTRACE=no

### LINKED ISSUE #1424: Few issues if WITH_RUNTIME_STACKTRACE=no
* [x] CMake should print the result of this variable
* [x] `lpython -g` then links functions from our C library that is disabled, so the Clang link command fails. Instead we should either give an error message earlier, or perhaps let's always implement these functions, so that they link, and just give a runtime message that stacktraces were not enabled.
* [x] The `./run_tests.py` fails as well, so we should probably disable those tests if `WITH_RUNTIME_STACKTRACE=no`.

Thirumalai-Shaktivel:
Fixes #1424


Thirumalai-Shaktivel:
> The ./run_tests.py fails as well, so we should probably disable those tests if WITH_RUNTIME_STACKTRACE=no

@certik would this: 1ae000c022af65ea0d9119187d98dc54d3ef8105 be a valid fix?

certik:
I posted my comments at https://github.com/lcompilers/lpython/commit/1ae000c022af65ea0d9119187d98dc54d3ef8105 but they do not appear here in the PR, so you have to read them there: To summarize: let's just add an option to skip the run_dbg tests and be done with it.

certik (approved):

PR #1427, opened 2023-01-14
ASR: Introduce `deterministic` and `side-effect-free` flags

akshanshbhatt:
See https://github.com/lcompilers/lpython/pull/1293#issuecomment-1382192012


certik (approved):

certik:
I think that this is good, merging.

PR #1428, opened 2023-01-14
Minor fixes/improvements

Shaikh-Ubaid:


certik (approved):
I think that's fine, I tested by hand locally, it seems to work.

Thanks for unifying the test runners.

PR #1431, opened 2023-01-16
[ASR PASS]: Fix printing of tuples

### LINKED ISSUE #1410: Bugs with `print_list` pass
The pass is not that robust, see with the current `main` branch the following fails in the LLVM backend:
```py
def test_print_list():
    a: list[str] = ["ab", "abc", "abcd"]

    print(a) # works
    print(a, a) # fails

test_print_list()
```
Also, that pass seems to be only working for single argument list type, not with nested tuple, see:
```py
def test_print_list():
    a: list[tuple[i32, i32]] = [(1,2), (3, 4), (5, 6)]

    print(a) # fails

test_print_list()
```

That will require much work on the pass too.

Moreover, `C`, at present, doesn't use any passes and already has a design that will need slight modifications to make all this work.

I will fix those issues with the pass in a new PR soon to make it work for LLVM.

_Originally posted by @Smit-create in https://github.com/lcompilers/lpython/issues/1405#issuecomment-1374442129_

Smit-create:
FIxes #1410


certik (approved):
I think this looks good. Thanks!

PR #1436, opened 2023-01-18
Renaming ``EnumMember`` and ``UnionRef`` nodes

### LINKED ISSUE #1183: Rename EnumMember -> EnumClassMember
I would still do `EnumClassMember`. I think the class and instance relationship is well established and people understand it. I had to lookup our discussion to figure out what `EnumMember` means.

Essentially, `Struct`, `Enum`, `Union`, etc. they all have a "class" and "instance" and one can access both (with different semantics and consequences). If we don't like the term "class" then let's figure out a different word, to make this clear.

gptsarthak:
Closes #1183 
Renamed the following expressions, as discussed in the issue:
EnumMember -> EnumStaticMember
EnumName -> EnumStaticName
EnumValue -> EnumStaticValue
UnionRef -> UnionStaticRef
(StructStaticMember already exists)

LPython is building successfully and all tests were passed after the changes.


czgdp1807 (changes requested):
Here's what I would recommend to do,

1. `EnumMember` -> `EnumStaticMember`
2. `UnionRef` -> `UnionInstanceMember` (because there is no use-case of `UnionStaticMember`, when we will find it then we will add this new node).

`EnumValue` and `EnumName` are applicable to both use cases, `Color.RED.name` and `color.name` where `color` is a variable of type `Color` (an enum). So renaming these two nodes doesn't make sense as of now.

Also, after doing a clean build of LPython, run, `python run_tests.py -u`, and commit the changes.

Thanks for the PR.

certik:
Looks good to me. @czgdp1807 can you please review & merge?

czgdp1807 (approved):

PR #1437, opened 2023-01-18
Add a test for struct initialization

certik:
This currently only works with the LLVM backend, does not yet work with CPython or C.


certik:
The following backends are not enabled as they fail:
* C: https://github.com/lcompilers/lpython/issues/1438
* CPython: https://github.com/lcompilers/lpython/issues/1439

certik:
It's interesting that the LLVM test works on macOS, but fails on Linux.

certik:
We first need to fix this PR on Linux to pass, then merge it. Then we need to fix LPython to work for this test via the C backend, as well as fix things so that CPython works.

Issue #1433, opened 2023-01-18
The @packed does not work in pure Python mode


certik:
One must use the following technique for packed structures:
```diff
--- a/src/runtime/ltypes/ltypes.py
+++ b/src/runtime/ltypes/ltypes.py
@@ -275,6 +275,7 @@ def convert_to_ctypes_Structure(f):
         fields.append((name, convert_type_to_ctype(ltype_)))
 
     class ctypes_Structure(ctypes.Structure):
+        _pack_ = 1
         _fields_ = fields
 
     ctypes_Structure.__name__ = f.__name__
```
Obviously this would then fail for non-packed structures. 

I have a large example where this is needed, so there is definitely a bug. We need to create a simple example that fails at the CI, and then fix it. The `_pack_` attribute must be set according to the presence of the `@packed` decorator.

certik:
Something like the following should exercise the issue:
```diff
diff --git a/integration_tests/structs_18.py b/integration_tests/structs_18.py
index 45ba10d1b..a28ba1fdf 100644
--- a/integration_tests/structs_18.py
+++ b/integration_tests/structs_18.py
@@ -5,6 +5,7 @@ from copy import deepcopy
 @packed
 @dataclass
 class buffer_struct:
+    tmp: i8
     buffer8: i8[32]
     buffer1: i32[32]
     buffer2: f32[32]
@@ -18,6 +19,7 @@ class buffer_struct:
 @packed
 @dataclass
 class buffer_struct_clink:
+    tmp: i8
     buffer8: i8[32]
     buffer1: i32[32]
     buffer2: f32[32]
diff --git a/integration_tests/structs_18b.c b/integration_tests/structs_18b.c
index fb24bd593..9a8cb93d8 100644
--- a/integration_tests/structs_18b.c
+++ b/integration_tests/structs_18b.c
@@ -10,6 +10,7 @@ int16_t sum_buffer_i16(int16_t data[], int32_t size) {
 }
 
 struct __attribute__((packed)) buffer_c {
+    int8_t tmp;
     int8_t buffer8[32];
     int32_t buffer1[32];
     float buffer2[32];
```
But then I found out that this test does not run in cpython, since this:
```diff
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -367,7 +367,7 @@ RUN(NAME structs_14          LABELS cpython llvm c)
 RUN(NAME structs_15          LABELS cpython llvm c)
[...]

PR #1441, opened 2023-01-19
Support structs originating in C and read/modified in C/Python

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1433
Closes https://github.com/lcompilers/lpython/pull/1435

Works for both packed and unpacked without any change in LPython's code base. Only `ltypes.py` has been changed to make it work with CPython.


certik (approved):
Excellent, beautiful, thanks for fixing it!

I am going to merge it, as this is a great improvement. I will then test it if things break by removing the `_pack_`, hopefully they do.

Issue #1439, opened 2023-01-18
p_c_pointer does not work in CPython


certik:
Here is an example:
```python
from ltypes import i32, CPtr, dataclass, c_p_pointer, p_c_pointer, \
        pointer, empty_c_void_p

@dataclass
class S:
    a: i32

def f(c: CPtr):
    # The following two lines are the actual test: we take an argument `c` of
    # type CPtr, convert to type Pointer[S] and access the member `a`. This
    # tests that the order of initialization is preserved in the generated LLVM
    # or C code.
    p: Pointer[S] = c_p_pointer(c, S)
    A: i32 = p.a

    # We check that we get the expected result:
    print(A)
    assert A == 3

def main():
    s: S = S(3)
    print(s.a)
    p: CPtr = empty_c_void_p()
    p_c_pointer(pointer(s, S), p)
    f(p)

main()
```
The above example has been committed in https://github.com/lcompilers/lpython/pull/1437 into `integration_tests/struct_19.py`. This produces:

```console
$ PYTHONPATH=../src/runtime/ltypes python structs_19.py
3
1
Traceback (most recent call last):
  File "/Users/ondrej/repos/lpython/integration_tests/structs_19.py", line 27, in <module>
    main()
  File "/Users/ondrej/repos/lpython/integration_tests/structs_19.py", line 25, in main
    f(p)
  File "/Users/ondrej/repos/lpython/integration_tests/structs_19.py", line 18, in f
    assert A == 3
AssertionError
```

But it works with LLVM:
```console
$ lpython structs_19.py 
3
3
```

The issue is that `p_c_pointer(pointer(s, S), p)` can't set the `p` variable correctly.

Issue #1438, opened 2023-01-18
C backend: Wrong order of initialization for structs


certik:
Here is a minimal example:
```python
from ltypes import i32, CPtr, dataclass, c_p_pointer, p_c_pointer, \
        pointer, empty_c_void_p

@dataclass
class S:
    a: i32

def f(c: CPtr):
    # The following two lines are the actual test: we take an argument `c` of
    # type CPtr, convert to type Pointer[S] and access the member `a`. This
    # tests that the order of initialization is preserved in the generated LLVM
    # or C code.
    p: Pointer[S] = c_p_pointer(c, S)
    A: i32 = p.a

    # We check that we get the expected result:
    print(A)
    assert A == 3

def main():
    s: S = S(3)
    print(s.a)
    p: CPtr = empty_c_void_p()
    p_c_pointer(pointer(s, S), p)
    f(p)

main()
```
This produces:
```c
void f(void* c)
{
    struct S* p;
    int32_t A = p->a;
    p = (struct S*) c;
    A = p->a;
    printf("%d\n", A);
    ASSERT(A == 3);
}
```
Which will segfault at the line `int32_t A = p->a;` because `p` is not initialized. Also notice that `A` is initialized twice, which also should not happen.

The above example has been committed in https://github.com/lcompilers/lpython/pull/1437 into `integration_tests/struct_19.py`.

There are two bugs to fix here:
* [ ] Wrong order of initialization (segfault)
* [ ] Double initializaiton

PR #1445, opened 2023-01-20
Fix initialisation issues in C and improve CPython support

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1439
Closes https://github.com/lcompilers/lpython/issues/1438
Closes https://github.com/lcompilers/lpython/pull/1437

Some points to be noted,

1. In C backend, we can only control what shouldn't be initialised at declaration. Double initialisation will always be there in C/LLVM backend because if we don't initialise after declaration then tests like https://github.com/lcompilers/lpython/blob/main/integration_tests/expr_07.py won't work. So, the issue #1438 is fixed by not initialising A at declaration (that's the only thing we can do to fix this issue). How we decide that? We see if all dependencies of A were initialised only then we initialise A. Otherwise we don't.
2. #1439 is fixed by generating `ctypes.Structure` for dataclasses decorated by `ccallable`. This also allows us to fix the segfaults reported in https://github.com/lcompilers/lpython/issues/1433#issuecomment-1396196635 (i.e., structs originating in CPython and accessed in C. I have enabled `cpython` for `structs_18` and it works).
3. The only thing which will have partial support is writing to indices of complex type array members of dataclasses decorated with callable. This is because ctypes doesn't have complex types and hence I defined a ctypes.Structure to represent those. Assigning a Python object to ctypes.Structure is not allowed, so that's a limitation of ctypes.  Rest everything will work (or can be made work).

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -367,13 +367,13 @@ RUN(NAME structs_14          LABELS cpython llvm c)
 RUN(NAME structs_15          LABELS cpython llvm c)
 RUN(NAME structs_16          LABELS cpython llvm c)
 RUN(NAME structs_17          LABELS cpython llvm c)
-RUN(NAME structs_18          LABELS llvm c
+RUN(NAME structs_18          LABELS cpython llvm c
     EXTRAFILES structs_18b.c)
 RUN(NAME structs_19          LABELS cpython llvm c
     EXTRAFILES structs_19b.c)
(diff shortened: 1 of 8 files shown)


certik:
I think this is good to merge. TODO in subsequent PRs:

* Make CPtrPointer an expression: https://github.com/lcompilers/lpython/issues/1449
* make p_c_pointer() a function, not a statement

certik (approved):

PR #1454, opened 2023-01-23
WASMDecoder: Replace while loop with individual decode intructions

### LINKED ISSUE #1453: WAT: Bug in WASM Decoding
**Example:**
```python
def main0():
    print("01234567891") # string of length 11
    print("hi")

main0()
```
**Output:**
```bash
(lp) lpython$ lpython examples/expr2.py --show-wat                                        
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/bin/lpython.cpp", line 1609
    return emit_wat(arg_file, runtime_library_dir, compiler_options);
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/bin/lpython.cpp", line 351
    LCompilers::Result<std::string> res = LCompilers::wasm_to_wat(r2.result, al, diagnostics);
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/libasr/codegen/wasm_to_wat.cpp", line 392
    std::string wat;
  File "../libasr/codegen/wasm_decoder.h", line 312
  File "../libasr/codegen/wasm_decoder.h", line 273
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/libasr/codegen/wasm_utils.cpp", line 76
    LCOMPILERS_ASSERT(offset < code.size());
AssertFailed: offset < code.size()
(lp) lpython$ 
```

Shaikh-Ubaid:
fixes #1453

### DIFF
--- a/src/libasr/codegen/wasm_decoder.h
+++ b/src/libasr/codegen/wasm_decoder.h
@@ -264,11 +264,15 @@ class WASMDecoder {
             data_segments.p[i].insts_start_index = offset;
 
             // read i32.const
-            assert(read_b8(wasm_bytes, offset) == 0x41);
+            if (read_b8(wasm_bytes, offset) != 0x41) {
+                throw CodeGenError("DecodeDataSection: Invalid byte for i32.const");
+            }
             // read the integer (memory location)
             read_i32(wasm_bytes, offset);
             // read expr end
-            assert(read_b8(wasm_bytes, offset) == 0x0B);
+            if (read_b8(wasm_bytes, offset) != 0x0B) {
+                throw CodeGenError("DecodeDataSection: Invalid byte for expr end");
+            }
 
             uint32_t text_size = read_u32(wasm_bytes, offset);
             data_segments.p[i].text.resize(
--- a/src/libasr/codegen/wasm_decoder.h
+++ b/src/libasr/codegen/wasm_decoder.h
@@ -262,8 +262,13 @@ class WASMDecoder {
             }
 
             data_segments.p[i].insts_start_index = offset;
-            while (read_b8(wasm_bytes, offset) != 0x0B)
-                ;
+
+            // read i32.const
+            assert(read_b8(wasm_bytes, offset) == 0x41);
+            // read the integer (memory location)
+            read_i32(wasm_bytes, offset);
+            // read expr end
+            assert(read_b8(wasm_bytes, offset) == 0x0B);
 
             uint32_t text_size = read_u32(wasm_bytes, offset);
             data_segments.p[i].text.resize(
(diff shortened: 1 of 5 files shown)


src/libasr/codegen/wasm_decoder.h
```diff
@@ -262,8 +262,13 @@ class WASMDecoder {
             }
 
             data_segments.p[i].insts_start_index = offset;
-            while (read_b8(wasm_bytes, offset) != 0x0B)
```
Shaikh-Ubaid, line 265: For the following example shared at #1453 
```python
def main0():
    print("01234567891") # string of length 11
    print("hi")

main0()
```
The crux of the issue is that strings length `11` when encoded using `leb128` signed integer is `0x0B` which is also the `byte` for expression/instructions end in `wasm`.

So, the above `while` loop terminates at the first `0x0B` encountered (which is of `11`'s encoding and does not represent expression/instructions end). As the loop terminates early, the remaining bytes of `wasm_bytes` are not meaningful and thus leads to unexpected errors like in https://github.com/lcompilers/lpython/issues/1453#issue-1553212840 and https://github.com/lcompilers/lpython/issues/1453#issuecomment-1400471921.

Shaikh-Ubaid:

czgdp1807 (changes requested):
Explicit code (if small enough) is better than a loop. However, could you please add a test where your fix works and `main` branch fails?

Shaikh-Ubaid:
>Explicit code (if small enough) is better than a loop. However, could you please add a test where your fix works and main branch fails?

Thank you so much for the valuable feedback. I just added a test in [WAT: Add and enable test for wasm decoding bug](https://github.com/lcompilers/lpython/pull/1454/commits/807d0b3d15780107b747f22db64d26f4a5e6d3f3).


tests/print_str.py
```diff
@@ -0,0 +1,13 @@
...
+   print("x > y")
+   print("x !> y")
+   print("x == y")
+   print("x != y")
+
+
+def main0():
+   printCompOperations()
+
+main0()
```
czgdp1807, line 13: Why we cannot have it as an integration test?

Shaikh-Ubaid: The bug is during decoding of `wasm`. The simplest case where decoding of `wasm` takes place is during `wat` generation. Hence, I added the test as `wat` reference test.

Another case where decoding of `wasm` takes place is in the `wasm_x64`/`wasm_x86` backends during `x64/x86` generation. It seems possible to add it as an integration test by adding the same test and enabling the `wasm_x64` backend. But it seems that might indicate that the issue/bug/corner-case (that the test file is testing) is in the `wasm_x64` backend during `x64` generation. In actual, the issue is in `wasm` decoding, so it seems more natural to add as a `wat` reference test.

certik (approved):
It's fine with me.

czgdp1807 (approved):

Shaikh-Ubaid:
Thank you for the approval.

PR #1455, opened 2023-01-24
C: Support generics

Smit-create:


src/libasr/codegen/asr_to_c.cpp
```diff
@@ -516,6 +516,9 @@ class ASRToCVisitor : public BaseCCPPVisitor<ASRToCVisitor>
                     sub = format_type_c("", "const " + const_underlying_type + " ",
                                         v.m_name, false, false);
                 }
+            } else if (ASR::is_a<ASR::TypeParameter_t>(*v_m_type)) {
```
czgdp1807, line 519: If I am not wrong backend shouldn't need to care if there are any generics in the frontend? The code for each type combination should be generated by the `instantiate_template` pass? Correct @certik?

certik: Yes, everything should be already instantiated, so the generics should work in all backends, no special logic needed.

czgdp1807: So this PR should never have been needed. @Smit-create Can you check what's there in ASR which is causing you to handle generics in C backend? I guess ASR is not correct that's why you created this PR for C backend.

Smit-create: Opened an issue for it's discussion: #1471

certik:
Converted to a Draft for now.

src/libasr/codegen/asr_to_c.cpp
Smit-create: As per the discussion in #1471, the current ASR is fine and this diff fixes the codegen of generics in C backend.

czgdp1807 (approved):

PR #1456, opened 2023-01-24
WASM: Support bitwise operations

Shaikh-Ubaid:
This PR adds support for bitwise operators in the `wasm` backend.


src/libasr/codegen/asr_to_wasm.cpp
```diff
@@ -941,6 +957,28 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
...
 
+    void visit_IntegerBitNot(const ASR::IntegerBitNot_t &x) {
+        if (x.m_value) {
+            visit_expr(*x.m_value);
+            return;
+        }
+        this->visit_expr(*x.m_arg);
+        ASR::Integer_t *i = ASR::down_cast<ASR::Integer_t>(x.m_type);
+        // there is no direct bit-invert inst in wasm,
+        // so xor-ing with -1 (sequence of 32/64 1s)
```
Shaikh-Ubaid, line 968: More Info: https://github.com/WebAssembly/design/issues/379

Shaikh-Ubaid:
For `integration_tests`, I am unsure if we should have all bit operations testing in a single file or if we should split testing in different files for each bitwise operator. I currently tested them all in same file and there are two files, one for `32-bit` integers and other for `64-bit` integers.

Shaikh-Ubaid:

certik (approved):

PR #1474, opened 2023-01-26
Synchronizing libasr's template instantiation with LFortran

ansharlubis:
Removed functionalities unique to LPython from libasr.


certik:
@ansharlubis is this ready for review?

czgdp1807:
@ansharlubis Please mark it ready for review whenever you want it to be reviewed. I review non-draft open PRs on a daily basis, so will go through yours as well when you mark it as ready for review.

ansharlubis:
> @ansharlubis Please mark it ready for review whenever you want it to be reviewed. I review non-draft open PRs on a daily basis, so will go through yours as well when you mark it as ready for review.

Got it! I was waiting for the test to pass before requesting a review. (cc @certik )

certik:
@czgdp1807 go ahead and review please.

czgdp1807 (approved):
My next feature request is to support accepting any rank arrays in a function. I need it to implement reduction function in NumPy such as `np.sum()`, `np.max()`, `np.argmax()`. See an example below,

```python
def sum(array: T[...]) -> T:
    result: T = 0
    for t in itertools.permutations(array.shape):
        result += array[t]
    return result
```

`T[...]` encodes the type of any rank array and the code should be generated according to the function call.

With an `axis` argument things would be slightly different but its doable with a logic similar to above.

certik:
See https://github.com/lfortran/lfortran/issues/1199#issuecomment-1400750193 for any rank. @czgdp1807 can you please merge this PR if it can be merged?

czgdp1807:
> See [lfortran/lfortran#1199 (comment)](https://github.com/lfortran/lfortran/issues/1199#issuecomment-1400750193) for any rank.

Yes, the above comment is for Fortran. For LPython we need a different syntax. One of which is I proposed in https://github.com/lcompilers/lpython/pull/1474#pullrequestreview-1272108760. I guess we should we have a multi-index type. So, we will be able to do `t: MultiIndex[...]` and then use it to index `array: T[...]`.

certik:
I am mainly concerned about the ASR details, I am not worried about syntax right now. So the work is the same for both LPython and LFortran.

PR #1476, opened 2023-01-27
C: Generalize handing of extra codegen

### LINKED ISSUE #1472: Fix C backend for List/Tuple Constants
This is one of the examples where the current implementation fails:
```py
from ltypes import i32

def f(y: list[i32]) -> list[i32]:
    y.append(4)
    return y

def g():
    print(f([1,2,3]))

g()
```

This will fail in the C backend because of the way we handle constant lists/tuples in it. We treat them as normal variables while at the same time generating the code for them whenever we come across them. This creates troubles in many `stmt` nodes and requires a good fix which is to generate them and declare at the top along with all other variables in the scope.

Smit-create:
Fixes #1472


src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -198,6 +199,27 @@ R"(#include <stdio.h>
         src = unit_src;
     }
 
+    void get_expr_src(ASR::expr_t *arg) {
```
certik, line 202: Can you document what this function does in a comment?

certik: When should one use it instead of `visit_expr` directly?

czgdp1807: Well this won't prevent direct usage of `visit_expr` in future. So we will have to keep collecting special cases again and add into this or another function. And the cycle will continue. 

Also, we won't know why some special case is handled here or which handling is for which node (thus making debugging difficult). Doing directly inside the node helps in figuring out the reasoning behind a particular check. 

IMO, such refactoring would make more sense in the long term future when LPython/LFortran will stabilise.

certik: What special case does it handle that has to be handled every time?

Smit-create: If we come across `ListConstant`, `TupleConstant`, `DictConstant`, etc. which have some extra code-generation, we need to handle it in every node using if-else. So to avoid those repetitions in each node, I tried to extract these cases and wrap `visit_expr`.

czgdp1807: Then probably make it explicit with a function like `is_constant_aggregate_data_structure` and then call your new function. It should be explicit that why we are calling different function than `visit_expr`. But again not all nodes have been implemented yet so probably repeating code that handles special cases for now is okay. We will at least know what happens when we add a new constant and how existing node implementations are affected due to it.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -198,6 +199,27 @@ R"(#include <stdio.h>
...
+        self().visit_expr(*arg);
+        if (ASR::is_a<ASR::ListConstant_t>(*arg) ||
+                ASR::is_a<ASR::TupleConstant_t>(*arg) ||
+                ASR::is_a<ASR::DictConstant_t>(*arg)) {
+            src = const_var_names[get_hash((ASR::asr_t*)arg)];
+        } else if (ASR::is_a<ASR::ListSection_t>(*arg)) {
+            ASR::ListSection_t *l_sec = ASR::down_cast<ASR::ListSection_t>(arg);
+            src = "*" + const_var_names[get_hash((ASR::asr_t*)l_sec)];
+        }
```
certik, line 211: I think these can go into visit_ListContant, visit_TupleConstant, visit_ListSection, etc. and at the end of the function simply put the line `src = const_var_names[get_hash((ASR::asr_t*)arg)];` etc. Then remove this `get_expr_src`.

Smit-create:
I have updated the PR and now the logic seems very intuitive and generalized to handle the extra code-generation of the constants.

czgdp1807 (approved):

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -116,6 +116,7 @@ class BaseCCPPVisitor : public ASR::BaseVisitor<Struct>
     std::string const_name;
     size_t const_vars_count;
     size_t loop_end_count;
+    int bracket_open;
```
certik, line 119: Can you please add documentation here what this variable is used for, where it needs to be incremented/decremented (essentially what is the "contract" that a developer has to follow when adding new functionality)?

certik (approved):
This is fine. Thanks for fixing it. 

The documentation can be added in a separate PR.

src/libasr/codegen/asr_to_c_cpp.h
Smit-create: Sure, will do that. Thanks

PR #1489, opened 2023-02-01
XX Try moving \r into whitespace

certik:


Thirumalai-Shaktivel:
Yup, this will work, but:

Consider an example
```
:\r\n
```
- with this PR changes (skipped in the whitespace):
```
(TOKEN ":") 1:1
(NEWLINE) 3:3
```
- with the previous changes
`\r\n` is considered one
```
(TOKEN ":") 1:1
(NEWLINE) 2:3
```

```
$ python -m tokenized a.py
...
1,0-1,1:            OP             ':'            
1,1-1,3:            NEWLINE        '\r\n'
...
```

certik:
So to much the current master with Python, we take first/last from AST, and do "first-1 : last" which gives the location printed by Python for a given token. So master prints:
```
(TOKEN ":") 1:1
(NEWLINE) 2:3
```
Which corresponds to the following in Python:
```
(TOKEN ":") 0:1
(NEWLINE) 1:3
```
Which is exactly what Python prints:
```
1,0-1,1:            OP             ':'            
1,1-1,3:            NEWLINE        '\r\n'
```

While this PR would give:
```
(TOKEN ":") 1:1
(NEWLINE) 3:3
```
which would correspond to:
```
(TOKEN ":") 0:1
(NEWLINE) 2:3
```


But is this a problem? All tests passed, so it didn't seem to break any error message.

I am fine leaving it as is in master. We can revisit this later.

PR #1491, opened 2023-02-02
[ASRPass] Wrap all the global symbols into a module

Thirumalai-Shaktivel:
This PR aims to move all the symbols from the global scope (except main_program) into the Module (_global_scope).
Now the TranslationUnit contains only two symbols: Module (_global_scope) and a Program (main_program).
For more details and discussion: https://github.com/lcompilers/lpython/pull/1384


src/libasr/codegen/asr_to_c.cpp
```diff
@@ -645,6 +645,32 @@ R"(
...
+            }
+
+            std::vector<std::string> struct_deps = ASRUtils::order_deps(struct_dep_graph);
+
+            for (auto &item : struct_deps) {
+                ASR::symbol_t* struct_sym = global_symbols->get_symbol(item);
+                visit_symbol(*struct_sym);
+                array_types_decls += src;
+            }
+        }
```
certik, line 672: Why do we need special handling of this global_symbols module? I was hoping that since it's just a regular module, it would not need any special handling like this.

src/libasr/asr_scopes.cpp
```diff
@@ -135,4 +135,104 @@ std::string SymbolTable::get_unique_name(const std::string &name) {
...
+                ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(a.second);
+                v->m_parent_symtab = module_scope;
+                module_scope->add_symbol(a.first, (ASR::symbol_t *) v);
+                syms.push_back(al, s2c(al, a.first));
+                break;
+            } default : {
+                throw LCompilersException("Moving the symbol:`" + a.first +
+                    "` from global scope is not implemented yet");
+            };
+        }
```
czgdp1807, line 234: This isn't scalable. I would do it by writing a visitor in `asdl_cpp.py`. Will you be able to try the class design and implement it? Give it a try for a couple of days, if not then we will merge this with a `TODO`.

czgdp1807 (changes requested):

src/libasr/codegen/asr_to_c.cpp
Thirumalai-Shaktivel: Yup, here in C backend, I think the visit_module is not yet implemented, except `lfortran_intrinsics` module.

czgdp1807: Correct. Probably this can be removed. I guess you can just do `if(ASR::is_a<ASR::Module_t>(*item.second))` and call `visit_Module(*ASR::down_cast<ASR::Module_t>(item.second))` inside the `if` block by creating a new loop similar to the following,

https://github.com/lcompilers/lpython/blob/8f62ac4ff1acebdeb621557d8008a5a8423455bb/src/libasr/codegen/asr_to_c.cpp#L613-L623

Please comment here if you face any blocker or issues while following this approach.

src/libasr/pass/array_op.cpp
```diff
@@ -182,6 +182,38 @@ class ArrayOpVisitor : public PassUtils::PassVisitor<ArrayOpVisitor>
         }
     }
 
+    void visit_Module(const ASR::Module_t &x) {
+        if (strcmp(x.m_name, "_global_symbols") == 0) {
+            // FIXME: this is a hack, we need to pass in a non-const `x`,
```
certik, line 187: try removing this if check. If it does not work, let's figure out the real bug why it doesn't work.

src/libasr/pass/subroutine_from_function.cpp
```diff
@@ -94,6 +94,38 @@ class CreateFunctionFromSubroutine: public PassUtils::PassVisitor<CreateFunction
             }
         }
 
+        void visit_Module(const ASR::Module_t &x) {
+            if (strcmp(x.m_name, "_global_symbols") == 0) {
+                // FIXME: this is a hack, we need to pass in a non-const `x`,
```
certik, line 99: Same here.

certik:
Let's see if we can just construct the new module, but then treat it like any other module, no special treatment. That would be the cleanest design.

src/libasr/asr_scopes.cpp
Thirumalai-Shaktivel: Added this in TODO

certik:
@Thirumalai-Shaktivel let me know when this is ready for review.

Thirumalai-Shaktivel:
Yup, I made the required changes. This is ready for review.

certik:
Thanks! This requires a thorough review, I'll try to do it soon. It looks good at first sight thought.

@czgdp1807 can you also please review this?

integration_tests/CMakeLists.txt
```diff
@@ -427,4 +427,5 @@ RUN(NAME comp_01 LABELS cpython llvm wasm wasm_x64)
 RUN(NAME bit_operations_i32 LABELS cpython llvm wasm wasm_x64)
 RUN(NAME bit_operations_i64 LABELS cpython llvm wasm)
 
-RUN(NAME test_argv_01    LABELS llvm) # TODO: Test using CPython
+RUN(NAME test_argv_01      LABELS llvm) # TODO: Test using CPython
+RUN(NAME global_syms_01    LABELS cpython)
```
czgdp1807, line 431: Why not `llvm` and `c` here?

Thirumalai-Shaktivel: In LLVM, the global lists are not yet handled. After merging this I will start working on it.
C: I didn't try it. How do I verify this works in C?
Using `lpython integration_tests/global_syms_01.py --show-c`?

czgdp1807: > In LLVM, the global lists are not yet handled. After merging this I will start working on it.

Any rough idea on how that will be managed. I think LLVM doesn't allow `builder->CreateAlloca` in global scope. Worth researching approaches (implementation in LLVM is not at all required just a rough idea that there is a possibility to make it work in LLVM) once this   before merging this because the success of this might rely on the possibility of actually using it in LLVM.

One approach in my mind is to create `void*` typed global variables in the global scope and then bitcast them into `list` LLVM type inside LLVM functions. That way any manipulation will happen inside LLVM functions.

Other approach is to see how clang++ does it for global variables like `std::vector` in C++. For example the below code,

```cpp
#include <iostream>
#include <vector>

std::vector<int> vec;

int main() {
    return 0;
}
```

`clang++`'s LLVM output,

```llvm
; ModuleID = '/Users/czgdp1807/lpython_project/debug.cpp'
source_filename = "/Users/czgdp1807/lpython_project/debug.cpp"
target datalayout = "e-m:o-i64:64-i128:128-n32:64-S128"
target triple = "arm64-apple-macosx13.0.0"

%"class.std::__1::vector" = type { %"class.std::__1::__vector_base" }
%"class.std::__1::__vector_base" = type { i32*, i32*, %"class.std::__1::__compressed_pair" }
%"class.std::__1::__compressed_pair" = type { %"struct.std::__1::__compressed_pair_elem" }
%"struct.std::__1::__compressed_pair_elem" = type { i32* }
%"struct.std::__1::nullptr_t" = type { i8* }
%"struct.std::__1::__default_init_tag" = type { i8 }
%"class.std::__1::__vector_base_common" = type { i8 }
%"struct.std::__1::__compressed_pair_elem.0" = type { i8 }
%"class.std::__1::allocator" = type { i8 }
%"struct.std::__1::__non_trivial_if" = type { i8 }

@vec = global %"class.std::__1::vector" zeroinitializer, align 8
@__dso_handle = external hidden global i8
@llvm.global_ctors = appending global [1 x { i32, void ()*, i8* }] [{ i32, void ()*, i8* } { i32 65535, void ()* @_GLOBAL__sub_I_debug.cpp, i8* null }]

; Function Attrs: noinline ssp uwtable
define internal void @__cxx_global_var_init() #0 section "__TEXT,__StaticInit,regular,pure_instructions" {
  %1 = call %"class.std::__1::vector"* @_ZNSt3__16vectorIiNS_9allocatorIiEEEC1Ev(%"class.std::__1::vector"* @vec)
  %2 = call i32 @__cxa_atexit(void (i8*)* bitcast (%"class.std::__1::vector"* (%"class.std::__1::vector"*)* @_ZNSt3__16vectorIiNS_9allocatorIiEEED1Ev to void (i8*)*), i8* bitcast (%"class.std::__1::vector"* @vec to i8*), i8* @__dso_handle) #2
  ret void
}
```

It can be seen that it creates global variables but initialises them inside a function. So you can do something similar for lists as well. Also, worth noting that we will have to create a function something like `global_module_function` in LLVM and then call inside `main`. But that should be doable.

> C: I didn't try it. How do I verify this works in C?
> Using `lpython integration_tests/global_syms_01.py --show-c`?

If LLVM is not enabled then C is not worth enabling. LLVM should always work with a given ASR design then other backends will automatically follow.

czgdp1807: I am approving it as I think this is clean enough at the ASR as well as LLVM/C level.

czgdp1807 (approved):

integration_tests/CMakeLists.txt
certik: @czgdp1807 the question here is how to handle module variables, as well as program variables. They are scoped to the module or program, but are essentially global. WASM and x86 have the same issue (@Shaikh-Ubaid and I discussed this a few times). It seems one must allocate the global variable (using the platform dependent mechanism), and then initialize it from the main function as needed.

czgdp1807: So seems like global statements are already being wrapped into a main function in the ASR itself? Well I would like to see what we will need to do when we enable `llvm` for this test. Let’s merge it for now (I have approved earlier today).

certik: Let me review this first.

certik (approved):
I think that this is fine. I reviewed it and didn't find anything major.

Thirumalai-Shaktivel:
Thanks for the approvals.

src/libasr/asr_verify.cpp
```diff
@@ -578,21 +578,23 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
...
                 "ExternalSymbol::m_module_name `" + x_m_module_name
                 + "` must match external's module name `" + asr_owner_name + "`");
-            ASR::symbol_t *s = nullptr;
-            if( m != nullptr && ((ASR::symbol_t*) m == ASRUtils::get_asr_owner(x.m_external)) ) {
-                s = m->m_symtab->find_scoped_symbol(x.m_original_name, x.n_scope_names, x.m_scope_names);
-            } else if( sm ) {
-                s = sm->m_symtab->resolve_symbol(std::string(x.m_original_name));
-            } else if( em ) {
-                s = em->m_symtab->resolve_symbol(std::string(x.m_original_name));
+            if (asr_owner_name != "_global_symbols") {
```
certik, line 581: Why is special handling required here?

I would document it why it is needed. I think we should check `_global_symbols` too.

Thirumalai-Shaktivel: The issue for which I made these changes, seems to be fixed. So, I'm reverting this back.
Sorry, I didn't notice this before.

certik: ok!

certik (approved):
I think this is fine to merge and iterate on later.

Issue #1461, opened 2023-01-24
Ensure the WASM output is compatible with wasmtime + WASI


certik:
We should ensure that the subset of WASM that we generate in the ASR->WASM is compatible with wasmtime + WASI, so that one can use wasmtime and other tools like (wasm2c) out of the box.

certik:
The goal here would be to do:
```console
$ lfortran --backend=wasm examples/expr2.f90 -o expr2.wasm
$ wasmtime expr2.wasm
Error: failed to run main module `expr2.wasm`

Caused by:
    0: failed to instantiate "expr2.wasm"
    1: unknown import: `js::print_i32` has not been defined
```
We should standardize the WASM output that we generate, and it seems the change is minor, in this example the WAT looks like:
```console
$ lfortran --show-wat examples/expr2.f90              
(module
    (type (;0;) (func (param i32) (result)))
    (type (;1;) (func (param i64) (result)))
    (type (;2;) (func (param f32) (result)))
    (type (;3;) (func (param f64) (result)))
    (type (;4;) (func (param i32 i32) (result)))
    (type (;5;) (func (param) (result)))
    (type (;6;) (func (param i32) (result)))
    (type (;7;) (func (param) (result)))
    (import "js" "print_i32" (func (;0;) (type 0)))
    (import "js" "print_i64" (func (;1;) (type 1)))
    (import "js" "print_f32" (func (;2;) (type 2)))
    (import "js" "print_f64" (func (;3;) (type 3)))
    (import "js" "print_str" (func (;4;) (type 4)))
    (import "js" "flush_buf" (func (;5;) (type 5)))
    (import "js" "set_exit_code" (func (;6;) (type 6)))
    (import "js" "memory" (memory (;0;) 100 100))
    (func $7 (type 7) (param) (result)
        (local i32)
        i32.const 25
        local.set 0
        local.get 0
        call 0
        call 5
        i32.const 0
        call 6
        return
    )
    (export "_lcompilers_main" (func $7))
)
```
And we just need to change the imports:
```wat
    (import "js" "print_i32" (func (;0;) (type 0)))
    (import "js" "print_i64" (func (;1;) (type 1)))
    (import "js" "print_f32" (func (;2;) (type 2)))
    (import "js" "print_f64" (func (;3;) (type 3)))
    (import "js" "print_str" (func (;4;) (type 4)))
    (import "js" "flush_buf" (func (;5;) (type 5)))
[...]

PR #1497, opened 2023-02-03
WASI: Initial Support

Shaikh-Ubaid:
This PR implements initial support of `WASI` in the `wasm` and `wasm_x64` backends.

### DIFF
--- a/src/libasr/codegen/asr_to_wasm.cpp
+++ b/src/libasr/codegen/asr_to_wasm.cpp
@@ -332,6 +332,12 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
             wasm::emit_b8(m_code_section, m_al, wasm::type::i64);
         }
 
+        // locals 0 is given parameter
+        // locals 1 is digits_cnt
+        // locals 2 is divisor (in powers of 10)
+        // locals 3 is loop counter (counts upto digits_cnt (which is decreasing))
+        // locals 4 is extra copy of given parameter
+
         emit_if_else([&](){
             wasm::emit_get_local(m_code_section, m_al, 0);
             wasm::emit_i64_const(m_code_section, m_al, 0);
@@ -368,11 +374,8 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
             wasm::emit_i64_add(m_code_section, m_al);
             wasm::emit_set_local(m_code_section, m_al, 1);
             wasm::emit_get_local(m_code_section, m_al, 0);
-            wasm::emit_f64_convert_i64_s(m_code_section, m_al);
             wasm::emit_i64_const(m_code_section, m_al, 10);
-            wasm::emit_f64_convert_i64_s(m_code_section, m_al);
-            wasm::emit_f64_div(m_code_section, m_al);
-            wasm::emit_i64_trunc_f64_s(m_code_section, m_al);
+            wasm::emit_i64_div_s(m_code_section, m_al);
             wasm::emit_set_local(m_code_section, m_al, 0);
         });
 
@@ -389,19 +392,12 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
             wasm::emit_i64_const(m_code_section, m_al, 1);
             wasm::emit_set_local(m_code_section, m_al, 2);
             wasm::emit_i64_const(m_code_section, m_al, 0);
-            wasm::emit_i64_const(m_code_section, m_al, 1);
-            wasm::emit_i64_sub(m_code_section, m_al);
             wasm::emit_set_local(m_code_section, m_al, 3);
 
             emit_loop([&](){
                 wasm::emit_get_local(m_code_section, m_al, 3);
-                wasm::emit_i64_const(m_code_section, m_al, 1);
-                wasm::emit_i64_add(m_code_section, m_al);
-
                 wasm::emit_get_local(m_code_section, m_al, 1);
-                wasm::emit_i64_const(m_code_section, m_al, 1);
-                wasm::emit_i64_sub(m_code_section, m_al);
-                wasm::emit_i64_le_s(m_code_section, m_al);
+                wasm::emit_i64_lt_s(m_code_section, m_al);
             }, [&](){
                 wasm::emit_get_local(m_code_section, m_al, 3);
                 wasm::emit_i64_const(m_code_section, m_al, 1);
@@ -415,11 +411,8 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
 
 
             wasm::emit_get_local(m_code_section, m_al, 4);
-            wasm::emit_f64_convert_i64_s(m_code_section, m_al);
             wasm::emit_get_local(m_code_section, m_al, 2);
-            wasm::emit_f64_convert_i64_s(m_code_section, m_al);
-            wasm::emit_f64_div(m_code_section, m_al);
-            wasm::emit_i64_trunc_f64_s(m_code_section, m_al);
+            wasm::emit_i64_div_s(m_code_section, m_al);
             wasm::emit_i64_const(m_code_section, m_al, 10);
             wasm::emit_i64_rem_s(m_code_section, m_al);
 
--- a/src/libasr/codegen/asr_to_wasm.cpp
+++ b/src/libasr/codegen/asr_to_wasm.cpp
@@ -89,6 +89,7 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
     uint32_t no_of_imports;
     uint32_t no_of_data_segments;
     uint32_t avail_mem_loc;
+    uint32_t digits_mem_loc;
 
     uint32_t min_no_pages;
     uint32_t max_no_pages;
@@ -96,6 +97,7 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
     std::map<uint64_t, uint32_t> m_var_name_idx_map;
     std::map<uint64_t, SymbolFuncInfo *> m_func_name_idx_map;
     std::map<std::string, ASR::asr_t *> m_import_func_asr_map;
+    std::map<std::string, uint32_t> m_string_to_iov_loc_map;
 
    public:
     ASRToWASMVisitor(Allocator &al, diag::Diagnostics &diagnostics)
@@ -180,10 +182,24 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
                                        ASR::down_cast<ASR::symbol_t>(variable));
             params.push_back(m_al, ASRUtils::EXPR(var));
         }
-
+        ASR::expr_t* return_var = nullptr;
+        if (import_func.result_types.size() > 0) {
+            var_idx = 0;
+            auto param = import_func.result_types[var_idx];
+            auto type = get_import_func_var_type(param);
+            auto variable = ASR::make_Variable_t(
+                m_al, global_scope_loc, nullptr,
+                s2c(m_al, std::to_string(var_idx)), nullptr, 0, ASR::intentType::Out,
+                nullptr, nullptr, ASR::storage_typeType::Default,
+                ASRUtils::TYPE(type), ASR::abiType::Source,
+                ASR::accessType::Public, ASR::presenceType::Required, false);
+            auto var = ASR::make_Var_t(m_al, global_scope_loc,
+                                       ASR::down_cast<ASR::symbol_t>(variable));
+            return_var = ASRUtils::EXPR(var);
+        }
         auto func = ASR::make_Function_t(
             m_al, global_scope_loc, global_scope, s2c(m_al, import_func.name),
-            nullptr, 0, params.data(), params.size(), nullptr, 0, nullptr,
+            nullptr, 0, params.data(), params.size(), nullptr, 0, return_var,
             ASR::abiType::Source, ASR::accessType::Public,
             ASR::deftypeType::Implementation, nullptr, false, false, false, false, false,
             nullptr, 0, nullptr, 0, false, false, false);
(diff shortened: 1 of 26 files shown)


Shaikh-Ubaid:
Currently, https://github.com/lcompilers/lpython/issues/1461#issuecomment-1410863607 seems to work.

```bash
(lp) lpython$ cat examples/expr2.py 
from sys import exit
def main0():
    a: i32 = 67
    exit(a)

main0()
(lp) lpython$ lpython examples/expr2.py --backend wasm -o tmp
(lp) lpython$ wasmtime tmp
(lp) lpython$ echo $?
67
(lp) lpython$ 
```

src/libasr/codegen/wasm_assembler.h
```diff
@@ -168,6 +168,20 @@ void emit_import_fn(Vec<uint8_t> &code, Allocator &al,
     emit_u32(code, al, type_idx);
 }
 
+void emit_declare_mem(Vec<uint8_t> &code, Allocator &al,
+                     uint32_t min_no_pages, uint32_t max_no_pages = 0) {
+    if (max_no_pages) {
```
certik, line 173: To make it clear:
```suggestion
    if (max_no_pages > 0) {
```

certik:
Beautiful. I think this approach will force us to "clean up" our code generation to make it "standardized".

src/libasr/codegen/wasm_assembler.h
Shaikh-Ubaid: Got it. Thank you for the guidance. I will update it in upcoming commits.

Shaikh-Ubaid:
**Current Output:**
```bash
(lp) lpython$ cat examples/expr2.py 
def main0():
    print("hi", "1234", "jldakfjs!jdlkasf#9u843*)(")
    print("abcdefgh1234")
    print(-1234)
    print(6579352)
    print("hello")

main0()
(lp) lpython$ lpython examples/expr2.py --backend wasm -o tmp
(lp) lpython$ wasmtime tmp
hi, 1234, jldakfjs!jdlkasf#9u843*)(
abcdefgh1234
-####
#######
hello
(lp) lpython$ 
```
- String printing currently works
- For integers, we are currently printing `#` instead of the exact digit.

src/libasr/codegen/asr_to_wasm.cpp
```diff
@@ -1623,17 +1787,35 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
...
+        wasm::emit_i32_store(m_code_section, m_al, wasm::mem_align::b8, 0);
+
+        m_string_to_iov_loc_map[str] = avail_mem_loc;
+
+        avail_mem_loc += 8;
+        wasm::emit_str_const(m_data_section, m_al, avail_mem_loc, str);
         avail_mem_loc += str.length();
         no_of_data_segments++;
+
+        avail_mem_loc = (int(ceil(avail_mem_loc / 4.0)) << 2); /* align mem location to 4 bytes */
```
Shaikh-Ubaid, line 1812: **Line 1812:** It seems that `fd_write` works only when its `iov` vector (it stores the start location of the string and the size of the string) is aligned by `4`. It doesn't seem to print any output when the location of the `iov` vector is not a multiple of `4`.

We previously had memory aligned by single byte. It seems we might need the memory alignment as `4` bytes in future.

Shaikh-Ubaid:
**doubts/concerns:**
1. Is the current approach the expected way to print integers?
2. Is the current way easily maintainable?
    - It seems that in case of `X86Assembler`, we have a good abstraction, i.e. if we have required function instructions, we can use them and we need not dive into bytes much.
    - In case of `wasm`, it seems that, as a programmer, we are needing to touch bytes. For example: when implementing `if-else`, `loop`, the `return` instruction or for defining a whole new function.  
     - To define a function in `x86`, it seems simpler as we add a label and start defining the function. In case of `wasm`, we are needing to define the signature of the function in the `type` section, then reference the signature in the `function` section, add the function code in `code` section and then export the function in `export` section. For each section, whenever we add a new element, we need to (without missing) increment the variable `no_of_element` corresponding to that section. Any minor change could lead to incorrect binary which could be tricky to debug. 
     - I am unsure if the current approach is easily maintainable/extendable.

Shaikh-Ubaid:
Printing integers seems to work:
```bash
(lp) lpython$ cat examples/expr2.py               
def main0():
    print("hi", "1234", "jldakfjs!jdlkasf#9u843*)(")
    print("abcdefgh1234")
    print(-1234)
    print(6579352, -123, "hi", 45897, "bye", 0, -1, 1, -10, 10, "done.")
    print("hello")
    print(2, -2, 10, -10, 1000, -10000)

main0()
(lp) lpython$ lpython examples/expr2.py --backend wasm -o tmp
(lp) lpython$ wasmtime tmp                                   
hi, 1234, jldakfjs!jdlkasf#9u843*)(
abcdefgh1234
-1234
6579352, -123, hi, 45897, bye, 0, -1, 1, -10, 10, done.
hello
2, -2, 10, -10, 1000, -10000
(lp) lpython$ 
```

src/libasr/codegen/asr_to_wasm.cpp
certik: Good point.

certik:
Great job.

I think this is maintainable for now. I think we might later need to add a class that handles the WASM as a format, and exposes the API using capabilities like "add a function" and it internally takes care of the details. That way one cannot make a mistake.

src/libasr/codegen/asr_to_wasm.cpp
```diff
@@ -243,6 +267,203 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
...
+            wasm::emit_u32(m_code_section, m_al, 1u); // count of local vars of this type
+            wasm::emit_b8(m_code_section, m_al, wasm::type::i64);
+        }
+
+        emit_string("-");
+        emit_string("#");
+        uint32_t digits_mem_loc = avail_mem_loc;
+        for (int i = 0; i < 10; i++) {
+            emit_string(std::to_string(i));
+        }
```
Shaikh-Ubaid, line 340: This is currently inefficient. The `emit_string` has some `i32.store` instructions inside which would be executed everytime the `print_i64` is called. Ideally, they are only needed to be called once when storing string into memory. (The subsequent calls to `print_i64` store the same values at the same locations in memory.)

Ideally, we should extract this out in some other function which we could call once. Since, it seems to work currently, I guess we can fix it in a separate PR.

tests/reference/wat-print_str-385e953.stdout
```diff
@@ -1,76 +1,484 @@
...
+    (data (;6;) (i32.const 84) "4")
+    (data (;7;) (i32.const 96) "5")
+    (data (;8;) (i32.const 108) "6")
+    (data (;9;) (i32.const 120) "7")
+    (data (;10;) (i32.const 132) "8")
+    (data (;11;) (i32.const 144) "9")
+    (data (;12;) (i32.const 156) ".")
+    (data (;13;) (i32.const 168) "x < y")
+    (data (;14;) (i32.const 184) "
+")
```
Shaikh-Ubaid, line 478: `newline` character (which is needed/printed by `print` after printing all arguments) is being shown as above in the generated  `WAT`.

Shaikh-Ubaid:
This might still fail at the CI.

**Currently:**
- All the `wasm` tests pass.
- Support for `wasm_x64` (and `wasm_x86`) is needed.

certik:
This looks good, just the WASM->x64 needs to be fixed so that the CI passes and we can merge this.

src/libasr/codegen/wasm_to_x64.cpp
```diff
@@ -309,6 +326,23 @@ class X64Visitor : public WASMDecoder<X64Visitor>,
...
+        // m_a.asm_mov_m64_r64(&base, nullptr, 1, 0, X64Reg::rbx);
+
+        // TODO: Memory values shared by wasm correspond to i32 integers
+        // Storing each integer using 8 bytes could lead
+        // to different memory location from that shared by wasm.
+        // Fix this in upcoming PRs.
+        m_a.asm_mov_r32_label(X86Reg::esi, "master_string"); // base_memory
+        m_a.asm_add_r32_r32(X86Reg::esi, X86Reg::eax); // base_memory + location
+        X86Reg base = X86Reg::esi;
+        m_a.asm_mov_m32_r32(&base, nullptr, 1, 0, X86Reg::ebx);
```
Shaikh-Ubaid, line 343: **Line 343:** The assembly code segment faults at this line. I am unsure, but it seems the memory area where we are storing strings (and their location, length information) probably is read only. If that is the case, we might need to search for other options.

I will make a small example and confirm if the memory area is actually read-only or not.

Shaikh-Ubaid: Yes, it seems it is read-only.

**Small Example:**
```asm
_start:
    mov r10, base_memory
    mov r8, 15
    mov [r10], r8 ; writing to base_memory location
    mov rdi, 23 ; custom exit code
    mov rax, 60
    syscall

base_memory:
    db 0x20
    db 0x20
    db 0x20
    db 0x20
```
```bash
lpython$ nasm -fbin simple.asm && chmod +x simple && ./simple
lpython$ echo $?
139
```

Shaikh-Ubaid: **PS:** I figured a better approach which could/would tackle it (and hopefully make the strings implementation simpler). It (hopefully) would also fix the inefficiency shared at https://github.com/lcompilers/lpython/pull/1497#discussion_r1097932266.

Shaikh-Ubaid: The better approach involves directly encoding the `iov` vector into a data segment (i.e. in the form of a string). This approach works in `wasm` (and hopefully would work with `wasm_x64`) but the `wat` generated doesn't seem to print any output.

**PS:** This was occurring due to incorrect conversion of `byte` to string. Using `sprintf(byte.data(), "%02x", uint8_t(ch))` instead of `sprintf(byte.data(), "%02x", int(ch))` seems to fix the issue.

Shaikh-Ubaid:
Now, printing partially works in `wasm_x64` backend (`string` printing works, integer and float printing yet to be supported/fixed).

certik:
Nice, great progress!

Shaikh-Ubaid:
It works now!

## WASM:

**Binary:**
```bash
(lp) lpython$ lpython integration_tests/print_05.py --backend wasm -o tmp
(lp) lpython$ wasmtime tmp
hi 1234 jldakfjs!jdlkasf#9u843*)(
abcdefgh1234
-1234
6579352 -123 hi 45897 bye 0 -1 1 -10 10 done.
hello
2 -2 10 -10 1000 -10000
3.14000000
-1000.10000000
-1000.00099999
abcdefgh1234
0 0.000000000 abcd 0 efghijk 0.000000000 1^3 -1.000000000 pqrswxyz 1.000000000 mno 2.000000000 p -5.09999999 -5 zz
pi = 3.14000000
```
**WAT Text:**
```bash
(lp) lpython$ lpython integration_tests/print_05.py --show-wat > main.wat
(lp) lpython$ wasmtime main.wat
hi 1234 jldakfjs!jdlkasf#9u843*)(
abcdefgh1234
-1234
6579352 -123 hi 45897 bye 0 -1 1 -10 10 done.
hello
2 -2 10 -10 1000 -10000
3.14000000
-1000.10000000
-1000.00099999
abcdefgh1234
0 0.000000000 abcd 0 efghijk 0.000000000 1^3 -1.000000000 pqrswxyz 1.000000000 mno 2.000000000 p -5.09999999 -5 zz
pi = 3.14000000
```

## ASM:

**Binary:**
```bash
(wasm_asm) lpython$ lpython integration_tests/print_05.py --backend wasm_x64 -o tmp > main.asm
(wasm_asm) lpython$ ./tmp
hi 1234 jldakfjs!jdlkasf#9u843*)(
abcdefgh1234
-1234
6579352 -123 hi 45897 bye 0 -1 1 -10 10 done.
hello
2 -2 10 -10 1000 -10000
3.14000000
-1000.10000000
-1000.00099999
abcdefgh1234
0 0.000000000 abcd 0 efghijk 0.000000000 1^3 -1.000000000 pqrswxyz 1.000000000 mno 2.000000000 p -5.09999999 -5 zz
pi = 3.14000000
```

**ASM Text:**
```bash
(wasm_asm) lpython$ nasm -fbin main.asm && chmod +x main && ./main
hi 1234 jldakfjs!jdlkasf#9u843*)(
abcdefgh1234
-1234
6579352 -123 hi 45897 bye 0 -1 1 -10 10 done.
hello
2 -2 10 -10 1000 -10000
3.14000000
-1000.10000000
-1000.00099999
abcdefgh1234
0 0.000000000 abcd 0 efghijk 0.000000000 1^3 -1.000000000 pqrswxyz 1.000000000 mno 2.000000000 p -5.09999999 -5 zz
pi = 3.14000000
```

Shaikh-Ubaid:
I disabled testing for the `wasm_x86` backend in [CI: Disable testing of WASM_X86 backend](https://github.com/lcompilers/lpython/pull/1497/commits/081926348ccbd454ae8ace59b6c0e4032779ce08). Our current `wasm-wasi` implementation uses `64-bit` integers (at some places) and the `wasm_x86` backend supports only `32-bit` integers.

src/libasr/codegen/wasm_assembler.h
Shaikh-Ubaid: This has been fixed in [WASM: Make memory max_no_of_pages condition clearer](https://github.com/lcompilers/lpython/pull/1497/commits/e46b457040140170ea7e44723891fa013843863a).

src/libasr/codegen/asr_to_wasm.cpp
Shaikh-Ubaid: This has been fixed now. The `emit_string` now works at compile-time. It does not use any `i32.store` instructions. The string `location` and `length` information is directly encoded in the form of a `wasm` data segment at compile-time.

tests/reference/wat-print_str-385e953.stdout
Shaikh-Ubaid: This has been fixed. The `newline` character is now **escaped** and is printed in its `byte` form (`\0a`) as follows:

```
(data (;3;) (i32.const 24) "\0a   ")
```

Shaikh-Ubaid:
I organised the changes in structured commits. This is ready. Please review and share feedback.

certik:
I reviewed all the changes I think this is beautiful, everything seems to work.

Is the `wasm_x86` completely broken, or do some tests work? I would keep the backend running, but maybe disable the particular tests that do not work.

Will it be difficult to update the online GUI to work with these changes?

Shaikh-Ubaid:
>Is the wasm_x86 completely broken, or do some tests work?

`wasm` always includes the `print_int` function (for printing to `stdout` using `fd_write()`) which uses variables of type `i64` which is not supported in the `wasm_x86` backend.

Shaikh-Ubaid:
>Will it be difficult to update the online GUI to work with these changes?

It could indeed present some challenges (and would thus be a great opportunity to learn and grow). We need to see if there are any `WASI` polyfills available for browser (https://github.com/bjorn3/browser_wasi_shim seems to be one).

I am currently concerned about making the `wasm` and `wasm_x64` backends more robust. Especially, simplifying the `wasm` backend and abstracting away the details. For example: When someone adds a new type, he/she should not need to increment/update the variable `no_of_types`. Another concern is related to  accessing memory in `assembly`. The linear memory in `wasm` is editable (we can store values in it). We need to see if there is such memory (apart from the execution stack) available in assembly.

certik:
I would get wasm_x86 working, not fully, but in some limited capacity, so that we can keep testing it and ensuring it works (in its limited capacity). The way I would do it is implement WASM 64bit integers in wasm_x86 (you have to use pair of registers or operate on memory in pairs of 32bit operations), then you can just use the existing WASM code. That can be done later. For now, why don't you get some simple tests that only print strings (but not numbers) working again? That way we have at least something. Or if that is too complicated, then at least returning zero / nonzero return value. Then we can merge this PR and implement the rest later.

Regarding the web UI, I would implement the functions that we use ourselves. That way the implementation will be small and clean.

Shaikh-Ubaid:
>I would get wasm_x86 working, not fully, but in some limited capacity, so that we can keep testing it and ensuring it works (in its limited capacity). The way I would do it is implement WASM 64bit integers in wasm_x86 (you have to use pair of registers or operate on memory in pairs of 32bit operations), then you can just use the existing WASM code. That can be done later.

Sure, Sir. Using pairs of `32-bit` registers/instructions is a great idea.

>For now, why don't you get some simple tests that only print strings (but not numbers) working again? That way we have at least something. Or if that is too complicated, then at least returning zero / nonzero return value. Then we can merge this PR and implement the rest later.

Got it.

Thank you for the guidance.

Shaikh-Ubaid:
If it is fine, let us tackle the WebUI and the assembly (heap) memory later.

src/libasr/codegen/wasm_to_x86.cpp
```diff
@@ -329,9 +348,92 @@ class X86Visitor : public WASMDecoder<X86Visitor>,
...
-    void visit_F64Const(double Z) {
-        float z = Z; // down cast 64-bit double to 32-bit float
-        std::string label = "float_" + std::to_string(z);
+    void visit_I64Const(int64_t value) {
+        m_a.asm_push_imm32(value);
+    }
+
+    void visit_I64ExtendI32S() {
+        // empty, since all i32's are already considered as i64's currently.
+    }
```
Shaikh-Ubaid, line 357: For `i64` types, we are currently storing and managing them as `i32`. This is way we simply get the backend working with `i64`'s and passing the `CI`. Please share if this approach is fine.

certik: Yes, that's perfect.

certik:
I think this looks good, let's see if tests pass.

Shaikh-Ubaid:
This PR is ready.

>I think this looks good, let's see if tests pass.

Sure.

Shaikh-Ubaid:
The `integration_tests` for `wasm_x86` seem to pass on my system locally.

There is an integer overflow when printing the value of `w` in the following.
https://github.com/lcompilers/lpython/blob/00ba081f0bf0bb52e9606ba7bfcbfbb0d195ff38/integration_tests/print_03.py#L6-L10

On my system locally,  it just prints the `-` symbol but does not fail/terminate with an error. I guess it could be failing at the `CI`.
```bash
(lp) lpython$ lpython integration_tests/print_03.py --backend wasm_x86 -o tmp
(lp) lpython$ ./tmp
-25
100
2147483647
-
Hi
Hello
21
Bye
```

Shaikh-Ubaid:
I am sorry about https://github.com/lcompilers/lpython/pull/1497#issuecomment-1426240561. I see it was the `x86` backend (instead of `wasm_x86`) that was failing.

What shall we do about [TEST: Slightly modify test to support it in wasm_x86](https://github.com/lcompilers/lpython/pull/1497/commits/84782a38610c30f83b9160ccf099c0270ce888e1)?

Shaikh-Ubaid:
This is ready.

certik (approved):
I think this is good to go in!

I reviewed most of it, I don't see any major issue now. Great job, this is excellent progress.

Shaikh-Ubaid:
Thank you for the approval and merging!

PR #1500, opened 2023-02-04
Fixed Printing of Escape Sequences

### LINKED ISSUE #1473: Escape sequence recognized but not ignored in printing.
The parser identifies backslash as an escape sequence and allows the printing but the `\` also gets printed.

```
    s: str
    s = 'I said 'Wow' '  # Without escape sequence 
    print(s)

  File "C:\Users\kunni\lpython\try.py", line 3
    s = 'I said 'Wow''
                 ^^^
SyntaxError: invalid syntax

LCompilersException: The string is not recognized by the parser.
```
```
    s: str
    s = 'I said \'Wow\' '  # With escape sequence
    print(s)

I said 'Wow'

I said \'Wow\'
```

gptsarthak:
Fixes #1473
Modified LPython semantics, C and Julia backend so that they do not print the backslash even after it is recognized.

When asked to print the following string: 
```
s: str
s = 'I said \'Wow\' '
print(s)
``` 
LPython now prints:
`I said 'Wow' `

### DIFF
--- a/src/lpython/parser/semantics.h
+++ b/src/lpython/parser/semantics.h
@@ -800,10 +800,10 @@ char* unescape(Allocator &al, LCompilers::Str &s) {
         if (s.p[idx] == '\\' && s.p[idx+1] == 'n') {
             x += "\n";
             idx++;
-        }else if (s.p[idx] == '\\' && s.p[idx+1] == '\'') {
+        } else if (s.p[idx] == '\\' && s.p[idx+1] == '\'') {
             x += "'";
             idx++;
-        }else {
+        } else {
             x += s.p[idx];
         }
     }
--- a/src/lpython/parser/semantics.h
+++ b/src/lpython/parser/semantics.h
@@ -800,7 +800,10 @@ char* unescape(Allocator &al, LCompilers::Str &s) {
         if (s.p[idx] == '\\' && s.p[idx+1] == 'n') {
             x += "\n";
             idx++;
-        } else {
+        }else if (s.p[idx] == '\\' && s.p[idx+1] == '\'') {
+            x += "'";
+            idx++;
+        }else {
             x += s.p[idx];
         }
     }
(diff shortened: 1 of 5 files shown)


src/lpython/parser/semantics.h
```diff
@@ -800,7 +800,10 @@ char* unescape(Allocator &al, LCompilers::Str &s) {
         if (s.p[idx] == '\\' && s.p[idx+1] == 'n') {
             x += "\n";
             idx++;
-        } else {
+        }else if (s.p[idx] == '\\' && s.p[idx+1] == '\'') {
```
certik, line 803:
```suggestion
        } else if (s.p[idx] == '\\' && s.p[idx+1] == '\'') {
```

src/lpython/parser/semantics.h
```diff
@@ -800,7 +800,10 @@ char* unescape(Allocator &al, LCompilers::Str &s) {
...
+            x += "'";
+            idx++;
+        }else {
```
certik, line 806:
```suggestion
        } else {
```

tests/reference/ast_new-string2-44323ea.stdout
```diff
@@ -8,5 +8,5 @@ class a:
 " ())] [])) (Expr (ConstantStr "
     "\\"
 " ())) (Expr (ConstantStr "
-    '\\'
+    '\'
```
certik, line 11: I checked against CPython and this change seems to be correct.

certik (approved):
I think this looks good, thanks for fixing it! I left some minor formatting comments.

certik:
Thanks @gptsarthak for the fix!

faze-geek:
The obvious edge case to be checked here was for `" "` instead of `' '` :)
Just as I was writing this comment the PR was merged.
```
def main():
	print("I said \"hi\" ")

main()

I said "hi"

I said \"hi\"
```
@gptsarthak Shouldn't be difficult to add this. Thanks !

gptsarthak:
Hey @faze-geek thanks for pointing it out! Now that this PR is merged, should I make another PR after adding it?

faze-geek:
> Hey @faze-geek thanks for pointing it out! Now that this PR is merged, should I make another PR after adding it?

Yes, you can do that!
I noticed you made the changes from your main. It's better to always make a new branch instead.

PR #1502, opened 2023-02-05
Increased Support for Printing Escape Sequences

### LINKED ISSUE #1473: Escape sequence recognized but not ignored in printing.

gptsarthak:
Closes #1473 

LPython now replicates python behavior with all types of combinations of escape sequences.
```
def test_new_line():
    print('I said \'Wow\' ')
    print("I said \"hi\" ")
    print("I said \\\"hi\" ")
    print("abc \\ ") 
    print("abc \\\ wow")

test_new_line()

(lp) sarthak@pop-os:~/lpython/examples$ python test.py
I said 'Wow' 
I said "hi" 
I said \"hi" 
abc \ 
abc \\ wow

(lp) sarthak@pop-os:~/lpython/examples$ lpython test.py
I said 'Wow' 
I said "hi" 
I said \"hi" 
abc \ 
abc \\ wow
```


src/lpython/parser/semantics.h
```diff
@@ -803,6 +803,15 @@ char* unescape(Allocator &al, LCompilers::Str &s) {
         } else if (s.p[idx] == '\\' && s.p[idx+1] == '\'') {
             x += "'";
             idx++;
+        } else if (s.p[idx] == '\\' && s.p[idx+1] == '\"') {
```
faze-geek, line 806: Could you recheck for this case -

A tricky one I came across while randomly trying out a bunch of strings. Although a very improbable input, I would like this pr to perfectly handle all cases.
```
print('I said \'\\\'hi\\''hey\' ')

I said '\'hi\hey'

I said '\'hi\hey\'
```

Thanks !

gptsarthak: I have observed that LPython generates incorrect output for the following cases:
```
print('  ''hey\' ')
print(" " " hey\" ")
```
Meaning, when there are 2 strings in a print function, which have no comma or any operator between them. This is accepted behavior in python.
Like you said this is an impossible input, since a user should put an operator like a comma between two strings, in which case lpython executes correctly.
I do not yet understand why lpython outputs that escape sequence after 'hey', but will try after my classes end. But right now I think that this might be a more parser related issue.

czgdp1807: > But right now I think that this might be a more parser related issue.

Try checking the AST using `--show-ast` command line argument in `lpython`.

gptsarthak: I was kinda right - lpython considers these both of the print statements I wrote as single strings and ignores the quotes in the middle, which somehow leads to the problem of printing the escape sequence.
I believe that in  ("a" " b") or ('a' 'b'), a and b should be considered two different strings by lpython, which it does not.  
Some AST results for further investigations:
```
print('a  b\' ')
[(ConstantStr "a  b' " ())]                           // Correct

print('a'' b\' ')
[(ConstantStr "a b\' " ())]                           // Incorrect (Same AST but '\' also printed)

print('a'              ' b\' ')
[(ConstantStr "a b\' " ())]                           // Incorrect (Whitespace ignored) 

print('a' , ' b\' ')
[(ConstantStr "a" ()) (ConstantStr " b' " ())]        // Correct (Two strings because of comma) 

print("a " " b\' ")
[(ConstantStr "a  b\' " ())]                          // Incorrect (Same issue with double quotes)

print("a' ' b\' ")
[(ConstantStr "a' ' b' " ())]                         // Correct (Nested quotes work)
```
I  think we should open a new issue for this, if it is a problem that needs to be solved.

Thirumalai-Shaktivel: Yup, thanks for investigating. You can create a new issue or directly open a new PR and work towards fixing it.
If you get any doubts regarding the parser, do ping me I can help you.

czgdp1807 (changes requested):
Please add a runnable test for this. See https://github.com/lcompilers/lpython/tree/main/integration_tests for the pattern.

integration_tests/test_str_03.py
```diff
@@ -5,6 +5,14 @@ def test_new_line():
     print("\ndef")
     print("x\nyz")
 
+def test_escape_sequences():
+    print('I said \'Wow\' ')
+    print("I said \"hi\" ")
+    print("I said \\\"hi\" ")
+    print('I said \'\\\'hi\\''hey\' ')
+    print('a''b')
+    print("a" "b")
```
Thirumalai-Shaktivel, line 14: Can you please add this test to `tests/parser/string2.py` as well? (so that we test the AST changes)

tests/reference/ast_new-string2-44323ea.stdout
```diff
@@ -1,10 +1,7 @@
-(Module [(Expr (ConstantStr "\"\\" ())) (Expr (ConstantStr "
-\\" ())) (Expr (ConstantStr "\\" ())) (Expr (ConstantStr "   \\" ())) (Expr (ConstantStr "\\ " ())) (Expr (ConstantStr "\
-" ())) (Expr (ConstantStr "\\|" ())) (Expr (ConstantStr "Text\\" ())) (Expr (ConstantStr "\
-" ())) (Expr (Call (Name x Load) [(ConstantStr "\
-class a:
+(Module [(Expr (ConstantStr ""\" ())) (Expr (ConstantStr "
+\" ())) (Expr (ConstantStr "\" ())) (Expr (ConstantStr "   \" ())) (Expr (ConstantStr "\ " ())) (Expr (ConstantStr "\" ())) (Expr (ConstantStr "\|" ())) (Expr (ConstantStr "Text\\" ())) (Expr (ConstantStr "\" ())) (Expr (Call (Name x Load) [(ConstantStr "\class a:
```
Thirumalai-Shaktivel, line 2: Here, `""\"`, `"\"`, ... seems to be wrong.

Can you cross check with Python's output and update this? you can use `python -m ast tests/parser/string2.py` to verify the Python's AST.

src/lpython/parser/semantics.h
```diff
@@ -774,9 +799,12 @@ static inline ast_t* concat_string(Allocator &al, Location &l,
             throw LCompilers::LPython::parser_local::ParserError(
                 "The byte and non-byte literals can not be combined", l);
         }
-    } else {
+    } else {
```
Thirumalai-Shaktivel, line 802:
```suggestion
    } else {
```

src/lpython/parser/semantics.h
```diff
@@ -720,6 +720,31 @@ static inline ast_t* BOOLOP_01(Allocator &al, Location &loc,
...
+            idx++;
+        } else if (s.p[idx] == '\\' && s.p[idx+1] == '\"') {
+            x += '"';
+            idx++;
+        } else if (s.p[idx] == '\\' && s.p[idx+1] == ' ') {
+            x += "\\ ";
+            idx++;
+        } else if (s.p[idx] == '\\') {
+            x += "\\";
+            idx++;
```
Thirumalai-Shaktivel, line 740: Comment this, check the output, uncomment or add new lines, ...; play around with this. Please make sure that you get the same output as CPython ast

gptsarthak:
In the latest commit :

- Added code to escape the strings in the AST. This required me to edit the ```asdl_cpp.py``` file to emit the correct code in the ```visit_ConstantStr``` function in the Pickle Visitor class.

- Made the strings in LPython AST be same as you would get in ```python -m AST {file-name}``` by escaping newline

- Fixed printing of backslash when used as a continuation character at the end of a line. We already had tests for this in string2.py and many other files, but they produced incorrect AST and output before. 

- Refactored code according to suggested changes.

- Added test in parser/string2.py

There is one issue left to solve. After making the changes mentioned in first point, I observed that ```visit_ConstantBytes``` is also getting edited along the ```visit_ConstantStr``` function. I could not find a way to differentiate between the two inside ```make_visitor```. If that is a problem, we can discuss and fix it asap. After that I believe we can merge this PR.

src/libasr/asdl_cpp.py
```diff
@@ -1445,6 +1445,19 @@ def visitField(self, field, cons):
...
+                    self.emit(  "for (size_t idx=0; idx < str.size(); idx++) {",2)
+                    self.emit(      "if (str[idx] == \'\\n\') {",3)
+                    self.emit(          "s += \"\\\\n\";",4)
+                    self.emit(      "} else if (str[idx] == '\\\\') {",3)
+                    self.emit(          "s += \"\\\\\\\\\";",4)
+                    self.emit(      "} else {",3)
+                    self.emit(          "s += str[idx];",4)
+                    self.emit(      "}",3)
+                    self.emit(  "}",2)
+                    self.emit(  "s.append(\"\\\"\");",2)
```
Thirumalai-Shaktivel, line 1460: I think we have to handle this in `semantics.h` itself.

tests/reference/ast_new-string2-44323ea.stdout
```diff
@@ -1,12 +1 @@
...
-" ())) (Expr (Call (Name x Load) [(ConstantStr "\
-class a:
-    def test():
-        return r"\\LaTeX"
-" ())] [])) (Expr (ConstantStr "
-    "\\"
-" ())) (Expr (ConstantStr "
-    '\'
-" ()))] [])
+(Module [(Expr (ConstantStr ""\\" ())) (Expr (ConstantStr "\n\\" ())) (Expr (ConstantStr "\\" ())) (Expr (ConstantStr "   \\" ())) (Expr (ConstantStr "\\ " ())) (Expr (ConstantStr "" ())) (Expr (ConstantStr "\\|" ())) (Expr (ConstantStr "I said 'Wow' " ())) (Expr (ConstantStr "I said "hi" " ())) (Expr (ConstantStr "I said \\"hi" " ())) (Expr (ConstantStr "I said '\\'hi\\hey' " ())) (Expr (ConstantStr "ab" ())) (Expr (ConstantStr "ab" ())) (Expr (ConstantStr "Text\\\\" ())) (Expr (ConstantStr "" ())) (Expr (Call (Name x Load) [(ConstantStr "class a:\n    def test():\n        return r"\\LaTeX"\n" ())] [])) (Expr (ConstantStr "\n    "\\\\"\n" ())) (Expr (ConstantStr "\n    '\\'\n" ()))] [])
```
Thirumalai-Shaktivel, line 1: Everything seems to be correct. Except `""\\"`, which should be `'"\\'` or `"\"\\"`. I think we don't unescape `"` in this PR. Instead, Let's create a new issue or work on another PR.

Thirumalai-Shaktivel:
Thank you, @gptsarthak for working on this. It was very much required.
I have left some comments, you can work on it if you know what to do. Otherwise, I will try to work on it coming weekend.

src/libasr/asdl_cpp.py
gptsarthak: I have tried handling this in the semantics alone, but could not do it because that changes the output of the code. It changes the ```value``` in the ```ConstantStr``` node. If we change that, we'll have to handle this again in the backend (```python_ast_to_asr.cpp``` i guess) to get the same output as python. If that is what we need I'll try to do that myself first, otherwise you should take over this PR. Thank you

tests/reference/ast_new-string2-44323ea.stdout
gptsarthak: Yes, please create a new issue for this for a separate discussion.
Right now what I think is, in the ```ConstantStr``` node we use ```string value``` while python use ```constant value```. I think this is why our strings are wrapped in double quotes instead of single quotes, which is the only difference I see here.

gptsarthak: Yes, please create a new issue to work on this.
Right now what i think is, in Lpython's ```ConstantStr``` node we use ```string value``` while python uses ```constant value```. I think this is why the strings are wrapped in double quotes instead of single quotes, which is the only difference I see here. If fixing this is simply not unescaping ```"```, that will be easy and I'll do it in my next commit in this PR only.

Thirumalai-Shaktivel:
@gptsarthak I made the required changes.
Now, you can review and test it. If you have any doubts or find any mistakes, please ping me.

czgdp1807 (approved):
Looks okay to me.

anutosh491:
LGTM ! I did some testing and it looks fine .

Thirumalai-Shaktivel:
Thanks for the approval and testing.

Thirumalai-Shaktivel:
@gptsarthak is it okay if I merge this?
Do you wanna test or make some other changes??

czgdp1807:
Conflicts should be resolved here.

gptsarthak:
I believe I resolved the conflict. Its my first time dealing with them so hopefully it went right. 
@Thirumalai-Shaktivel thanks for working on this! I did some tests and they ran fine. Your commits taught me a lot and gave an insight on how the issue was supposed to be solved. It can be merged now.
I believe I can help out with https://github.com/lcompilers/lpython/pull/1529#discussion_r1111224014 after this is merged.

certik:
I think this PR will have to be fixed. It seems it keeps the strings escaped in ASR. I think that's not a good design, because then we need to be unescaping the strings in almost all the backends. Also, the escaping and unescaping will be different in every surface language, in particular it is different in Fortran and Python.

Rather, a proper design is what we had in master before this PR, that is:

* Strings should be unescaped, represented as utf-8 strings, directly. I think they can be null terminated, that's probably fine. 
* The frontend's job is to unescape the strings, based on the surface level language rules.
* The backend's job is to properly escape the string as needed, based on the language, it will in general be different for Julia, Python, Fortran, C, etc. Most lowering backends like LLVM and WASM will not be escaping the string at all.

Issue #1475, opened 2023-01-26
For loop should be implemented for strings without declaration 


faze-geek:
Related to #206 which was closed couple of days ago.

```
a :str
for a in "abc":
      print (a)

(lp) C:\Users\kunni\lpython>python try.py
a
b
c

(lp) C:\Users\kunni\lpython>src\bin\lpython try.py
semantic error: Only function call `range(..)` supported as for loop iteration for now
 --> try.py:3:5 - 4:17
  |
3 |        for a in "abc":
  |        ^^^^^^^^^^^^^^^...
...
  |
4 |            print (a)
  | ...^^^^^^^^^^^^^^^^^
```
Considering `range` is the only supported function, a string input is not another function like `reverse(str_name)` or `enumerate(str_name)` so a better error message can be provided if this is not in the implementation plan as of now.

certik:
I think we should implement this. You can change the error message if you want in the meantime.

faze-geek:
> I think we should implement this. You can change the error message if you want in the meantime.

Yes sure! I'll be attempting a few PR's once `run_tests` starts correctly working with windows. If I'm not wrong #1463 is a shot at that.

anutosh491:
I've been trying to address this issue , but not sure what approach could be used for the same . Let's say we want to address
```
i: i32
for i in [1, 2, 3]:
     print(i)
```
Here the iterable we have is of type `AST::List_t` and we currently can't address this case but we can address cases where the list has been assigned to a variable .
```
i: i32
lst: list[i32] = [1, 2, 3]
for i in list:
      print(i)
```
Here the iterable is of type `AST::Name_t`  . So what comes to my mind for addressing the first case is ... that why don't we assign a dummy name to the list and create an object of `AST::Name_t` type and repeat the procedure what's being done already . Not sure if this is the way to go here @Thirumalai-Shaktivel !?

PR #1503, opened 2023-02-05
Updated error message related to for loops being used on undeclared iterables

anutosh491:
Addresses https://github.com/lcompilers/lpython/issues/1475#issuecomment-1405246377

On the comitted branch we now get `semantic error: reverse(..) is not supported as for loop iteration for now` rather than `semantic error: Only range(..) supported as for loop iteration for now` which we get on main for the following
```
def main0():
    s : str
    for s in reverse("abc"):
        print(s)

main0()
```

I'll be working on the implementation part of things soon , but being new to lpython , I wanted to get familiar with the development workflow with a relatively smaller pr first.

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -4361,7 +4361,7 @@ public:
                     x.base.base.loc);
             }
             if (call_name != "range") {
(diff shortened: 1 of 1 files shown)


czgdp1807 (approved):
Its fine. Makes sense to me.

anutosh491:
Thanks for the review !

PR #1506, opened 2023-02-06
LLVM: Support logical binop for strings, int, real

### LINKED ISSUE #1487: Logical operation of two integers gives wrong answers.
I haven't done a proper sleuthing before opening this issue. Are logical operations between `integers` (or even `strings`) implemented separately in `lpython` ? 
```
def main0():
    a : i32
    b : i32
    x : i32 
    y : i32 
    or_op1 : i32
    or_op2 : i32
    a = 1
    b = 2
    x = 3
    y = 4
    or_op1 = a or b
    or_op2 = x or y
    print(or_op1,or_op2)

1 3

3 7
```

I'll do a quick similar check for logical `and`,`not` etc. and report if they give different results too !

Smit-create:
Fixes #1487


czgdp1807 (approved):

certik (approved):
Thanks!

integration_tests/logical_binop1.py
```diff
@@ -0,0 +1,50 @@
...
+    x : i32
+    y : i32
+    or_op1 : i32
+    or_op2 : i32
+    a = 1
+    b = 2
+    x = 0
+    y = 4
+    or_op1 = a or b
+    or_op2 = x or y
```
Shaikh-Ubaid, line 15: I think we should explicitly be casting the operands to`bool()` (when are not already of `bool` type). For example:

```python
res: bool
res = bool(a) or bool(b)
```

For `or_op1 = a or b`, since `or_op1` is of type `i32`, I think it should be
```console
or_op1 = i32(bool(a) or bool(b))
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -4651,14 +4651,58 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            cond = builder->CreateFCmpUEQ(left_val, zero);
+        } else if (ASRUtils::is_character(*x.m_type)) {
+            zero = llvm::Constant::getNullValue(character_type);
+            cond = lfortran_str_cmp(left_val, zero, "_lpython_str_compare_eq");
+            result = builder->CreateAlloca(character_type, nullptr);
+        } else if (ASRUtils::is_logical(*x.m_type)) {
+            zero = llvm::ConstantInt::get(context,
+                            llvm::APInt(1, 0));
+            cond = builder->CreateICmpEQ(left_val, zero);
+            result = builder->CreateAlloca(llvm::Type::getInt1Ty(context), nullptr);
```
Shaikh-Ubaid, line 4683: I think this operation should not be done at the backend level. A backend should exactly follow what the ASR says. In this case, it should always return a logical result. This logical result can/should later be casted to the desired type.

Shaikh-Ubaid: Opened an issue here https://github.com/lfortran/lfortran/issues/3620.

integration_tests/logical_binop1.py
certik: It seems the result of `x or y` is an integer, I think we are fine in this case.

src/libasr/codegen/asr_to_llvm.cpp
certik: Yes, this should be explicitly done in ASR, not in the backend.

integration_tests/logical_binop1.py
Shaikh-Ubaid: >It seems the result of x or y is an integer, I think we are fine in this case.

It looks like `or` here is not a logical operation, but instead a selective operation. I think we should have a separate node for this `selective` `or` operation.

Shaikh-Ubaid: Maybe we can name it as `SelectiveOr` which takes two arguments both of any type (but equal types). The return value of this `SelectiveOr` is the same as that of the type of one of the operands.

Issue #1287, opened 2022-11-10
`for-loop` does not work in lpython


akshanshbhatt:
While trying to work on https://github.com/lcompilers/lpython/issues/1228, I tried some of the simplest for-loop statements in lpython such as -
```py
j: i32
for i in range(5):
    j = i
print(j)
```

```py
for i in range(5):
    ...
```

and it didn't work. So, before working on #1228, we need to work on making these work first.

Shaikh-Ubaid:
I think/guess simple `for` loops work in `lpython`. To some extent, `for-in` loop is also supported https://github.com/lcompilers/lpython/pull/775.

akshanshbhatt:
> I think/guess simple `for` loops work in `lpython`. To some extent, `for-in` loop is also supported #775.

Oh, I see- Can you once confirm that the examples I mentioned work on your setup?

Shaikh-Ubaid:
Example:
```python
def main0():
    i: i32
    j: i32
    for i in range(5):
        j = i
        print("Integer is ", j)
main0()

# Not implemented yet in LPython:
#if __name__ == "__main__":
#    main()
```

Output:
```bash
(lp) ubaid@ubaid-Lenovo-ideapad-330-15ARR:~/OpenSource/lpython$ lpython examples/expr2.py 
Integer is  0
Integer is  1
Integer is  2
Integer is  3
Integer is  4
(lp) ubaid@ubaid-Lenovo-ideapad-330-15ARR:~/OpenSource/lpython$ 
```
It seems to work fine on my system.

akshanshbhatt:
> It seems to work fine on my system.

Yeah, I discussed this with @certik yesterday. It works inside function def and not independently.

czgdp1807:
I guess there are many features that don’t work in global scope.

czgdp1807:
I see what's going on here. The code is calling `visit_stmt` of `SymbolTableVisitor` and `visit_For` is only defined in `BodyVisitor`. Seems like its possible to make loops work in global scope by minor changes in how we visit statements in Module in `SymbolTableVisitor`.

czgdp1807:
The following diff makes things work,

```diff
diff --git a/src/lpython/semantics/python_ast_to_asr.cpp b/src/lpython/semantics/python_ast_to_asr.cpp
index 7ad2ca315..77b589e78 100644
--- a/src/lpython/semantics/python_ast_to_asr.cpp
[...]

PR #1509, opened 2023-02-06
Added empty `visit_For` function in the `SymbolTableVisitor` class

anutosh491:
Addresses the first part of https://github.com/lcompilers/lpython/issues/1287#issue-1444377844
```
j: i32
i: i32
for i in range(5):
    j = i
    print("j: ", j)
```
Something like this doesn't give us the expected output on main . It appears that the code is calling the `visit_stmt` function of the `SymbolTableVisitor`, but the `visit_For` function is only defined in the `BodyVisitor`. Hence an empty `visit_For` function in the `SymbolTableVisitor` class should do the job. Once this is done , we have the correct output
```
j:  0
j:  1
j:  2
j:  3
j:  4
```


czgdp1807 (approved):
Its okay to have it. I personally don't see any downsides. @certik Thoughts?

certik:
I think that's fine. Should we add a test for this?

czgdp1807 (changes requested):
Please add a test for this. See the existing pattern of integration_tests and add your own. Thanks.

anutosh491:
> Please add a test for this. See the existing pattern of integration_tests and add your own. Thanks.

Sure , I'll do the needful soon !

integration_tests/loop_04.py
```diff
@@ -0,0 +1,6 @@
+from ltypes import i32
+j: i32
+i: i32
+for i in range(5):
+    j = i
+    print("j: ", j)
```
czgdp1807, line 6: Register it in https://github.com/lcompilers/lpython/blob/main/integration_tests/CMakeLists.txt with `cpython`, `llvm`.

czgdp1807 (changes requested):

integration_tests/loop_04.py
Thirumalai-Shaktivel: @anutosh491 was there any issue?

czgdp1807 (approved):

anutosh491:
Thanks for the reviews on this !

PR #1512, opened 2023-02-11
Fixed compatibility of sep, end keyword with list/tuple containers

### LINKED ISSUE #1477: Variable of list[i32] type has unnecessary newline character while printing.
A new line character is encountered before and after the `list[i32] ` type. I checked integers, floats, strings which give expected answers when printed together.
```
def main1():
    a : i32 = 1
    s : str ="abc"
    l : list[i32] = [1,2,3,4,5]
    print(a,l,s)

main1()

1 [1, 2, 3, 4, 5] abc

1
[1, 2, 3, 4, 5]
abc
```

anutosh491:
Fixes #1477

Some issues that were fixed are
1) `sep` and `end` keyword were assumed to be functioning correctly beforehand with data types like `i32`, `str` etc but trying out some cases showed that they aren't fully compatible . Pointed out in https://github.com/lcompilers/lpython/issues/1477#issuecomment-1422083691 . This has been addressed.

2) Secondly all sorts of variations of print statements involving the `list`/`tuple` containers have been fixed . 
```
def main0():
    a: i32 = 1
    b: i32 = 2
    c: str = "abc"
    d: str = "def"
    e: list[i32] = [1, 2]
    f: list[i32] = [3, 4]
    g: tuple[i32, i32, i32] = (1, 2, 3)
    h: tuple[i32, i32] = (4, 5)
    l : list[i32] = [1,2,3]
    l2 : list[i32] = [4, 5, 6]
    s : str = "abc"
    print(a, g, c,sep = "xyz", end = "pqr")
    print(g, a, c, sep = "xyz", end = "pqr")
    print(a, e, g, sep = "xyz", end = "pqr")
    print(a, l, s)
    print(l , a, s)
    print(a, s, l)
    print(a, l, s, end = " pqr ")
    print(l, a, s, end = " pqr ")
    print(a, s, l, end = " pqr ")
    print(a, l, s, sep = "xyz")
    print(l, a, s, sep = "xyz")
    print(a, s, l, sep = "xyz")
    print(a, l, s, sep = "xyz", end = "pqr")
    print(l, a, s, sep = "xyz", end = "pqr")
    print(a, s, l, sep = "xyz", end = "pqr")
    print(e, f, a, sep = "xyz")
    print(g, e, f, end = "pqr")

main0()
```
```
1xyz(1, 2, 3)xyzabcpqr(1, 2, 3)xyz1xyzabcpqr1xyz[1, 2]xyz(1, 2, 3)pqr1 [1, 2, 3] abc
[1, 2, 3] 1 abc
1 abc [1, 2, 3]
1 [1, 2, 3] abc pqr [1, 2, 3] 1 abc pqr 1 abc [1, 2, 3] pqr 1xyz[1, 2, 3]xyzabc
[1, 2, 3]xyz1xyzabc
1xyzabcxyz[1, 2, 3]
1xyz[1, 2, 3]xyzabcpqr[1, 2, 3]xyz1xyzabcpqr1xyzabcxyz[1, 2, 3]pqr[1, 2]xyz[3, 4]xyz1
(1, 2, 3) [1, 2] [3, 4]pqr(lf) anutosh491@spbhat68:~/lpython/lpython$ python examples/expr2.py 
1xyz(1, 2, 3)xyzabcpqr(1, 2, 3)xyz1xyzabcpqr1xyz[1, 2]xyz(1, 2, 3)pqr1 [1, 2, 3] abc
[1, 2, 3] 1 abc
1 abc [1, 2, 3]
1 [1, 2, 3] abc pqr [1, 2, 3] 1 abc pqr 1 abc [1, 2, 3] pqr 1xyz[1, 2, 3]xyzabc
[1, 2, 3]xyz1xyzabc
1xyzabcxyz[1, 2, 3]
1xyz[1, 2, 3]xyzabcpqr[1, 2, 3]xyz1xyzabcpqr1xyzabcxyz[1, 2, 3]pqr[1, 2]xyz[3, 4]xyz1
(1, 2, 3) [1, 2] [3, 4]pqr(lf) anutosh491@spbhat68:~/lpython/lpython$
```
These are like some random cases I thought and none of these work on master !


anutosh491:
I'll pick some tests from the above shown function and add an integration file soon @Thirumalai-Shaktivel !

src/libasr/pass/print_list.cpp
```diff
@@ -299,26 +299,45 @@ class PrintListVisitor
...
                     tmp_vec.reserve(al, print_tmp.size());
                     for (auto &e: print_tmp) {
                         tmp_vec.push_back(al, e);
                     }
-                    ASR::stmt_t *print_stmt = ASRUtils::STMT(
+                    if(x.m_separator){
+                    print_stmt = ASRUtils::STMT(
+                        ASR::make_Print_t(al, x.base.base.loc, nullptr, tmp_vec.p, tmp_vec.size(),
+                                    x.m_separator, x.m_separator));
+                    } else{
```
Thirumalai-Shaktivel, line 320:
```suggestion
                    if (x.m_separator) {
                        print_stmt = ASRUtils::STMT(ASR::make_Print_t(al,
                            x.base.base.loc, nullptr, tmp_vec.p, tmp_vec.size(),
                            x.m_separator, x.m_separator));
                    } else {
```
Do the same formatting for else as well

src/libasr/pass/print_list.cpp
```diff
@@ -299,26 +299,45 @@ class PrintListVisitor
...
                     pass_result.push_back(al, print_stmt);
-
                 }
-                if (ASR::is_a<ASR::List_t>(*ASRUtils::expr_type(x.m_values[i])))
-                    print_list_helper(x.m_values[i], x.m_separator, nullptr, x.base.base.loc);
-                else
-                    print_tuple_helper(x.m_values[i], x.m_separator, nullptr, x.base.base.loc);
+                if (ASR::is_a<ASR::List_t>(*ASRUtils::expr_type(x.m_values[i]))){
+                    if (x.m_separator){
+                        print_list_helper(x.m_values[i], x.m_separator, i == x.n_values - 1 ? x.m_end : x.m_separator, x.base.base.loc);
```
Thirumalai-Shaktivel, line 330: Here and following can you please use if-else instead of ternary expression

Thirumalai-Shaktivel:
I think we have to add runtime tests.
@certik @czgdp1807 What do you say?

Thirumalai-Shaktivel:
Great job, @anutosh491. Thanks for fixing this.

czgdp1807:
> I think we have to add runtime tests.

Could you please elaborate more? If you are talking about adding a file in `integration_test`, then I guess @anutosh491 has said he will do it soon (https://github.com/lcompilers/lpython/pull/1512#issuecomment-1426619755). Is there anything more to this? If not then I am marking it as draft.

Thirumalai-Shaktivel:
I was talking about saving the runtime output using `tests.toml`

czgdp1807:
I see. That's very much needed. For printing we can't use `assert` statements. However I guess, a better way is that `print` function should call `str` over its inputs and then just send the output of `str` for printing. That way we will be able to do `assert` with `str` calls and no need to store reference files for outputs.

src/libasr/pass/print_arr.cpp
```diff
@@ -122,7 +122,7 @@ class PrintArrVisitor : public PassUtils::PassVisitor<PrintArrVisitor>
             }
             print_stmt = ASRUtils::STMT(ASR::make_Print_t(
                 al, x.base.base.loc, nullptr, body.p, body.size(),
-                nullptr, nullptr));
+                x.m_separator, x.m_end));
```
czgdp1807, line 125: Test with `print(a, b)` where `a` and `b` both are arrays.

czgdp1807 (changes requested):

src/libasr/pass/print_arr.cpp
faze-geek, line 125: Could you add consideration for `sep` and `end `as keyword arguments.
Check that `lpython` does not give an error when order is not followed -
```
def main():
    print(sep = "++","c", 3, "d", 4, [4, 5, 6])

main()

  File "C:\Users\kunni\lpython\try.py", line 2
    print(sep = "++","c", 3, "d", 4, [4, 5, 6])
                                              ^
SyntaxError: positional argument follows keyword argument

c++3++d++4++[4, 5, 6] 
```

faze-geek: Could be corrected in another pull request if this is close to merging. I would always prefer that the issue is fixed completely. Thanks !

Thirumalai-Shaktivel: But this should have been caught in AST. I'm surprised that we missed handling this.
Thanks, @faze-geek for reporting this. Can you please report this as a new Parser issue?

faze-geek: Sure I'll do the needful quick !

anutosh491:
> I think we have to add runtime tests.
> I was talking about saving the runtime output using tests.toml: run
> I see. That's very much needed. For printing we can't use assert statements. However I guess, a better way is that print function should call str over its inputs and then just send the output of str for printing. That way we will be able to do assert with str calls and no need to store reference files for outputs.

I think I am not fully sure how do I address these comments . I get that we could use the `assert` keyword if we make use of `str()` function , but where and what sort of tests am I suppossed to add is not very clear. Any references I could look into ?

anutosh491:
> I think I am not fully sure how do I address these comments . I get that we could use the assert keyword if we make use of str() function , but where and what sort of tests am I suppossed to add is not very clear. Any references I could look into ?

cc @Thirumalai-Shaktivel

Thirumalai-Shaktivel:
ping @czgdp1807

src/libasr/pass/print_arr.cpp
anutosh491: Yeah I can see that this might fail . Though I am not sure how are arrays expected to be printed . For eg 
```
def main0():
    x: i32 = 2
    arr1: i32[x]
    arr1[0] = 5
    arr1[1] = 10
    print(arr1)

main0()

5
10
```
Are the elements expected to be printed out on seperate lines as shown above ?

anutosh491: Currently we have the following 
```
def main0():
    x: i32 = 2
    arr1: i32[x]
    arr2: i32[x]
    arr1[0] = 5
    arr1[1] = 10
    arr2[0] = 20
    arr2[1] = 40
    print(arr1, arr2, sep = "abc", end = "xyz")

main0()
#if __name__ == "__main__":
#    main()
5
10

20
40
```

czgdp1807: Personally I would like to see the output as,

```zsh
5 10
20 40
```

i.e., individual elements of an array should be printed with spaces between them and each array should be printed in a separate line.

czgdp1807:
> I think I am not fully sure how do I address these comments . I get that we could use the `assert` keyword if we make use of `str()` function , but where and what sort of tests am I suppossed to add is not very clear. Any references I could look into ?

Nothing to worry about here. Just add normal `print` statements with constant strings and register only in `tests.toml` with `asr=true`, `ast=true`. That's all. No need to worry about printing of runtime strings.

src/libasr/pass/print_arr.cpp
anutosh491: This the output for something like `print(arr1, arr2)` right ? This tells me the sep and the end keyword by default would be `"\n"`
If we maybe stick to what's default in python , we might expect something like the following . I wouldn't prefer this too !
```
5 1020 40
```

anutosh491: I've addressed printing of arrays in my latest comment ( mostly taken 1d arrays into consideration for now) . Here are few examples of printing arrays together and also a random combination at the end 
```
def main0():
    x: i32 = 2
    y: i32 = 4
    z: i32 = 6
    arr1: i32[x]
    arr2: i32[x]
    arr1[0] = 5
    arr1[1] = 10
    arr2[0] = 20
    arr2[1] = 40
    print(arr1, arr2)
    print(arr1, arr2, sep = "abc")
    print(arr1, arr2, end = "xyz\n")
    print(arr1, arr2, sep = "abc", end = "xyz\n")
    print(x, "xyz", arr1, y, z, arr2, "xyz", z, sep = "abc", end = "pqr\n")

main0()

5 10 
20 40 
5 10abc20 40 
5 10 
20 40xyz
5 10abc20 40xyz
2abcxyzabc5 10abc4abc6abc20 40abcxyzabc6pqr
```

anutosh491: @czgdp1807 Could you approve of this output , once done I'll add these print statements in the integration test file I've pushed !

Thirumalai-Shaktivel:
LGTM! But, Let @czgdp1807 also review and approve this.
Also, It seems you made changes to `print_arr.cpp`, do you wanna add tests (in test_sep_end_kws.py) for it?

anutosh491:
> Also, It seems you made changes to print_arr.cpp

Yess I'll be updating my integration test file with tests for printing arrays . Just want to get this output confirmed (https://github.com/lcompilers/lpython/pull/1512#discussion_r1111518263) before I go on with that ! The changes makes printing arrays compatible with other data types/containers including arrays along with the end and sep keywords !

src/libasr/pass/print_arr.cpp
Thirumalai-Shaktivel:
```
5 10 
```
extra space at the end of the line? Was there a problem fixing it?

src/libasr/pass/print_arr.cpp
```diff
@@ -100,15 +111,38 @@ class PrintArrVisitor : public PassUtils::PassVisitor<PrintArrVisitor>
...
+                            al, x.base.base.loc, nullptr, body.p, body.size(),
+                            nullptr, space));
+                    }
                     pass_result.push_back(al, print_stmt);
-                    pass_result.push_back(al, empty_print_endl);
                     print_body.clear();
                 }
                 print_stmt = print_array_using_doloop(x.m_values[i], x.base.base.loc);
                 pass_result.push_back(al, print_stmt);
+                pass_result.push_back(al, back);
```
Thirumalai-Shaktivel, line 128: Also, I was a little worried about this backspace!

anutosh491: yeah . I'll give a brief of why I had to use a `backsapce`. So on main , if we print an array we get this 
```
def main0():
    x: i32 = 2
    arr1: i32[x]
    arr1[0] = 5
    arr1[1] = 10
    print(arr1)

main0()
5
10

```
So each element is being printed on a newline and then there's an empty line .
Now first I've fixed this through my changes in the `print_array_using_doloop` function. I've added the following 
```
+        ASR::ttype_t *str_type_len_1 = ASRUtils::TYPE(ASR::make_Character_t(
+       al, loc, 1, 1, nullptr, nullptr, 0));
+        ASR::expr_t *space = ASRUtils::EXPR(ASR::make_StringConstant_t(
+       al, loc, s2c(al, " "), str_type_len_1));
....
                ASR::stmt_t* print_stmt = ASRUtils::STMT(ASR::make_Print_t(al, loc, nullptr,
-                                                                print_args.p, print_args.size(), nullptr, nullptr));
-                                                                print_args.p, print_args.size(), nullptr, space));
```
This function basically supports printing of the elements of the array . And the following change will change the structure from something like 
```
5
10

```
to 
```
5 10 # where 10 is followed by a space 
```
Hence I introduced a `backspace` character there to get rid of the space after 10

src/libasr/pass/print_arr.cpp
czgdp1807: What's the CPython output for the same code?

anutosh491: Not sure if CPython supports arrays as lpython. 
```
def main0():
    x: i32 = 2
    arr1: i32[x]
    arr2: i32[x]
    arr1[0] = 5
    arr1[1] = 10
    arr2[0] = 20
    arr2[1] = 40
    print(arr1, arr2)

main0()

5 10 
20 40 

  File "/home/anutosh491/lpython/lpython/examples/expr2.py", line 11, in <module>
    main0()
  File "/home/anutosh491/lpython/lpython/examples/expr2.py", line 5, in main0
    arr1[0] = 5
    ^^^^
UnboundLocalError: cannot access local variable 'arr1' where it is not associated with a value
```

Thirumalai-Shaktivel: try it using NumPy arrays

czgdp1807: See the pattern of how we enable `cpython` with array tests. You should be able to figure out.

anutosh491: > try it using NumPy arrays

For something like the following
```
import numpy as np

def main0():
    x: np.int32 = 2
    arr1: np.ndarray[np.int32_t, np.ndim[x]] = np.zeros(x, dtype=np.int32)
    arr2: np.ndarray[np.int32_t, np.ndim[x]] = np.zeros(x, dtype=np.int32)
    arr1[0] = 100
    arr1[1] = 200
    arr2[0] = 300
    arr2[1] = 400
    print(arr1, arr2)
    print(arr1, arr2, sep="abc")
    print(arr1, arr2, end="pqr\n")
    print(arr1, arr2, sep="abc", end="pqr\n")

main0()
```
We get the following result 
```
[100 200] [300 400]
[100 200]abc[300 400]
[100 200] [300 400]pqr
[100 200]abc[300 400]pqr
```
This tells me 
1) We might have to use square brackets to enclose the elements .
2) Also default seperator would be a space rather than the end line which I have used here .

Should I go forward with this @Thirumalai-Shaktivel @czgdp1807 ? (Also would it make sense to dedicate a completely new Pr for arrays as I would like to incorporate these changes .... but at the same time , also check for cases other than 1d arrays which currently my changes might not satisfy . The Pr currently would support sep and end keyword with all basic data types and other containers like list/tuple . If we could keep it till this ... i'll make a subsequent pr for arrays !)

Thirumalai-Shaktivel: this works in both CPython and LPython:
```py
from ltypes import i32
from numpy import empty, int32

def main0():
    x: i32 = 2
    arr1: i32[x] = empty(x, dtype=int32)
    arr2: i32[x] = empty(x, dtype=int32)
    arr1[0] = 100
    arr1[1] = 200
    arr2[0] = 300
    arr2[1] = 400
    print(arr1, arr2)
    print(arr1, arr2, sep="abc")
    print(arr1, arr2, end="pqr\n")
    print(arr1, arr2, sep="abc", end="pqr\n")

main0()
```

Thirumalai-Shaktivel: I see. @certik @czgdp1807, what do you think about this?
Should we enclose the array values within SquareBrackets while printing?
CPython output:
```
[100 200] [300 400]
[100 200]abc[300 400]
[100 200] [300 400]pqr
[100 200]abc[300 400]pqr
```

anutosh491: I would say enclosing arrays with SquareBrackets isn't a bad addition . If all of us are in .... I'll address it !

anutosh491: My lpython output for the following shared program above is 
```
100 200 
300 400 
100 200abc300 400 
100 200 
300 400pqr
100 200abc300 400pqr
```
Basically my default seperator is a newline rather than a space and obviously I haven't includes braces.

czgdp1807 (approved):

tests/tests.toml
```diff
@@ -279,6 +279,11 @@ filename = "c_interop1.py"
 asr = true
 c = true
 
+[[test]]
+filename = "test_end_sep_keywords.py"
+asr = true
+ast = true
```
certik, line 285: Unless this is testing some bug in the parser, we don't need to test AST:
```suggestion
```

certik:
Does CPython and LPython print exactly the same thing, or does it differ?

anutosh491:
1) For containers like lists/tuples and other fundamental data types, both outputs are the same while using end and sep keyword . For example https://github.com/lcompilers/lpython/pull/1512#issue-1580661104 as shown in PR description

2) For arrays though , they do differ

For something like 
```
from ltypes import i32
from numpy import empty, int32

def main0():
    x: i32 = 2
    arr1: i32[x] = empty(x, dtype=int32)
    arr2: i32[x] = empty(x, dtype=int32)
    arr1[0] = 100
    arr1[1] = 200
    arr2[0] = 300
    arr2[1] = 400
    print(arr1, arr2)
    print(arr1, arr2, sep="abc")
    print(arr1, arr2, end="pqr\n")
    print(arr1, arr2, sep="abc", end="pqr\n")

main0()
```
The CPython output is 
```
[100 200] [300 400]
[100 200]abc[300 400]
[100 200] [300 400]pqr
[100 200]abc[300 400]pqr
```
But the lpython output is
```
100 200 
300 400 
100 200abc300 400 
100 200 
300 400pqr
100 200abc300 400pqr
```
Differences are 
1) We haven't incorporated square brackets 
2) Our default separator between 2 arrays is a newline rather than enclosing the array elements with brackets and using space in between !

You coudl go through the thread (https://github.com/lcompilers/lpython/pull/1512#discussion_r1103529589) where we have been discussing this . If you feel something more needs to be done for arrays , I can maybe open up a subsequent pr for it because I am quite confident about the initial changes which were mentioned in the issue description (https://github.com/lcompilers/lpython/pull/1512#issue-1580661104)

anutosh491:
cc @certik

tests/reference/ast-test_end_sep_keywords-87ad1b5.json
```diff
@@ -0,0 +1,13 @@
...
+    "infile": "tests/test_end_sep_keywords.py",
+    "infile_hash": "5ea30711228d4ebb64266988c1a706a3d64f196457b939ed3bf15ecf",
+    "outfile": null,
+    "outfile_hash": null,
+    "stdout": "ast-test_end_sep_keywords-87ad1b5.stdout",
+    "stdout_hash": "de3720f0800ee1e94c99f22c5e13d9d4ef224aada410a8028ff5e480",
+    "stderr": null,
+    "stderr_hash": null,
+    "returncode": 0
+}
```
certik, line 13: these AST reference tests are not used anymore, so they can be removed.

tests/reference/ast-test_end_sep_keywords-87ad1b5.stdout
```diff
@@ -0,0 +1,3 @@
+(Module [(FunctionDef testEndSepKeywords ([] [] [] [] [] [] []) [(Expr (Call (Name print Load) [(ConstantStr "abc" ()) (ConstantStr "lmn" ()) (ConstantStr "pqr" ())] [])) (Expr (Call (Name print Load) [(ConstantStr "abc" ()) (ConstantStr "lmn" ()) (ConstantStr "pqr" ())] [(sep (ConstantStr "+" ()))])) (Expr (Call (Name print Load) [(ConstantStr "abc" ()) (ConstantStr "lmn" ()) (ConstantStr "pqr" ())] [(end (ConstantStr "xyz
+" ()))])) (Expr (Call (Name print Load) [(ConstantStr "abc" ()) (ConstantStr "lmn" ()) (ConstantStr "pqr" ())] [(sep (ConstantStr "+" ())) (end (ConstantStr "xyz
+" ()))]))] [] () ()) (Expr (Call (Name testEndSepKeywords Load) [] []))] [])
```
certik, line 3: This can be removed.

src/libasr/pass/print_arr.cpp
certik: Yes, we should use square brackets for LPython to be compatible with NumPy, and for LFortran we should not be compatible with other compilers. This can be done in subsequent PRs. Also for both LFortran and LPython having a nice printing of arrays, especially matrices, using unicode, etc. would be awesome, like Julia does it.

certik:
I left some more comments. Otherwise I think this can be merged, and subsequent PRs can improve the rest.

src/libasr/pass/print_arr.cpp
anutosh491: Thanks for the feedback , I try to cover most of what you've said in a following pr !

anutosh491:
I've removed the ast based reference files which wouldn't be required anymore .

certik (approved):
I think that this is fine for now. Squashing and merging.

anutosh491:
Thanks @Thirumalai-Shaktivel @czgdp1807 and @certik for the great reviews on this one !

PR #1516, opened 2023-02-13
C: Initialize lists

Smit-create:


integration_tests/test_list_02.py
```diff
@@ -55,8 +55,18 @@ def test_list_02():
         assert x[i] == y[i]
 
 
+def test_list_03():
+    x: list[i32]
+    x.append(2)
+    assert len(x) == 1
+    x.append(10)
+    assert len(x) == 2
```
Smit-create, line 63: Currently, we need to initialize the list after the declaration in `cpython` as well as in the `C` backend. I'm trying to fix that in the `C` backend where we will automatically initialize the list.

This fails in CPython as :
```py
>>> x: list[int]
>>> type(x)
NameError: name 'x' is not defined
```
I'm not sure if we want to support this or how should we make it compatible with the current diff.

cc @certik @czgdp1807

czgdp1807: Well the new example will not work with `cpython`. So supporting it shouldn't be a priority.

PR #1517, opened 2023-02-14
Completed test suite for logical binary operators

faze-geek:
Related to recently closed #1506 
Earlier the tests were specific to the issue. I have added all fundamental and derived logical binary operators in python.
@Smit-create Could to have a look at it.


Smit-create:
Thanks! It works so I'm fine.

Smit-create (approved):

PR #1522, opened 2023-02-15
Fix CPtr to array casting in ltypes.py

### LINKED ISSUE #1521: pure Python assigning to pointer fails
This works with LPython, but fails with CPython:
```python
def initialize_A(n: i32, a: CPtr) -> None:         
    A: Pointer[i16[n]] = c_p_pointer(a, i16[n]) 
    i: i32
    for i in range(n):
        A[i] = i16(i);                   
```
where I get:
```console
  File "a.py", line 5, in initialize_A
    A[i] = i16(i);
TypeError: expected LP_c_short instance, got int
```

This operation is tested in `bindc_07.py`, but I can see that it is not tested by CPython.

certik:
Fixes #1521.


certik:
+1 to merge.

Issue #1519, opened 2023-02-14
WASM: complex numbers


certik:
Add complex numbers and operations on them.

certik:
There are several ways to implement it. 

### In the backend

Just create a WASM function complex_add(), which accepts the two operands a,b on stack and returns the result on stack (this function is implemented in `asr_to_wasm` directly using wasm instructions). Each complex number is represented by two floating point numbers on stack.

### Using ASR->ASR pass

This pass would work similarly to `pass_compare.cpp` which inserts new ASR functions (to global scope) to implement a given operation (and only inserts it once and reuses it everywhere).

PR #1527, opened 2023-02-15
WASM: Initial support for complex numbers

Shaikh-Ubaid:
This PR adds support for complex numbers and operations on them in the `wasm` backend.

Fixes https://github.com/lcompilers/lpython/issues/1519.


Shaikh-Ubaid:
We were hoping to use Array approach to store a complex number. In the Array approach, we had to reset the memory at the end of the function. To maintain where a function started (from which memory location), we are needing to store the memory offset in a (new/extra) local variable. Also, we currently have support only for memory allocation at compile-time. With these constraints, supporting the array approach seems tricky.

For the moment, to have something which just works I currently used two local variables to store a `complex` number.

Shaikh-Ubaid:
Some concerns:
- the code for `emit_complex_add_32()`, `emit_complex_add_64()`, `emit_complex_sub_32()`, `emit_complex_sub_64()`, `emit_complex_mul_32()` and `emit_complex_mul_64()` altogether seems lengthy. Is it maintainable enough?
- the function indices have been **hard-coded** when calling `add_c32()`, `add_c64()`, `sub_32()`, `sub_64()`, `mul_32()`, `mul_64()`.

src/libasr/codegen/asr_to_wasm.cpp
```diff
@@ -2376,6 +2808,22 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
...
+                }
+                wasm::emit_call(m_code_section, m_al, 3  /* print_f64 */);
+                emit_call_fd_write(1, ",", 1, 0);
+                this->visit_expr(*x.m_values[i]);
+                if (a_kind == 4) {
+                    wasm::emit_f64_promote_f32(m_code_section, m_al);
+                }
+                wasm::emit_call(m_code_section, m_al, 3  /* print_f64 */);
+                emit_call_fd_write(1, ")", 1, 0);
+                wasm::emit_drop(m_code_section, m_al); // drop real part
```
Shaikh-Ubaid, line 2826: **Concern:**
**Line 2814 & Line 2826:** When printing a complex number we print the real part and then the imaginary part. It currently happens that, when we fetch a complex number on stack top, the imaginary part of the complex is at the stack top followed by the real part. Thus, we are needing to drop the top stack so that we can print the appropriate real/imag part of the complex number.

Shaikh-Ubaid: This concern is now solved/fixed in [WASM: handle_print(): Use globals and fix revisiting expr](https://github.com/lcompilers/lpython/pull/1527/commits/f3d56291abd67b6c207e77fbeaed13dd657077f7).

Shaikh-Ubaid:
From https://github.com/lcompilers/lpython/pull/1527#issuecomment-1432117376,
>Concern: the function indices have been hard-coded when calling add_c32(), add_c64(), sub_32(), sub_64(), mul_32(), mul_64().

This is now solved/fixed in [WASM: Refactor: Define and use m_self_func_name_idx_map](https://github.com/lcompilers/lpython/pull/1527/commits/8f96beccd2dbc3d5e1f4c955c18f2d031dcc0166).

src/libasr/codegen/asr_to_wasm.cpp
```diff
@@ -579,18 +717,31 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
...
         emit_print_int();
         emit_print_float();
+        emit_complex_add_32();
+        emit_complex_add_64();
+        emit_complex_sub_32();
+        emit_complex_sub_64();
+        emit_complex_mul_32();
+        emit_complex_mul_64();
+        emit_complex_abs_32();
+        emit_complex_abs_64();
```
Shaikh-Ubaid, line 744: Ideally, these should be emitted into `wasm` code only when the particular operations are being used in the source code. 

**Concern/Challenge:** If we do not define theses functions initally and add them at the very end (that is after defining all the other functions), then we do not have any function index to refer to these functions when they are called during a call (while defining the other functions).

certik: I think I would keep a running index of all functions and just appending these at the end as needed; then once we are done, we convert the index to the function index at the start of the WASM binary file.

Shaikh-Ubaid: I attempted to support it in [WASM: Support emitting functions based on requirement](https://github.com/lcompilers/lpython/pull/1527/commits/24b943aa2cd9cf3ef90e55af98765259b3dcfb51), and I'd love to get your feedback on it. Specifically, I'm wondering if this is what you had in mind, and if there are any changes you would like me to make. I'm open to any suggestions you have and would love to discuss this further with you.

Shaikh-Ubaid:

czgdp1807:
> * the code for `emit_complex_add_32()`, `emit_complex_add_64()`, `emit_complex_sub_32()`, `emit_complex_sub_64()`, `emit_complex_mul_32()` and `emit_complex_mul_64()` altogether seems lengthy. Is it maintainable enough?

Can we somehow convert,

```cpp
void emit_complex_add_32(int fn_idx = -1) {
        using namespace wasm;
        define_emit_func({f32, f32, f32, f32}, {f32, f32}, {}, "add_c32", [&](){
            wasm::emit_get_local(m_code_section, m_al, 0);
            wasm::emit_get_local(m_code_section, m_al, 2);
            wasm::emit_f32_add(m_code_section, m_al);

            wasm::emit_get_local(m_code_section, m_al, 1);
            wasm::emit_get_local(m_code_section, m_al, 3);
            wasm::emit_f32_add(m_code_section, m_al);
        }, fn_idx);
    }
```

to 

(assuming type of `f32`, `f64` is `wasm::type` and `callback_type` is say the type of `wasm::emit_f32_add`, `wasm::emit_64_add` and so on)

```cpp
void emit_complex_op_util(wasm::type& ftype, std::string op_name, callback_type callback, int fn_idx = -1) {
        using namespace wasm;
        define_emit_func({ftype, ftype, ftype, ftype}, {ftype, ftype}, {}, name, [&](){
            wasm::emit_get_local(m_code_section, m_al, 0);
            wasm::emit_get_local(m_code_section, m_al, 2);
            callback(m_code_section, m_al);

            wasm::emit_get_local(m_code_section, m_al, 1);
            wasm::emit_get_local(m_code_section, m_al, 3);
            callback(m_code_section, m_al);
        }, fn_idx);
    }
```

and call it whereever needed.

Shaikh-Ubaid:
The `callback` seems to work for `emit_complex_add_32` and `emit_complex_add_64`.

@czgdp1807  how do we use the similar `callback` approach in `emit_complex_mul_32`?

```cpp
void emit_complex_mul_32(int fn_idx = -1) {
        using namespace wasm;
        define_emit_func({f32, f32, f32, f32}, {f32, f32}, {}, "mul_c32", [&](){
            wasm::emit_get_local(m_code_section, m_al, 0);
            wasm::emit_get_local(m_code_section, m_al, 2);
            wasm::emit_f32_mul(m_code_section, m_al);

            wasm::emit_get_local(m_code_section, m_al, 1);
            wasm::emit_get_local(m_code_section, m_al, 3);
            wasm::emit_f32_mul(m_code_section, m_al);

            wasm::emit_f32_sub(m_code_section, m_al);

            wasm::emit_get_local(m_code_section, m_al, 0);
            wasm::emit_get_local(m_code_section, m_al, 3);
            wasm::emit_f32_mul(m_code_section, m_al);

            wasm::emit_get_local(m_code_section, m_al, 1);
            wasm::emit_get_local(m_code_section, m_al, 2);
            wasm::emit_f32_mul(m_code_section, m_al);

            wasm::emit_f32_add(m_code_section, m_al);
        }, fn_idx);
    }
```

czgdp1807:
> how do we use the similar `callback` approach in `emit_complex_mul_32/64` or `emit_complex_abs_32/64`?

I guess we also need to support division of complex numbers. So some similar refactoring can be done for `emit_complex_mul_32/64`, `emit_complex_div_32/64`. `emi_complex_abs_32/64` doesn't seem to have any common logic with `add`, `mul`, `div`, `sub`, so I guess it can be written independently without calling any utility function accepting callbacks.

Shaikh-Ubaid:
@czgdp1807 Refactoring `emit_complex_mul_32` and `emit_complex_mul_64` and combining them into one common function is not very clear to me. Please, could you share an example implementation where we refactor these two functions and combine them into one, similar to the transformation of `emit_complex_add_32` as you shared in https://github.com/lcompilers/lpython/pull/1527#issuecomment-1435485516?

As a later change, I wonder if we can/should keep the functions `emit_complex_mul_32/64`, `emit_complex_div_32/64`. `emi_complex_abs_32/64` in a separate file (maybe with some name like `wasm_rt_lib.cpp`). For the moment, shall we keep these as per the current implementation?

czgdp1807:
If you need it urgently then no need to refactor anything because that can be done later. If you have time to merge this, then may be figure out the patterns and create utility functions. The boilerplate code is high but that's okay. Just add TODOs before merging. I am approving it in its current state.

czgdp1807 (approved):
Resolve conflicts then it should be good to go. Add TODOs for refactoring whevever you think refactoring can be done to reduce the boilerplate code. Thanks. cc: @certik

Shaikh-Ubaid:
>If you have time to merge this, then may be figure out the patterns and create utility functions. The boilerplate code is high but that's okay.

Yes, these functions would later be implemented as `ASR->ASR` pass. Thus, I am unsure if we should refactor them now.

src/libasr/codegen/asr_to_wasm.cpp
```diff
@@ -579,18 +724,32 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
...
+        m_self_funcs_map["print_i64"] = &ASRToWASMVisitor::emit_print_int;
+        m_self_funcs_map["print_f64"] = &ASRToWASMVisitor::emit_print_float;
+        m_self_funcs_map["add_c32"] = &ASRToWASMVisitor::emit_complex_add_32;
+        m_self_funcs_map["add_c64"] = &ASRToWASMVisitor::emit_complex_add_64;
+        m_self_funcs_map["sub_c32"] = &ASRToWASMVisitor::emit_complex_sub_32;
+        m_self_funcs_map["sub_c64"] = &ASRToWASMVisitor::emit_complex_sub_64;
+        m_self_funcs_map["mul_c32"] = &ASRToWASMVisitor::emit_complex_mul_32;
+        m_self_funcs_map["mul_c64"] = &ASRToWASMVisitor::emit_complex_mul_64;
+        m_self_funcs_map["abs_c32"] = &ASRToWASMVisitor::emit_complex_abs_32;
+        m_self_funcs_map["abs_c64"] = &ASRToWASMVisitor::emit_complex_abs_64;
```
certik, line 752: Let's use enums here for performance.

certik: You can also then just use std::vector.

src/libasr/codegen/asr_to_wasm.cpp
```diff
@@ -1437,6 +1657,81 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
...
+        this->visit_expr(*x.m_left);
+        this->visit_expr(*x.m_right);
+        LCOMPILERS_ASSERT(ASRUtils::is_complex(*x.m_type));
+        int a_kind = ASR::down_cast<ASR::Complex_t>(ASRUtils::type_get_past_pointer(x.m_type))->m_kind;
+        switch (x.m_op) {
+            case ASR::binopType::Add: {
+                if (a_kind == 4) {
+                    if (m_self_func_name_idx_map.find("add_c32") == m_self_func_name_idx_map.end()) {
+                        m_self_func_name_idx_map["add_c32"] = no_of_types++;
+                    }
```
certik, line 1674: This just becomes a quick check, like `if (m_self_complex_fn_present[wasm_add_c32])`.

src/libasr/codegen/asr_to_wasm.cpp
certik: This can be done in a subsequent PR.

certik (approved):
I think this is good to merge as is, after resolving conflicts / polishing history as needed.

In subsequent PRs we can refine.

PR #1529, opened 2023-02-16
Extended escape sequence (tab and backspace) support

anutosh491:
This PR extends escape sequence support for tabs (through \t) and backspaces (through \b) .

On main
```
def main0():
    s1: str
    s2: str
    s1 = 'Hello\tWorld'
    s2 = 'Hello    \bWorld'
    print(s1)
    print(s2)

main0()

Hello   World
Hello   World
Hello\tWorld
Hello    \bWorld
```
On branch
```
Hello   World
Hello   World
Hello   World
Hello   World
```


src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -849,6 +849,10 @@ R"(#include <stdio.h>
                 src += "\\";
             } else if (s[idx] == '\"') {
                 src += "\"";
+            } else if (s[idx] == '\t') {
```
faze-geek, line 852: Along with this, you can add `\v` too, it is the `vertical tab` escape sequence in python. Should be an easy addition.
https://stackoverflow.com/questions/3380538/what-is-a-vertical-tab

anutosh491:
cc @Thirumalai-Shaktivel

Thirumalai-Shaktivel:
LGTM!

czgdp1807 (changes requested):
Add a test, register it in `tests.toml` file with `ast = true` and `asr = true`. We need to make sure the generated ASR is as expected always.

anutosh491:
I've addressed the requested changes here !

tests/reference/ast-escape_character1-eb0f575.stdout
```diff
@@ -0,0 +1 @@
+(Module [(FunctionDef testEscapeCharacters ([] [] [] [] [] [] []) [(AnnAssign (Name s1 Store) (Name str Load) (ConstantStr "Hello	World" ()) 1) (AnnAssign (Name s2 Store) (Name str Load) (ConstantStr "Hello    World" ()) 1) (AnnAssign (Name s3 Store) (Name str Load) (ConstantStr "HelloWorld" ()) 1) (Expr (Call (Name print Load) [(Name s1 Load)] [])) (Expr (Call (Name print Load) [(Name s2 Load)] [])) (Expr (Call (Name print Load) [(Name s3 Load)] []))] [] () ()) (FunctionDef main0 ([] [] [] [] [] [] []) [(Expr (Call (Name testEscapeCharacters Load) [] []))] [] () ()) (Expr (Call (Name main0 Load) [] []))] [])
```
czgdp1807, line 1: @certik `"Hello World"` and `"HelloWorld"` make sense?

anutosh491: When I print the AST locally , I get the following (which seems like something we would expect) ... not sure what the `""` character stands for 
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ ./src/bin/lpython --show-ast tests/escape_character1.py --indent
(Module
    [(FunctionDef
        testEscapeCharacters
        ([]
        []
        []
        []
        []
        []
        [])
        [(AnnAssign
            (Name
                s1
                Store
            )
            (Name
                str
                Load
            )
            (ConstantStr
                "Hello  World"
                ()
            )
            1
        )
        (AnnAssign
            (Name
                s2
                Store
            )
            (Name
                str
                Load
            )
            (ConstantStr
                "Hello   World"
                ()
            )
            1
        )
        (AnnAssign
            (Name
                s3
                Store
            )
            (Name
                str
                Load
            )
            (ConstantStr
                "Hello
                      World"
                ()
            )
            1
        )
        (Expr
            (Call
                (Name
                    print
                    Load
                )
                [(Name
                    s1
                    Load
                )]
                []
            )
        )
        (Expr
            (Call
                (Name
                    print
                    Load
                )
                [(Name
                    s2
                    Load
                )]
                []
            )
        )
        (Expr
            (Call
                (Name
                    print
                    Load
                )
                [(Name
                    s3
                    Load
                )]
                []
            )
        )]
        []
        ()
        ()
    )
    (FunctionDef
        main0
        ([]
        []
        []
        []
        []
        []
        [])
        [(Expr
            (Call
                (Name
                    testEscapeCharacters
                    Load
                )
                []
                []
            )
        )]
        []
        ()
        ()
    )
    (Expr
        (Call
            (Name
                main0
                Load
            )
            []
            []
        )
    )]
    []
)
```

Thirumalai-Shaktivel: I think the AST and ASR should store the escape sequences info in the form of `\t`, `\v`, ... and then handle them later in the backends. CPython does the same thing:
```clojure
python -m ast examples/expr2.py
Module(
   body=[
      Expr(
         value=Constant(value='Hello\tWorld')),
      Expr(
         value=Constant(value='Hello  \x08World')),
      Expr(
         value=Constant(value='Hello\x0bWorld'))],
   type_ignores=[])
```

anutosh491: Note sure , how I could address the above comment @Thirumalai-Shaktivel . Locally my thing doesn't store these escape sequences !

anutosh491: @Thirumalai-Shaktivel As we have tested out and merged #1502 , I can see that the AST now returns something which is similar to what CPython returns . I'll resolve conflicts here and modify the tests/references accordingly for the same . Then this could go in !

anutosh491: I've updated the referneces accordingly . Should be good to go once the tests pass !

tests/tests.toml
```diff
@@ -1045,4 +1050,4 @@ run_with_dbg = true
 
 [[test]]
 filename = "runtime_errors/test_raise_01.py"
-run_with_dbg = true
+run_with_dbg = true
```
czgdp1807, line 1053:
```suggestion
run_with_dbg = true

```

czgdp1807 (approved):
Other than the suggestion LGTM.

anutosh491:
> Other than the suggestion LGTM.

Thank you . I've made the requested change !

certik:
I think the string should be de-escaped in the parser, and stored in AST and ASR directly, so `\t` (tab) would be represented using the corresponding byte for tab, not as a slash and "t" (two bytes). Then when printing, we can convert these characters such as a tab to `\t`.

certik:
It might already be fixed. If not, it needs to be rebased. I would not test AST, only ASR.

anutosh491:
> It might already be fixed. If not, it needs to be rebased. I would not test AST, only ASR.

I'll test it out of this has been fixed already !

anutosh491:
The snippets being pointed out in this PR now work perfectly on master and hence the PR can be closed.

PR #1543, opened 2023-02-22
Fixing local increments for the target variable in for loops

anutosh491:
This pr aims to fix the following issue 
```
def main():
    i: i32
    for i in range(5):
        print(i)
        i=i+1

main()

(lp) C:\Users\kunni\lpython>python try.py     # expected
0
1
2
3
4

(lp) C:\Users\kunni\lpython>src\bin\lpython try.py         # given output 
0
2
4
```

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -4327,6 +4327,8 @@ public:
         std::string loop_src_var_name = "";
         ASR::expr_t *loop_end = nullptr, *loop_start = nullptr, *inc = nullptr;
         ASR::expr_t *for_iter_type = nullptr;
+        ASR::stmt_t *assign_to_j = nullptr;
+        ASR::stmt_t *i_update = nullptr;
         if (AST::is_a<AST::Call_t>(*x.m_iter)) {
             AST::Call_t *c = AST::down_cast<AST::Call_t>(x.m_iter);
             std::string call_name;
@@ -4442,13 +4444,27 @@ public:
 
             head.m_v = ASRUtils::EXPR(explicit_iter_var);
         } else {
-            body.reserve(al, x.n_body);
+            body.reserve(al, x.n_body + 2);
             head.m_v = target;
         }
 
         SymbolTable *parent_scope = current_scope;
         current_scope = al.make_new<SymbolTable>(parent_scope);
+        std::string name = "j";
+        ASR::expr_t *j = ASR::down_cast<ASR::expr_t>(ASR::make_Var_t(
+            al, x.base.base.loc, ASR::down_cast<ASR::symbol_t>(ASR::make_Variable_t(
+                al, x.base.base.loc, current_scope, s2c(al, name), nullptr, 0, ASR::intentType::Local,
+                nullptr, nullptr, ASR::storage_typeType::Default,
+                ASRUtils::TYPE(ASR::make_Integer_t(al, x.base.base.loc, 4, nullptr, 0)),
+                ASR::abiType::Source, ASR::accessType::Public,
+                ASR::presenceType::Required, false))));
+        assign_to_j = ASRUtils::STMT(ASR::make_Assignment_t(al, x.base.base.loc, j, target, nullptr));
+        i_update = ASRUtils::STMT(ASR::make_Assignment_t(al, x.base.base.loc, target,
+                    ASRUtils::EXPR(ASR::make_IntegerBinOp_t(al, x.base.base.loc, j,
+                    ASR::binopType::Add, constant_one, a_type, nullptr)), nullptr));
+        body.push_back(al, assign_to_j);
         transform_stmts(body, x.n_body, x.m_body);
+        body.push_back(al, i_update);
         int32_t total_syms = current_scope->get_scope().size();
         if( total_syms > 0 ) {
             std::string name = parent_scope->get_unique_name("block");


anutosh491:
I guess step 1 here should be creating an temporary variable and registering it in the current scope . Maybe we should address this in the block which catches if the iterable is of type `range` ( this if check in particular `if (AST::is_a<AST::Call_t>(*x.m_iter)) {`) 
```
index 305642f87..0d924610f 100644
@@ -4363,6 +4363,24 @@ public:
                     x.base.base.loc);
             }
 
+            auto loop_src_var_symbol = current_scope->resolve_symbol(call_name);
+            LCOMPILERS_ASSERT(loop_src_var_symbol!=nullptr);
+            ASR::ttype_t* int_type = ASRUtils::TYPE(ASR::make_Integer_t(al, loc, 4, nullptr, 0));
+            std::string j = current_scope->get_unique_name("j");
+            Vec<char*> variable_dependencies_vec;
+            variable_dependencies_vec.reserve(al, 1);
+            ASRUtils::collect_variable_dependencies(al, variable_dependencies_vec, int_type);
+            auto j_variable = ASR::make_Variable_t(al, loc,
+                current_scope, s2c(al, j),
+                variable_dependencies_vec.p, variable_dependencies_vec.size(),
+                ASR::intentType::Local, nullptr, nullptr, storage_type,
+                int_type, ASR::abiType::Source, ASR::accessType::Public,
+                ASR::presenceType::Required, false
+            );
+            current_scope->add_symbol(j,
+                            ASR::down_cast<ASR::symbol_t>(j_variable));
+            auto loop_src_var = ASR::make_Var_t(al, loc, loop_src_var_symbol);
+
         } else if (AST::is_a<AST::Name_t>(*x.m_iter)) {
             loop_src_var_name = AST::down_cast<AST::Name_t>(x.m_iter)->m_id;
             loop_end = for_iterable_helper(loop_src_var_name, x.base.base.loc);
```
Is the following diff sufficient to address step 1 @Thirumalai-Shaktivel

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4442,13 +4444,27 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
         SymbolTable *parent_scope = current_scope;
         current_scope = al.make_new<SymbolTable>(parent_scope);
+        std::string name = "j";
+        ASR::expr_t *j = ASR::down_cast<ASR::expr_t>(ASR::make_Var_t(
+            al, x.base.base.loc, ASR::down_cast<ASR::symbol_t>(ASR::make_Variable_t(
+                al, x.base.base.loc, current_scope, s2c(al, name), nullptr, 0, ASR::intentType::Local,
+                nullptr, nullptr, ASR::storage_typeType::Default,
+                ASRUtils::TYPE(ASR::make_Integer_t(al, x.base.base.loc, 4, nullptr, 0)),
+                ASR::abiType::Source, ASR::accessType::Public,
+                ASR::presenceType::Required, false))));
```
Thirumalai-Shaktivel, line 4460: Here you have created the Variable, but it is not registered in the current scope.
See: https://github.com/lcompilers/lpython/blob/main/src/libasr/pass/pass_utils.cpp#L503

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4442,13 +4444,27 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+        assign_to_j = ASRUtils::STMT(ASR::make_Assignment_t(al, x.base.base.loc, j, target, nullptr));
+        i_update = ASRUtils::STMT(ASR::make_Assignment_t(al, x.base.base.loc, target,
+                    ASRUtils::EXPR(ASR::make_IntegerBinOp_t(al, x.base.base.loc, j,
+                    ASR::binopType::Add, constant_one, a_type, nullptr)), nullptr));
```
Thirumalai-Shaktivel, line 4464: Can you check where `i` value is incremented, and replace this with that or vice-versa?
 I think this will make `i` to be incremented twice.

anutosh491: Hmm correct , maybe we could just assign `j` back to `i` after the for loop body ?

Thirumalai-Shaktivel: I think so

src/lpython/semantics/python_ast_to_asr.cpp
anutosh491: Thanks for letting me know about this function , I've updated my code accordingly . Resolving !

src/lpython/semantics/python_ast_to_asr.cpp
anutosh491: Thanks for the prompt , here my latest commit fixes all sorts of local increments . Few cases are shown below
```
def main():
    i: i32
    for i in range(5):
        print(i)
        i=i+1

main()


0
1
2
3
4
```
```
def main():
    i: i32
    for i in range(5):
        i = i*10
        print(i)

main()

0
10
20
30
40
```
```
def main():
    i: i32
    lst: list[i32] = [1, 2, 3]
    for i in lst:
        print(i)
        i = i*10

main()

1
2
3
```

anutosh491: But as we can see tests fail , I guess this is because of reference files being updated . Once you go through the code and give me a green flag , I'll update the references too . cc @Thirumalai-Shaktivel  (Is this all ? maybe this would need an integration test or something added in tests.toml , you could let me know ...)

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4442,13 +4445,21 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
             head.m_v = ASRUtils::EXPR(explicit_iter_var);
         } else {
-            body.reserve(al, x.n_body);
+            body.reserve(al, x.n_body + 2);
             head.m_v = target;
         }
 
         SymbolTable *parent_scope = current_scope;
         current_scope = al.make_new<SymbolTable>(parent_scope);
+        std::string name = "j";
```
Thirumalai-Shaktivel, line 4454:
```py
from ltypes import i32
def main():
    i: i32
    for i in range(5):
        print(j) # not declared but this might compile
        i=i+1
        print(i)
main()
```

For that we have to mangle the name, something like `1_i_j` or `_i_j`.

anutosh491: Yess it does unfortunately , do I need to use something like the following 
```
std::string j = current_scope->get_unique_name("__j");
```

Thirumalai-Shaktivel:
Rest looks great! Thanks for fixing this.
I think you can add integration_tests, instead of printing just use assert (in and after the loop.).

src/lpython/semantics/python_ast_to_asr.cpp
Thirumalai-Shaktivel: Yup, you can do that.

anutosh491:
>  I think you can add integration_tests, instead of printing just use assert (in and after the loop.).

Sure .Once you address this comment https://github.com/lcompilers/lpython/pull/1543#discussion_r1115225933, I'll do all needful here .

src/lpython/semantics/python_ast_to_asr.cpp
anutosh491: Yes , I tried this ... works as expected !

anutosh491: I see the example program you've shared throwing a Semantic error when I run it 
```
semantic error: Variable 'j' not declared
 --> examples/expr2.py:5:15
  |
5 |         print(j) # `j` not declared but this might compile
  |               ^ 


```

Thirumalai-Shaktivel: I think `_j` might also be caught as well, do `1_j` instead (as `1_j` is not allowed as a variable in Python)
```
from ltypes import i32
def main():
    i: i32
    for i in range(5):
        print(_j) # `_j` or `_j1` not declared but this might compile
        i=i+1
        print(i)
main()
```

Thirumalai-Shaktivel: `1_for_tmp_var` or `1_for_dummy` or `1_<variable_name>_copy` or something. So that the developer or user can understand where it comes from

anutosh491: Sure yeah , `j` doesn't meet the purpose completely !

anutosh491:
> I think you can add integration_tests,

I've added one and registered the same with `cpython` and `llvm` for now  . You could have a look at it !

Thirumalai-Shaktivel:
Make a clean build, and update tests.
```
git clean -dfx
./build0.sh && ./build1.sh
```

anutosh491:
> Make a clean build, and update tests.
> 
> ```
> git clean -dfx
> ./build0.sh && ./build1.sh
> ```

Ohhh do I do this after reverting my last 2 commits ?

Thirumalai-Shaktivel:
I think the only issue here is with the BlockCall not being implemented in wat.

Thirumalai-Shaktivel:
You can try implementing `visit_BlockCall` in `asr_to_wasm.cpp`. If you get any doubts, please ask.

anutosh491:
Sure I'll study what `visit_BlockCall` stands for ! Would be great if you could educate me on how to found this out ( I mean the code already fixed all wrongly working cases and also all tests would run locally , so I am not sure how you realized this !)

Thirumalai-Shaktivel:
Block example in C++:
```cpp
{		// new block
	int x;
	x = 0;
	std::court << x << "\n";
}		// block ends
```

Thirumalai-Shaktivel:
BlockCall is already implemented in C (asr_to_c_cpp.h). You can see it for understanding and further implementations.
Basically, you should declare the variables (by visit_variables) and visit all the stmts.

Thirumalai-Shaktivel:
In this PR, we create a `dummy` variable to store the `i` value: 
so the following
```py
for I in range(5):
	print(i)
	i=i+1
```
becomes:
```py
for i in range(5):
	dummy: type(i) = i # ASR
	print(i)
	i=i+1
	i=dummy  # ASR
	i=i+1    # ASR increment
```

In visit_for, there was already a block node used:
https://github.com/lcompilers/lpython/blob/ba525e69ad328d28a492134b1d824a9e44a898d0/src/lpython/semantics/python_ast_to_asr.cpp#L4455
So, we thought to make use of it, by declaring the dummy variable inside the block scope. But this seems to be a problem in wasm, as Shaikh mentioned here 2.: https://github.com/lcompilers/lpython/pull/1543#issuecomment-1441542720 
Now we also have another option to do this without block: we have to create this dummy variable in for_loop's parent scope.

@certik @czgdp1807 @Shaikh-Ubaid What are your thoughts about this? Declare dummy in loop scope (using block) or loop's parent scope?

anutosh491:
Hello @Thirumalai-Shaktivel , now that #1548 (merged) implements the basic structure for `visit_BlockCall` , would it make sense to update this branch to the current main and test out things . Is there anything left to address ?

anutosh491:
> Now we also have another option to do this without block: we have to create this dummy variable in for_loop's parent scope.

By this do mean , incorporating the following diff ?
```

@@ -4451,10 +4451,10 @@ public:

         SymbolTable *parent_scope = current_scope;
         current_scope = al.make_new<SymbolTable>(parent_scope);
-        std::string name = current_scope->get_unique_name("_1_for_dummy");
+        std::string name = parent_scope->get_unique_name("_1_for_dummy");
         ASR::expr_t *dummy;
         dummy = PassUtils::create_auxiliary_variable(x.base.base.loc,
-                name, al, current_scope, a_type);
+                name, al, parent_scope, a_type);
```
As per what I can see & test , even this would do the job for us , but not sure if this is the way to go !?

certik:
I think we should not allow assigning to loop variables, we should do the same approach as in https://github.com/lfortran/lfortran/pull/1345.

anutosh491:
> I think we should not allow assigning to loop variables, we should do the same approach as in [lfortran/lfortran#1345](https://github.com/lfortran/lfortran/pull/1345).

Ohh , do you mean we raise an error for any re definition/local assignment of the loop variable ? Why is this though , I mean we do expect the following through CPython right ?
```
def main():
    i: i32
    for i in range(5):
        print(i)
        i=i+1

main()

0
1
2
3
4
```

certik:
> Ohh , do you mean we raise an error for any re definition/local assignment of the loop variable ?

Yes.

> Why is this though , I mean we do expect the following through CPython right ?

The contract is that if it compiles and runs with LPython, it will work with CPython, but not the other way around. So we should restrict it, just like in LFortran. That fixes the immediate problem.

As a separate topic later, we can investigate if we should relax the condition. I think assigning to a loop variable is a bad programming practice, so I think we should not do it, but we can discuss it.

certik (approved):
Looks great!

PR #1546, opened 2023-02-23
feat: add bit count method

diivi:
Adds the bit_count() method for integers to lpython.
Related to #281


Shaikh-Ubaid:
I am unsure if the approach in the PR is best way to solve this. Maybe we could try to implement it as part of pure runtime library.

Currently, I am just trying to get the tests to pass here.

integration_tests/CMakeLists.txt
```diff
@@ -400,6 +400,7 @@ RUN(NAME logical_binop1      LABELS cpython llvm)
 RUN(NAME vec_01              LABELS cpython llvm c)
 RUN(NAME test_str_comparison LABELS cpython llvm c)
 RUN(NAME test_bit_length     LABELS cpython llvm c)
+RUN(NAME test_bit_count      LABELS cpython llvm c)
```
Shaikh-Ubaid, line 403: @diivi the labels here enable that particular test for the specified backend. Here, you enabled the `test_bit_count` test case for `llvm` as well as `c`. The tests passed for `llvm`, but were failing for `c` as `visit_IntegerBitCount` was not implemented in the `c` backend.

diivi: Ah, okay thanks! I referred the int.bit_length() method so kept it the same as that. Could you help me understand the difference/benefit of adding the visit_IntegerBitCount function in the c backend too? And why this wasn't done for visit_IntegerBitLength? Thanks!

Shaikh-Ubaid: >And why this wasn't done for visit_IntegerBitLength?

I think `visit_IntegerBitLength()` is already supported in the `c` backend. It is defined here:
https://github.com/lcompilers/lpython/blob/ba525e69ad328d28a492134b1d824a9e44a898d0/src/libasr/codegen/asr_to_c_cpp.h#L1337-L1348

>Could you help me understand the difference/benefit of adding the visit_IntegerBitCount function in the c backend too?

The more backends support a feature the better for us. (Ideally, every backend would/should support every feature). In your PR, since you implemented most of the required portion and the CI tests were failing for `c` backend for the test case `test_bit_count` (due to missing `visit_IntegerBitCount`), we went ahead and added support for it so that the CI passes. 

What we could have also done is to not enable the `c` backend for that specific test (since, the `c` backend didn't have support for it). And mayber later (in subsequent PRs) we could have added support for that test case in the `c` backend.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -1347,6 +1347,19 @@ R"(#include <stdio.h>
...
+    void visit_IntegerBitCount(const ASR::IntegerBitCount_t& x) {
+        self().visit_expr(*x.m_a);
+        int arg_kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);
+        switch (arg_kind) {
+            case 1: src = "_lpython_bit_count1(" + src + ")"; break;
+            case 2: src = "_lpython_bit_count2(" + src + ")"; break;
+            case 4: src = "_lpython_bit_count4(" + src + ")"; break;
+            case 8: src = "_lpython_bit_count8(" + src + ")"; break;
+            default: throw CodeGenError("Unsupported Integer Kind: " + \
+                            std::to_string(arg_kind));
```
czgdp1807, line 1359:
```suggestion
        src = "_lpython_bit_count" + std::to_string(arg_kind) + "(" + src + ")";
```

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -808,6 +808,46 @@ LFORTRAN_API int32_t _lpython_bit_length8(int64_t num)
...
+    return res;
+}
+
+LFORTRAN_API int32_t _lpython_bit_count2(int16_t num)
+{
+    int32_t res = 0;
+    num = abs(num);
+    for(; num; num >>= 1){
+        res += num & 1;
+    }
```
czgdp1807, line 827: This is fine but implementing it in C runtime library doesn't feel good. Its so simple that can be done directly in the ASR itself. Replace `IntegerBitCount` with the above loop using ASR APIs, so the whole implementation will be inlined in each backend and no overhead of any function call.

I am leaning towards the ASR approach for now. I don't think we need this feature urgently, so if we do it correctly then it would be better.

src/libasr/codegen/asr_to_c_cpp.h
diivi: Should I do this for the `visit_IntegerBitLen` function above too?

diivi: Also, once I do this directly in the ASR, I can remove the entire function right?

czgdp1807: I meant doing it as an ASR to ASR pass. We should use `BaseExprReplacer`. @certik Do you think https://github.com/lcompilers/lpython/pull/1546#discussion_r1116489030 is a good idea?

certik:
These auxiliary builtin methods will be cleaned up with: https://github.com/lfortran/lfortran/pull/1045

src/libasr/codegen/asr_to_c_cpp.h
certik: Yes.

PR #1548, opened 2023-02-23
WASM: Initial support for BlockCall

Shaikh-Ubaid:
This PR implements initial support for `visit_BlockCall` in the `wasm` backend.
- Varaibles and statements inside a block are supported (their support is yet to be verfied by adding a test case)
    - For variables declared inside the block scope, we are declaring them as local variables of the function (thus they are now available throughout the whole function scope).
- GoTo and GoToTarget are not yet supported

### DIFF
--- a/src/libasr/codegen/asr_to_wasm.cpp
+++ b/src/libasr/codegen/asr_to_wasm.cpp
@@ -897,42 +897,30 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
         }
     }
 
-    template <typename T>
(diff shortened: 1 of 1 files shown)


Shaikh-Ubaid:
There is a test case to be added here.

I think/guess python does not have block or goto statements. @czgdp1807 How is a `block` defined/represented in `lpython`?

certik:
LPython doesn't use blocks, and if this is tested by LFortran, then we don't need to test it here.

Shaikh-Ubaid:
I am merging this as its related PR https://github.com/lfortran/lfortran/pull/1354 has been merged.

PR #1550, opened 2023-02-25
feat: add title and upper methods to strings

diivi:
Add the `title` and `upper` methods to strings.


src/runtime/lpython_builtin.py
```diff
@@ -619,6 +619,31 @@ def _lpython_str_capitalize(x: str) -> str:
     x = chr(val) + x[1:]
     return x
 
+@overload
```
faze-geek, line 622: Hi, existing reference files get modified and new reference files get generated when you make an addition like this to the code base. You should run `./run_tests.py -u` to do the following.

diivi: oh okay, I did that but that changed a lot of files so I asked on Zulip, thanks, I'll do that again.

certik:
@czgdp1807 I think it's fine to get these functions working in this way. Later as we get IntrinsicFunction working better, we can refactor this and remove it from the C library.

PR #1551, opened 2023-02-26
Treating Integer and Real types as immutable

faze-geek:
The concept of immutable has a deeper meaning than being a variable that is easy to modify. You can checkout this post of reference. - https://stackoverflow.com/questions/62177372/is-integer-immutable-or-mutable

```
def f():
    i: f64
    i = 3.14
    print(i)
    i = 3.1415926359
    print(i)

f()

3.14
3.1415926359

3.14000000000000012e+00
3.14159263589999993e+00
```
But assignment is not correct in python -
Such an error is not given on master
```

def f():
    i: f64
    i = 3.14
    i[0] = 3.1415926359
    print(i)

f()
  File "C:\Users\kunni\lpython\try.py", line 14, in <module>
    f()
  File "C:\Users\kunni\lpython\try.py", line 11, in f
    i[0] = 3.1415926359
TypeError: 'float' object does not support item assignment

semantic error: 'f64' object does not support item assignment
  --> try.py:11:5
   |
11 |     i[0] = 3.1415926359
   |     ^^^^^^^^^^^^^^^^^^^


```

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -1145,7 +1145,7 @@ static inline std::string binop_to_str_python(const ASR::binopType t) {
 
 static inline bool is_immutable(const ASR::ttype_t *type) {
     return ((ASR::is_a<ASR::Character_t>(*type) || ASR::is_a<ASR::Tuple_t>(*type)
-        || ASR::is_a<ASR::Complex_t>(*type)));
+        || ASR::is_a<ASR::Complex_t>(*type)) || ASR::is_a<ASR::Integer_t>(*type) || ASR::is_a<ASR::Real_t>(*type));
 }
 
 // Returns a list of values
--- a/tests/errors/test_assign10.py
+++ b/tests/errors/test_assign10.py
@@ -1,0 +1,8 @@
+from ltypes import f64
+
+def f():
+    i: f64
+    i = 3.14
+    i[0] = 3.1415926359
+
+f()
--- a/tests/errors/test_assign9.py
+++ b/tests/errors/test_assign9.py
@@ -1,0 +1,8 @@
+from ltypes import i32
+
+def f():
+    i: i32
+    i = 100
+    i[0] = 500
+
+f()
--- a/tests/tests.toml
+++ b/tests/tests.toml
@@ -753,6 +753,14 @@ filename = "errors/test_assign8.py"
 asr = true
 
 [[test]]
+filename = "errors/test_assign9.py"
+asr = true
+
+[[test]]
+filename = "errors/test_assign10.py"
+asr = true
+
+[[test]]
 filename = "errors/test_binop1.py"
 asr = true
 


faze-geek:
@Thirumalai-Shaktivel I have separate commits for the code and the updated reference files, so that in case of any mistake the 2nd commit can be reverted straight away. I would still appreciate if you check the 2nd commit once.

tests/reference/asr-array_01_decl-39cf894.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: 'i16[:]' object does not support item assignment
```
certik, line 1: I think this error is not correct, `i16[:]` does support item assignment.

faze-geek: > I think this error is not correct, `i16[:]` does support item assignment.

Yes, agreed.
I'm not sure how to handle this, my change was only intended to catch pure integer and floating types like `i8, i16, i32, i64, f32, f64`. I'll look into this soon.

faze-geek: I think this diff does the job - basically we want to check the length of the structure. Like integers are dimensionless -
```
+    bool is_integer = ASR::is_a<ASR::Integer_t>(*type);
+    bool is_dimensionless = false;
+    bool is_real = ASR::is_a<ASR::Real_t>(*type);
+    if(is_integer){
+        ASR::Integer_t *integer = ASR::down_cast<ASR::Integer_t>(type);
+        is_dimensionless = integer->n_dims == 0;
+    }
+    if(is_real){
+        ASR::Real_t *real = ASR::down_cast<ASR::Real_t>(type);
+        is_dimensionless = real->n_dims == 0;
+    }
     return ((ASR::is_a<ASR::Character_t>(*type) || ASR::is_a<ASR::Tuple_t>(*type)
-        || ASR::is_a<ASR::Complex_t>(*type)) || ASR::is_a<ASR::Integer_t>(*type) || ASR::is_a<ASR::Real_t>(*type));
+        || ASR::is_a<ASR::Complex_t>(*type)) || is_dimensionless );
 }
```

faze-geek: So let us checkout a failing test case -

```
from numpy import empty
def loop_vec():
    a: f64[5] = empty(5)
    i: i32 = 0
    for i in range(5):
        a[i] = 5.0

loop_vec()

(lp) C:\Users\kunni\lpython>src\bin\lpython try.py  # No longer catches the error

```
But as soon as I make the array into a floating point variable -
```
def loop_vec():
    a: f64 = 5.0
    i: i32 = 0
    for i in range(5):
        a[i] = 5.1

loop_vec()

  File "C:\Users\kunni\lpython\try.py", line 8, in <module>
    loop_vec()
  File "C:\Users\kunni\lpython\try.py", line 6, in loop_vec
    a[i] = 5.1
TypeError: 'float' object does not support item assignment

semantic error: 'f64' object does not support item assignment
 --> try.py:6:9
  |
6 |         a[i] = 5.1
  |         ^^^^^^^^^^


```

faze-geek: cc @certik

faze-geek:
I believe the change does what I intended earlier. I just see 1 issue here-
New reference files were generated with an earlier commit which you can see in the Files changed section. They should not be present.
I've checked all the possible test files eg- `tests/../integration_tests/array_01_decl.py` , `tests/../integration_tests/bindc_02.py` , `tests/../integration_tests/expr_12.py `etc. The error is not caught in my branch and all these test work as expected.

Should I manually delete them and commit ?

faze-geek:
@czgdp1807 Can you have a look at it . Thanks!

src/libasr/asr_utils.h
```diff
@@ -1144,8 +1144,19 @@ static inline std::string binop_to_str_python(const ASR::binopType t) {
...
+        ASR::Integer_t *integer = ASR::down_cast<ASR::Integer_t>(type);
+        is_dimensionless = integer->n_dims == 0;
+    }
+    if(is_real){
+        ASR::Real_t *real = ASR::down_cast<ASR::Real_t>(type);
+        is_dimensionless = real->n_dims == 0;
+    }
     return ((ASR::is_a<ASR::Character_t>(*type) || ASR::is_a<ASR::Tuple_t>(*type)
-        || ASR::is_a<ASR::Complex_t>(*type)));
+        || ASR::is_a<ASR::Complex_t>(*type)) || is_dimensionless);
```
czgdp1807, line 1159:
```suggestion
        || ASR::is_a<ASR::Complex_t>(*type)) || !ASRUtils::is_array(type));
```

czgdp1807 (changes requested):

src/libasr/asr_utils.h
faze-geek: Didn't realize there was a function for this :) I'll try out this suggestion. Thanks.

faze-geek: I tried this diff directly but it doesn't seem to work -
```
const LCompilers::ASR::ttype_t *type
argument of type "const LCompilers::ASR::ttype_t *" is incompatible with parameter of type "LCompilers::ASR::ttype_t *"C/C++(167)
```
I will do some more investigation into this.

faze-geek: @Thirumalai-Shaktivel @certik This is the only roadblock to this pull request, rest the results are as expected. Could you have a look at it. Let's get this in!

Thirumalai-Shaktivel: Pass `type` using an `explicit` cast: `!ASRUtils::is_array((ASR::ttype_t *) type)`

PR #1553, opened 2023-02-27
Check for type mismatches in c_p_pointer and its target

czgdp1807:
Also, allow `IntegerBinOp` in dimension specification i.e., `i16[n*k]` is doable now.


certik (approved):
This is good, I opened up an issue for another problem to fix.

PR #1558, opened 2023-03-01
Overloaded mode function to support i64

tareknaser360:
# Overloaded mode function to support i64

## Description

Currently, the mode function only supports i32. This PR adds support for i64.

## Type of change

- New feature (non-breaking change which adds functionality)

## Notes

- I tried adding support for f64 but I encountered this error when I initialized the count dict with `count: dict[f64, i32] = {f64(0.0): 0}`  
```
semantic error: 'dict' key type cannot be float/complex because resolving collisions by exact comparison of float/complex values will result in unexpected behaviours. In addition fuzzy equality checks 
with a certain tolerance does not follow transitivity with float/complex values.
```

**Test Configuration**:
* Firmware version: Ubuntu 20.04.5 LTS

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings


tareknaser360:
I think we can overcome the error with f64 by sorting the array first and then getting the max count as a variable to be updated whenever we find a number with more occurrences. 

However, this would make this function less optimized depending on the sorting algorithm of choice.

certik:
@tareknaser360 thanks. Can you please add a test for this new function?

tareknaser360:
> @tareknaser360 thanks. Can you please add a test for this new function?

The tests are already implemented in `/integration_tests/test_statistics.py` 
I noticed that there are no separate tests for i64

czgdp1807:
Supporting mode of floating points doesn't seem reasonable to me. Here's the reasoning - https://github.com/lcompilers/lpython/pull/958#discussion_r960188424. However you can support them but without using `dict` for now.

czgdp1807:
> I noticed that there are no separate tests for i64

Please add the tests for `i64` in the same file.

src/runtime/statistics.py
```diff
@@ -220,8 +220,7 @@ def harmonic_mean(x: list[f64]) -> f64:
 
     return k / sum
 
-
-# TODO: Use generics to support other types.
```
czgdp1807, line 224: Please restore this `TODO`.

czgdp1807 (changes requested):

tareknaser360:
> > I noticed that there are no separate tests for i64
> 
> Please add the tests for `i64` in the same file.

Done

tareknaser360:
> Supporting mode of floating points doesn't seem reasonable to me. Here's the reasoning - [#958 (comment)](https://github.com/lcompilers/lpython/pull/958#discussion_r960188424). However you can support them but without using `dict` for now.

I think the best solution is to sort the list first. As in here:
```
@overload
def mode(x: list[f64]) -> f64:
    k: i32 = len(x)
    c: i32
    x.sort()

    ans: f64
    ans = x[0]

    max_count: f64 = 1.0
    curr_count: f64 = 1.0

    for c in range(1, k):
        if x[c] == x[c - 1]:
            curr_count = curr_count + 1
        else:
            curr_count = 1
        if curr_count > max_count:
            max_count = curr_count
            ans = x[c - 1]
    return ans
```

but  list.sort is not implemented yet. 
Is there any clean way to implement this?

rebcabin:
iirc, list.sort modifies a list in-place so might not be a safe way to sort on the way to computing the mode. We'll probably need some library function to sort a copy. This is a continual source of surprise and annoyance in Python.

also, sfaik, sort is a standard way to get the mode.

tareknaser360:
Should we leave out the support for float64 in the mode function until we either have dictionaries with key type float or list.sort implemented?

@czgdp1807  mentioned that supporting mode of floating points doesn't seem reasonable because for example `1.5` and `1.49999999` are approximately equal but while hashing they will generate totally different hash values.

tareknaser360:
> iirc, list.sort modifies a list in-place so might not be a safe way to sort on the way to computing the mode. We'll probably need some library function to sort a copy. This is a continual source of surprise and annoyance in Python.
> 
> also, sfaik, sort is a standard way to get the mode.

What kind of library function can we use to achieve this?

integration_tests/test_statistics.py
```diff
@@ -208,10 +208,11 @@ def test_mode():
...
 
-    d: list[i32]
-    d = [-1, 2, -3, -5, -3, -1, 4, -2, 4, -5, -3, 4, -3]
-    k = mode(d)
-    assert k == -3
+    d: list[i64]
+    d = [i64(-1), i64(2), i64(-3), i64(-5), i64(-3), i64(-1), i64(4), i64(-2), i64(4), i64(-5), i64(-3), i64(4), i64(-3)]
+    l: i64
+    l = mode(d)
+    assert l == i64(-3)
```
certik, line 215: Keep the i32 test, but add the i64 one in addition.

src/runtime/statistics.py
```diff
@@ -220,8 +220,8 @@ def harmonic_mean(x: list[f64]) -> f64:
 
     return k / sum
 
-
```
certik, line 223: Keep the empty line here I think.

certik:
For now just use the same algorithm as for i32. Later we can upgrade it with sort, once we implement sort.

integration_tests/test_statistics.py
```diff
@@ -207,12 +207,18 @@ def test_mode():
     k: i32
     k = mode(c)
     assert k == 2
-
+
```
certik, line 210: Please remove white space.

integration_tests/test_statistics.py
```diff
@@ -207,12 +207,18 @@ def test_mode():
...
+    
     d: list[i32]
     d = [-1, 2, -3, -5, -3, -1, 4, -2, 4, -5, -3, 4, -3]
     k = mode(d)
     assert k == -3
 
+    e: list[i64]
+    e = [i64(-1), i64(2), i64(-3), i64(-5), i64(-3), i64(-1), i64(4), i64(-2), i64(4), i64(-5), i64(-3), i64(4), i64(-3)]
+    l: i64
+    l = mode(e)
```
certik, line 219:
```suggestion
    l: i64 = mode(e)
```

tareknaser360:
> A function that deepcopies a list, sorts the deepcopy, then returns the deepcopy.


I am sorry but can you elaborate further?
What function exactly should I use?
I found out that `sorted()` creates a copy of the list before sorting but it doesn't create a deep-copy.
Also,
Using this code `y: list[f64] = sorted(x)` generates this error when testing on my machine `Function 'sorted' is not declared and not intrinsic`

rebcabin:
Perhaps a shallow copy suffices. I'm glad you found out that `sorted` makes a copy. I had the opposite impression. 

By "library function" I meant that we write our own utility function.

tareknaser360:
> By "library function" I meant that we write our own utility function. 

Sure, let me do that.
Should I make a separate PR for this?
I think this PR has done its job. 

What do you recommend?

rebcabin:
it depends on what you're trying to do. if list.sorted already makes a
copy, then it's not necessary to make a new function. Otherwise, you should
IMO.

tareknaser360:
Yes, `y: list[f64] = sorted(x)` makes a copy of the list. 
but on my machine I get this error when I try to build.
`Function 'sorted' is not declared and not intrinsic`

Up till now, in this PR we added the support for i64. 
Should we leave the work for f64 support until we have `list.sort()` or `sorted(list)` ?

rebcabin:
how are you finding mode for other types without sorting the list?

tareknaser360:
Using a dictionary.
This was used for the i32 type and I overloaded the function to support i64.

rebcabin:
I see. Have you benchmarked the dictionary solution versus an alternative
sort solution? (you could do so in ordinary CPython).

tareknaser360:
Sure, Here is the code I used to benchmark 
```
import timeit, sys
print(sys.version)

def mode_using_dict():
    l = [3, 4, 4, 5, 6, 4]
    count = {}
    for i in range(len(l)):
        count[l[i]] = 0
    for i in range(len(l)):
        count[l[i]] = count[l[i]] + 1

    max_count = 0
    ans = l[0]
    for i in range(len(l)):
        if max_count < count[l[i]]:
            max_count = count[l[i]]
            ans = l[i]
    return ans

def mode_using_sort():
    l = [3, 4, 4, 5, 6, 4]
    l_sorted = sorted(l)
    ans = l[0]
    max_count = 1
    curr_count = 1
    for i in range(1, len(l)):
        if l_sorted[i] == l_sorted[i - 1]:
            curr_count = curr_count + 1
        else:
            curr_count = 1
        if curr_count > max_count:
            max_count = curr_count
            ans = l_sorted[i - 1]
    return ans

time_using_dict = timeit.Timer(mode_using_dict).timeit(number=1000000)
print ("python can do %d mode_using_dict per second", time_using_dict)

time_using_sort = timeit.Timer(mode_using_sort).timeit(number=1000000)
print ("python can do %d mode_using_sort per second", time_using_sort)
```

and this is the output on my machine
```
3.9.0 (tags/v3.9.0:9cf6752, Oct  5 2020, 15:34:40) [MSC v.1927 64 bit (AMD64)]
python can do %d mode_using_dict per second 1.8609775
python can do %d mode_using_sort per second 0.9502808000000003
```

I think the sort method is computationally better as it takes approximately half the time to run

tareknaser360:
Sure, Here is the code I used to benchmark 
```
import timeit, sys
print(sys.version)

def mode_using_dict():
    l = [3, 4, 4, 5, 6, 4]
    count = {}
    for i in range(len(l)):
        count[l[i]] = 0
    for i in range(len(l)):
        count[l[i]] = count[l[i]] + 1

    max_count = 0
    ans = l[0]
    for i in range(len(l)):
        if max_count < count[l[i]]:
            max_count = count[l[i]]
            ans = l[i]
    return ans

def mode_using_sort():
    l = [3, 4, 4, 5, 6, 4]
    l_sorted = sorted(l)
    ans = l[0]
    max_count = 1
    curr_count = 1
    for i in range(1, len(l)):
        if l_sorted[i] == l_sorted[i - 1]:
            curr_count = curr_count + 1
        else:
            curr_count = 1
        if curr_count > max_count:
            max_count = curr_count
            ans = l_sorted[i - 1]
    return ans

time_using_dict = timeit.Timer(mode_using_dict).timeit(number=1000000)
print ("python can do %d mode_using_dict per second", time_using_dict)

time_using_sort = timeit.Timer(mode_using_sort).timeit(number=1000000)
print ("python can do %d mode_using_sort per second", time_using_sort)
```

and this is the output on my machine
```
3.9.0 (tags/v3.9.0:9cf6752, Oct  5 2020, 15:34:40) [MSC v.1927 64 bit (AMD64)]
python can do %d mode_using_dict per second 1.8609775
python can do %d mode_using_sort per second 0.9502808000000003
```

I think the sort method is computationally better as it takes approximately half the time to run

czgdp1807:
Input size is too small. Can you increase it to 1000 or higher?

tareknaser360:
Yes, I ran this test locally
```
import timeit, sys, random
random.seed(0)
print(sys.version)

def mode_using_dict():
    l = [random.randint(1, 20) for _ in range(1000)]
    count = {}
    for i in range(len(l)):
        count[l[i]] = 0
    for i in range(len(l)):
        count[l[i]] = count[l[i]] + 1

    max_count = 0
    ans = l[0]
    for i in range(len(l)):
        if max_count < count[l[i]]:
            max_count = count[l[i]]
            ans = l[i]
    return ans

def mode_using_sort():
    l = [random.randint(1, 20) for _ in range(1000)]
    l_sorted = sorted(l)
    ans = l[0]
    max_count = 1
    curr_count = 1
    for i in range(1, len(l)):
        if l_sorted[i] == l_sorted[i - 1]:
            curr_count = curr_count + 1
        else:
            curr_count = 1
        if curr_count > max_count:
            max_count = curr_count
            ans = l_sorted[i - 1]
    return ans

time_using_dict = timeit.Timer(mode_using_dict).timeit(number=100000)
print ("python can do {0} mode_using_dict".format(str(time_using_dict)))

time_using_sort = timeit.Timer(mode_using_sort).timeit(number=100000)
print ("python can do {0} mode_using_sort".format(str(time_using_sort)))
```

This is the output now 
```
3.9.0 (tags/v3.9.0:9cf6752, Oct  5 2020, 15:34:40) [MSC v.1927 64 bit (AMD64)]
python can do 99.44749569999999 mode_using_dict
python can do 92.26956440000001 mode_using_sort
```

rebcabin:
I hypothesize that sort will be better than dict for all data types and all sizes, but it's just a hypothesis :) thanks for the dilligent benchmarking to prove/refute the hypothesis!

czgdp1807:
I see. It depends on the internal implementation of `dict`. Theoretically, `dict` should perform better. because its O(N) and sort is O(NlgN). However, the constant factor for CPython `dict` seems higher, so that's causing a problem. LPython might show up different results because its fully compiled and its `dict` implementation is fully inlined. However same applies for sorting. Anyways, `f64` mods may be delayed (for now) and for `i64` you can use `dict` or sort approach after benchmarking.

tareknaser360:
> I hypothesize that sort will be better than dict for all data types and all sizes, but it's just a hypothesis :) thanks for the dilligent benchmarking to prove/refute the hypothesis!

Happy to help!

czgdp1807:
Any progress on this?

czgdp1807 (approved):

Smit-create (approved):

certik (approved):

PR #1559, opened 2023-03-02
Adding built-in functions `all()` and `any()`

tareknaser360:
# Add built-in `all()` and `any()`

## Description

`all(iterable)`: returns True if all elements of the iterable are true (or if the iterable is empty) 
`any(iterable)`: returns True if any element of the iterable is true. If the iterable is empty, returns False

## Type of change

- New feature (non-breaking change which adds functionality)

**System Configuration**:
* Firmware version: Ubuntu 20.04.5 LTS


tareknaser360:
How come the CI tests are only failing for macos?

src/runtime/lpython_builtin.py
```diff
@@ -100,6 +100,122 @@ def abs(c: c64) -> f64:
...
+@overload
+def any(arr: list[bool]) -> bool:
+    """
+    Return True if any element of the iterable is true. If the iterable is empty, return False.
+    """
+    i: i32
+    for i in range(len(arr)):
+        if arr[i]:
+            return True
+    return False
```
czgdp1807, line 217: I feel generics usage can prevent a lot of code repetition here. @certik Can we use generics for these kind of purposes? The other way is to generate these functions on demand using ASR APIs (that's the most optimal) because `all`, `any` can take any iterable as input (list, set, tuple, etc.).

tareknaser360: I tried using generics. However, I think this will only work for lists but I am not sure.

czgdp1807: Its cool. My above comment is for long term approach. If @certik agrees then we will have to add a `TODO` for all the implementations here so that we don't forget to replace them with a better one.

tareknaser360:
Any updates on this PR?
@czgdp1807  @certik

certik:
These have to be implemented using IntrinsicFunction. I think they might already be implemented.

tareknaser360:
Yes thanks!
I will go ahead and close this one now.

certik:
@tareknaser360 thanks for your work!

PR #1561, opened 2023-03-03
Remove chr() and ord() functions from runtime library

Shaikh-Ubaid:


Shaikh-Ubaid:

src/runtime/lpython_builtin.py
```diff
@@ -2,29 +2,22 @@
...
     """
     Returns the string representing a unicode character from
     the given Unicode code point. This is the inverse of `ord()`.
     """
-    if i < 32 or i > 126:
-        return "Not yet supported"
-    all_chars: str
-    all_chars = ' !"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~'
-    return all_chars[i - 32]
+    return str(i)
```
certik, line 20: Do these changes work with CPython also? It seems it does not:
```
In [1]: a = "x"

In [2]: int(a)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 int(a)

ValueError: invalid literal for int() with base 10: 'x'
```

certik: Both int and str convert decimal digits. But we want their ASCII / unicode values instead. So something is fundamentally wrong here. I think we need to fix LPython for str/int.

Shaikh-Ubaid: >Both int and str convert decimal digits. But we want their ASCII / unicode values instead. So something is fundamentally wrong here. I think we need to fix LPython for str/int.

I see. For this PR to proceed further, what should we do?

certik: I think we would need to implement ASR nodes or IntrinsicFunctions for chr and ord.

Shaikh-Ubaid: Got it. Thank you for the guidance.

src/runtime/lpython_builtin.py
```diff
@@ -2,29 +2,22 @@
...
     Returns an integer representing the Unicode code
     point of a given unicode character. This is the inverse of `chr()`.
     """
     if len(s) != 1:
         return -1 # not a character
-    i: i32
-    for i in range(32, 127):
-        if chr(i) == s:
-            return i
+    return i32(int(s))
```
czgdp1807, line 12: Good one.

certik: Since this does not work in CPython, it's a bug in LPython that we should fix.

src/runtime/lpython_builtin.py
Shaikh-Ubaid: >I think we would need to implement ASR nodes or IntrinsicFunctions for chr and ord.

Shall we implement the `chr` and `ord` functions as ASR nodes or as IntrinsicFunctions? In case we implement as ASR nodes, do we need to call into `C` or do we implement it using `llvm` instructions?

Shaikh-Ubaid:
We already have `chr` and `ord` ASR nodes.
https://github.com/lcompilers/lpython/blob/3c53ffb1c217405224ff53f0b3a7a5d306fb07d7/src/libasr/ASR.asdl#L270-L271

It seems the runtime implementations of `chr` and `ord` are unused. I removed them in this PR.

PR #1563, opened 2023-03-04
C: Add docs for `bracket_open`

Smit-create:
https://github.com/lcompilers/lpython/pull/1476#pullrequestreview-1324140235

### DIFF
--- a/src/libasr/codegen/asr_to_c_cpp.h
+++ b/src/libasr/codegen/asr_to_c_cpp.h
@@ -116,6 +116,17 @@ public:
     std::string const_name;
     size_t const_vars_count;
     size_t loop_end_count;
+
+    // This is used to track if during the codegeneration whether or not
+    // the source is inside any bracket. bracket_open is always >= 0. We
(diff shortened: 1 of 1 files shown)


certik (approved):
Thanks! I still don't understand why it is needed, but I now understand how to use it and maintain it, so that's good.

Smit-create:
> I still don't understand why it is needed

Consider the following example:
```py
def f():
    print([1, 2, 3])

f()
```
ASR:

```console
                                    [(Print
                                        ()
                                        [(ListConstant
                                            [(IntegerConstant 1 (Integer 4 []))
                                            (IntegerConstant 3 (Integer 4 []))]
                                            (List
                                                (Integer 4 [])
                                            )
                                        )]
                                        ()
                                        ()
                                    )]
```

So while visiting `Print` node, `C` will have `printf(..`, then we will visit `ListConstant`. So since `ListConstant` has some extra codegen, we want to prevent that extra code inside `printf`. `bracket_open` helps to keep the track of open brackets and once when `bracket_open==0`, we will add this extra code there.

Issue #1533, opened 2023-02-19
Negative variable indexing do not work for lists


anutosh491:
What works 
```
def main0():
    l: list[i32] = [1, 2, 3]
    x: i32 = 2
    print(l[1], l[-1], l[x])

main0()

(lf) anutosh491@spbhat68:~/lpython/lpython$ ./src/bin/lpython examples/expr2.py 
2 3 3
```

What doesn't work 
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ cat examples/expr2.py 
def main0():
    l: list[i32] = [1, 2, 3]
    x: i32 = -2
    print(l[x])

main0()

(lf) anutosh491@spbhat68:~/lpython/lpython$ ./src/bin/lpython examples/expr2.py 
IndexError: List index is out of range. Index range is (0, 2), but the given index is -2

(lf) anutosh491@spbhat68:~/lpython/lpython$ python examples/expr2.py 
2
```
So basically what's happening is , it follows the expected code flow and through the code block responsible for handling negative indexing . But here the `val` comes out to be a null pointer when indexing is done through a variable .

https://github.com/lcompilers/lpython/blob/8f62ac4ff1acebdeb621557d8008a5a8423455bb/src/lpython/semantics/python_ast_to_asr.cpp#L3109-L3114

The positive index remains unaffected though (it skips this block ) and still works out well through the code which follows 
```
tmp = make_ListItem_t(al, loc, value, index,
                                      ASR::down_cast<ASR::List_t>(type)->m_type, nullptr);
                return false;
```

anutosh491:
I am not fully sure how we could approach this . Indexing for strings are pretty similar 
```
            if (ASRUtils::is_character(*type)) {
                ASR::expr_t* val = ASRUtils::expr_value(index);
                if (val && ASR::is_a<ASR::IntegerConstant_t>(*val)) {
                    if (ASR::down_cast<ASR::IntegerConstant_t>(val)->m_n < 0) {
                        // Replace `x[-1]` to `x[len(x)+(-1)]`
                        ASR::ttype_t *int_type = ASRUtils::TYPE(ASR::make_Integer_t(
                                                        al, loc, 4, nullptr, 0));
                        ASR::expr_t *list_len = ASRUtils::EXPR(ASR::make_StringLen_t(
                                    al, loc, value, int_type, nullptr));
                        ASR::expr_t *neg_idx = ASRUtils::expr_value(index);
[...]

PR #1565, opened 2023-03-06
LLVM: Fix `list.insert()` when given index is greater than length of the list

faze-geek:
Consider `list.insert(position,element)`, In python if `position `is greater than the length of list then by default the `element `is inserted at `position = len(list)`. 
On master - 
```
def f():
    l:list[str] = ["a","b","c","d"]
    l.insert(6,"e")
    print(l)
f()

(lp) faze-geek@DESKTOP-497JDCU:~/lpython/lpython$ python try.py
['a', 'b', 'c', 'd', 'e']
(lp) faze-geek@DESKTOP-497JDCU:~/lpython/lpython$ ./src/bin/lpython try.py
['a', 'b', 'c', 'd', '(null)']    #lpython adds a garbage value

def f():
    l:list[i32] = [1,2,3]
    l.insert(8,4)
    print(l)
f()

(lp) faze-geek@DESKTOP-497JDCU:~/lpython/lpython$ python try.py
[1, 2, 3, 4]
(lp) faze-geek@DESKTOP-497JDCU:~/lpython/lpython$ ./src/bin/lpython try.py
[1, 2, 3, 0]    #lpython adds a garbage value

```

On branch - 
```
def f():
    l:list[str] = ["a","b","c","d"]
    l.insert(6,"e")
    print(l)
f()

['a', 'b', 'c', 'd', 'e']
['a', 'b', 'c', 'd', 'e']
```

### DIFF
--- a/integration_tests/test_list_03.py
+++ b/integration_tests/test_list_03.py
@@ -26,15 +26,20 @@ def test_list_insert_02(x: list[i32], n: i32) -> list[i32]:
 
 def test_list_insert_03():
     l1:list[str] = ["a", "b", "c", "d"]
-    l1.insert(6,"e")
+    l2:list[str] = []
+    l1.insert(5,"e")
     s:str = "a"
     i:i32 = 0
     for i in range(len(l1)):
         assert l1[i] == chr(ord(s)+i)
 
-    l1.insert(18,"f")
-    for i in range(len(l1)):
-        assert l1[i] == chr(ord(s)+i)
+def test_list_insert_04():
+    l:list[i32] = [1, 2, 3, 4, 5]
+    i:i32
+    l.insert(6,6)
+    l.insert(100,7)
+    for i in range(1,8):
+        assert l[i-1] == i
 
 def test_list_02(n: i32) -> i32:
     x: list[i32] = [50, 1]
@@ -73,5 +78,6 @@ def verify():
     assert test_list_02(50) == 3628
     test_list_02_string()
     test_list_insert_03()
+    test_list_insert_04()
 
 verify()


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2365,8 +2365,18 @@ namespace LCompilers {
...
+            builder->CreateICmpSGT(pos, current_end_point),
+            current_end_point,
+            pos
+        );
+        adjusted_pos = builder->CreateSelect(
+            builder->CreateICmpSGE(adjusted_pos, llvm::ConstantInt::get(context, llvm::APInt(32, 0))),
+            adjusted_pos,
+            llvm::ConstantInt::get(context, llvm::APInt(32, 0))
+        );
         llvm::AllocaInst *tmp_ptr = builder->CreateAlloca(el_type, nullptr);
```
Thirumalai-Shaktivel, line 2378: The TODO at line 2365 is for AllocaInst, I think. Why don't you move this above?

faze-geek: Okay sure. You are right.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2344,7 +2344,17 @@ namespace LCompilers {
...
+            builder->CreateICmpSGT(pos, current_end_point),
+            current_end_point,
+            pos
+        );
+        adjusted_pos = builder->CreateSelect(
+            builder->CreateICmpSGE(adjusted_pos, llvm::ConstantInt::get(context, llvm::APInt(32, 0))),
+            adjusted_pos,
+            llvm::ConstantInt::get(context, llvm::APInt(32, 0))
+        );
+
```
czgdp1807, line 2357: I guess so for this, we can use, [CreateMinimum](https://llvm.org/doxygen/classllvm_1_1IRBuilderBase.html#ada46861dd8cf60fa7e1ac61bbd385b49). Its just like, `min(len(list), position)`. If `position` will be greater than `len(list)` then automatically `position` will not be used and otherwise `position` will be used.

faze-geek: Sorry I have not got very familiar with `llvm` , I tried - 
`llvm::Value* adjusted_pos = builder->CreateMinimum(pos, current_end_point);
`
But it leads to an error while running above test cases - 
```
attributes #0 = { argmemonly nounwind willreturn }
attributes #1 = { nounwind readnone speculatable willreturn }
Intrinsic has incorrect return type!
i32 (i32, i32)* @llvm.minimum.i32
```
Could you help me out here !

czgdp1807: I see. Why are we having, `adjusted_pos = builder->CreateSelect(
             builder->CreateICmpSGE(adjusted_pos, llvm::ConstantInt::get(context, llvm::APInt(32, 0))),
             adjusted_pos,
             llvm::ConstantInt::get(context, llvm::APInt(32, 0))
         );`?  For handling negative indices?

faze-geek: @czgdp1807  Sorry I could not respond to this earlier, was busy with exams at university. 

Yes, that's correct. The code you mentioned is responsible for handling negative indices. Basically, it ensures new item is inserted at the beginning of the list if a negative index is provided.

faze-geek: I remember a comment regarding not allowing negative indexing in the list ( allowed as of now ). You can check it [here](https://github.com/lcompilers/lpython/issues/1533#issuecomment-1445144248)
 
It thats the case I'll just remote the snippet checking for negative indices. Rest the pr looks good to me.

czgdp1807 (changes requested):
Please add a test for it or point me to any existing tests which verify that your change works correctly.

faze-geek:
> Please add a test for it or point me to any existing tests which verify that your change works correctly.

I have added a test case now. Basically, it handles cases when first argument is greater than list length.

integration_tests/test_list_03.py
```diff
@@ -24,6 +24,11 @@ def test_list_insert_02(x: list[i32], n: i32) -> list[i32]:
 
     return x
 
+def test_list_insert_03():
+    l1:list[str] = ["a","b","c","d"]
+    l1.insert(6,"e")
+    assert l1 == ['a', 'b', 'c', 'd', 'e']
+
```
czgdp1807, line 31:
```suggestion
    l1:list[str] = ["a", "b", "c", "d"]
    l1.insert(6,"e")
    assert l1 == ['a',  'b',  'c',  'd',  'e']
    l1.insert(18, "f")
    assert l1 == ['a', 'b', 'c', 'd', 'e', 'f']
```

czgdp1807 (changes requested):

integration_tests/test_list_03.py
faze-geek: Okay I'll add this quick. 
Also the logs fail, does asserting like this work or should I try asserting each element in a for loop ?

czgdp1807: Asserting each element should be better option for making sure our list algorithms are working as expected. But as a feature direct comparison of lists should also work (but that's a separate issue).

faze-geek: As of now, all the integration tests do element wise comparison (possibly because direct assertion is an issue). I'll use a for loop now.

faze-geek: I have updated the test accordingly.

faze-geek:
@czgdp1807 
I investigated the failing logs now. Firstly the running `integration_tests/test_list_03.py` (where tests have been added) does not give a segmentation error in windows MSVC unlike my WSL setup. This is a seperate issue.
Secondly the `c backend` in WSL raises the Segmentation Fault when I test indexes greater than the length of the list and as I use an index within the length the error does not occur.
But then this case is a valid edge case as I have shown above, comparing with Cpython -
```
def test_list_insert_03():
    l1:list[str] = ["a", "b", "c", "d"]
    l1.insert(5,"e")
    s:str = "a"
    i:i32 = 0
    # for i in range(len(l1)):
    #     assert l1[i] == chr(ord(s)+i)
    print(l1)

test_list_insert_03()

['a', 'b', 'c', 'd', 'e']

['a', 'b', 'c', 'd', 'e'] 
```

src/libasr/codegen/llvm_utils.cpp
czgdp1807, line 2357: Seems like an issue with edge case. Note that `current_end_point` equals `len(list)` all the times.

https://github.com/lcompilers/lpython/blob/57ebbc81e23854d3af50c2d5042f2f1b184941b5/src/libasr/codegen/llvm_utils.cpp#L2252-L2254

Issue #1562, opened 2023-03-04
% operator gives incorrect results compared to Cpython


faze-geek:
Got this while testing the % operator alongside `math.mod ` in Lpython - 

```
def f():
    print((8)%3)
    print((8)%(3))
    print(-(8)%3)          #wrong
    print((8)%-3)          #wrong
    print(-(8)%-3)
    print((-8)%3)          #wrong
    print(-8%3)            #wrong
    
f()

(lp) C:\Users\kunni\lpython>python try.py
2
2
1
-1
-2
1
1

(lp) C:\Users\kunni\lpython>src\bin\lpython try.py
2
2
-2
2
-2
-2
-2
```

faze-geek:
`LPython` gives correct results with `math.mod` but still I feel % should be corrected because of its use at some instances in the codebase.

```
import math
def f():
    print(math.mod((8),3))
    print(math.mod((8),(3)))
    print(math.mod(-(8),3))
    print(math.mod((8),-3))
    print(math.mod(-(8),-3))
    print(math.mod((-8),3))
    print(math.mod(-8,3))
    
f()

(lp) C:\Users\kunni\lpython>src\bin\lpython try.py
2
2
1
-1
-2
1
1
```

seifashraf1:
Hello, I would like to work on this issue. Where should I find the implementation of the % operation? 

I searched and found this function in `/lpython/src/runtime/lpython_builtin.py`

`@overload`
`def _mod(a: i32, b: i32) -> i32:`
    `return a - _lpython_floordiv(a, b)*b`

I am not sure though that I am looking at the right part.

rebcabin:
For mathematical background, please see this Wikipedia article https://en.wikipedia.org/wiki/Modulo?wprov=sfti1. The article clarifies the several ambiguities in the choice of outputs of mod when arguments are negative. It might explain some of the answers above.

seifashraf1:
Thank you! While this Wikipedia article will help me figure out the logic behind the right answers, I am still concerned about which parts of the codebase I should tackle to address the issue.

Smit-create:
Just replace this:
https://github.com/lcompilers/lpython/blob/cc02ba9fd21c3fa04c6b342400280c9412651429/src/runtime/lpython_builtin.py#L525-L539

with this:
https://github.com/lcompilers/lpython/blob/cc02ba9fd21c3fa04c6b342400280c9412651429/src/runtime/math.py#L469-L473

Please modify it according to the types.

[...]

PR #1566, opened 2023-03-07
Fix some edge cases for compile time operations of % operator

faze-geek:
Fixes #1562 
On master - 

```
def f():
    print((8)%3)
    print((8)%-3)         
    print((-8)%3)
    print((8.0)%3.0)
    print((8.0)%-3.0)         
    print((-8.0)%3.0)            
    
f()

(lp) C:\Users\kunni\lpython>python try.py
2
-1
1
2.0
-1.0
1.0

(lp) C:\Users\kunni\lpython>src\bin\lpython try.py
2
2
-2
2.00000000000000000e+00
2.00000000000000000e+00
-2.00000000000000000e+00
```

On branch - 
```
def f():
    print((8)%3)
    print((8)%-3)         
    print((-8)%3)
    print((8.0)%3.0)
    print((8.0)%-3.0)         
    print((-8.0)%3.0)            
    
f()

(lp) C:\Users\kunni\lpython>python try.py
2
-1
1
2.0
-1.0
1.0

(lp) C:\Users\kunni\lpython>src\bin\lpython try.py
2
-1
1
2.00000000000000000e+00
-1.00000000000000000e+00
1.00000000000000000e+00
```

PR #1567, opened 2023-03-07
Fix _mod function to fix the % operator results

seifashraf1:
Closing [1562](https://github.com/lcompilers/lpython/issues/1562)

**Task List:**

- Replaced the `_lpython_floordiv` function in all the `_mod` function with `_floor` function I created based on the `math.floor` function. 

On Branch: 
```
def f():
    print((8)%3)
    print((8)%(3))
    print(-(8)%3)       
    print((8)%-3)          
    print(-(8)%-3)
    print((-8)%3)          
    print(-8%3)            
    
f()

(base) seif@seif:~/Documents/GSoC/lpython$ ./src/bin/lpython examples/try.py
2
2
1
-1
-2
1
1

```

### DIFF
--- a/src/lpython/semantics/python_comptime_eval.h
+++ b/src/lpython/semantics/python_comptime_eval.h
@@ -211,7 +211,7 @@ struct PythonIntrinsicProcedures {
             int64_t a = ASR::down_cast<ASR::IntegerConstant_t>(arg1)->m_n;
             int64_t b = ASR::down_cast<ASR::IntegerConstant_t>(arg2)->m_n;
             return ASR::down_cast<ASR::expr_t>(
-                ASR::make_IntegerConstant_t(al, loc, a%b, type));
+                ASR::make_IntegerConstant_t(al, loc, ((a%b)+b)%b, type));
         } else if (ASRUtils::is_real(*type)) {
             double a = ASR::down_cast<ASR::RealConstant_t>(arg1)->m_r;
             double b = ASR::down_cast<ASR::RealConstant_t>(arg2)->m_r;
--- a/src/runtime/lpython_builtin.py
+++ b/src/runtime/lpython_builtin.py
@@ -470,19 +470,32 @@ def _lpython_floordiv(a: bool, b: bool) -> bool:
 
 @overload
 def _mod(a: i32, b: i32) -> i32:
-    return a - _lpython_floordiv(a, b)*b
+    """
+    Returns a%b
+    """
+    return a - i32(_floor(a/b))*b
 
 @overload
 def _mod(a: f32, b: f32) -> f32:
-    return a - _lpython_floordiv(a, b)*b
+    """
+    Returns a%b
+    """
+    return a - f32(_floor(a/b))*b
 
 @overload
 def _mod(a: i64, b: i64) -> i64:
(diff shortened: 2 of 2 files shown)


src/lpython/semantics/python_comptime_eval.h
```diff
@@ -212,7 +212,7 @@ struct PythonIntrinsicProcedures {
             int64_t a = ASR::down_cast<ASR::IntegerConstant_t>(arg1)->m_n;
             int64_t b = ASR::down_cast<ASR::IntegerConstant_t>(arg2)->m_n;
             return ASR::down_cast<ASR::expr_t>(
-                ASR::make_IntegerConstant_t(al, loc, a%b, type));
+                ASR::make_IntegerConstant_t(al, loc, ((a%b)+b)%b, type));
```
faze-geek, line 215: Hi, I have handled the compile time evaluations in my pr #1566 and got reviews on it. Could you please remove it from here to avoid clashes.
Also `_mod` function in `lpython_builtin.py` works on runtime evaluations, something like this -
```
def f():
    a:i32 = 8
    b:i32 = -3
    print(a%b)

f()
```
I hope you are aware of that, since all the test cases you refer in the OP are not in that format. Checkout if outputs of the `print(a%b)` format are actually getting corrected by the changes you made in `lpython_builtin.py` . On face value I can see that the below case is handled incorrectly - 
```
def f():
    a:i32 = 8
    b:i32 = 0
    print(a%b)

f()
  File "C:\Users\kunni\lpython\try.py", line 6, in <module>
    f()
  File "C:\Users\kunni\lpython\try.py", line 4, in f
    print(a%b)
ZeroDivisionError: integer division or modulo by zero

8
```

seifashraf1:
My bad I didn't see your PR, I removed it. 

However, thank you for letting me know about the format of the test case. The test case in the OP is the one used in [#1562](https://github.com/lcompilers/lpython/issues/1562) so I was just trying to test against it again.

 I did some edits to handle the case mentioned in your comment: 

```
def f():
    a:i32 = 8
    b:i32 = 0
    print(a%b)

f()
```
I made `_mod` function to raise ZeroDivisionError if the divisor is 0. Also, I got the `_lpython_floordiv` back as the `_floor` function made the tests to fail. 

```
@overload
def _mod(a: i32, b: i32) -> i32:
    if b == i32(0):
        raise ZeroDivisionError("Integer division or modulo by zero not possible")
    return a - _lpython_floordiv(a, b)*b
```

I got this: 

```
(lp) seif@seif:~/Documents/GSoC/lpython$ ./src/bin/lpython examples/try.py
ERROR STOP
```
I am not sure yet why it is behaving like this but I am looking into it.

faze-geek:
> I got this:
> 
> ```
> (lp) seif@seif:~/Documents/GSoC/lpython$ ./src/bin/lpython examples/try.py
> ERROR STOP
> ```
> 
> I am not sure yet why it is behaving like this but I am looking into it.

Yes, a bit strange. I got the same message on trying out your diff. @Thirumalai-Shaktivel I had pointed this out on zulip recently, Can you look into it !?

czgdp1807 (changes requested):
Add a test in `integration_tests` folder for this.

PR #1570, opened 2023-03-09
Fixes handling of while loop during pass

### LINKED ISSUE #1569: Incorrect result of print inside do loop
Reference - https://github.com/lfortran/lfortran/issues/1395 
```
def f():                                            # Equivalent python code of #1395 lfortran
    j :i32 = 0
    while True:
        j = j + 1
        print(j)
        if(j>4):
            break
    print("out of outer loop")
f()

1
2
3
4
5
out of outer loop

1
2
3
4
5
5
5
out of outer loop
```
I think https://github.com/lfortran/lfortran/pull/1401 is the correct direction for a similar fix here .

anutosh491:
Fixes #1569 
```
def f():                                            
    j :i32 = 0
    while True:
        j = j + 1
        print(j)
        if(j>4):
            break
    print("out of outer loop")
f()
```
On Branch
```
(lp) C:\Users\kunni\lpython>lpython try.py
1
2
3
4
5
out of outer loop
```
On Master
```
1
2
3
4
5
5
5
out of outer loop
```


integration_tests/loop_05.py
```diff
@@ -0,0 +1,10 @@
+def f():
+    j :i32 = 0
+    while True:
+        j = j + 1
+        print(j)
+        if(j>4):
```
czgdp1807, line 6:
```suggestion
        if j > 4:
```

integration_tests/loop_05.py
```diff
@@ -0,0 +1,10 @@
...
+            break
+    print("out of outer loop")
```
czgdp1807, line 8:
```suggestion
    print("out of outer loop")
    assert j == 5
```

czgdp1807 (changes requested):

integration_tests/CMakeLists.txt
```diff
@@ -258,6 +258,7 @@ RUN(NAME loop_01             LABELS cpython llvm c)
 RUN(NAME loop_02             LABELS cpython llvm c wasm wasm_x86 wasm_x64)
 RUN(NAME loop_03             LABELS cpython llvm c wasm wasm_x64)
 RUN(NAME loop_04             LABELS cpython llvm)
+RUN(NAME loop_05             LABELS cpython llvm)
```
czgdp1807, line 261: Works for `c` backend as well?

anutosh491: Yeah should work !

czgdp1807 (approved):

Issue #1575, opened 2023-03-13
C: Prospective Ideas


Smit-create:
The `C` backend currently supports almost all the integration tests and I think it's time to introduce some additional features to`C`. Some of the possible extensions and improvements to this backend can be:
- [ ] Support ASR optimization passes. I'd wait until we finish ASR Passes refactoring.
- [ ] Using the `--fast` flag to optimize the code by let's say using the `x.m_value` wherever possible and any other optimizations wherever possible.
- [ ] Directly run and output when we do `% lpython --backend=c a.py`, Presently we have:
```console
% lpython --backend=c a.py
The backend must be one of: llvm, cpp, x86, wasm, wasm_x86, wasm_x64.
```

Please free to edit/suggest any other ideas that you would like to have @certik @czgdp1807 @rebcabin.

PR #1580, opened 2023-03-15
Support `fast` option in C

Smit-create:
#1575

### DIFF
--- a/src/libasr/codegen/asr_to_c.cpp
+++ b/src/libasr/codegen/asr_to_c.cpp
@@ -17,6 +17,11 @@
 #include <map>
 #include <utility>
 
+#define CHECK_FAST_C(compiler_options, x)                         \
+        if (compiler_options.fast && x.m_value != nullptr) {    \
+            visit_expr(*x.m_value);                             \
+            return;                                             \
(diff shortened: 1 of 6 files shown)


src/libasr/codegen/asr_to_c.cpp
```diff
@@ -1203,6 +1227,10 @@ R"(
     }
 
     void visit_ArrayItem(const ASR::ArrayItem_t &x) {
+        if (compiler_options.fast && x.m_value != nullptr) {
+            visit_expr(*x.m_value);
+            return;
+        }
```
czgdp1807, line 1233: This is exactly the same block repeated in every node. So what about creating macro to generate this and placing the macro everywhere. 

In fact I would like to call it automatically so that one doesn't have to add this to every node. I don't know if C++ supports a way to always insert a code block in subclass method (other than macro one which one has to do by hand in every node). Anyways, let's leave this automation for future, right now just create a macro and use it everywhere.

czgdp1807 (changes requested):

certik (approved):

czgdp1807 (approved):

Issue #1539, opened 2023-02-21
WASM->x86: add support for global variables


certik:
WASM has a section that allows to define global variables that we then access from any function. We use it for internal WASM backend implementation of things like stack manipulation and memory allocator pointer.

certik:
Options how to implement this:

* Reserve registers. Not ideal, we need registers for computation
* Allocate on stack in the main program, then remember the location somehow and use it (challenge: the location changes in nested functions relative to the current runtime stack location)
* Use heap. Challenge: you get a pointer, where do you store it.

certik:
I think in the ELF format we have to use a `.data` section (that should be read+write). We have to figure out how to reference data from it.

Shaikh-Ubaid:
**doubt:** What should be the virtual address for the `data` segment? Or how do we compute the virtual address for the `data` segment? I looked up websites but could not any exact approach (which works) to compute the virtual address for the `data` segment.

certik:
Post the NASM example that creates a RW data section, that works. Then we can see how NASM creates the ELF file and figure out the algorithm.

certik:
The `text` section seems read-only, which makes sense for code. The `data` section seems read/write, so it should be possible to modify the string and everything should work. So that's what we need. Can you try to modify your NASM file to actually modify the string and print it again, ensuring that everything works as expected?

If so, then we just need to reproduce this with our backend.

Shaikh-Ubaid:
>Can you try to modify your NASM file to actually modify the string and print it again, ensuring that everything works as expected?

The following example works. We already checked it as marked in https://github.com/lcompilers/lpython/issues/1539#issuecomment-1438904011.

```bash
(wasm_asm) lpython$ cat main.asm
          global    _start

          section   .text
_start:
          ; print the string before modification
          mov       rax, 1                  ; system call for write
          mov       rdi, 1                  ; file handle 1 is stdout
          mov       rsi, message            ; address of string to output
          mov       rdx, 13                 ; number of bytes
          syscall                           ; invoke operating system to do the write

          ; modify the string
          mov rax, message
          mov byte [rax], 0b1000001              ; store A inplace of H

          ; print the string after modification
          mov       rax, 1                  ; system call for write
          mov       rdi, 1                  ; file handle 1 is stdout
          mov       rsi, message            ; address of string to output
          mov       rdx, 13                 ; number of bytes
          syscall                           ; invoke operating system to do the write

          ; exit the program with exit code 0
[...]

PR #1585, opened 2023-03-16
WASM_X64: Support Data Segment

Shaikh-Ubaid:
towards #1539

### DIFF
--- a/src/libasr/codegen/x86_assembler.h
+++ b/src/libasr/codegen/x86_assembler.h
@@ -470,7 +470,7 @@ public:
         for (size_t i = 0; i < padding_size; i++) {
             m_code.push_back(m_al, 0);
         }
-        EMIT("times " + std::to_string(padding_size) + " db 0");
+        EMIT("\n\talign " + std::to_string(alignment) + ", db 0");
     }
 
     void define_symbol(const std::string &name, uint32_t value) {
--- a/src/libasr/codegen/x86_assembler.h
+++ b/src/libasr/codegen/x86_assembler.h
@@ -40,7 +40,7 @@ Old Link: https://www.systutorials.com/go/intel-x86-64-reference-manual/
 #ifdef LFORTRAN_ASM_PRINT
 #    define EMIT(s) emit("    ", s)
 #    define EMIT_LABEL(s) emit("", s)
-#    define EMIT_VAR(a, b) emit("\n", a + " equ " + i2s(b) + "\n")
+#    define EMIT_VAR(a, b, c) emit("    ", a + " equ " + c + " - " + b)
 #    define EMIT_VAR_SIZE(a) emit("\n", a + " equ $ - $$\n") // $ is current addr, $$ is start addr
 #else
 #    define EMIT(s)
@@ -555,15 +555,17 @@ public:
         EMIT_VAR_SIZE(var);
     }
 
-    void add_var64(const std::string &var, uint64_t val) {
+    void add_var64(const std::string &var, const std::string &start, const std::string &end) {
         // TODO: Support 64-bit or 8 byte parameter val in define_symbol()
+        uint64_t val = get_defined_symbol(end).value - get_defined_symbol(start).value;
         define_symbol(var, val);
-        EMIT_VAR(var, val);
+        EMIT_VAR(var, start, end);
     }
 
-    void add_var(const std::string &var, uint32_t val) {
+    void add_var(const std::string &var, const std::string &start, const std::string &end) {
+        uint32_t val = get_defined_symbol(end).value - get_defined_symbol(start).value;
         define_symbol(var, val);
-        EMIT_VAR(var, val);
+        EMIT_VAR(var, start, end);
     }
 
     uint32_t pos() {
--- a/src/libasr/codegen/x86_assembler.h
+++ b/src/libasr/codegen/x86_assembler.h
@@ -30,6 +30,7 @@ Old Link: https://www.systutorials.com/go/intel-x86-64-reference-manual/
 #include <sstream>
 #include <fstream>
 #include <map>
+#include <cmath>
 
 #include <libasr/alloc.h>
 #include <libasr/containers.h>
@@ -463,6 +464,15 @@ public:
 
     Vec<uint8_t>& get_machine_code() {
         return m_code;
+    }
+
+    void align_by_byte(uint64_t alignment) {
+        uint64_t code_size = m_code.size() ;
+        uint64_t padding_size = (alignment * ceil(code_size / (double)alignment)) - code_size;
+        for (size_t i = 0; i < padding_size; i++) {
+            m_code.push_back(m_al, 0);
+        }
+        EMIT("times " + std::to_string(padding_size) + " db 0");
     }
 
     void define_symbol(const std::string &name, uint32_t value) {
--- a/src/libasr/codegen/x86_assembler.h
+++ b/src/libasr/codegen/x86_assembler.h
@@ -1542,7 +1542,7 @@ void emit_print_float(X86Assembler &a, const std::string &name);
 // Generate an ELF 64 bit header and footer
 // With these two functions, one only must generate a `_start` assembly
 // function to have a working binary on Linux.
-void emit_elf64_header(X86Assembler &a, uint32_t p_flags=5);
+void emit_elf64_header(X86Assembler &a);
 void emit_elf64_footer(X86Assembler &a);
 
 void emit_exit_64(X86Assembler &a, std::string label, int exit_code);
--- a/src/libasr/codegen/x86_assembler.h
+++ b/src/libasr/codegen/x86_assembler.h
@@ -42,12 +42,10 @@ Old Link: https://www.systutorials.com/go/intel-x86-64-reference-manual/
 #    define EMIT(s) emit("    ", s)
 #    define EMIT_LABEL(s) emit("", s)
 #    define EMIT_VAR(a, b, c) emit("    ", a + " equ " + c + " - " + b)
-#    define EMIT_VAR_SIZE(a) emit("\n", a + " equ $ - $$\n") // $ is current addr, $$ is start addr
 #else
 #    define EMIT(s)
 #    define EMIT_LABEL(s)
 #    define EMIT_VAR(a, b)
-#    define EMIT_VAR_SIZE(a)
 #endif
 
 namespace LCompilers {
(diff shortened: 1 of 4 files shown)


Shaikh-Ubaid:
**Example:**
```bash
(lp) lpython$ cat integration_tests/test_complex_01.py 
from ltypes import i32, i64, f32, f64, c32, c64

def test_real_imag():
    x: c64
    x = c64(2) + 3j
    a: f64
    b: f64
    eps: f64
    eps = 1e-12
    a = x.real
    b = x.imag
    assert abs(a - 2.0) <= eps
    assert abs(b - 3.0) <= eps

    print(x)

def test_complex():
    x: c64
    x = complex(4.5, 6.7)
    eps: f64
    eps = 1e-12
    assert abs(x.real - 4.5) <= eps
    assert abs(x.imag - 6.7) <= eps

    x = complex(-4, 2)
    assert abs(x.real - (-4.0)) <= eps
    assert abs(x.imag - 2.0) <= eps

    x = complex(4, 7.89)
    assert abs(x.real - 4.0) <= eps
    assert abs(x.imag - 7.89) <= eps

    x = complex(5.6, 0)
    assert abs(x.real - 5.6) <= eps
    assert abs(x.imag - 0.0) <= eps

    a: f64
    a = 534.6
    x = complex(a, -a) # (f64, f64)

    assert abs(x.real - 534.60000000000002274) <= eps
    assert abs(x.imag - (-534.60000000000002274)) <= eps

    a2: f32
    a2 = -f32(423.5430806348152437)
    a3: f32
    a3 = f32(34.5)
    x2: c32
    x2 = c32(complex(a2, a3)) # (f32, f32)

    assert f64(abs(x2.imag - f32(34.5))) <= eps

    i1: i32
    i1 = -5
    i2: i64
    i2 = -i64(6)

    x = complex(a3, a) # (f32, f64)
    x = complex(a, a3) # (f64, f32)
    x = complex(i1, i2) # (i32, i64)
    x = complex(i1, -i1) # (i32, i32)
    x = complex(-i2, -i2) # (i64, i64)
    x = complex(i2, -i1) # (i64, i32)

    print(x)

def test_complex_unary_minus():
    c: c32
    c = c32(complex(3, 4.5))
    _c: c32
    _c = -c
    assert abs(f64(_c.real) - (-3.0)) <= 1e-12
    assert abs(f64(_c.imag) - (-4.5)) <= 1e-12
    _c = c32(complex(5, -78))
    _c = -_c
    assert abs(f64(_c.real) - (-5.0)) <= 1e-12
    assert abs(f64(_c.imag) - 78.0) <= 1e-12
    c2: c64
    c2 = complex(-4.5, -7.8)
    c2 = -c2
    assert abs(c2.real - 4.5) <= 1e-12
    assert abs(c2.imag - 7.8) <= 1e-12
    c2 = c64(3) + 4j
    c2 = -c2
    assert abs(c2.real - (-3.0)) <= 1e-12
    assert abs(c2.imag - (-4.0)) <= 1e-12

    print(c, _c, c2)

def test_complex_not():
    c: c32
    c = c32(complex(4, 5))
    b: bool
    b = not c
    assert not b

    c2: c64
    c2 = complex(0, 0)
    b = not c2
    assert b

    print(c,c2, b)

def check():
    test_real_imag()
    test_complex()
    test_complex_unary_minus()
    test_complex_not()

check()
```
**ASM Binary Format:**
```
(lp) lpython$ lpython integration_tests/test_complex_01.py --backend wasm_x64 -o tmp > main.asm
(lp) lpython$ ./tmp
(2.000000000,3.000000000)
(-6.000000000,5.000000000)
(3.000000000,4.50000000) (-5.000000000,78.000000000) (-3.000000000,-4.000000000)
(4.000000000,5.000000000) (0.000000000,0.000000000) 1
(lp) lpython$ readelf -h tmp
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              EXEC (Executable file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x804e2ea
  Start of program headers:          64 (bytes into file)
  Start of section headers:          0 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         3
  Size of section headers:           0 (bytes)
  Number of section headers:         0
  Section header string table index: 0
(lp) lpython$ readelf -l tmp

Elf file type is EXEC (Executable file)
Entry point 0x804e2ea
There are 3 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000000000 0x0000000008048000 0x0000000008048000
                 0x0000000000001000 0x0000000000001000  R      0x1000
  LOAD           0x0000000000001000 0x0000000008049000 0x0000000008049000
                 0x0000000000006425 0x0000000000006425  R E    0x1000
  LOAD           0x0000000000008000 0x0000000008050000 0x0000000008050000
                 0x0000000000000100 0x0000000000000100  RW     0x1000
(lp) lpython$ 
```
**ASM Text Format:**
```bash
(wasm_asm) lpython$ nasm -fbin main.asm && chmod +x main
(wasm_asm) lpython$ ./main
(2.000000000,3.000000000)
(-6.000000000,5.000000000)
(3.000000000,4.50000000) (-5.000000000,78.000000000) (-3.000000000,-4.000000000)
(4.000000000,5.000000000) (0.000000000,0.000000000) 1
(wasm_asm) lpython$ readelf -h main
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              EXEC (Executable file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x804cba9
  Start of program headers:          64 (bytes into file)
  Start of section headers:          0 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         3
  Size of section headers:           0 (bytes)
  Number of section headers:         0
  Section header string table index: 0
(wasm_asm) lpython$ readelf -l main

Elf file type is EXEC (Executable file)
Entry point 0x804cba9
There are 3 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000000000 0x0000000008048000 0x0000000008048000
                 0x0000000000001000 0x0000000000001000  R      0x1000
  LOAD           0x0000000000001000 0x0000000008049000 0x0000000008049000
                 0x00000000000047c9 0x00000000000047c9  R E    0x1000
  LOAD           0x00000000000063a4 0x000000000804e3a4 0x000000000804e3a4
                 0x0000000000000100 0x0000000000000100  RW     0x1000
(wasm_asm) lpython$ 
```

src/libasr/codegen/x86_assembler.h
```diff
@@ -465,6 +464,15 @@ class X86Assembler {
         return m_code;
     }
 
+    void align_by_byte(uint64_t alignment) {
+        uint64_t code_size = m_code.size() ;
+        uint64_t padding_size = (alignment * ceil(code_size / (double)alignment)) - code_size;
+        for (size_t i = 0; i < padding_size; i++) {
+            m_code.push_back(m_al, 0);
+        }
+        EMIT("times " + std::to_string(padding_size) + " db 0");
```
Shaikh-Ubaid, line 473: **Line 473:** I think the `padding_size` computed here is valid only for the binary that we generate. For assembly text format, while encoding the `text_segment`, since `nasm` can encode instructions using different number of bytes than our encoding, the padding size required by `nasm` binary could be different from that of our binary. 

I am looking into the possibilities of computing the padding size dynamically in assembly text format.

Shaikh-Ubaid:

czgdp1807 (approved):

Shaikh-Ubaid:
>Should we add some tests for this new feature?

Sure, we can add tests. I think the test `integration_tests/test_complex_01.py` tests reading as well as writing data to our new data segment (It reads and writes to global variables which in assembly are located/allocated in the new data segment).

```bash
(lp) lpython$ lpython integration_tests/test_complex_01.py --backend wasm_x64 -o tmp > main.asm
(lp) lpython$ ./tmp
(2.000000000,3.000000000)
(-6.000000000,5.000000000)
(3.000000000,4.50000000) (-5.000000000,78.000000000) (-3.000000000,-4.000000000)
(4.000000000,5.000000000) (0.000000000,0.000000000) 1
(lp) lpython$ 
(wasm_asm) lpython$ nasm -fbin main.asm && chmod +x main
(wasm_asm) lpython$ ./main
(2.000000000,3.000000000)
(-6.000000000,5.000000000)
(3.000000000,4.50000000) (-5.000000000,78.000000000) (-3.000000000,-4.000000000)
(4.000000000,5.000000000) (0.000000000,0.000000000) 1
(wasm_asm) lpython$ 
```
(In the above example, I added some print statements to the test `integration_tests/test_complex_01.py` so that we have some output and can verify that the assembly binary and text generated actually works.)

Shaikh-Ubaid:
As this PR seems approved, I am merging  this. I will work on further changes/suggestions (if any) in another PR.

certik:
I see, perfect!

I don't see any global variables in https://github.com/lcompilers/lpython/blob/1dc74c6200cac7c983f3ae15d3dce25f37c55f3f/integration_tests/test_complex_01.py, which line is a global variable?

What capability does this new RW data segment allows us to do? One is that the code segment becomes read-only, which is good. The other might be that now we should be able to resize it, thus allow arbitrary length heap allocation. Anything else?

Shaikh-Ubaid:
https://github.com/lcompilers/lpython/blob/1dc74c6200cac7c983f3ae15d3dce25f37c55f3f/src/libasr/codegen/asr_to_wasm.cpp#L74-L79

We currently use global variable for temporary operations like stack manipulation (for example: to swap the top and top-second stack elements). (The `cur_mem_loc` global variable is not used yet I think.)

For example, the global get/set operations are used here
https://github.com/lcompilers/lpython/blob/1dc74c6200cac7c983f3ae15d3dce25f37c55f3f/src/libasr/codegen/asr_to_wasm.cpp#L1699-L1710

and here

https://github.com/lcompilers/lpython/blob/1dc74c6200cac7c983f3ae15d3dce25f37c55f3f/src/libasr/codegen/asr_to_wasm.cpp#L2615-L2621

The `integration_tests/test_complex_01.py` utilizes the above code lines. Thus, it tests getting and setting globals.

Shaikh-Ubaid:
>The other might be that now we should be able to resize it, thus allow arbitrary length heap allocation

I think it can allow us compile-time memory allocation (For example: we can have a label and can emit `0` bytes for as much space we require. This way I think we have a zero initialised memory of the required size. But this could increase the file size.)

Using the above approach, I think we can now support arrays by initializing memory for them in the data segment. For dynamic/runtime memory allocation (which could be needed for allocatable arrays), I think we should try to see/use system calls for memory allocation.

certik:
I see, perfect, thanks for the explanation!

PR #1589, opened 2023-03-17
Fix edge cases for `int()` casting of strings 

faze-geek:
All of these cases below throw the `invalid literal` semantic error on master as of now.

On branch
```
def foo():        
    print(int("  3   "))
    print(int("+3"))
    print(int("\n3"))
    print(int("3\n"))
    print(int("\r\t\n3\r\t\n"))


foo()

3
3
3
3
3

3
3
3
3
3
```

### DIFF
--- a/integration_tests/test_str_to_int.py
+++ b/integration_tests/test_str_to_int.py
@@ -20,4 +20,10 @@ def f():
     i = i32(int(s))
     assert i == -1234
 
+    assert int("  3   ") == 3
+    assert int("+3") == 3
+    assert int("\n3") == 3
+    assert int("3\n") == 3
+    assert int("\r\t\n3\r\t\n") == 3
+
 f()
--- a/src/lpython/semantics/python_intrinsic_eval.h
+++ b/src/lpython/semantics/python_intrinsic_eval.h
@@ -77,20 +77,34 @@ struct IntrinsicNodeHandler {
                 char *c = ASR::down_cast<ASR::StringConstant_t>(
                                     ASRUtils::expr_value(arg))->m_s;
                 int ival = 0;
+                std::string str;
                 char *ch = c;
-                if (*ch == '-') {
+                if (*ch == '-' || *ch == '+') {
                     ch++;
                 }
                 while (*ch) {
+                    if(*ch == ' '){
+                        ch++;
+                        continue;
+                    }
+                    if(*ch == '\\'){
+                        ch++;
+                        if (*ch == 'n' || *ch == 'r' || *ch == 't') {
+                            ch++;
+                            continue;
+                        }
+                        throw SemanticError("invalid literal for int() with base 10: '"+ std::string(c) + "'", arg->base.loc);
+                    }
                     if (*ch == '.') {
                         throw SemanticError("invalid literal for int() with base 10: '"+ std::string(c) + "'", arg->base.loc);
                     }
                     if (*ch < '0' || *ch > '9') {
                         throw SemanticError("invalid literal for int() with base 10: '"+ std::string(c) + "'", arg->base.loc);
                     }
+                    str+=*ch;
                     ch++;
                 }
-                ival = std::stoi(c);
+                ival = std::stoi(str);
                 return (ASR::asr_t *)ASR::down_cast<ASR::expr_t>(ASR::make_IntegerConstant_t(al,
                                 loc, ival, to_type));
             }


src/lpython/semantics/python_intrinsic_eval.h
```diff
@@ -77,20 +77,34 @@ struct IntrinsicNodeHandler {
...
                         throw SemanticError("invalid literal for int() with base 10: '"+ std::string(c) + "'", arg->base.loc);
                     }
                     if (*ch < '0' || *ch > '9') {
                         throw SemanticError("invalid literal for int() with base 10: '"+ std::string(c) + "'", arg->base.loc);
                     }
+                    str+=*ch;
                     ch++;
                 }
-                ival = std::stoi(c);
+                ival = std::stoi(str);
```
faze-geek, line 107: Edge case to be looked into :
```
def f():
    print(int(' '))    # Only whitespace without any numeric character.
f()

  File "C:\Users\kunni\lpython\try.py", line 3, in <module>
    f()
  File "C:\Users\kunni\lpython\try.py", line 2, in f
    print(int(' '))
ValueError: invalid literal for int() with base 10: ' '

std::exception: invalid stoi argument
```

Been a bit busy with university exams, will handle this soon.

faze-geek:
@czgdp1807 @Thirumalai-Shaktivel Kindly consider this pull request ready for review. The CI logs fail as of now, but I'll handle it at once after having collective reviews.

src/lpython/semantics/python_intrinsic_eval.h
Thirumalai-Shaktivel: Does this case throw an error?

Thirumalai-Shaktivel:
LGTM!

src/lpython/semantics/python_intrinsic_eval.h
faze-geek: Yes this case throws an error - The `stoi` catches invalid input and throws an error. I'll have to add a if condition and throws a corresponding semantic error. Thats it right?

Thirumalai-Shaktivel: Yes, I think, we have to check and throw a SemanticError.

faze-geek: Other than this edge case I feel all cases are handled !

faze-geek:
> LGTM!

Have a look at the CI logs once please. I'm not sure if I had updated the references after the latest commit.

integration_tests/test_str_to_int.py
```diff
@@ -20,4 +20,10 @@ def f():
     i = i32(int(s))
     assert i == -1234
 
+    assert i32(int("  3   ")) == 3
+    assert i32(int("+3")) == 3
+    assert i32(int("\n3")) == 3
+    assert i32(int("3\n")) == 3
+    assert i32(int("\r\t\n3\r\t\n")) == 3
```
Thirumalai-Shaktivel, line 27: There is an `AssertionError` among these tests.
Check all the cases if it gives the correct output.

Shaikh-Ubaid: We have the error at the following place.
```bash
(lp) lpython$ git diff                                    
diff --git a/integration_tests/test_str_to_int.py b/integration_tests/test_str_to_int.py
index 765dd1e0f..b67dcab32 100644
--- a/integration_tests/test_str_to_int.py
+++ b/integration_tests/test_str_to_int.py
@@ -3,21 +3,27 @@ from ltypes import i32
 def f():
     i: i32
     i = i32(int("314"))
+    print(i)
     assert i == 314
     i = i32(int("-314"))
     i = i32(int("-314"))
+    print(i)
     assert i == -314
     s: str
     s = "123"
     i = i32(int(s))
+    print(i)
     assert i == 123
     s = "-123"
     i = i32(int(s))
+    print(i)
     assert i == -123
     s = "    1234"
     i = i32(int(s))
+    print(i)
     assert i == 1234
     s = "    -1234   "
     i = i32(int(s))
+    print(i)
     assert i == -1234
 
     assert i32(int("  3   ")) == 3
@@ -25,5 +31,6 @@ def f():
     assert i32(int("\n3")) == 3
     assert i32(int("3\n")) == 3
     assert i32(int("\r\t\n3\r\t\n")) == 3
+    print("Done")
 
 f()
(lp) lpython$ lpython integration_tests/test_str_to_int.py
314
314
AssertionError
```
It seems currently `+` and `-` signs are being ignored.

src/lpython/semantics/python_intrinsic_eval.h
```diff
@@ -77,20 +77,34 @@ struct IntrinsicNodeHandler {
                 char *c = ASR::down_cast<ASR::StringConstant_t>(
                                     ASRUtils::expr_value(arg))->m_s;
                 int ival = 0;
+                std::string str;
                 char *ch = c;
-                if (*ch == '-') {
+                if (*ch == '-' || *ch == '+') {
                     ch++;
                 }
```
Shaikh-Ubaid, line 84: We have the error at the following place.
```bash
(lp) lpython$ git diff                                    
diff --git a/integration_tests/test_str_to_int.py b/integration_tests/test_str_to_int.py
index 765dd1e0f..b67dcab32 100644
--- a/integration_tests/test_str_to_int.py
+++ b/integration_tests/test_str_to_int.py
@@ -3,21 +3,27 @@ from ltypes import i32
 def f():
     i: i32
     i = i32(int("314"))
+    print(i)
     assert i == 314
     i = i32(int("-314"))
     i = i32(int("-314"))
+    print(i)
     assert i == -314
     s: str
     s = "123"
     i = i32(int(s))
+    print(i)
     assert i == 123
     s = "-123"
     i = i32(int(s))
+    print(i)
     assert i == -123
     s = "    1234"
     i = i32(int(s))
+    print(i)
     assert i == 1234
     s = "    -1234   "
     i = i32(int(s))
+    print(i)
     assert i == -1234
 
     assert i32(int("  3   ")) == 3
@@ -25,5 +31,6 @@ def f():
     assert i32(int("\n3")) == 3
     assert i32(int("3\n")) == 3
     assert i32(int("\r\t\n3\r\t\n")) == 3
+    print("Done")
 
 f()
(lp) lpython$ lpython integration_tests/test_str_to_int.py
314
314
AssertionError
```
It seems currently `+` and `-` signs are being ignored.

Shaikh-Ubaid: Also, since we check for `+`/`-` sign for only at the start, I think the following could fail (it fails currently):
```bash
(lp) lpython$ python3 examples/expr2.py
-3
3
(lp) lpython$ lpython examples/expr2.py
semantic error: invalid literal for int() with base 10: '    -3    '
 --> examples/expr2.py:4:19
  |
4 |     print(i32(int("    -3    ")))
  |                   ^^^^^^^^^^^^ 


```

src/lpython/semantics/python_intrinsic_eval.h
```diff
@@ -77,20 +77,34 @@ struct IntrinsicNodeHandler {
...
+                            continue;
+                        }
+                        throw SemanticError("invalid literal for int() with base 10: '"+ std::string(c) + "'", arg->base.loc);
+                    }
                     if (*ch == '.') {
                         throw SemanticError("invalid literal for int() with base 10: '"+ std::string(c) + "'", arg->base.loc);
                     }
                     if (*ch < '0' || *ch > '9') {
                         throw SemanticError("invalid literal for int() with base 10: '"+ std::string(c) + "'", arg->base.loc);
                     }
```
Shaikh-Ubaid, line 103: I think the above could be simplified as follows:
```suggestion
                    if(*ch == ' ' || *ch == '\n' || *ch == '\r' || *ch == '\t'){
                        ch++;
                        continue;
                    }
                    if (*ch < '0' || *ch > '9') {
                        throw SemanticError("invalid literal for int() with base 10: '"+ std::string(c) + "'", arg->base.loc);
                    }
```

src/lpython/semantics/python_intrinsic_eval.h
```diff
@@ -77,20 +77,34 @@ struct IntrinsicNodeHandler {
...
+                    str+=*ch;
```
Shaikh-Ubaid, line 104: (Minor: I guess two spaces here would look good).
```suggestion
                    str += *ch;
```

src/lpython/semantics/python_intrinsic_eval.h
Shaikh-Ubaid: I think `ch` holds the whole escaped character. So, we need not check for `\n` (or others) in two steps. The above approach performs the check in one step.

Example:
```cpp
#include <iostream>
#include <cstring>
using namespace std;

int main()
{
    char *mystr = "abc def \n\r\t pqr xyz";
    char *ch = mystr;
    while (*ch) {
        if (*ch == ' ') {
            // do not print anything
        } else if (*ch == '\n') {
            std::cout << "newline character" << std::endl;
        } else if (*ch == '\r') {
            std::cout << "carriage return" << std::endl;
        } else if (*ch == '\t') {
            std::cout << "tab character" << std::endl;
        } else {
            std::cout << *ch << std::endl;
        }
        ch++;
    }
    return 0;
}
```
Output:
```bash
a
b
c
d
e
f
newline character
carriage return
tab character
p
q
r
x
y
z
```

PR #1591, opened 2023-03-20
Initial implementation of Symbolic type

Thirumalai-Shaktivel:
cc @anutosh491

### DIFF
--- a/examples/expr2.py
+++ b/examples/expr2.py
@@ -1,10 +1,6 @@
-def main0():
-    x: i32
-    x = (2+3)*5
+def main():
+    x: symbolic = Symbolic('x')
     print(x)
 
-main0()
+main()
 
-# Not implemented yet in LPython:
-#if __name__ == "__main__":
-#    main()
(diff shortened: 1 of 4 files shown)


examples/expr2.py
```diff
@@ -1,10 +1,6 @@
-def main0():
-    x: i32
-    x = (2+3)*5
+def main():
+    x: symbolic = Symbolic('x')
     print(x)
 
-main0()
+main()
```
Thirumalai-Shaktivel, line 5: Currently, this doesn't work in CPython. I think we handle this ltypes and make sure this works with CPython

examples/expr2.py
```diff
@@ -1,10 +1,6 @@
-def main0():
-    x: i32
-    x = (2+3)*5
+def main():
+    x: symbolic = Symbolic('x')
```
certik, line 2:
```suggestion
    x: Symbolic = Symbol('x')
```

PR #1592, opened 2023-03-20
ASR Pass: Fix variables names to be used in C backend

Smit-create:
#1575


certik (approved):
I think this is fine for now, although a more general approach might be needed regarding what is allowed as a symbol name in ASR, and then "escape" it properly in the backends.

PR #1598, opened 2023-03-21
Allow Backend::c in the main driver

certik:


czgdp1807 (approved):
I feel this should be done in a registry styled approach i.e., use a map to store the strings ("c", "llvm", etc to the appropriate backend and raise error if the backend name string is not present in the map). Just like what we do for passes and intrinsic functions.

certik:
Yes, it can be refactored, but we can do that later.

gptsarthak:
@czgdp1807  Do we not need to actually implement the C backend like we have done with C++,wasm and others?
Something like this:
```
 else if (backend == Backend::c) {
        std::string CC = "gcc";
        std::string options, post_options;
        if (static_executable) {
            options += " -static ";
        }
        // if (compiler_options.openmp) {      // Not sure if we can use openmp with C
            // options += " -fopenmp ";
        // }
        std::string cmd = CC + options + " -o " + outfile + " ";
        for (auto &s : infiles) {
            cmd += s + " ";
        }
        cmd += + " -L";
        cmd += " " + post_options + " -lm";
        int err = system(cmd.c_str());
        if (err) {
            std::cout << "The command '" + cmd + "' failed." << std::endl;
            return 10;
        }
        return 0;
} else if (backend == Backend::x86) {
```

Smit-create:
Yep, we need to do that. I'm working on it and will open a PR to fix this.

Issue #1608, opened 2023-03-23
Supporting Callbacks in LPython


Smit-create:
For providing functionalities like: https://github.com/lcompilers/lpython/issues/992#issuecomment-1478215958 we need to support callbacks. A simple example is:
```python
from ltypes import i32, Callable

def f(x: i32) -> i32:
    return x + 1

def g(func: Callable[[i32], i32], arg: i32):
    ret: i32
    ret = func(arg)
    print(ret)

g(f, 10)
```

Currently, this throws the following error:
<details>
<summary> Error </summary>
<p>

```console
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/thebigbool/repos/lpython/src/bin/lpython.cpp", line 1602
    return emit_asr(arg_file, lpython_pass_manager, runtime_library_dir,
  File "/Users/thebigbool/repos/lpython/src/bin/lpython.cpp", line 194
    r = LCompilers::LPython::python_ast_to_asr(al, lm, *ast, diagnostics,
  File "/Users/thebigbool/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 6237
    if (res.ok) {
  File "/Users/thebigbool/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 3814
    v.visit_Module(ast);
  File "/Users/thebigbool/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 3302
    visit_stmt(*x.m_body[i]);
  File "/Users/thebigbool/repos/lpython/src/lpython/python_ast.h", line 1882
    void visit_stmt(const stmt_t &b) { visit_stmt_t(b, self()); }
  File "/Users/thebigbool/repos/lpython/src/lpython/python_ast.h", line 1750
    case stmtType::FunctionDef: { v.visit_FunctionDef((const FunctionDef_t &)x); return; }
  File "/Users/thebigbool/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 3389
    ASR::ttype_t *arg_type = ast_expr_to_asr_type(x.base.base.loc, *x.m_args.m_args[i].m_annotation);
  File "/Users/thebigbool/repos/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 1540
    this->visit_expr(*s->m_slice);
  File "/Users/thebigbool/repos/lpython/src/lpython/python_ast.h", line 1909
[...]

PR #1609, opened 2023-03-23
Add test for a small newton raphson package

Smit-create:
This tests only on an already-defined function. We will modify this test to accept callbacks once we complete working on #1608 

Related to #992

### DIFF
--- a/integration_tests/nrp/nr.py
+++ b/integration_tests/nrp/nr.py
@@ -1,0 +1,20 @@
+from ltypes import i32, f64
+
+
+def func(x: f64, c: f64) -> f64:
+    return x**2.0 - c**2.0
+
+
+def func_prime(x: f64) -> f64:
+    return 2.0*x
+
+
+def newton_raphson(x: f64, c: f64, maxiter: i32) -> f64:
+    h: f64
+    err: f64 = 1e-5
+    i: i32 = 0
+    while abs(func(x, c)) > err and i < maxiter:
+        h = func(x, c) /  func_prime(x)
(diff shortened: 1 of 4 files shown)


integration_tests/nrp/nr.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def newton_raphson(x: f64, c: f64, maxiter: i32) -> f64:
+	h: f64 = func(x, c) / func_prime(x, c)
+	err: f64 = 1e-5
+	i: i32 = 0
+	while abs(func(x, c)) > err and i < maxiter:
+		h = func(x, c) / func_prime(x, c)
+		x = x - h
+		i += 1
```
certik, line 19: Convert tabs to spaces.

certik:
Beautiful, that's exactly what we need. Thanks!

integration_tests/nrp/nr.py
Smit-create: Ah, some wierd issue. I have used a tab for indenting and locally looks fine.

ADV1K: We should probably set some pre-commit hooks to autoformat the code when commiting. we can use python black as the formatter.

certik: We should use LPython as the formatter. :)

certik (approved):

integration_tests/nrp/nr.py
Thirumalai-Shaktivel: Yup!

PR #1615, opened 2023-03-24
Implementation of `str.partition()` with tuples 

harshsingh-24:
- `Functionality`: Split the string at the first occurrence of sep, and return a 3-tuple containing the part before the separator, the separator itself, and the part after the separator. If the separator is not found, return a 3-tuple containing the string itself, followed by two empty strings.
- Added sufficient test cases for integration testing along with proper comments for understanding


src/runtime/lpython_builtin.py
```diff
@@ -775,6 +775,45 @@ def _lpython_str_startswith(s: str ,sub: str) -> bool:
         res = res and (j == len(sub))
     return res
 
+@overload
+def _lpython_str_endswith(s: str, suffix: str) -> bool:
```
Thirumalai-Shaktivel, line 779: These changes are already in the `main`, right?

If there is some problem,
- `Rebase` this branch on top of the updated main branch.
- Then do `git reset main` and commit the changes.

harshsingh-24:
Ignore this PR.

PR #1617, opened 2023-03-24
Import Global statements

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
TODO
- [ ] Enclose all the global statements into a function: `<module_name>@global_statements`
- [ ] Call the above function using SubroutineCall
- [ ] Global lists initialization has to be handled
- [ ] check multiple `import a` statements
- [ ] Pass all the tests

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2239,13 +2244,17 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
             ASR::expr_t* v_expr = ASRUtils::EXPR(ASR::make_Var_t(al, loc, v_sym));
             cast_helper(v_expr, init_expr, true);
             ASR::asr_t* assign = ASR::make_Assignment_t(al, loc, v_expr,
                                                         init_expr, nullptr);
-            current_body->push_back(al, ASRUtils::STMT(assign));
+            if (current_body) {
+                current_body->push_back(al, ASRUtils::STMT(assign));
+            } else if (ASR::is_a<ASR::List_t>(*type)) {
+                global_init.push_back(al, assign);
+            }
```
Thirumalai-Shaktivel, line 2257: I assumed `current_body` contains `true` value only when `create_add_variable_to_scope` (this function) is visited inside the `Function` or `Program` nodes. In the module or global scope, `current_body` would be `nullptr`.
So, I have added a check for `global_lists`.
Have I handled this correctly? @czgdp1807

Thirumalai-Shaktivel:
This PR is ready for review.

src/lpython/semantics/python_ast_to_asr.cpp
czgdp1807: Make sense to me.

czgdp1807:
Please resolve the conflicts.

Thirumalai-Shaktivel:
Done!

integration_tests/global_syms_03_a.py
```diff
@@ -4,4 +4,8 @@
 
 def populate_lists() -> list[i32]:
     return [10, -20]
-l_2: list[i32] = populate_lists()
+l_2: list[i32]
+l_2 = populate_lists()
```
czgdp1807, line 8: Why this change?

Thirumalai-Shaktivel: I wanted to add an assignment statement for the initializer. Instead of adding a new test, I just used the existing one 😅.

src/libasr/asr_scopes.cpp
```diff
@@ -206,47 +203,6 @@ void SymbolTable::move_symbols_from_global_scope(Allocator &al,
                 es->m_parent_symtab = module_scope;
                 ASR::symbol_t *s = ASRUtils::symbol_get_past_external(a.second);
                 LCOMPILERS_ASSERT(s);
-                if (ASR::is_a<ASR::Variable_t>(*s)) {
```
Thirumalai-Shaktivel, line 209: All the initializers of list types are moved into the global_initializer, so I removed all the changes which do the same thing.

integration_tests/global_syms_03_a.py
czgdp1807: I would recommend to have both tests, initialiser and assignment statement for the initialiser.

Thirumalai-Shaktivel: Got it! I will update the tests.

certik (approved):
I think this looks good to me. Thanks @Thirumalai-Shaktivel !

PR #1620, opened 2023-03-24
Rename ltypes.py to lpython.py

### LINKED ISSUE #1602: Rename ltypes.py to lpython.py
We should have just one simple Python API, and I think it should go via `lpython`. So:
```python
import lpython

@lpython.jit
def f(..)
    ...
```
and
```python
from lpython import i32, f64, Pointer, ...
```
Similar to `numpy.int32`.

ADV1K:
Fixes #1602 

I have changed all the imports and related instances of ltypes to lpython.  
However I do have some doubts regarding the changes inside `python_ast_to_asr.cpp` as changing the variable names from ltypes to lpython might be confusing. Please do let me know if i should revert that change.

I have also tested that importing lpython works and updated the test references.


certik (approved):
Thanks for fixing this!

I think that this is fine, I didn't see any issues, but I would like @czgdp1807, @Smit-create and @Thirumalai-Shaktivel to also review, to ensure we didn't miss anything.

Smit-create (approved):
I think it looks good, just a rename so shouldn't have any issues.

czgdp1807 (approved):

Thirumalai-Shaktivel (approved):

certik:
Ok, let's merge it then. If anything breaks, let's fix it as we go.

Thanks @ADV1K for fixing this!

PR #1622, opened 2023-03-25
Fix undeclared variable error in array_01.py

Manali0806:
Add static type to variable 'foo' in array_01.py

### DIFF
--- a/integration_tests/array_01.py
+++ b/integration_tests/array_01.py
@@ -4,7 +4,7 @@ from numpy import empty
 def main0():
     Nx: i32 = 600; Ny: i32 = 450
     arr: i32[450, 600] = empty([Ny, Nx])
-    i: i32
+    i: i32 = 42
     j: i32
     for i in range(Ny):
         for j in range(Nx):


integration_tests/array_01.py
```diff
@@ -4,7 +4,7 @@
 def main0():
     Nx: i32 = 600; Ny: i32 = 450
     arr: i32[450, 600] = empty([Ny, Nx])
-    i: i32
+    i: i32 = 42
```
Smit-create, line 7: Any reason for initializing the variable? I don't think we need this.

Manali0806: Initializing variables is important in general because it can prevent issues related to uninitialized variables, which can lead to unexpected behavior in the program. If a variable is not initialized, its value can be unpredictable and may result in errors or bugs in the program.

czgdp1807 (commented):
Can you show what the error actually is? I don't see any error in `main` branch due to this test.

integration_tests/array_01.py
ditsuke, line 7:
```suggestion
    i: i32 = 69
```

Might as well make it a justifiable

ditsuke, line 7:
```suggestion
    i: i32 = 69
```

Might as well make it justifiable

certik:
That variable is a loop variable, so you don't want to initialize it like this.

LPython should check for uninitialized variables, but that's a different issue.

PR #1623, opened 2023-03-25
Added Sets

AarnavNagariya:
I have added sets detection in this file.


src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -731,6 +731,8 @@ R"(#include <stdio.h>
         bool is_value_tup = ASR::is_a<ASR::Tuple_t>(*m_value_type);
         bool is_target_dict = ASR::is_a<ASR::Dict_t>(*m_target_type);
         bool is_value_dict = ASR::is_a<ASR::Dict_t>(*m_value_type);
+        bool is_targe_set = ASR::is_a<ASR::Set_t>(*m_target_type);
```
certik, line 734:
```suggestion
        bool is_target_set = ASR::is_a<ASR::Set_t>(*m_target_type);
```

certik:
Thanks!

This doesn't seem to compile. Test should be added (ideally enabling one that we already have), and CI needs to pass.

Issue #1626, opened 2023-03-25
C backend: variable is initialized twice


certik:
Example:
```python
from ltypes import f64

def f() -> f64:
    return 5.5

def main():
    t1: f64 = f()*1e6
    print(t1)

main()
```
This generates:
```console
$ lpython --show-c a.py
...
void _xx_lcompilers_changed_main_xx()
{
    double t1 = f()*  1.00000000000000000e+06;
    t1 = f()*  1.00000000000000000e+06;
    printf("%lf\n", t1);
}
...
```
Which is wrong, it should only initialize it once.

Shaikh-Ubaid:
I have also experienced this issue with the `wat` generation.

For example, the same example at https://github.com/lcompilers/lpython/issues/1626#issue-1640614146 produces the following `wat` which has duplicated assignment:
```wat
...
(func $4 (type 4) (param) (result)
        (local f64)
        call 3
        f64.const 1000000.000000
        f64.mul
        local.set 0 // first assignment
        call 3
        f64.const 1000000.000000
        f64.mul
        local.set 0 // second assignment
...
```

I think the issue is at the `ASR` level. The following example
```python
from ltypes import i32

def main():
    t1: i32 = 5

main()
```
generates the `ASR`:
```bash
main:
...
(Function
    (SymbolTable
        2
        {
            t1:
                (Variable
                    2
                    t1
                    []
                    Local
                    (IntegerConstant 5 (Integer 4 [])) // symbolic value for the variable
                    ()
                    Default
                    (Integer 4 [])
                    Source
                    Public
                    Required
                    .false.
                )
        })
    main
    (FunctionType
        []
        ()
        Source
        Implementation
        ()
        .false.
        .false.
        .false.
        .false.
        .false.
        []
        []
        .false.
    )
    []
    []
    [(=                                            // assignment for the variable
        (Var 2 t1)
        (IntegerConstant 5 (Integer 4 []))
        ()
    )]
    ()
    Public
    .false.
    .false.
)
...
```
In the above `asr`, we see that the variable `t1` has a `symbolic` value of `IntegerConstant 5` and in the function body, the same variable is also being assigned the value of `IntegerConstant 5`. This leads to duplicate assignment of values to variables initialised during declaration.

I think ideally the `asr` should either contain the symbolic value and no initialisation statements in function body, or it should contain no symbolic value and initialisation statements in function body.

certik:
[...]

PR #1629, opened 2023-03-26
Fix double initialisation of variables in ASR

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1626


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -2210,24 +2237,30 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
                 current_procedure_abi_type, s_access, s_presence,
                 value_attr);
         ASR::symbol_t* v_sym = ASR::down_cast<ASR::symbol_t>(v);
+        ASR::Variable_t* v_variable = ASR::down_cast<ASR::Variable_t>(v_sym);
 
-        if( init_expr && current_body) {
+        if( init_expr && current_body &&
+            (is_runtime_expression || !is_variable_const)) {
```
czgdp1807, line 2243: Only those variables will be initialised by converting to assignments which are non-constant or have a runtime expressions. Also an error is thrown if a constant variable is not initialised at declaration (see lines 2215 to 2219).

src/libasr/asr_verify.cpp
```diff
@@ -516,6 +516,15 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
             "Variable::m_parent_symtab must be present in the ASR ("
                 + std::string(x.m_name) + ")");
 
+        if( symtab->parent == nullptr ) {
+            // For nowRestrict this check only to variables which are present
```
certik, line 520:
```suggestion
            // For now restrict this check only to variables which are present
```

certik (approved):
I think that this is good, thanks for fixing it and adding a test!

I suggest we merge it, and if there are more issues, we'll tackle it as we discover it.

czgdp1807:
Okay. Let me check if verify changes work with LFortran. Then I will merge both the PRs.

certik (approved):

PR #1630, opened 2023-03-26
Corrected typos and grammar mistake

sayan-36:
Corrected typos and grammar mistake  on line 135 & 342.

### DIFF
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -132,7 +132,7 @@ presence = Required | Optional
 
 -- abi=LFortranModule/GFortranModule/BindC: the symbol's implementation is
 -- stored as machine code in some object file that must be linked in. It
--- uses the specified ABI (one of LFortran module, GFortran module or C ABI).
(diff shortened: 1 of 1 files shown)


src/libasr/ASR.asdl
```diff
@@ -339,7 +339,7 @@ expr
 --     and Fortran: in Python "Booleans in Python are implemented as a subclass
 --     of integers", in Fortran the "default logical kind has the same storage
 --     size as the default integer"; we currently use kind=4 as default
---     integer, so we also use kind=4 for the default logical.)
+--     integer, so we also use kind=4 for the default logic.)
```
czgdp1807, line 342:
```suggestion
--     integer, so we also use kind=4 for the default logical.)
```

Here logical refers to logical type and not a logic.

czgdp1807 (changes requested):

src/libasr/ASR.asdl
rebcabin: why not "... for the default logical type" ?

PR #1634, opened 2023-03-27
New functions added

### LINKED ISSUE #1633: Functions missing from cmath module
`polar`, `isfinite`, `isinf` functions are missing from cmath library

suhanigarg29:
fixes #1633

### DIFF
--- a/src/runtime/cmath.py
+++ b/src/runtime/cmath.py
@@ -250,3 +250,51 @@ def tanh(x: c64) -> c64:
 @overload
 def tanh(x: c32) -> c32:
     return _lfortran_ctanh(x)
+
+@ccall
+def _lfortran_zpolar(x: c64) -> c64:
+    pass
+
+@ccall
+def _lfortran_cpolar(x: c32) -> c32:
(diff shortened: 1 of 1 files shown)


src/runtime/cmath.py
```diff
@@ -250,3 +250,51 @@ def tanh(x: c64) -> c64:
...
+def _lfortran_cpolar(x: c32) -> c32:
+    pass
+
+@overload
+def polar(x: c64) -> c64:
+    return _lfortran_zpolar(x)
+
+@overload
+def polar(x: c32) -> c32:
+    return _lfortran_cpolar(x)
```
Smit-create, line 268: These won't work as we don't have these functions in `lfortran_intrinsic.h`.

suhanigarg29: Where can I find the functions included in lfortran_intrinsic.h?

certik: In `src/libasr/runtime/lfortran_intrinsic.h/cpp`.

PR #1636, opened 2023-03-28
C: Support passes

Smit-create:


tests/reference/c-expr_01-28f449f.stdout
```diff
@@ -36,30 +36,34 @@ struct dimension_descriptor
...
-    return _lpython_return_variable;
-}
-
-inline __attribute__((always_inline)) int32_t and_op(int32_t x, int32_t y)
-{
-    int32_t _lpython_return_variable;
-    _lpython_return_variable = x&y;
-    return _lpython_return_variable;
-}
-
```
czgdp1807, line 52: This is not good. Please skip inline function calls pass in C. `inline __attribute__((always_inline))` directs the C compiler to do the same and since C language is a high level language so the generated code should be as readable as possible.

czgdp1807 (changes requested):

czgdp1807 (approved):
LGTM.

certik:
Looks good.

PR #1642, opened 2023-03-29
Catch unsupported binary operators on some data structures

faze-geek:
On main - 

```
def d():
    s:str = "a"
    print(s*"3")
d()

  File "C:\Users\kunni\lpython\try.py", line 7, in <module>
    d()
  File "C:\Users\kunni\lpython\try.py", line 3, in d
    print(s*"3")
TypeError: can't multiply sequence by non-int of type 'str'

 --> try.py:3:11
  |
3 |     print(s*"3")
  |           ^ ^^^ type mismatch (str and str)
```

This has been corrected for data structures like strings and list on this branch for a couple of operations.


faze-geek:
If there are other unsupported operators, then I would like to know and add them here. Thanks!

faze-geek:
Here is a minimal use case for this pull request -
```
def d():
    s:str = "a"
    print(s*"3")

d()

semantic error: Unsupported operand type
 --> try.py:3:11
  |
3 |     print(s*"3")
  |           ^ ^^^ for ('str' and 'str')

```

Not sure why the logs fail as of now.

faze-geek:
@Thirumalai-Shaktivel Can you have a look at this. Some CL logs fail.

Issue #941, opened 2022-08-09
Implement other List methods


Thirumalai-Shaktivel:
Currently we support: append(), insert(), remove() and pop().
TODO:
* [ ] extend()
* [ ] clear()
* [ ] index()
* [ ] count()
* [ ] reverse()
* [ ] copy()
* [ ] sort()

Source: https://docs.python.org/3/tutorial/datastructures.html#more-on-lists

certik:
Yes, we should implement all these. Given the requirement for list to work for all types, it might be easier to just implement in the backend directly, until we have working generics.

certik:
Other operations that we can think about adding later, taken from [std::vector](https://en.cppreference.com/w/cpp/container/vector) (once we have a C++ frontend, we'll need these):
* reserve()
* capacity()
* resize()
Also in order for our ASR `list` to be usable as an `std::vector` implementation, we probably need some way to construct a list from a pointer and a size (we have to figure out who owns what), so that an existing C++ function that accepts or returns an `std::vector`, say an external function, can be interfaced with ASR without any need for a copy. Although this might be possible to solve with arrays, we need the same mechanism for arrays as well, so that ASR can just use arrays, but one can pass in `std::vector` as an argument (with some cast), that does not copy data. Anyway, that's for later.

namannimmo10:
> Given the requirement for a list to work for all types, it might be easier to just implement in the backend directly, until we have working generics

Do you mean possibly adding an ASR node for each method?

certik:
> Do you mean possibly adding an ASR node for each method?

That path is clear and we might start with that.

I am hoping to keep the ASR minimal, but we can I think figure it out later how to implement some of this functionality another way.

namannimmo10:
Hmm... It's still not clear to me what we _really_ want.
Quoting from what you wrote in the "Intrinsic Functions Design" doc:

> Next step from a "minimal design":
> 
> * Add to ASR all nodes that are shared across all frontends. That includes all nodes that must be implemented by the backend, but it also includes nodes that could have been implemented in the frontend, such as `StringRepeat`.
> * This "broader scope" can be restated as: each frontend only has to maintain its "runtime library" of functions that are specific to that frontend. If it is shared with another frontend, it can go into ASR itself as a node. In another words: we do not need to maintain a set of ASR functionality/library shared across languages. Everything is either as ASR nodes, or in the frontend (and not shared with other frontends).
[...]

PR #1644, opened 2023-03-29
Attempt to Impelent list.count() method without generics

gptsarthak:
I read https://github.com/lcompilers/lpython/issues/941#issuecomment-1211092508 but soon realized that generics do not seem to be working yet, so this is my attempt to implement list.count() method without generics

I want to know if I am going in the correct direction, and whether this is the correct way of implementing this method. This PR is not ready to merge as it only supports ```list[i32]``` yet.

I have not added tests yet, but here is one performed on my branch:

```
def test_list():
    arr_i32 :list[i32]
    arr_i32 = [6,12,6,18,6]
    res_i32 :i32 = arr_i32.count(6)
    print(res_i32)
    
test_list()
```
```
(base) sarthak@pop-os:~/lpython$ python examples/list.py
3
(base) sarthak@pop-os:~/lpython$ lpython examples/list.py
3
```


rsaba5612:
It looks like you're attempting to implement the count() method for lists in LPython without using generics, which is a good idea given that generics aren't working yet. Your approach seems to be correct, since the count() method simply loops over the list and counts the number of times the given element appears.

However, it's important to note that your implementation currently only supports lists of i32 type. In order to fully implement the count() method for any type of list, you'll need to add support for all the built-in types, such as i64, f32, f64, bool, str, etc.

Also, as you mentioned, tests are missing from your implementation. It's important to add thorough tests for all edge cases and potential errors to ensure that your implementation is correct and robust.

src/runtime/lpython_builtin.py
```diff
@@ -743,6 +743,17 @@ def _lpython_str_find(s: str, sub: str) -> i32:
...
+@overload
+def _lpython_list_count(l: list[i32], x: i32) -> i32:
+    count : i32
+    count = 0
+    i : i32
+    i = 0
+    for i in l:
+        if(i == x):
+            count = count + 1
+    return count
```
czgdp1807, line 755: Well three approaches,

1. Use generics - Very messy and not worth it at the moment.
2. Use IntrinsicFunction infrastructure to programmatically generate the code for these functions and call it in `intrinsic_functions.cpp` pass.
3. Insert a new node, `ListCount(expr list, expr x)` in ASR.asdl and implement it directly in LLVM backend.

Approach 3 is most doable right now and will generate the most efficient LLVM code. However to scale well across all backends, approach 2 is the best one but should be done later. There is no disadvantage in going for approach 3. Once approach 2 is implemented we can remove `ListCount` node easily. @certik What do you say?

certik: Yes, 3 is fine, later we'll move to 2.

czgdp1807:
Closing in favour of https://github.com/lcompilers/lpython/pull/1676

PR #1639, opened 2023-03-28
Add test package lpdraw

Shaikh-Ubaid:
This PR adds a test package `lpdraw`.

```bash
python integration_tests/test_pkg_lpdraw.py > cpy_tmp.pgm
```
cpy_tmp

```bash
lpython integration_tests/test_pkg_lpdraw.py --backend wasm > wasm_tmp.pgm
```
wasm_tmp

(Note: Images converted to `jpeg` before uploading here. It seems GitHub does not support uploading `.pgm` image files)

PR #1647, opened 2023-03-30
Add test package lpdraw

Shaikh-Ubaid:
details: https://github.com/lcompilers/lpython/pull/1639

### DIFF
--- a/integration_tests/test_pkg_lpdraw.py
+++ b/integration_tests/test_pkg_lpdraw.py
@@ -1,11 +1,18 @@
 from lpdraw import Line, Circle, Clear, Display, DisplayTerminal
 
+from lpython import i32
+from numpy import empty, int32
+
 def main():
-    Clear()
-    Line(2, 4, 99, 11)
-    Line(0, 39, 49, 0)
-    Circle(52, 20, 6.0)
-    Display()
-    DisplayTerminal()
+    Width: i32 = 100 # x-axis limits [0, 99]
+    Height: i32 = 40 # y-axis limits [0, 39]
+    Screen: i32[40, 100] = empty((40, 100), dtype=int32)
+
+    Clear(Height, Width, Screen)
+    Line(Height, Width, Screen, 2, 4, 99, 11)
+    Line(Height, Width, Screen, 0, 39, 49, 0)
+    Circle(Height, Width, Screen, 52, 20, 6.0)
+    Display(Height, Width, Screen)
+    DisplayTerminal(Height, Width, Screen)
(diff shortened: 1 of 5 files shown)


Shaikh-Ubaid:

integration_tests/test_pkg_lpdraw.py
```diff
@@ -0,0 +1,34 @@
...
+W = TypeVar("W")
+
+def test_screen(H: i32, W: i32, Screen: i32[H, W]):
+    i: i32
+    j: i32
+    cnt: i32 = 0
+    for i in range(H):
+        for j in range(W):
+            if Screen[i, j] == 255:
+                cnt += 1
```
czgdp1807, line 16:
```suggestion
        cnt += (Screen[i, j] - 256):
```

integration_tests/test_pkg_lpdraw.py
```diff
@@ -0,0 +1,34 @@
...
+
+    assert cnt == 175
```
czgdp1807, line 18:
```suggestion
    assert cnt == -175
```

czgdp1807 (changes requested):

integration_tests/test_pkg_lpdraw.py
Shaikh-Ubaid: If we do `cnt += (Screen[i, j] - 256):`, I guess the count would be some negative number. Since, for each pixel (not switched on), we add `-256`.

czgdp1807: Ah! Yes. What I am trying to suggest is to just sum up all the values without any filter. And then make sure that the sum is as expected.

czgdp1807: Right now, you are ignoring all values `Screen[i, j] != 255` which will not reflect any error for such values in `cnt`. So better to just sum up all the values without any preference.

Shaikh-Ubaid: Got it. Done. Thank you so much for the guidance!

integration_tests/test_pkg_lpdraw.py
Shaikh-Ubaid: Updated. Thank you.

czgdp1807 (approved):

Shaikh-Ubaid:
I am adding this to auto-merge as other the related PR https://github.com/lcompilers/lpython/pull/1639 seems approved. This PR just improves the test strength.

PR #1652, opened 2023-03-31
Add types and decorators to imports in the integration tests

khairalanam:
This PR adds missing imports to many of the integration tests.

I have also updated the reference tests.


certik:
Thanks! I think tests have to be updated.

khairalanam:
> Thanks! I think tests have to be updated.

I have updated the reference tests. Did I miss something?

PR #1657, opened 2023-04-01
WASMAssembler: Initial support

Shaikh-Ubaid:
towards https://github.com/lfortran/lfortran/issues/589

It seems our existing `wasm` assembly functions and the `ASRToWASMVisitor` Class are tightly coupled together. In this PR, I am hoping to decouple them and have a separate `WASMAssembler` Class responsible for `wasm` generation.

### DIFF
--- a/src/libasr/codegen/asr_to_wasm.cpp
+++ b/src/libasr/codegen/asr_to_wasm.cpp
@@ -98,33 +98,13 @@ std::string import_fn_to_str(IMPORT_FUNC fn) {
 class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
    public:
     Allocator &m_al;
+    WASMAssembler m_wa;
     diag::Diagnostics &diag;
 
     SymbolFuncInfo *cur_sym_info;
-    uint32_t nesting_level;
-    uint32_t cur_loop_nesting_level;
     bool is_prototype_only;
     bool is_local_vars_only;
     ASR::Function_t* main_func;
-
-    Vec<uint8_t> m_type_section;
-    Vec<uint8_t> m_import_section;
-    Vec<uint8_t> m_func_section;
-    Vec<uint8_t> m_memory_section;
-    Vec<uint8_t> m_global_section;
-    Vec<uint8_t> m_export_section;
-    Vec<uint8_t> m_code_section;
-    Vec<uint8_t> m_data_section;
-
-    uint32_t no_of_types;
-    uint32_t no_of_functions;
-    uint32_t no_of_memories;
-    uint32_t no_of_globals;
-    uint32_t no_of_exports;
-    uint32_t no_of_imports;
-    uint32_t no_of_data_segments;
-    uint32_t avail_mem_loc;
-    uint32_t digits_mem_loc;
 
     uint32_t min_no_pages;
     uint32_t max_no_pages;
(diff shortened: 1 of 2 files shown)


Shaikh-Ubaid:
I would like to request if someone could review (at a high level) the current state of `WASMAssembler` and `ASRToWASMVisitor` in this PR and share if we are heading in proper direction. I have some concerns that I am sharing below.

src/libasr/codegen/wasm_assembler.h
```diff
@@ -51,1198 +59,1387 @@ void emit_leb128_i64(Vec<uint8_t> &code, Allocator &al, int64_t n) {  // for i64
...
+    // no_of_types indicates total (imported + defined) no of functions
+    uint32_t no_of_types;
+    uint32_t no_of_functions;
+    uint32_t no_of_memories;
+    uint32_t no_of_globals;
+    uint32_t no_of_exports;
+    uint32_t no_of_imports;
+    uint32_t no_of_data_segments;
+    uint32_t avail_mem_loc;
+    uint32_t digits_mem_loc;
```
Shaikh-Ubaid, line 184: I would wish the above variables to be private and not accessible by outer entities.

src/libasr/codegen/asr_to_wasm.cpp
```diff
@@ -1116,42 +962,42 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
...
             m_var_idx_map[get_hash((ASR::asr_t *)arg)] = s->no_of_variables;
-            emit_var_type(m_type_section, arg, s->no_of_variables, false);
+            emit_var_type(m_wa.m_type_section, arg, s->no_of_variables, false);
             if (isRefVar(arg)) {
                 s->referenced_vars.push_back(m_al, arg);
             }
         }
-        wasm::fixup_len(m_type_section, m_al,
+        m_wa.fixup_len(m_wa.m_type_section, m_al,
                             len_idx_type_section_param_types_list);
```
Shaikh-Ubaid, line 981: The variables noted in https://github.com/lcompilers/lpython/pull/1657#discussion_r1155150521 are needed at several places in the `ASRToWASMVisitor` Class. Thus, I am unable to make them private.

It seems that without those variables being private, we may/might not have truly decoupled `WASMAssembler` from `ASRToWASMVisitor` Class.

certik:
It looks like the assembler itself only writes to one section in the WASM binary, correct?

In that case, why not make the WASMAssembler just to return the binary instructions, and have another class that handles the WASM format? That way at least the WASMAssembler will be very simple.

Overall this direction is good I think.

Shaikh-Ubaid:
>It looks like the assembler itself only writes to one section in the WASM binary, correct?

We have several sections in the wasm binary. After completion of wasm generation, we combine these section into one single array and return it. Currently, as per the main branch, `wasm assembler` is more like a collection of several functions that help us generate wasm. These functions accept any `wasm` section (`code`, `type`, `global`, etc) and write bytes into it.

>In that case, why not make the WASMAssembler just to return the binary instructions, and have another class that handles the WASM format? That way at least the WASMAssembler will be very simple.

One class that handles instructions (like `emit_i32_add()`, `emit_f32_div()`, etc) only. And another class that handles the `WASM` format. Sure, that can be done. For the second class which handles the format, it seems that we might come across the same issue shared here https://github.com/lcompilers/lpython/pull/1657#discussion_r1155151073.

certik:
It seems to me the assembler class should only construct the "code" section. It seems we only use the assembler to construct bytes in the global and type sections, but we should rather use the WasmFormat class for that I think, it would be more readable and robust. Please correct me if I missed something.

Shaikh-Ubaid:
>It seems to me the assembler class should only construct the "code" section.

I implemented a `WASMInstsAssembler` Class here https://github.com/lcompilers/lpython/pull/1666. Please share if it is in the direction we are thinking.

src/libasr/codegen/asr_to_wasm.cpp
Shaikh-Ubaid: This is fixed in https://github.com/lcompilers/lpython/pull/1675.

Shaikh-Ubaid:
Closing in favour of https://github.com/lcompilers/lpython/pull/1675.

PR #1663, opened 2023-04-03
Add cmath polar functions

virendrakabra14:
Adds `phase`, `polar`, and `rect` functions to `cmath` module.


virendrakabra14:
I tried using (LPython) `math` functions in `cmath`, but the import failed with this message
```
Module __main__ dependencies must contain math because a function present in it is getting called in __main__
```

I wasn't sure on where to add this dependency, so I created `rect` function in `lfortran_intrinsics` files.

virendrakabra14:
The error was due to an inconsistent return type.

Running these commands locally did not give any error:
```
python run_tests.py -u
./run_tests.py
cd ./integration_tests
./run_tests.py
```

However the tests here (GitHub workflow) pointed that.
Can these tests be run locally as well?

czgdp1807:
> I tried using (LPython) `math` functions in `cmath`, but the import failed with this message
> 
> ```
> Module __main__ dependencies must contain math because a function present in it is getting called in __main__
> ```
> 
> I wasn't sure on where to add this dependency, so I created `rect` function in `lfortran_intrinsics` files.

Can you share the exact LPython command you followed?

czgdp1807:
@virendrakabra14 I pushed some changes, let's see if they work with Windows CI or not.

virendrakabra14:
> > I tried using (LPython) `math` functions in `cmath`, but the import failed with this message
> > ```
> > Module __main__ dependencies must contain math because a function present in it is getting called in __main__
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I wasn't sure on where to add this dependency, so I created `rect` function in `lfortran_intrinsics` files.
> 
> Can you share the exact LPython command you executed?

I was using `import math` in `cmath.py` and using sin, cos functions. Test file `test_cmath.py` was same as the current one. Then
```
lpython ./integration_tests/test_cmath.py
```
gave that error - which I think is due to importing `math` in `cmath`.

czgdp1807:
Also, eventually we might implement all these intrinsics using `IntrinsicFunction` infrastructure and provide direct ASR implementations for their functions. So if calling into C can be avoided then it should be. Functions like, `_lfortran_dsin`, `_lfortran_dcos` can also be done via ASR directly (by using some approximation algorithms to produce outputs for `sin` and `cos` calls).

czgdp1807 (approved):

certik (approved):

PR #1673, opened 2023-04-06
Register `Partition` in IntrinsicFunction

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
We have to discuss, how this has to be handled in the Pass.
Currently, I thought to do: `str.partition(x)` -> `partition(str, x)`

Should I implement the following code into the ASR?
https://github.com/lcompilers/lpython/blob/b35e744dce5e435d04624bfbed157c9cb50befb2/src/runtime/lpython_builtin.py#L810-L825
`_lpython_str_find` function seems to be very big function!!

czgdp1807:
> Should I implement the following code into the ASR?

Yes, definitely. It doesn't matter to me how big the function is. Initially it will be painful to program in ASR directly but then we will observe patterns and create abstract APIs to reduce the boilerplate. Like while implementing `any` intrinsic in LFortran, I noticed the need for an API to create `if-else-if` ladder in ASR directly. So there will be more abstractions figured out later.

Thirumalai-Shaktivel:
Got it! I understood and agree on this, we will create boilerplate for everything (like loops, if-else, ...).

Thirumalai-Shaktivel:
Almost ready! Everything works as expected!
I will try to cleanup repeatative code in `intrinsic_functionc_registry.h`.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -511,6 +812,207 @@ static inline ASR::asr_t* create_ListIndex(Allocator& al, const Location& loc,
...
+            body.push_back(al, STMT(ASR::make_If_t(al, loc, a_test,
+                if_body.p, if_body.n, else_body.p, else_body.n)));
+            body.push_back(al, STMT(ASR::make_Return_t(al, loc)));
+        }
+        ASR::symbol_t *fn_sym = make_Function_t(fn_name, fn_symtab, dep, args,
+            body, return_var, Source, Implementation, nullptr);
+        scope->add_symbol(fn_name, fn_sym);
+        return EXPR(ASR::make_FunctionCall_t(al, loc, fn_sym, fn_sym,
+            new_args.p, new_args.n, return_type, compile_time_value, nullptr));
+    }
```
certik, line 1014: This function is not "generic", correct? The signature is always the same.

It takes about 100 lines of ASR construction to do what can be done in about 11 lines of Python:

https://github.com/lcompilers/lpython/blob/aa1f1952676d66608b75b18c1c89c66978c87d88/src/runtime/lpython_builtin.py#L827-L838

I suggest we merge this, but why don't we try to figure out how to make these 100 lines become 11 lines?

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -175,6 +181,301 @@ static inline ASR::asr_t* create_UnaryFunction(Allocator& al, const Location& lo
...
+        }
+        body.push_back(al, STMT(ASR::make_Return_t(al, loc)));
+
+    }
+    ASR::symbol_t *fn_sym = make_Function_t(fn_name, fn_symtab, dep, args,
+        body, return_var, Source, Implementation, nullptr);
+    global_scope->add_symbol(fn_name, fn_sym);
+    return fn_sym;
+}
+
```
certik, line 478: Similarly here, it takes about 282 lines of C++ to do what Python can do in 44 lines:

https://github.com/lcompilers/lpython/blob/aa1f1952676d66608b75b18c1c89c66978c87d88/src/runtime/lpython_builtin.py#L713-L757

src/libasr/pass/intrinsic_function_registry.h
Shaikh-Ubaid: I have similar thoughts. The idea of moving functions from runtime library (`pure` surface language implementation) to the intrinsic function registry (C++ ASR implementation) is not very clear to me yet. What benefits does the transition bring us? (Or what is the motivation behind the transition?) Are we trying to maintain a unique copy of intrinsic functions for every frontend language? I would be appreciate it and be grateful if someone could clarify this.

certik: The main motivation for this is for generic functions like minloc() or any() that require very special loop structure based on the array rank and must work for any array type, so doing it here is less code than overloading 100+ functions in the front end language.

The secondary motivation is to have these implementations self-contained in libasr, and shared with lpython/lfortran (and all future frontends), given that they take less code than in the front end language.

The above case unfortunately takes more code than in the frontend language. So I see two ways forward:

* Either keep these in the frontend langauge (much easier to maintain, but will have to be duplicated for all frontends)
* Or figure out how to make the ASR as easy to write as the frontend language. I would say maximum 2x penalty in terms of lines of code, so 11 lines of Python could be maximum of 22 lines of C++, but ideally around 11 lines also. Either way 100 lines is too much.

I would like to see how to figure out the second option, as it would fix everything. To make progress on that, we need some kind of ASR Builder that has the most common defaults that effectively will allow to write code as if it was Python or Fortran.

CC @czgdp1807 and @rebcabin. This is a blocker. We don't need to figure it out this week, but we do need a solution say within a month or so.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -49,11 +49,17 @@ enum class IntrinsicFunctions : int64_t {
...
     // ...
 };
 
 namespace UnaryIntrinsicFunction {
 
+#define create_var_expr(var_expr, name, intent, abi, value_attr, symtab, type)\
+    ASR::expr_t *var_expr;                                                      \
+    { create_variable(sym, name, intent, abi, value_attr, symtab, type)         \
+    var_expr = EXPR(ASR::make_Var_t(al, loc, sym)); }
+
```
czgdp1807, line 62: Actually we don't need this now. Please update `create_variable` macro and its usage in LPython.

https://github.com/lfortran/lfortran/blob/6eeb126cde6dcc4a1e73371e0fa3c7a010324b7f/src/libasr/pass/intrinsic_function_registry.h#L120-L126

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -175,6 +181,301 @@ static inline ASR::asr_t* create_UnaryFunction(Allocator& al, const Location& lo
...
+
+        body.push_back(al, STMT(ASR::make_Assignment_t(al, loc, s_len, EXPR(
+            ASR::make_StringLen_t(al, loc, args[0], int_type, nullptr)), nullptr)));
+        body.push_back(al, STMT(ASR::make_Assignment_t(al, loc, pat_len, EXPR(
+            ASR::make_StringLen_t(al, loc, args[1], int_type, nullptr)), nullptr)));
+        body.push_back(al, STMT(ASR::make_Assignment_t(al, loc, result,
+            EXPR(ASR::make_IntegerUnaryMinus_t(al, loc,
+            EXPR(ASR::make_IntegerConstant_t(al, loc, 1, int_type)), int_type,
+            EXPR(ASR::make_IntegerConstant_t(al, loc, -1, int_type)))), nullptr)));
```
czgdp1807, line 245: You can write an API to create constants like the way I am doing in, https://github.com/lfortran/lfortran/pull/1577

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -175,6 +181,301 @@ static inline ASR::asr_t* create_UnaryFunction(Allocator& al, const Location& lo
...
+                            int_type, nullptr)), nullptr)));
+                    else_body_2.push_back(al, STMT(ASR::make_Assignment_t(al, loc,
+                        i, EXPR(ASR::make_IntegerBinOp_t(al, loc, i,
+                            ASR::binopType::Add,
+                            EXPR(ASR::make_IntegerConstant_t(al, loc, 1, int_type)),
+                            int_type, nullptr)), nullptr)));
+                    else_body_1.push_back(al, STMT(ASR::make_If_t(al, loc, a_test,
+                        if_body_2.p, if_body_2.n, else_body_2.p, else_body_2.n)));
+                }
+                loop_body.push_back(al, STMT(ASR::make_If_t(al, loc, a_test,
```
czgdp1807, line 346: This loop creation can be abstracted into an API just like (just overload `create_do_loop` as you like) 

https://github.com/lfortran/lfortran/blob/6eeb126cde6dcc4a1e73371e0fa3c7a010324b7f/src/libasr/pass/intrinsic_function_registry.h#L55-L116

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -175,6 +181,301 @@ static inline ASR::asr_t* create_UnaryFunction(Allocator& al, const Location& lo
...
+            ASR::make_IntegerConstant_t(al, loc, 0, int_type)), nullptr)));
+        body.push_back(al, STMT(ASR::make_Assignment_t(al, loc, i, EXPR(
+            ASR::make_IntegerConstant_t(al, loc, 0, int_type)), nullptr)));
+        {
+            ASR::expr_t *a_test = EXPR(ASR::make_LogicalBinOp_t(al, loc,
+                EXPR(ASR::make_IntegerCompare_t(al, loc,
+                    EXPR(ASR::make_IntegerBinOp_t(al, loc, s_len,
+                        ASR::binopType::Sub, i, int_type, nullptr)),
+                    ASR::cmpopType::GtE,
+                    EXPR(ASR::make_IntegerBinOp_t(al, loc, pat_len,
```
czgdp1807, line 363: This can be simplified as well. See `create_ElementalAdd` and `create_ElementalBinOp` in https://github.com/lfortran/lfortran/pull/1577.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -175,6 +181,301 @@ static inline ASR::asr_t* create_UnaryFunction(Allocator& al, const Location& lo
...
+                        ASR::binopType::Sub, j, int_type, nullptr)),
+                    logical_type, nullptr)),
+                ASR::logicalbinopType::And,
+                EXPR(ASR::make_LogicalNot_t(al, loc, flag, logical_type, nullptr)),
```
czgdp1807, line 367: Same here. Should be as simple as `not(al, loc, flag)`.

czgdp1807 (changes requested):
I see lot of patterns in this PR which can be abstracted out into simpler APIs. The expression can be as simple as `not(al, loc, input)` or `add(al, loc, left, right)`, etc. So the code size will reduce by huge margins. I think we should try it in this PR before merging.

src/libasr/pass/intrinsic_function_registry.h
czgdp1807: > * Either keep these in the frontend langauge (much easier to maintain, but will have to be duplicated for all frontends)

I think for common functions like `exp`, `sum`, `abs`, etc I won't write code for it in the frontend. It requires duplicating efforts in all the compilers and also increased code to be tested since each frontend will have code in its own syntax so higher chances of errors. By using intrinsic functions we implement and test only once so its only a one time pain which can be reduced with time. See my earlier comment https://github.com/lcompilers/lpython/pull/1673#issuecomment-1499936663 as well.

I can already see a lot of patterns here as pointed out in my latest review of this PR.

certik: I see no reason why we could not program in ASR as easily as in Python. Once we do, then the second approach is preferable, since it would be as easy to maintain, as in Python.

Here is the meat of the function:
```python
 res : tuple[str, str, str] 
 ind : i32 
 ind = _lpython_str_find(s, sep) 
 if ind == -1: 
     res = (s, "", "") 
 else:  
     res = (s[0:ind], sep, s[ind+len(sep): len(s)])     
 return res 
```
So how could this look like in C++ building ASR directly? Here is one idea:

```c++
auto res = Variable("res", Tuple({Str, Str, Str}))
auto ind = Variable("ind", i32)
body = {
    If(Eq(ind, i32(-1)), {
        Assignment(res, Tuple({s, "", ""});
    }, {
        Assignment(res, Tuple({StringSection(s, {0, ind}), sep, StringSection(s, {Add(ind, StringLen(sep)), StringLen(s))}));
    }),
    Assignment(result, res);
}
// Do something with "body"
```

The guiding principle: Let it look like the Clojure style S-expression of `--show-asr`, such as `(Assignment result res)`, but use the syntax of M-expressions: `Assignment(result, res)`. Let's use C++ variables like `res` and `ind` to store intermediate results. Let `Assignment` and other "heads" have the right defaults, so that one can use it like I did above. Under the hood, it probably needs to keep track of a few things (allocator, location info, the current active symbol table scope, etc.). Let's assume these things can be figured out. Wouldn't the above be very maintainable and readable?

Thirumalai-Shaktivel: Yup, I had the same idea, I started implement one by one. 
`ASR::make_IntegerConstant` -> `i32(1)`
`ASR::make_Integer_t` -> i32 or int32

certik: Create a new namespace, like Builder, and just put all these things in.

czgdp1807: Yes. We just need to remove ASR's verbosity. I would like Builder to be a class because I don't want to keep passing `al, loc` to all the creator functions. Also, the ideas in [this comment](https://github.com/lcompilers/lpython/pull/1673#discussion_r1172122992) can be the base for builder API.

certik: Yes, let's make Builder a class and make the instance very simple, perhaps `b.Assignment(...)`, to make it very easy to type and visual. Yes, `al` should be hidden. The location should actually be the location in the C++ code I think. This would have to be done with a macro, to automatically infer the line number and filename. So maybe we need `Assignment` macro, not sure if it should be `ASSIGNMENT`, or `XAssignment`. Something short, but making sure it does not clash.

czgdp1807 (commented):
I have added an ASR builder class in https://github.com/lfortran/lfortran/pull/1577. Please feel free to use it here as well. We will sync later.

Thirumalai-Shaktivel:
Prefect, thank you! I will start working on it.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -814,51 +1179,150 @@ static inline ASR::expr_t* instantiate_Any(Allocator &al, const Location &loc,
...
+                        StringSection(args[0], iAdd(index, StringLen(args[1])),
+                            StringLen(args[0]))}, return_type))
+                }));
+            body.push_back(al, STMT(ASR::make_Return_t(al, loc)));
+        }
+        ASR::symbol_t *fn_sym = make_Function_t(fn_name, fn_symtab, dep, args,
+            body, result, Source, Implementation, nullptr);
+        scope->add_symbol(fn_name, fn_sym);
+        return b.Call(fn_sym, new_args, return_type, compile_time_value, loc);
+    }
```
Thirumalai-Shaktivel, line 1324: I think this is the max we can simplify here.
What do you think about this?

@certik @czgdp1807

Thirumalai-Shaktivel: I used this as a reference: https://github.com/lcompilers/lpython/pull/1673#discussion_r1172122992

src/libasr/pass/intrinsic_function_registry.h
czgdp1807, line 1: I think there are several `BinOp` nodes which can be made less verbose to use. Like we can create `ElementalSub`, `ElementalEq`, etc and then use them instead of verbose ASR APIs. Also to create variables I would like to use, `b.declare("variable_name", variable_type)`.

Probably better to have `ASRBuilder` in a separate file and then use it in all the passes we have (should be done after this PR is merged).

czgdp1807 (changes requested):

src/libasr/pass/intrinsic_function_registry.h
certik: This is good. We can do better with changing `body` to just use `{}`, but let's do that later. For now this looks good.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -248,6 +475,101 @@ static inline ASR::asr_t* create_UnaryFunction(Allocator& al, const Location& lo
...
+            b.If(loc, sEq(StringItem(args[1], iAdd(i, i32(1))),
+                    StringItem(args[1], iAdd(pi_len, i32(1)))), {
+                Assign(pi_len, iAdd(pi_len, i32(1))),
+                Assign(ListItem(lps, i, int32), pi_len),
+                Assign(i, iAdd(i, i32(1)))
+            }, {
+                b.If(loc, iNotEq(pi_len, i32(0)), {
+                    Assign(pi_len, ListItem(lps, iSub(pi_len, i32(1)),
+                        int32))
+                }, {
```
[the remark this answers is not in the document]
Thirumalai-Shaktivel: It wasn't required, as `lps` is already filled with `0`;

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -814,51 +1038,147 @@ static inline ASR::expr_t* instantiate_Any(Allocator &al, const Location &loc,
...
+                Assignment(result, b.TupleConstant({
+                    StringSection(args[0], i32(0), index), args[1],
+                    StringSection(args[0], iAdd(index, StringLen(args[1])),
+                        StringLen(args[0]))}, return_type))
+            }));
+        body.push_back(al, Return());
+        ASR::symbol_t *fn_sym = make_Function_t(fn_name, fn_symtab, dep, args,
+            body, result, Source, Implementation, nullptr);
+        scope->add_symbol(fn_name, fn_sym);
+        return b.Call(fn_sym, new_args, return_type, compile_time_value);
```
Thirumalai-Shaktivel, line 1179: @certik, during our last discussion, you suggested renaming `Assign` to `Assignment`.
Should I do the same for `Call` as well? `Call` -> `FunctionCall`?

Thirumalai-Shaktivel:
I think I addressed all the comments!
Kindly review this PR and provide feedback.

src/libasr/pass/intrinsic_function_registry.h
certik: I would. This can be done later too.

src/libasr/pass/intrinsic_function_registry.h
certik: Beautiful, it is now about 20 lines, but with different formatting it would be around 14 or so. This is very good, thanks!

certik (approved):
I think this is good to merge as is.

@czgdp1807 can you please also review it?

Thirumalai-Shaktivel:
@harshsingh-24, this PR is ready.
After @czgdp1807 reviews it, I will clean up the commits and merge it. You can use this PR as a reference for your future implementations.

harshsingh-24:
@Thirumalai-Shaktivel Thank you so much. I will look into it. 😄

src/libasr/pass/intrinsic_function_registry.h
czgdp1807, line 1: The new APIs are awesome. Once I merge this here, can you update LFortran's `ASRUtils::ASRBuilder` as well? No need to do the full sync, just this ASRBuilder. Probably in both the LCompilers shift `ASRBuilder` to `asr_utils.h`, so that we can use it in all the ASR passes as well. Do it after this PR.

Thirumalai-Shaktivel: Cool, I will submit a PR there. Thanks for the review!

PR #1674, opened 2023-04-07
Restrict step size 0 in for loop

faze-geek:
On master :
```
def f():
    i:i32 = 0
    for i in range(1,5,0):
        print(i)
    
f()

  File "C:\Users\kunni\lpython\try.py", line 6, in <module>
    f()
  File "C:\Users\kunni\lpython\try.py", line 3, in f
    for i in range(1,5,0):
ValueError: range() arg 3 must not be zero

1
1
1
1
.... #Goes into infinite loop
```

On branch :
```
semantic error: For loop increment should not be zero.
 --> try.py:3:5 - 4:16
  |
3 |        for i in range(1,5,0):
  |        ^^^^^^^^^^^^^^^^^^^^^^...
```

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -4493,6 +4493,10 @@ public:
         int64_t inc_int = 1;
         bool is_value_present = ASRUtils::extract_value(inc_value, inc_int);
         if (is_value_present) {
+            if (inc_int == 0) {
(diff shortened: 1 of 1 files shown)


faze-geek:
This is ready for reviews.

czgdp1807:
This makes sense but should be merged only after ``libasr`` sync is complete.

certik (approved):

PR #1676, opened 2023-04-07
Add list.count

virendrakabra14:
Adds `list.count` method. Completes a part of #941.

1. For incompatible types, an error is raised. Should we return count 0 instead?
2. I am not sure in which file to add the test-cases, but I tried the following test locally:
```python
def test_List():
    a: list[i32]
    a = [1, 2, 3]

    # print(a.count(1.0)) # semantic error: Type mismatch in 'count', the types must be compatible
    print(a.count(-1))
    print(a.count(2))
    a.append(2)
    print(a.count(2))
    a.remove(2)
    print(a.count(2))

test_List()
```
```
0
1
2
1
```


czgdp1807 (changes requested):
Nice. Please add an integration test as well for `list.count`. See `test_list_*.py` in https://github.com/lcompilers/lpython/tree/main/integration_tests for the pattern of testing lists and use your creativity for writing strong test for `list.count` (for example `l.count` where `l` can be empty, check `l.count` after every insertion/deletion in a loop, etc.)

czgdp1807:
> 1. For incompatible types, an error is raised. Should we return count 0 instead?

I would prefer an error. For testing error see https://github.com/lcompilers/lpython/tree/main/tests/errors and following pattern for registering error tests,

https://github.com/lcompilers/lpython/blob/b35e744dce5e435d04624bfbed157c9cb50befb2/tests/tests.toml#L66-L68

Shaikh-Ubaid:
I think your commits are including previous changes (already merged/present in latest `main` branch).

Shaikh-Ubaid:
To remove extra commits (already present in `main`), you can rebase over `main` branch.

For example, first update your `main` branch:
```bash
git checkout main
git pull
```

Now switch to your branch and rebase:
```bash
git checkout scratch
git rebase main
```

Now force push it
```bash
git push your_remote scratch --force-with-lease
```

virendrakabra14:
Thanks @Shaikh-Ubaid. Will use this in future.

integration_tests/test_list_count.py
```diff
@@ -0,0 +1,33 @@
...
+    assert x == [-5, -4, -3, -2, -1]
+
+    for i in range(0, 5):
+        assert x.count(i) == 0
+        x.append(i)
+        assert x.count(i) == 1
+
+    assert x == [-5, -4, -3, -2, -1, 0, 1, 2, 3, 4]
+    
+    while len(x)>0:
```
czgdp1807, line 25:
```suggestion
    while len(x) > 0:
```

integration_tests/test_list_count.py
```diff
@@ -0,0 +1,33 @@
+from lpython import i32
+
+def test_list_count():
+    i: i32
+    x: list[i32] = []
```
czgdp1807, line 5: What about testing with `list[str]`, `list[tuple[i32, str, i64]`, etc.?

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2498,6 +2498,83 @@ namespace LCompilers {
...
+        {
+            // if occurrence found, increment cnt
+            llvm::Function *fn = builder->GetInsertBlock()->getParent();
+            llvm::BasicBlock *thenBB = llvm::BasicBlock::Create(context, "then", fn);
+            llvm::BasicBlock *elseBB = llvm::BasicBlock::Create(context, "else");
+            llvm::BasicBlock *mergeBB = llvm::BasicBlock::Create(context, "ifcont");
+
+            llvm::Value* left_arg = read_item(list, LLVM::CreateLoad(*builder, i),
+                false, module, LLVM::is_llvm_struct(item_type));
+            llvm::Value* cond = llvm_utils->is_equal_by_value(left_arg, item, module, item_type);
```
czgdp1807, line 2550: Good that you figured it out on your own.

czgdp1807 (changes requested):

czgdp1807 (approved):

integration_tests/test_list_count.py
czgdp1807, line 1: Register this in the following file,

https://github.com/lcompilers/lpython/blob/7d08405d0a8e87a60467837be250bc0fbf0d659e/integration_tests/CMakeLists.txt#L282

czgdp1807 (changes requested):

integration_tests/test_list_count.py
czgdp1807: Ignore `c` for now, `cpython` and `llvm` should be fine.

integration_tests/test_list_count.py
certik: I think this is done.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2498,6 +2498,83 @@ namespace LCompilers {
...
+                        llvm::ConstantInt::get(context, llvm::APInt(32, 1)));
+            LLVM::CreateStore(*builder, tmp, i);
+        }
+        builder->CreateBr(loophead);
+
+        // end
+        llvm_utils->start_new_block(loopend);
+
+        return LLVM::CreateLoad(*builder, cnt);
+    }
```
certik, line 2576: For now this is good. Later we should lift this into ASR and implement there, it will be a lot easier to maintain.

certik (approved):
This is good, thanks!

PR #1680, opened 2023-04-07
Initial implementation of the time module

gptsarthak:
For timing the benchmarks, we currently use the bash ```time``` command. But that only provides the time in millisecond. Python's time.time() method outputs time in a much higher precision.

I have implemented the ```time.time()``` method here using ```Unix Time``` after understanding [timemodule.c](https://github.com/python/cpython/blob/main/Modules/timemodule.c)

Here is an example:
```Python
import time

def fibonacci(n : i32) -> i32:
    if n <= 1:
        return n
    return fibonacci(n-1) + fibonacci(n-2)

def main():
    print("Current time : ", time.time())
    start_time : f64 = time.time()
    a : i32 = fibonacci(35)
    print("Answer: ", a)
    print("Time taken : ", time.time()-start_time)

main()
```
```console
(base) sarthak@pop-os:~/lpython$ python examples/timetest.py
Current time :  1680876878.893159
Answer:  9227465
Time taken :  2.6417429447174072
(base) sarthak@pop-os:~/lpython$ lpython examples/timetest.py
Current time :  1.68087688629948306e+09          // Decimal : 1680876886.29948306
Answer:  9227465
Time taken :  7.29780197143554688e-02            // Decimal : 0.0729780197143554688
```
I will add tests after a review.

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -9,6 +9,7 @@
 #include <float.h>
 #include <limits.h>
 #include <ctype.h>
+#include <sys/time.h>
 
 #include <libasr/runtime/lfortran_intrinsics.h>
 #include <libasr/config.h>
@@ -1185,6 +1186,16 @@ LFORTRAN_API void _lfortran_i64sys_clock(
         *max = 0;
     }
 #endif
+}
+
+LFORTRAN_API double _lfortran_time()
+{
+    struct timeval tv;
+    if (gettimeofday(&tv, NULL) != 0) {
+        /* Error occurred, return -1 */
+        return -1.0;
+    }
+    return (double)tv.tv_sec + (double)tv.tv_usec / 1000000.0;
 }
 
 LFORTRAN_API void _lfortran_sp_rand_num(float *x) {
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -10,7 +10,7 @@
 #include <limits.h>
 #include <ctype.h>
 
-#if defined(MSC_VER) || defined(MACH_)
+#if defined(_MSC_VER)
 #  include <winsock2.h>
 #endif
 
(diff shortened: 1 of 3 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -9,6 +9,7 @@
 #include <float.h>
 #include <limits.h>
 #include <ctype.h>
+#include <sys/time.h>
```
certik, line 12: This fails on Windows:
```
D:\a\lpython\lpython\lpython-0.12.0-241-gd3a1a54f4\src\libasr\runtime\lfortran_intrinsics.c(12): fatal error C1083: Cannot open include file: 'sys/time.h': No such file or directory
```
So we need to do whatever CPython does on Windows.

src/runtime/time.py
```diff
@@ -0,0 +1,8 @@
+from lpython import f64, ccall
+
+def time() -> f64:
+    return _lfortran_time()
```
certik, line 4: Why not using the `_lfortran_cpu_time` function that is already implemented? It should work and at least it compiles on Windows, even if not implemented yet.

certik:
@gptsarthak perfect, thanks for this! I left some comments.

gptsarthak:
I have removed the use of ```<sys/time>``` and added the implementation of time for Windows. However, this needed us to include a new header <winsock2.h> (a part of <windows.h>).

CPython Reference used for implementation : [pytime.c#L871](https://github.com/python/cpython/blob/0ba0ca05d2b56afa0b055db02233e703fe138918/Python/pytime.c#L871)
> Why not using the `_lfortran_cpu_time` function that is already implemented? It should work and at least it compiles on Windows, even if not implemented yet.


I could not use that function becuase it uses ```clock()``` from ```time.h``` which does not give the same output as CPython.
Same example ran in Windows:
```console
(lp) C:\Users\gptsa\lpython>python examples\timetest.py
Current time :  1680962365.4241385
Answer:  9227465
Time taken :  6.094292402267456

(lp) C:\Users\gptsa\lpython>src\bin\lpython examples\timetest.py
Current time :  1.68096238567989349e+09
Answer:  9227465
Time taken :  1.15024566650390625e-01
```

certik (approved):

certik:
@gptsarthak do you want to add some tests in a subsequent PR?

PR #1686, opened 2023-04-08
Show Type Mismatch Error for Division

### LINKED ISSUE #1685: Bug: Division between different types does not show an error
Division between any combination of ```i32,i64,f32,f64``` and ```bool``` does not show any type mismatch error and returns a ```f64``` value.
```Python
def main():
    a : i32 = 10
    b : f64 = 20.0
    c : f64 = a / b
    d : f64 = b / a
    print(c)          # 0.5
    print(d)          # 2.0

    e : bool = True
    f : i32 = 15
    g : f64 = e / f
    h : f64 = f / e
    print(g)          # 0.066
    print(h)          # 15.0

main()
```
```console
(base) sarthak@pop-os:~/lpython$ python examples/div.py
0.5
2.0
0.06666666666666667
15.0
(base) sarthak@pop-os:~/lpython$ lpython examples/div.py
5.00000000000000000e-01
2.00000000000000000e+00
6.66666666666666657e-02
1.50000000000000000e+01
```
Though it has the same output as CPython, I don't think this is the intended behavior for LPython. If we do not support multiplication between these types, we should not support division either.

gptsarthak:
Fixes #1685


tests/reference/asr-test_c_interop_01-e374f43.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Type mismatch in binary operator; the types must be compatible
+  --> tests/../integration_tests/test_c_interop_01.py:27:31
+   |
+27 |     assert abs(_lfortran_dsin(pi/2) - 1.0) < 1e-12
```
certik, line 4: Update the test to read:
```suggestion
27 |     assert abs(_lfortran_dsin(pi/f64(2)) - 1.0) < 1e-12
```
Then this error will disappear.

tests/reference/asr-generics_list_01-39c4044.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Type mismatch in binary operator; the types must be compatible
+  --> tests/../integration_tests/generics_list_01.py:33:12
+   |
+33 |     return x / k
```
certik, line 4: Update the test to cast things correctly.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1818,6 +1818,19 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
                     int right_kind = ASRUtils::extract_kind_from_ttype_t(right_type);
                     bool is_left_f32 = ASR::is_a<ASR::Real_t>(*left_type) && left_kind == 4;
                     bool is_right_f32 = ASR::is_a<ASR::Real_t>(*right_type) && right_kind == 4;
+                    if((left_kind != right_kind) ||
```
faze-geek, line 1821: Also one thing to checkout here that is integer division by `//` still allows compatibility between these types on this branch, which I'm not sure should be allowed or not. Maybe @certik can tell if that should be allowed.
On face value I think it should not be allowed, as you can see these different types are not overloaded in the `lpython_floordiv() `function too.
```
@overload
def _lpython_floordiv(a: f64, b: f64) -> f64:
@overload
def _lpython_floordiv(a: f32, b: f32) -> f32:
@overload
def _lpython_floordiv(a: i8, b: i8) -> i8:
@overload
def _lpython_floordiv(a: i16, b: i16) -> i16:
@overload
def _lpython_floordiv(a: i32, b: i32) -> i32: 
@overload
def _lpython_floordiv(a: i64, b: i64) -> i64:
@overload
def _lpython_floordiv(a: bool, b: bool) -> bool:
```

certik: I think `//` should also be required to have the same type.

gptsarthak: Done!

tests/reference/asr-generics_list_01-39c4044.stderr
gptsarthak: Updated.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1745,6 +1745,20 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                std::string rtype = ASRUtils::type_to_str_python(ASRUtils::expr_type(right));
+                diag.add(diag::Diagnostic(
+                    "Type mismatch in binary operator; the types must be compatible",
+                    diag::Level::Error, diag::Stage::Semantic, {
+                        diag::Label("type mismatch (" + ltype + " and " + rtype + ")",
+                                {left->base.loc, right->base.loc})
+                    })
+                );
+                throw SemanticAbort();
+            }
```
czgdp1807, line 1761: Can we use the following logic here as well?

https://github.com/lcompilers/lpython/blob/57b497137b48eb0adb01e6f2b8dd6f3873b82c32/src/lpython/semantics/python_ast_to_asr.cpp#L1995-L2008

gptsarthak: Done. Thanks for pointing it out!

czgdp1807 (changes requested):
Can you add an error test as well? See, https://github.com/lcompilers/lpython/tree/main/tests/errors and 

https://github.com/lcompilers/lpython/blob/1efcc7f0db39ed7cfb502575a4d00f1fa44c0146/tests/tests.toml#L66-L68

Make a new file say, `test_type_mismatch_01.py`.

gptsarthak:
Added test.

tests/errors/test_type_mismatch_01.py
```diff
@@ -0,0 +1,14 @@
+from lpython import i32, i64, f64
+
+def test():
+    a : i32 = 10
+    b : f64 = 20.0
+    a *= b
+    a /= b
```
czgdp1807, line 7: From here onwards, no errors would be tested.

tests/errors/test_type_mismatch_01.py
```diff
@@ -0,0 +1,14 @@
...
+
+    e : i64 = i64(15)
+    f : bool = True
+    e *= f
+    e /= f
+
+test()
```
czgdp1807, line 14:
```suggestion
from lpython import i32, f64

def test():
    a : i32 = 10
    b : f64 = 20.0
    a /= b
```

czgdp1807 (changes requested):

certik (approved):
I think that this is good. Thanks for finishing the PR!

PR #1690, opened 2023-04-09
Nested Dictionaries Support

harshsingh-24:
- Adds to #983 checklist by giving support for nested dictionaries
- Added Integration tests for the same


czgdp1807 (commented):
We need to complete https://github.com/lcompilers/lpython/pull/1111 for supporting nested dictionaries. The tests you have added are very easy to support. We need to support all the tests in #1111 simultaneously to make sure that the changes aren't fragile.

harshsingh-24:
Yes. You are right @czgdp1807. The changes made here does not run for all possible combinations of string, integers, list in nested dicts. I just opened this PR for further discussions on implementations of it.

Issue #1139, opened 2022-09-23
Clarify the "LPython contract" in the docs


certik:
The contract is that if LPython compiles your code and there are no runtime errors (in Debug mode), then it will behave exactly as in Python (and you can also use Release mode and it will be very fast and still give exactly the same answer on the same input). You have to use bounds checking in Debug mode.

The same issue is with any runtime indexing, especially arrays. It can be out of bounds and we can't be checking it all the time in Release mode. The same with integer overflow. We have to be using smaller integers like i32 or i64, but they can overflow. There is nothing we can do about it, and we should not be checking it all the time, it will be slow. But in Debug mode I think we can check it reasonably fast.

We can have a ReleaseSafe mode which keeps all these runtime checks in place, but otherwise optimizes everything out as much as it can. In this mode, it would indeed not fail silently. But the ReleaseSafe mode will almost for sure be slower than just Release mode.

The default option in LPython should be Debug mode. The Debug mode must have all runtime safety options on. In this Debug mode, you get very fast compilation, and safe runtime behavior. No silent surprises. Beautiful runtime error messages with stack traces. If you don't get any runtime error and things compiles (obviously), then it will behave just like in CPython.

Then we have two more options: Release and ReleaseSafe. ReleaseSafe should never get any segfaults or undefined/random behavior, assuming there are no bugs in our compiler, but it might not be as fast as Release. The Release mode is the fastest, but you have to check each of your inputs in Debug mode (or ReleaseSafe mode). This mode is great for computational tasks where (or if) you don't mind it might segfault. As a backend in a webbrowser code you should use ReleaseSafe to prevent exploits.

We can call these modes: Debug, ReleaseSafe, ReleaseFast.

certik:
This needs to be explained and documented very early in our documentation.

PR #1700, opened 2023-04-12
C/LLVM: Fixes with dict get

Smit-create:


integration_tests/CMakeLists.txt
```diff
@@ -293,7 +293,8 @@ RUN(NAME test_dict_06        LABELS cpython llvm c)
 RUN(NAME test_dict_07        LABELS cpython llvm)
 RUN(NAME test_dict_08        LABELS cpython llvm c)
 RUN(NAME test_dict_09        LABELS cpython llvm c)
-RUN(NAME test_dict_10        LABELS cpython llvm) # TODO: Add support of dict with string in C backend 
+RUN(NAME test_dict_10        LABELS cpython llvm) # TODO: Add support of dict with string in C backend
+RUN(NAME test_dict_11        LABELS cpython c) # TODO: Add LLVM support
```
czgdp1807, line 297: We can't move ahead without enabling LLVM backend for any test. Please make it work with LLVM before C.

czgdp1807 (changes requested):

integration_tests/CMakeLists.txt
certik: That's a good point, the LLVM backend is a "lowering" backend, so it ensures our design is good. The C/C++/Julia backends are high level backends, so they don't reveal all issues. WASM is also a lowering backend.

Smit-create: Yep, that sounds right. Thanks, I fixed it.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1601,6 +1601,10 @@ namespace LCompilers {
         builder->CreateBr(mergeBB);
         llvm_utils->start_new_block(elseBB);
         {
+            if (def_value != nullptr) {
+                llvm_utils->start_new_block(mergeBB);
+                return def_value;
+            }
```
czgdp1807, line 1607: Shouldn't this be placed where we are raising the error like if `def_value` is not `nullptr` then generate code for using it, if it is `nullptr` then generate code for raising error. As of now I see that it will always return and never the generate code for reading the item if `def_value` is not `nullptr`.

integration_tests/test_dict_11.py
```diff
@@ -0,0 +1,8 @@
+from lpython import i32
+
+def test_dict_11():
+    num : dict[i32, i32]
+    num = {11: 22, 33: 44, 55: 66}
+    assert num.get(7, -1) == -1
```
czgdp1807, line 6: Test for the case following case,

```suggestion
    assert num.get(11, -1) == 22
    assert num.get(33, -1) == 44
    assert num.get(55, -1) == 66
    assert num.get(72, -110) == -110
```

faze-geek:
Did some testing on this branch. The default value should only be considered when the key is not present in the dictionary.

```
def test_dict_11():
    d : dict[i32, i32]
    d = {1: 1, 2: 2, 3: 3}
    print(d.get(1, -1))
    print(d.get(3,100)) 

test_dict_11()

1
3

-1
100
```

faze-geek:
Also another strange case I came across -

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1564,6 +1564,12 @@ namespace LCompilers {
                                         llvm_utils->create_ptr_gep(key_mask, key_hash));
         llvm::Value* is_prob_not_neeeded = builder->CreateICmpEQ(key_mask_value,
             llvm::ConstantInt::get(llvm::Type::getInt8Ty(context), llvm::APInt(8, 1)));
+        std::pair<std::string, std::string> llvm_key = std::make_pair(
+            ASRUtils::get_type_code(key_asr_type),
+            ASRUtils::get_type_code(value_asr_type)
+        );
+        llvm::Type* value_type = std::get<2>(typecode2dicttype[llvm_key]).second;
+        llvm::Value* result = builder->CreateAlloca(value_type, nullptr);
```
czgdp1807, line 1572: I think `result` should only be allocated when `def_value` is not `nullptr`. I think you should not mix `read_item` with the logic of `dict.get`. Make a new method `get_item` and call it to implement `dict.get`. Direct access via `[]` should be done by `read_item` (which should be left untouched in this PR).

czgdp1807 (changes requested):

src/libasr/codegen/llvm_utils.cpp
Smit-create: Presently, `[]` and `dict.get` are implemented with the same `DictItem` node in the ASR and we can re-use the same implementation just to extend it with default. I am not sure if it is a good idea to re-implement the same logic for `get_item` and `read_item` which effectively are doing the same job.

> I think `result` should only be allocated when `def_value` is not `nullptr`. 

It's also used on line 1597 which also makes it more generalized in the case when key_hash is matching but not the keys. This was the bug with the previous implementation where it raised an error for the case where key_hash matches but not the key while it didn't raise any error when the key is totally absent. This is also fixed using the current PR. So, I don't think separating two methods for implementing same logic is a good idea.

czgdp1807: Well in this PR itself we have seen (as pointed in https://github.com/lcompilers/lpython/pull/1700#discussion_r1169024218 and https://github.com/lcompilers/lpython/pull/1700#issuecomment-1512464866) how `read_item` is also affected when you implement `dict.get` inside it. So de-coupling `if-else-if` checks 



> I am not sure if it is a good idea to re-implement the same logic for `get_item` and `read_item` which effectively are doing the same job.

If code repetition is the issue then you can solve it by using a macro, or a utility method and call it from `read_item` and `get_item`.



> Presently, `[]` and `dict.get` are implemented with the same `DictItem` node in the ASR and we can re-use the same implementation just to extend it with default

Won't be the case very soon. We will use `IntrinsicFunction` to implement, `dict.item` and `dict.get` separately.

> It's also used on line 1597 which also makes it more generalized in the case when key_hash is matching but not the keys.

In https://github.com/lcompilers/lpython/blob/c935c96b50abeb0553213904a9f96247f6f78421/src/libasr/codegen/llvm_utils.cpp, AFAICT, its only used when `def_value != nullptr`,

https://github.com/lcompilers/lpython/blob/c935c96b50abeb0553213904a9f96247f6f78421/src/libasr/codegen/llvm_utils.cpp#L1596-L1598

Again, my POV is still the same `read_item` and `get_item` should be different methods. However they can call common utility methods or macros to execute the common logic. `if (some_arg_is_present)` and then `else` always indicate that de-coupling is needed.



> This was the bug with the previous implementation where it raised an error for the case where key_hash matches but not the key while it didn't raise any error when the key is totally absent.

Please add a test for it and fix the `read_item` and `get_item` methods accordingly.

czgdp1807 (approved):
LGTM. Before merging, please ensure our dict benchmarks don't degrade due to this PR.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1636,8 +1636,24 @@ namespace LCompilers {
...
+            llvm::Value *fmt_ptr2 = builder->CreateGlobalStringPtr(message);
+            print_error(context, module, *builder, {fmt_ptr, fmt_ptr2});
+            int exit_code_int = 1;
+            llvm::Value *exit_code = llvm::ConstantInt::get(context,
+                    llvm::APInt(32, exit_code_int));
+            exit(context, module, *builder, exit_code);
+        });
         llvm::Value* item = llvm_utils->list_api->read_item(value_list, pos,
-                                                        false, module, true);
+                                                        false, module, true);;
```
czgdp1807, line 1656: This should only happen in safe or debug modes of LPython. In fast mode one shouldn't catch these kind of errors. See, https://github.com/lcompilers/lpython/issues/1139.

I think following the same pattern as raising index errors in lists should work.

Smit-create: Thanks, updated this with even dedicated methods for default and bound checking which leaves the original method untouched (removed some if-else checks) and so its performance will be >= existing performance.

czgdp1807 (approved):

Issue #1701, opened 2023-04-12
Segmentation fault in `list.remove`


virendrakabra14:
With an element not present in the list, the error-message printing of `list.remove` segfaults.

```python
from lpython import i32, f64

def test_list_remove():
    list1: list[tuple[i32,f64]] = [(1,2.0)]
    # list1.remove((2,2.0)) # ValueError: The list does not contain the element: -698939408

    list2: list[tuple[i32, str, f64]] = [(1,'a',2.0)]
    list2.remove((2,'a',2.0)) # Segmentation Fault

test_list_remove()
```

This appears to be due to the `%d` in `("ValueError: %s%d\n")` [here](https://github.com/lcompilers/lpython/blob/ae6c5474d50f34a632704d8a75abc3e0ad3cf089/src/libasr/codegen/llvm_utils.cpp#L2485) (assumes an integer).

I think the solution is to split this into cases, but I am not exactly sure on how to handle `item` being list, tuple, etc.

czgdp1807:
I guess if an element is not present in the list then `remove` shouldn't do anything to the `list`, but Python gives me the following error,

```python
Python 3.9.6 (default, Mar 10 2023, 20:16:38) 
[Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> l = [1, 2, 3]
>>> l.remove(4)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: list.remove(x): x not in list
```

czgdp1807:
> This appears to be due to the `%d` in `("ValueError: %s%d\n")` [here](https://github.com/lcompilers/lpython/blob/ae6c5474d50f34a632704d8a75abc3e0ad3cf089/src/libasr/codegen/llvm_utils.cpp#L2485) (assumes an integer).

You can follow the same pattern as done by Python. Checked with Python 3.10.4 and its still the same. We can worry about beautification later (nice to have for usability but something which should happen later).

virendrakabra14:
> You can follow the same pattern as done by Python. Checked with Python 3.10.4 and its still the same. We can worry about beautification later (nice to have for usability but something which should happen later).

Agreed.

I tried the suggested change (to just print `x`). This resolves the first error, but the second (segfault) persists.

I think this is due to control going ahead of the `find_item_position` call in `remove` or `index`, even if the element was not found. Not sure why this is happening, since we call `exit` in the former.

PR #1703, opened 2023-04-12
Add list.index

virendrakabra14:
This reuses `LLVMList::find_item_position` in `llvm_utils.cpp`. Part of #941.

Issues

1. #1701 is common to both `list.remove` and `list.index`.
2. `tests/errors/test_list_index.py` doesn't appear to give an error, but putting the same snippet at the end of `integration_tests/test_list_index.py` correctly outputs `ValueError: The list does not contain the element: 0`. (Is this related to the above point?)

CPython allows specifying a range for finding the first occurrence as `list.index(x[, start[, end]])`. This would be easier to implement using the existing implementations of list-slicing and `index`. Could someone please point to where we implement such overloads for data-structure methods?


czgdp1807:
> This would be easier to implement using the existing implementations of list-slicing and `index`.

It won't be efficient because slicing a list will create a copy which is not needed at all. I will start the loop from `max(0, start)` and go till `min(end, len(x))`. Its the same scene as `list.count`. Add a new node for `list.index` as, `ListIndex(expr list, expr x, expr? start, expr? end)` and then implement in LLVM directly. Later on we can use `IntrinsicFunction` infrastructure. 

For long term, **How to use IntrinsicFunction for implementing methods like, `list.index`, `list.count`, etc.?**

1. Register `ListIndex` in the same way as we do for `Sin`, `Cos` in https://github.com/lcompilers/lpython/blob/main/src/libasr/pass/intrinsic_function_registry.h.
2. Now inside namespace `ListIndex` while writing the `instantiate_ListIndex` you can create a `Function` depending on the input arguments. Its like overloading (mangling the function's name) depending on the presence of `start` and `end` arguments. The do loop can be generated inside the `Function`'s body using ASR APIs directly.  

Point 2 is for the far future.

In fact if I think more you don't need to add `ListIndex` as a new ASR node. Just use `IntrinsicFunction` and register `list.index` in the same way as we do for `Sin`, `Cos`. Then you can do `visit_IntrinsicFunction` in `asr_to_llvm.cpp` and implement it there directly (like check the ID of the intrinsic and generate the LLVM implementation for that intrinsic). In https://github.com/lcompilers/lpython/blob/main/src/libasr/pass/intrinsic_function.cpp we can make the presence of `instantiate_function` optional by adding an `if` check around,

https://github.com/lcompilers/lpython/blob/ae6c5474d50f34a632704d8a75abc3e0ad3cf089/src/libasr/pass/intrinsic_function.cpp#L63-L71

Also, you can ignore implementing `instantiate_ListIndex` and do not register it in the following map,

https://github.com/lcompilers/lpython/blob/ae6c5474d50f34a632704d8a75abc3e0ad3cf089/src/libasr/pass/intrinsic_function_registry.h#L456-L466

Also modify the following by returning a `nullptr` for those IDs which are not present in `intrinsic_function_by_id_db`.

https://github.com/lcompilers/lpython/blob/ae6c5474d50f34a632704d8a75abc3e0ad3cf089/src/libasr/pass/intrinsic_function_registry.h#L489-L491

By following the above approach you won't be required to add new ASR nodes for each intrinsic feature of Python language like, `list.index`, `dict.get`, `list.remove`, etc. Later on once ASR is mature enough we can add `instantiate_ListIndex`, `instantiate_DictGet`, etc and make using LLVM implementations optional.

cc: @certik What do you say?

czgdp1807:
If @certik agrees and if you want, I can try the above approach here in your PR and we will see if its doable or not. If it is then we can remove a large number of unnecessary nodes from ASR and just register everything in https://github.com/lcompilers/lpython/blob/main/src/libasr/pass/intrinsic_function.cpp.

certik (approved):
Yes, @czgdp1807's plan is the way to go I think. However, this PR is small and clean, so I think it's ok to merge it and then do the refactoring in a new PR.

@czgdp1807 I'll leave the decision up to you. If it's easy to do the redesign here, then let's do it, otherwise in a new PR.

virendrakabra14:
Thanks for the reviews.

> 2. `tests/errors/test_list_index.py` doesn't appear to give an error, but putting the same snippet at the end of `integration_tests/test_list_index.py` correctly outputs `ValueError: The list does not contain the element: 0`. (Is this related to the above point?)

Was this reproducable?

tests/tests.toml
```diff
@@ -676,6 +676,10 @@ asr = true
 filename = "errors/test_list_count.py"
 asr = true
 
+[[test]]
+filename = "errors/test_list_index.py"
+asr = true
```
czgdp1807, line 681:
```suggestion
run = true
```

See, https://github.com/lcompilers/lpython/blob/a1fff8b7460fe350b3edbf837b7ff79c76192781/tests/tests.toml#L1025-L1040

czgdp1807:
> If it's easy to do the redesign here, then let's do it,

I am doing it here. Let's see how it goes.

czgdp1807:
@virendrakabra14 I have pushed the changes (which I described above). Please update your branch locally.

@certik Please let me know if they look good to you. I will apply them to LFortran as well. Initially keeping IntrinsicFunction in sync between LPython/LFortran will help in avoiding problems. I will not apply list.index changes, just the API changes in `IntrinsicFunctionRegistry` namespace will be applied to LFortran.

certik (approved):
I think this looks pretty good, thanks!

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -453,6 +455,54 @@ namespace Abs {
...
+    arg_values.reserve(al, args.size());
+    for( size_t i = 0; i < args.size(); i++ ) {
+        arg_values.push_back(al, ASRUtils::expr_value(args[i]));
+    }
+    ASR::expr_t* compile_time_value = eval_list_index(al, loc, arg_values);
+    ASR::ttype_t *to_type = ASRUtils::TYPE(ASR::make_Integer_t(al, loc,
+                            4, nullptr, 0));
+    return ASR::make_IntrinsicFunction_t(al, loc,
+            static_cast<int64_t>(ASRUtils::IntrinsicFunctions::ListIndex),
+            args.p, args.size(), 0, to_type, compile_time_value);
```
certik, line 501: Is the idea that `create_ListIndex` is to be used when we need to check the argument types and compute compile time value (if available), but if we already have everything ready (such as in some transformation passes), we just use make_IntrinsicFunction_t directly?

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -453,6 +455,54 @@ namespace Abs {
...
+    ASR::expr_t* list_expr = args[0];
+    ASR::ttype_t *type = ASRUtils::expr_type(list_expr);
+    ASR::ttype_t *list_type = ASR::down_cast<ASR::List_t>(type)->m_type;
+    ASR::ttype_t *ele_type = ASRUtils::expr_type(args[1]);
+    if (!ASRUtils::check_equal_type(ele_type, list_type)) {
+        std::string fnd = ASRUtils::type_to_str_python(ele_type);
+        std::string org = ASRUtils::type_to_str_python(list_type);
+        err(
+            "Type mismatch in 'index', the types must be compatible "
+            "(found: '" + fnd + "', expected: '" + org + "')", loc);
```
certik, line 489: I am a bit worried that we have a frontend dependency here (`ASRUtils::type_to_str_python`).

certik:
I think this is ok to merge. I have some design questions, but I think we can iterate on this after it is merged.

src/libasr/pass/intrinsic_function_registry.h
czgdp1807: Ah! Yes. I would do `get_type_code` here.

src/libasr/pass/intrinsic_function_registry.h
czgdp1807: Yes. All the error checking related to an intrinsic function resides in one place. Also there is no restriction on using `make_IntrinsicFunction_t` directly (its a public API and if you few feel you can avoid calling, `create_ListIndex` or any other create function then feel free to do so).

certik: Later we might need to add some checks to verify() to ensure all arguments to IntrinsicFunctions are correct. For now this is good.

czgdp1807:
I will make some updates here before merging.

src/libasr/pass/intrinsic_function_registry.h
czgdp1807: We now use `get_type_code` since its just an error message so our frontend agnostic type codes should convey the error with same clarity.

src/libasr/pass/intrinsic_function_registry.h
czgdp1807: We can easily do this by registering `verify_args` function in `IntrinsicFunctionRegistry`. We can define, `LogGamma::verify_args`, `Abs::verify_args`, etc. Then, in asr_verify.cpp we can call these methods in the same way as we call `instantiate_function` in `intrinsic_function.cpp`.

PR #1707, opened 2023-04-13
[C] Initialize the list struct type after declaration

Thirumalai-Shaktivel:


integration_tests/test_list_11.py
```diff
@@ -0,0 +1,9 @@
+def get_string() -> str:
+    return "1"
+
+def test() -> list[str]:
+    x: list[str]
+    x.append(get_string())
+    return x
```
Thirumalai-Shaktivel, line 7: I see, this doesn't seem to be a valid CPython code!
`x: list[str] = []` or `x: list[str] = f()` works without any seg_fault in the C backend.

@certik I think we have to test our LPython's modules like `sys`, `math`, ... using `CPython` (in CI maybe). So that we don't add code that doesn't compile with CPython.

Also, I don't think we can throw an error message saying `x is not initialized`.

Thirumalai-Shaktivel:
I think this PR can be closed.

Issue #703, opened 2022-06-28
Interface with Python (CPython)


certik:
We need to design some kind of a decorator similar to `ccall` and `ccallable` (which interfaces to and from C) but to/from Python. Since our surface language *is* Python, the design can be different. Such as:
```python
@cpython
def f(n: i32, x: str) -> str:
    import sympy
    x_ = sympy.Symbol(x)
    e = ((x_+5)**i).expand()
    return str(e)
```
LPython will take the contents of the decorated function and turn it into a series of Python C/API calls in ASR, and it will convert the arguments to / from ASR's i32/str into CPython correctly.

This approach will allow calling any Python code naturally from LPython. With `@ccall` we can naturally call any C code.

The decorator can be just called `@python`, but since LPython is Python (subset at first, but we are keeping the door open to possibly support all of Python later), it might be confusing what exactly is meant by it.

The second decorator we need is to call a function from Python. Something like:
```
@cpython_callable
def fast_sum(x: f64[n]) -> f64:
    s: f64 = 0
    for i in range(n):
        s += x[i]
    return s
```
The contents of the function is regular LPython code that will get optimized by our optimizers to produce top performing (vectorized) machine code. But the interface will get exposed to CPython, so LPython will do the conversion from a NumPy array into `f64[n]`, and it will convert the return value `f64` from ASR to CPython. Not sure about all the details yet in ASR, but with the C backend it will produce code that when compiled would create a CPython extension module (shared library) that you just import in Python and it will just work.

With these two decorartors (`@cpython` and `cpython_callable`) we can use LPython to create extension modules to CPython. We can easily call any C library if we want to (thus from this perspective it is very similar to Cython), and we can also just implement things in regular LPython code which gets highly optimized. Everything is just Python, so you can take any such LPython code and if you run it with CPython, it will work (just slower). 

This is a very powerful design that allows to 100% stay in Python. One starts with CPython, implements some code, then takes parts of it, extract into a separate module that LPython can compile, decorate the API with `@cpython_callable` and use LPython to compile. Since `@ccall` works from both CPython and LPython, one can interface any C code easily from both. It's a very simple gradual approach, and it allows to achieve the top performance, as we deliver more on LPython as a regular production optimizing compiler, that is very good with arrays.

czgdp1807:
So, here's what I understand from our discussions and the issue above,

1. `@cpython_callable` - Makes the function callable from CPython. For that you will be required to create a CPython extension module which can be imported from CPython code. 
[...]

PR #1708, opened 2023-04-13
`@JIT` interface for LPython

harshsingh-24:
- Solves #703 for adding `@JIT` interface to LPython

PS: Work in Progress


src/runtime/lpython/lpython.py
```diff
@@ -299,6 +299,31 @@ def __call__(self, *args, **kwargs):
             else:
                 new_args.append(arg)
         return self.cf(*new_args)
+    
+class jit:
+    def __init__(self, func):
+        self.func = func
+
+    def __call__(self, *args, **kwargs):
+        # Get the source code of the function
```
Smit-create, line 308: This will be used by just CPython tests and has no relation with LPython's LLVM backend. Can CPython not ignore the JIT and return a regular function? As we will need to handle `lpython.jit` at ASR or LLVM backend.

harshsingh-24: I don't know. I was just trying to follow the steps from what @certik wrote.

harshsingh-24:
Hey @Thirumalai-Shaktivel . How can I make `jit` recognize in backend? I do understand that we need to make some changes in backend.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3405,6 +3405,8 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                         is_inline = true;
                     } else if (name == "static") {
                         is_static = true;
+                    } else if (name == "jit") {
+                        is_jit = true;
```
certik, line 3409: I think `jit` should not compile with LPython, but let's raise some informative message, like "@lpython.jit` decorator must be run from CPython, not compiled using LPython".

src/runtime/lpython/lpython.py
certik: It looks like this does steps 1. and 2. from https://github.com/lcompilers/lpython/issues/703#issuecomment-1478344669.

Now we need to do step 3.

harshsingh-24:
Can someone explain on generate C code with Python wrappers? @Thirumalai-Shaktivel @certik

Smit-create:
@harshsingh-24 Any updates on this?

harshsingh-24:
I am working on it. @Smit-create Is this an urgent issue? I was having my exams that is why I am a little slow on progress here.

Thirumalai-Shaktivel:
@harshsingh-24 I will take over this implementation. You can work on other issues like adding other string built-in functions.

harshsingh-24:
> @harshsingh-24 I will take over this implementation. You can work on other issues like adding other string built-in functions.

I am really sorry that I delayed it this much @Thirumalai-Shaktivel. I know this was a deliverable MVP but I have my end-semester going and I was not able to keep up. 😓

Thirumalai-Shaktivel:
No issues, @harshsingh-24. Finish up your exams (it's very important. All the best with your exams!).
You can work here anytime.

certik:
Thanks for all your help @harshsingh-24 ! No worries, school is always higher priority.

src/runtime/lpython/lpython.py
```diff
@@ -498,3 +498,120 @@ def ccallable(f):
...
+        with open("a.c", "w") as file:
+            # Write the C source code to the file
+            file.write(template)
+
+        r = os.system("lpython --show-c --disable-main a.py > a.h")
+        assert r == 0, "Failed to create C file"
+        gcc_flags = " -bundle -flat_namespace -undefined suppress "
+        python_path = "-I" "$CONDA_PREFIX/include/python3.10 "
+        numpy_path = "-I" "$CONDA_PREFIX/lib/python3.10/site-packages/" \
+            "numpy/core/include/"
```
Thirumalai-Shaktivel, line 605: Is this okay? or is there any other way to specify the path for Numpy?

Thirumalai-Shaktivel:
This PR is ready for testing!
@certik if you have time, do you mind testing this? see if you can break something.
Meanwhile, I will try to clean up and add the required comments to explain what the code does.

src/runtime/lpython/lpython.py
certik: For NumPy use `import numpy; print numpy.get_include()`.

certik: For the Python path, I think this works: `from distutils.sysconfig import get_python_inc; print get_python_inc()`.

certik:
I would fix at the CI and merge this. This provides the foundation and we can then keep adding more tests and start fixing it until it is more usable.

I think it works on macOS, but we need to fix it on Linux:
```console
212/215 Test #215: test_jit_01 ......................***Failed    0.47 sec
gcc: error: unrecognized command-line option ‘-bundle’
gcc: error: unrecognized command-line option ‘-flat_namespace’
gcc: error: unrecognized command-line option ‘-rpath’
  File "/home/runner/work/lpython/lpython/integration_tests/test_jit_01.py", line 5, in <module>
    def fast_sum(n: i32, x: f64[:]) -> f64:
  File "/home/runner/work/lpython/lpython/src/runtime/lpython/lpython.py", line 722, in __init__
    assert r == 0, "Failed to create the shared library"
AssertionError: Failed to create the shared library
```
There you just use `-shared` as an option.

certik:
It looks like the CI might pass. If it does, then let's rebase the history (squash @harshsingh-24's commits into some logical commits, and then add your commits, create some nice sequence) and mark this "Ready for review". I'll then do one final review and let's merge.

I think you can keep the paths hardwired for now, if they work on both Linux and macOS, that's good enough. We will then make it completely general and Python version independent with subsequent PRs.

Thirumalai-Shaktivel:
I think this is ready! 
@certik @czgdp1807 Do you have any suggestions? Otherwise I will clean up the history and merge this.

src/runtime/lpython/lpython.py
```diff
@@ -4,6 +4,8 @@
 import platform
 from dataclasses import dataclass as py_dataclass, is_dataclass as py_is_dataclass
 from goto import with_goto
+from numpy import get_include
```
certik, line 7: I think we don't want to include numpy here, since it would be imported with every "import lpython" in every file and things will be slow. Python's imports are slow, and we want "import lpython" to be immediate. For now I would "import numpy" when the decorator is processed, not here. This can be done in subsequent PR.

certik (approved):
I think this is good enough.

PR #1709, opened 2023-04-13
Never inline intrinsic functions

### LINKED ISSUE #1698: lfortran_intrinsics do not work with --fast
Currently any of the functions/imports that use the lfortran_intrinsics architecture do not work with the ```--fast``` option.
Functions that are supposed to return integer/float values, return 0 or garbage values.

For example:
```Python
import time
import random
import math

print(time.time())
print(random.randrange(10,20))
print(random.random())
print(math.atanh(0.90))
```
```console
(lp) sarthak@pop-os:~/lpython$ lpython examples/time.py
1.68124988955907702e+09
13
3.94382926819093038e-01
1.47221948958322035e+00
(lp) sarthak@pop-os:~/lpython$ lpython --fast examples/time.py
0.00000000000000000e+00
0
9.53282412436823800e-130
9.53282412436823800e-130

```
Update : The issue is somewhere in ```inline_function_calls``` pass.

gptsarthak:
Fixes #1698 
```Python
import time
import random
import math

print(time.time())
print(random.randrange(10,20))
print(random.random())
print(math.atanh(0.90))
```
Main:
```console
(lp) sarthak@pop-os:~/lpython$ lpython --fast examples/time.py
0.00000000000000000e+00
0
9.53282412436823800e-130
9.53282412436823800e-130
```
On branch:
```console
(lp) sarthak@pop-os:~/lpython_new/lpython$ src/bin/lpython --fast examples/hi.py
1.68139046414960361e+09
13
3.94382926819093038e-01
1.47221948958322035e+00
```

### DIFF
--- a/src/libasr/pass/inline_function_calls.cpp
+++ b/src/libasr/pass/inline_function_calls.cpp
@@ -153,12 +153,6 @@ public:
                 ASR::ExternalSymbol_t* called_sym_ext = ASR::down_cast<ASR::ExternalSymbol_t>(x.m_name);
                 ASR::symbol_t* f_sym = ASRUtils::symbol_get_past_external(called_sym_ext->m_external);
                 ASR::Function_t* f = ASR::down_cast<ASR::Function_t>(f_sym);
-
-                // Never inline intrinsic functions
-                if( ASRUtils::is_intrinsic_function2(f) ) {
-                    return ;
-                }
-
                 ASR::symbol_t* called_sym = x.m_name;
 
                 // TODO: Handle later
@@ -217,6 +211,12 @@ public:
             return ;
         }
 
+        // Never inline intrinsic functions
+        if( ASRUtils::is_intrinsic_function2(func) ||
+            startswith(func->m_name,"_lfortran_") ) {
+            return ;
+        }
+        
         if( !is_fast && !ASRUtils::get_FunctionType(func)->m_inline ) {
             return ;
         }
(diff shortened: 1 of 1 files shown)


certik:
Thanks for the workaround!

I would rather fix this properly. The runtime library has a few hacks in the code base and this PR feels to me like it is adding another hack. I would rather remove some hacks, or add them into one place and wrap them in some functions.

If the problem is that we can't reliably identify intrinsic functions, then let's fix that.

gptsarthak:
I guess we just have have to find a way so that functions that are coming from imported modules do not get inlined. @czgdp1807 do you know anything about that?

czgdp1807:
> I guess we just have have to find a way so that functions that are coming from imported modules do not get inlined.

Why so? If we are able to access the body of the function in ASR then I think inlining should be possible and we should do it. Probably that would be the proper fix instead of avoiding inlining them. Let's do it here and see what blocks us from properly inlining them. The only case which needs avoiding inlining is those functions which contain fallback implementations of optimisations such as flip_sign, etc, specifically, https://github.com/lfortran/lfortran/blob/main/src/runtime/builtin/lfortran_intrinsic_optimization.f90.

gptsarthak:
But we cannot inline functions which are implemented in C using ```@ccall``` am I wrong?

certik:
For `BindC` functions, we can inline the function call. Obviously we can't inline the function itself, since we don't have the source code.

gptsarthak:
Is this alright? We stop inlining BindC function after we have inlined its fuction call.
```--fast``` works alright with this fix.

czgdp1807 (approved):

Issue #1442, opened 2023-01-20
For-else configuration not supported 


faze-geek:
Came across this while testing basic loops in `lpython. `
```
from ltypes import i32

def f():
    i: i32
    for i in range(0,5,1):
        print(i)
    else:
        print(100)

f()

(lp) C:\Users\kunni\lpython>python try.py
0
1
2
3
4
100  # Should be printed because there is no break statement in loop

(lp) C:\Users\kunni\lpython>src\bin\lpython try.py
   Creating library try.lib and object try.exp
0
1
2
3
4
```

anutosh491:
The ASR being generated here is not correct but the AST is ... which tells me that some ast to asr conversion file might be where this should be addressed . Taking the example above ... the ASR lacks the following block (responsible for the

```
                        [(Print
                            ()
                            [(IntegerConstant 100 (Integer 4 []))]
                            ()
                            ()
                        )]
```
Once we handle this correctly ... this should be done !

anutosh491:
Okay so the function `visit_For` in `python_ast_to_asr.cpp` file takes in a `const AST::For_t &x` variable . Now the struct `AST::For_t` respects the else construct of the for-else configuration as shown in the grammar .
```
stmt = FunctionDef(identifier name, arguments args,
                       stmt* body, expr* decorator_list, expr? returns,
                       string? type_comment)
         ....................
          -- use 'orelse' because else is a keyword in target languages
          | For(expr target, expr iter, stmt* body, stmt* orelse, string? type_comment)
```
But while converting the AST to an ASR , we use either the `ASR::make_DoConcurrentLoop_t` or the `ASR::make_DoLoop_t` both which do not take into consideration the else body . Does this demand for changing the structure of `DoLoop`
```
    struct DoLoop_t // Constructor
    {
        const static stmtType class_type = stmtType::DoLoop;
        typedef stmt_t parent_type;
        stmt_t base;
        do_loop_head_t m_head;
        stmt_t** m_body; size_t n_body; // Sequence
        
    };
```
Cause this takes into account only the loop body through `m_body/n_body `

czgdp1807:
AFAICT, the semantics of `for-else` is simply that if the loop is stopped by a `break` statement then else is never executed, otherwise it is always executed.

Now, there are two ways to represent the above semantics in ASR,

[...]

PR #1711, opened 2023-04-14
Add support for for-else

thebesttv:
This is my attempt at adding support for for-else (issue #1442, according to [czgdp1807's comment](https://github.com/lcompilers/lpython/issues/1442#issuecomment-1434764718)).

A new ASR node `ForElse` is added.  In `visit_For()` at `python_ast_to_asr.cpp`, loops with non-zero `n_orelse` are converted to `ForElse`.

An ASR pass `"for_else", &pass_replace_forelse}` is added, which consists of two steps:
1. `ForElseVisitor`: `ForElse` is visited and converted to `DoLoop` with an initial assignment to a boolean flag variable (`_no_break_N`), and an `If` block at the end that executes `orelse` as long as the flag remains true.  The flag variable corresponding to the `DoLoop` is recorded in `doLoopFlagMap`.
2. `ExitVisitor`: `Exit` is visited, and its enclosing `DoLoop` is recorded in a stack.  If an `Exit` is found, and the enclosing `DoLoop` is originally a `ForElse`, the corresponding flag is set to false.

Some tests are added, under `tests/for_else`.


certik:
@thebesttv I am really sorry, we somehow totally missed this PR. Next time please don't hesitate to ping us if we forget. We usually are pretty good about that. Also when the PR is ready, as this one seems, go ahead and mark it "Ready for review". I think what you implemented one approach --- a separate node that gets transformed via ASR->ASR pass.

Why not adding a new member to `DoLoop`: https://github.com/lcompilers/lpython/issues/2353#issuecomment-1742743488?

kmr-srbh:
@certik I think this PR can be closed as #2353 was merged.

PR #1594, opened 2023-03-20
Added cbrt() and exp2() functions in Math module

harshsingh-24:
Math Module enhancement:

- Added `cbrt()` or `cube root` function associated with Math module of LPython and it's associated integration tests
- Added `exp2()` or `exponential of 2` function associated with Math module of LPython and it's associated integration tests

Issue #92, opened 2022-02-01
Should `mod` be supported at ASR level or at compile time?


Smit-create:
I implemented the `gcd` function in the following way:
<details>
<summary> Code </summary>

<p>

```py
def gcd(a: i32, b: i32):
    temp: i32
    while b!=0:
        a = a % b
        temp = a
        a = b
        b = temp
    return a
```
</p>
</details>
<details>
<summary> AST </summary>

<p>

```bash
(Module [
   (FunctionDef 
      gcd 
      ([] [
         (
            a 
            (Name 
               i32 
               Load) ()) 
         (
            b 
            (Name 
               i32 
               Load) ())] [] [] [] [] []) [
      (AnnAssign 
         (Name 
            temp 
            Store) 
         (Name 
            i32 
            Load) () 1) 
      (While 
         (Compare 
            (Name 
               b 
               Load) 
            NotEq [
            (ConstantInt 0 ())]) [
         (Assign [
            (Name 
               a 
               Store)] 
            (BinOp 
               (Name 
                  a 
                  Load) 
               Mod 
               (Name 
                  b 
                  Load)) ()) 
         (Assign [
            (Name 
               temp 
               Store)] 
            (Name 
               a 
               Load) ()) 
         (Assign [
            (Name 
               a 
               Store)] 
            (Name 
               b 
               Load) ()) 
         (Assign [
            (Name 
               b 
               Store)] 
            (Name 
               temp 
               Load) ())] []) 
      (Return 
         (Name 
            a 
            Load))] [] () ())] [])

```
</p>
</details>

<details>
<summary> ASR </summary>

<p>

```bash
semantic error: Binary operator type not supported
 --> ser.txt:1:1
  |
1 | 0 0 1 0 3 gcd 0 2 1 a 1 26 3 i32 0 0 1 b 1 26 3 i32 0 0 0 0 0 0 0 3 7 26 4 temp 1 26 3 i32 0 0 1 10 15 26 1 b 0 1 1 20 0 0 4 5 1 26 1 a 1 2 26 1 a 0 5 26 1 b 0 0 5 1 26 4 temp 1 26 1 a 0 0 5 1 26 1 a 1 26 1 b 0 0 5 1 26 1 b 1 26 4 temp 0 0 0 3 1 26 1 a 0 0 0 0 0 
  | ^ 


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).

```
</p>
</details>

[...]

PR #1712, opened 2023-04-16
Improve performance of exp() and exp2()

gptsarthak:


gptsarthak:
>159/202 Test  #92: elemental_04 .....................***Exception: SegFault  0.20 sec

Cannot figure out whats causing the segfault. It does not give any error locally on my linux system.

czgdp1807 (commented):
Instead of this I would recommend to implement `exp` using `IntrinsicFunction` API as we have done for `abs`, `sin`, `cos` already. Then you can implement `generate_Exp` (similarly as `generate_ListIndex`)  in `visit_IntrinsicFunction` method in `asr_to_llvm.cpp`. Same for C backend. Then we won't have to use `lfortran_intrinsics.c` runtime library. Each backend will be able to use its preferred way of implementing exponentiation. 

References,

For `IntrinsicFunction` - https://github.com/lcompilers/lpython/blob/main/src/libasr/pass/intrinsic_function_registry.h (see `ListIndex` specifically as we will be implementing `exp` in the backend as we have done for `list.index`).
https://github.com/lcompilers/lpython/blob/57ebbc81e23854d3af50c2d5042f2f1b184941b5/src/libasr/codegen/asr_to_llvm.cpp#L1934-L1971
https://llvm.org/docs/LangRef.html#llvm-exp-intrinsic, https://llvm.org/docs/LangRef.html#llvm-exp2-intrinsic, https://llvm.org/doxygen/IRBuilder_8cpp_source.html#l00956

certik:
exp and exp2 should be implemented using the IntrinsicFunction mechanism.

gptsarthak:
@certik My exams got over today and I just started working right now, that was just a merge commit. I'll implement them now.

gptsarthak:
Implemented ```exp()``` and ```exp2``` using ```IntrinsicFunction```
```python
# no import

def test():
    x : f64 = 0.5
    print(exp(x))
    print(exp2(x))

test()
```
```console
(lp) sarthak@pop-os:~/lpython$ src/bin/lpython --fast examples/test.py
1.64872127070012819e+00
1.41421356237309515e+00
```

gptsarthak:
One thing I observed after implementing them is that Python does not actually have ```exp2()``` in the ```math``` module. Only Numpy has it. On the other hand, we have it in the math module, and since we have implemented them using IntrinsicFunction it will work without any import for now.

certik (approved):
I think it looks good. Yes, we should simplify using macros, but in subsequent PRs.

gptsarthak:
If we do 
```python
from math import exp
print(exp(0.5))
``` 
or do 
```python 
import math 
print(math.exp(0.5))
``` 
that still uses the old implementation. 
Do we need to fix this? Most users would import the modules.
One solution I can think of is:
In ```intrinsic_function_registry.h```
```cpp
{"math_exp", {&Exp::create_Exp, &Exp::eval_Exp}},
```
In ```math.py``` as well as ```lpython_intrinsic_numpy.py```
```python
def exp(x: f64) -> f64:
    """
    Return `e` raised to the power `x`.
    """
    return math_exp(x)
```

czgdp1807 (changes requested):
While doing AST to ASR transition, we also need to create `IntrinsicFunction` node in ASR as well. See,

https://github.com/lcompilers/lpython/blob/b28a44046c9f375ce6b8f39b00226cdb36e1ccfe/src/lpython/semantics/python_ast_to_asr.cpp#L6446-L6460

Probably adding "exp" should work. Also, we should raise an error if `from math import exp` or `from numpy import exp` isn't there. For "exp2" we should raise an error if `from numpy import exp2` is absent. Also `math.exp` should only accept scalars (error otherwise). `numpy.exp` and `numpy.exp2` can accept arrays.

gptsarthak:
@czgdp1807 So I did this
```c
std::set<std::string> not_cpython_builtin = {
                "sin", "cos", "gamma", "tan", "asin", "acos", "atan", "exp", "exp2"
            };
```
and it raised the error as expected.
```console
semantic error: Function 'exp' is not declared and not intrinsic
  --> examples/hi.py:12:7
   |
12 | print(exp(0.90))
   |       ^^^^^^^^^ 
```
But the problem is that when we do import, it is using the old implementation inside ```math.h```. (I benchmark them every time and also put debugging flags)
How to deal with this? We need to use the intrinsic functions for the performance increase.

czgdp1807:
Well try to fix this. Remove Python implementations of `exp` and `exp2` from LPython and then try. This needs debugging.

gptsarthak:
> Well try to fix this. Remove Python implementations of `exp` and `exp2` from LPython and then try. This needs debugging (and is doable).

@czgdp1807 Please check if this is the correct way to handle this. 
> Also math.exp should only accept scalars (error otherwise). numpy.exp and numpy.exp2 can accept arrays.

I'm working on this. How to vectorize these functions for arrays in case of numpy? I could not figure it out. I know its something related to ```elemental```.

src/runtime/math.py
```diff
@@ -695,7 +683,7 @@ def atanh(x: f64) -> f64:
 
 
 def expm1(x: f64) -> f64:
```
czgdp1807, line 685: May be, `expm1` can also become an `IntrinsicFunction`. It should just re-use `Exp` intrinsic.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -6428,15 +6428,41 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            std::set<std::string> module_math = {
+                "sin", "cos", "gamma", "tan", "asin", "acos", "atan", "exp"
             };
-            if (ASRUtils::IntrinsicFunctionRegistry::is_intrinsic_function(call_name)
-             && not_cpython_builtin.find(call_name) == not_cpython_builtin.end()) {
+            std::set<std::string> module_numpy = {
+                "exp", "exp2"
+            };
+            // Complex conditions because of functions which are in both modules
+            bool is_math_imported = (current_scope->resolve_symbol("math") != nullptr) ? true : false;
```
czgdp1807, line 6438: The condition is also boolean so you don't need to use a conditional operator which again produces boolean.

czgdp1807: Though the condition is fragile. What if I as a user also create a module named `math`? It would be true in that case as well.

czgdp1807:
I don't see either `Exp` or `Exp2` in the following function,

https://github.com/lcompilers/lpython/blob/f4a0a6d329421ba37fe032eafd4dc9b8f1815c0a/src/libasr/pass/intrinsic_function_registry.h#L1007-L1014

> If we use `exp(array[0])`:

So probably its related to incorrect processing of `ArrayItem` in `generate_Exp`. C code needs pointers but LLVM code needs value.

gptsarthak:
> I don't see either `Exp` or `Exp2` in the following function,

I meant that I tried that in my local repo and it didnt work. I'll push that now along with some changes.
Also, I observed that numpy in python actually accepts scalar values and does not throw an error. I think we can remove the "does not accept scalar values" error.

> So probably its related to incorrect processing of ArrayItem in generate_Exp. C code needs pointers but LLVM code needs value.

Yes. I tried the following in ```asr_to_llvm.cpp```
```cpp
if(ASR::is_a<ASR::ArrayItem_t>(*m_arg)){
    ASR::ArrayItem_t *item = ASR::down_cast<ASR::ArrayItem_t>(m_arg);
    m_arg = item->m_value;
}
generate_Exp(m_arg);
break ;
```
Didnt work, threw segfault. ```m_value``` isnt probably what I think it is.

gptsarthak:
@czgdp1807 Finally figured it out, just had to use ```this->visit_expr_wrapper(m_arg, true);```
I dont know why tests are failing here, integration tests pass on my system.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -536,6 +538,73 @@ namespace Abs {
...
+namespace Exp {
+
+    static ASR::expr_t* eval_Exp(Allocator &al, const Location &loc,
+            Vec<ASR::expr_t*> &args) {
+        LCOMPILERS_ASSERT(ASRUtils::all_args_evaluated(args));
+        ASR::expr_t* arg = args[0];
+        ASR::ttype_t* t = ASRUtils::expr_type(args[0]);
+        if (ASRUtils::is_real(*t)) {
+            double val = ASR::down_cast<ASR::RealConstant_t>(arg)->m_r;
+            return ASR::down_cast<ASR::expr_t>(ASR::make_RealConstant_t(
```
Thirumalai-Shaktivel, line 550:
```suggestion
            return EXPR(ASR::make_RealConstant_t(
```

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -536,6 +538,73 @@ namespace Abs {
...
+namespace Exp2 {
+
+    static ASR::expr_t* eval_Exp2(Allocator &al, const Location &loc,
+            Vec<ASR::expr_t*> &args) {
+        LCOMPILERS_ASSERT(ASRUtils::all_args_evaluated(args));
+        ASR::expr_t* arg = args[0];
+        ASR::ttype_t* t = ASRUtils::expr_type(args[0]);
+        if (ASRUtils::is_real(*t)) {
+            double val = ASR::down_cast<ASR::RealConstant_t>(arg)->m_r;
+            return ASR::down_cast<ASR::expr_t>(ASR::make_RealConstant_t(
```
Thirumalai-Shaktivel, line 584:
```suggestion
            return EXPR(ASR::make_RealConstant_t(
```

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -902,6 +971,10 @@ namespace IntrinsicFunctionRegistry {
             "atan"},
         {static_cast<int64_t>(ASRUtils::IntrinsicFunctions::Abs),
             "abs"},
+        {static_cast<int64_t>(ASRUtils::IntrinsicFunctions::Exp),
+            "exp"},
+            {static_cast<int64_t>(ASRUtils::IntrinsicFunctions::Exp2),
```
Thirumalai-Shaktivel, line 976:
```suggestion
        {static_cast<int64_t>(ASRUtils::IntrinsicFunctions::Exp2),
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -6428,15 +6428,36 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                "exp", "exp2"
+            };
+            // Complex conditions because of functions which are in both modules
+            bool is_math_imported = current_scope->resolve_symbol("math") != nullptr;
+            bool is_numpy_imported = current_scope->resolve_symbol("numpy") != nullptr;
+            bool is_math_module = module_math.find(call_name)!= module_math.end();
+            bool is_numpy_module = module_numpy.find(call_name)!= module_numpy.end();
+            if (( is_math_module && !is_numpy_module && !is_math_imported ) ||
+                ( is_numpy_module && !is_math_module && !is_numpy_imported ) ||
+                ( is_math_module && is_numpy_module && !is_math_imported && !is_numpy_imported )) {
```
Thirumalai-Shaktivel, line 6444: The following seems to work, can you please check what's going on here?
```py
from math import exp
def main0():
    x: f64 =  1.0
    print(sin(x))

main0()
```

gptsarthak: Actually error handling with imported modules is something I'm having trouble with. I could not exactly understand the code related to it so came up with this. Can you help me out with this and https://github.com/lcompilers/lpython/pull/1712#discussion_r1175426581 ? How should I approach this?
What we have to do is raise an error if ```from math import exp``` or ```from numpy import exp2``` is not present, for both functions that are unique to these module (like ```exp2``` in numpy) and ones that are common to them. Also functions imported from ```math``` should only accept scalars, but I hope that I implemented that error correctly.

Thirumalai-Shaktivel:
```py
from math import sin
```
The above generates the following ASR:
<details>
<summary>ASR</summary>

```Clojure
(TranslationUnit
    (SymbolTable
        1
        {
            _global_symbols:
                (Module
                    (SymbolTable
                        116
                        {
                            sin:
                                (ExternalSymbol
                                    116
                                    sin
                                    3 sin
                                    math
                                    []
                                    sin
                                    Public
                                )
                        })
                    _global_symbols
                    [math]
                    .false.
                    .false.
                ),
            main_program:
                (Program
                    (SymbolTable
                        115
                        {

                        })
                    main_program
                    []
                    []
                ),
            math:
                (Module
                    ...
                )
        })
    []
)
```
</details>

As you can see the ExternalSymbol is created for the `Sin` in the scope, if it is not present then you can throw an error.
Also, you can see the `module_name` (math) in the ExternalSymbol. you can use it to choose the modules like `math` or `numpy`.

Thirumalai-Shaktivel: You can use the same approach for functions accepts scalar or vectors as arguments.

Any issues, let me know.

Thirumalai-Shaktivel:
> I dont know why tests are failing here, integration tests pass on my system.

The failure is in C backend, Use `./run_tests.py -b c` to test or reproduce it.

Thirumalai-Shaktivel:
LGTM!

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -536,6 +538,71 @@ namespace Abs {
...
+namespace Exp {
+
+    static ASR::expr_t* eval_Exp(Allocator &al, const Location &loc,
+            Vec<ASR::expr_t*> &args) {
+        LCOMPILERS_ASSERT(ASRUtils::all_args_evaluated(args));
+        ASR::expr_t* arg = args[0];
+        ASR::ttype_t* t = ASRUtils::expr_type(args[0]);
+        if (ASRUtils::is_real(*t)) {
+            double val = ASR::down_cast<ASR::RealConstant_t>(arg)->m_r;
+            return EXPR(ASR::make_RealConstant_t(al, loc, val, t));
```
Thirumalai-Shaktivel, line 550: Here, we have to compute the value or evaluate the compile time value, right?

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -536,6 +538,71 @@ namespace Abs {
...
 
+namespace Exp {
+
+    static ASR::expr_t* eval_Exp(Allocator &al, const Location &loc,
+            Vec<ASR::expr_t*> &args) {
+        LCOMPILERS_ASSERT(ASRUtils::all_args_evaluated(args));
+        ASR::expr_t* arg = args[0];
+        ASR::ttype_t* t = ASRUtils::expr_type(args[0]);
+        if (ASRUtils::is_real(*t)) {
+            double val = ASR::down_cast<ASR::RealConstant_t>(arg)->m_r;
```
Thirumalai-Shaktivel, line 549:
```suggestion
        ASR::ttype_t* t = ASRUtils::expr_type(args[0]);
        if (ASRUtils::is_real(*t)) {
            double val = ASR::down_cast<ASR::RealConstant_t>(arg[0])->m_r;
```

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -536,6 +538,71 @@ namespace Abs {
...
+namespace Exp2 {
+
+    static ASR::expr_t* eval_Exp2(Allocator &al, const Location &loc,
+            Vec<ASR::expr_t*> &args) {
+        LCOMPILERS_ASSERT(ASRUtils::all_args_evaluated(args));
+        ASR::expr_t* arg = args[0];
+        ASR::ttype_t* t = ASRUtils::expr_type(args[0]);
+        if (ASRUtils::is_real(*t)) {
+            double val = ASR::down_cast<ASR::RealConstant_t>(arg)->m_r;
+            return EXPR(ASR::make_RealConstant_t(al, loc, val, t));
```
Thirumalai-Shaktivel, line 583: Make the above changes here as well

integration_tests/test_math_03.py
```diff
@@ -1,4 +1,5 @@
-from math import (cbrt, exp2)
+from math import cbrt
+from numpy import exp2
```
Thirumalai-Shaktivel, line 2: It seems CPython 3.11 supports exp2, see: https://github.com/lcompilers/lpython/pull/1594

You can add new tests for exp and exp2 in this or a new file.

certik:
Just do what the error is saying:
```
/Users/runner/work/lpython/lpython/integration_tests/_lpython-tmp-test-c/elemental_08.c:100:36: error: implicitly declaring library function 'exp' with type 'double (double)' [-Werror,-Wimplicit-function-declaration]
        ASSERT(_lcompilers_abs_f32(exp(array->data[(i - array->dims[0].lower_bound)]) - result->data[(i - result->dims[0].lower_bound)]) <= eps);
                                   ^
/Users/runner/work/lpython/lpython/integration_tests/_lpython-tmp-test-c/elemental_08.c:100:36: note: include the header <math.h> or explicitly provide a declaration for 'exp'
```
I think we need to inlude `math.h` in the generated file.

gptsarthak:
I believe this PR is ready. Let me know if there's anything else that needs to be done.

src/runtime/math.py
gptsarthak: I cant think of a way to solve this in ```IntrinsicFunctionRegistry``` itself, without changing ```asr_to_llvm```.
Should we do we do something like this?
```cpp
llvm::Value* expm1(llvm::Value* x) {
    llvm::Value* exp_x = builder->CreateUnaryIntrinsic(llvm::Intrinsic::exp, x);
    llvm::Value* one = llvm::ConstantFP::get(builder->getFloatTy(), 1.0);
    llvm::Value* result = builder->CreateFSub(exp_x, one);
    return result;
}
```
If there is too much code in  ```IntrinsicFunctionRegistry```, maybe I can create a macro for ```exp()``` like you did for trigonometric functions.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -6548,14 +6552,20 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
-             && not_cpython_builtin.find(call_name) == not_cpython_builtin.end()) {
+            if (ASRUtils::IntrinsicFunctionRegistry::is_intrinsic_function(call_name) &&
+                (not_cpython_builtin.find(call_name) == not_cpython_builtin.end() ||
+                imported_functions.find(call_name) != imported_functions.end() )) {
                 ASRUtils::create_intrinsic_function create_func =
                     ASRUtils::IntrinsicFunctionRegistry::get_create_function(call_name);
                 Vec<ASR::expr_t*> args_; args_.reserve(al, x.n_args);
                 visit_expr_list(x.m_args, x.n_args, args_);
+                if (ASRUtils::is_array(ASRUtils::expr_type(args_[0])) && 
+                    current_scope->resolve_symbol("numpy") == nullptr ) {
```
certik, line 6565: this seems to be the wrong check, I think it doesn't matter if `import numpy` is present, but rather if `exp` is imported from numpy.

src/libasr/codegen/asr_to_c.cpp
```diff
@@ -585,6 +585,7 @@ R"(
 #include <stdio.h>
 #include <string.h>
 #include <lfortran_intrinsics.h>
+#include <math.h>
```
certik, line 588: I think this should be only included when needed, I think there is a mechanism for that, based on the intrinsic function.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3994,6 +3994,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
     std::map<std::string, std::tuple<int64_t, bool, Location>> goto_name2id;
     int64_t gotoids;
     std::vector<ASR::symbol_t*> do_loop_variables;
+    std::set<std::string> imported_functions;
```
certik, line 3997: What is the purpose of this variable? Let's document it as a comment.

Is it to keep track which functions were imported? Don't we need to also track which modules they were imported from?

certik:
Thanks, I think it looks pretty good overall, there are just some smaller things to resolve.

src/lpython/semantics/python_ast_to_asr.cpp
certik: @czgdp1807 we already have this issue with `sin` and other functions I think, so we can probably fix this in a separate PR.

gptsarthak:
Changes:

- Implemented expm1() as a IntrinsicFunction
- Added a macro to create namespace for exponent related functions in ```intrinsic_function_registry```. This avoids repetition of code.
- Modified C backend to import ```math``` header only for IntrinsicFunction. 
- Added a map structure to keep track of imported functions and the name of the modules they are imported from.
- Error is thrown if function imported from ```math``` module is having a vector parameter.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -536,6 +539,39 @@ namespace Abs {
...
 
+#define create_exp_macro(X)                                                               \
+namespace X {                                                                             \
+    static inline ASR::expr_t* eval_##X(Allocator &al, const Location &loc,               \
+            Vec<ASR::expr_t*> &args) {                                                    \
+        LCOMPILERS_ASSERT(ASRUtils::all_args_evaluated(args));                            \
+        ASR::ttype_t* t = ASRUtils::expr_type(args[0]);                                   \
+        if (ASRUtils::is_real(*t)) {                                                      \
+            double val = ASR::down_cast<ASR::RealConstant_t>(args[0])->m_r;               \
+            return EXPR(ASR::make_RealConstant_t(al, loc, val, t));                       \
```
Thirumalai-Shaktivel, line 550: Here you are just extracting the value and returning the same thing. Instead, we have to compute the compile time value. 
For example, if you have extracted the value, compute `exp()` or `exp2()` or `expm1` of that value and then return it.
See: https://github.com/lcompilers/lpython/blob/268cec63c8c0a4efc8dc05876ae727106a4623a3/src/libasr/pass/intrinsic_function_registry.h#L306-L308

gptsarthak: Thanks for pointing it out, fixed it in the latest commit.

Thirumalai-Shaktivel:
LGTM!

czgdp1807 (approved):
Let's wait for #1673 to be merged.

PR #1616, opened 2023-03-24
Added IntrinsicFunction in grammar

anutosh491:
As I've been working on https://github.com/lcompilers/lpython/issues/1607#issue-1636596553 , which requires LPython to support `IntrinsicFunction` as `expr` . It seems it would also help us out at other places like https://github.com/lcompilers/lpython/pull/1606#pullrequestreview-1355786110  .

PR #1718, opened 2023-04-19
Try to register `Ord`  as IntrinsicFunction

faze-geek:
This is an initial implementation of the intrinsic ordinal function - `Ord()`.
As only compile time constants are handled on this branch -
```
def f():
    print(ord('A'))
    print(ord('2'))
    print(ord('z'))
    print(ord('!'))
f()

65
50
122
33

65
50
122
33
```
This is incomplete, I will add reference tests later on.


faze-geek:
Hey @Thirumalai-Shaktivel, made the pr as you suggested. We can collectively debug now -

My doubts from this [commit](https://github.com/lcompilers/lpython/pull/1616/commits/6920f6d6343c3acaf40272cfc59509b976532e58) for registration of `IntrinsicAbs()` -
What does the` instantiate_func()` do? How does it handle the runtime cases -

```
def f(x:str):
    print(ord(x))

f('a')
f('2')

97
50

^C
```

The case hangs with my current implementation (same for `--show-asr` and -`-show-llvm`) so probably a null pointer is returned.

faze-geek:
@Pranavchiku this pull request has some modifications since what we had discussed yesterday. Let's have our further discussions here. Kindly go through it once !

czgdp1807 (commented):
This looks okay to me logic wise. However some clean ups should be done.

czgdp1807:
The CI fails due to some Windows related to issues.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -463,6 +464,94 @@ namespace Abs {
...
+        if (ASRUtils::is_character(*type)) {
+            char* c = ASR::down_cast<ASR::StringConstant_t>(ASRUtils::expr_value(args[0]))->m_s;
+            if ((ASRUtils::expr_value(args[0]) != nullptr) && std::string(c).size() != 1) {
+                    err("ord() is only supported for String of length 1", args[0]->base.loc);
+                }
+        }
+        if (!ASRUtils::is_character(*type)) {
+            err("Argument of the ord function must be String of length 1",
+                args[0]->base.loc);
+        }
```
Thirumalai-Shaktivel, line 502: Move this `if check` above the another `if check`.

Thirumalai-Shaktivel, line 502: Move this `if` above the other `if` check
```suggestion
        if (!ASRUtils::is_character(*type)) {
            err("`x` argument of `ord()` must be a string",
                args[0]->base.loc);
        }
```
Or something like that and check for the compile time value

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -463,6 +464,94 @@ namespace Abs {
...
+        if (args.size() != 1) {
+            err("Intrinsic ord function accepts exactly 1 argument", loc);
+        }
+        ASR::ttype_t *type = ASRUtils::expr_type(args[0]);
+        if (ASRUtils::is_character(*type)) {
+            char* c = ASR::down_cast<ASR::StringConstant_t>(ASRUtils::expr_value(args[0]))->m_s;
+            if ((ASRUtils::expr_value(args[0]) != nullptr) && std::string(c).size() != 1) {
+                    err("ord() is only supported for String of length 1", args[0]->base.loc);
+                }
+        }
```
Thirumalai-Shaktivel, line 498: Here don't check for character type instead check for StringConstant and length of the character
For `StringConstant` the `expr_value` returns `StringConstant` itself, so instead check that args[0] is a StringConstant.

Also, as this PR only compute value for compile-time value, throw error for runtime input.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -463,6 +464,94 @@ namespace Abs {
...
 } // namespace Abs
 
+namespace Ord{
+    static ASR::expr_t* eval_Ord(Allocator &al, const Location &loc,
+            Vec<ASR::expr_t*> &args) {
+        LCOMPILERS_ASSERT(ASRUtils::all_args_evaluated(args));
+        ASR::ttype_t *type = ASRUtils::TYPE(ASR::make_Integer_t(al, loc, 4, nullptr, 0));
+        ASR::expr_t *expr = args[0];
+        ASR::ttype_t *t = ASRUtils::expr_type(expr);
+        if (ASRUtils::is_character(*t)) {
```
Thirumalai-Shaktivel, line 474: Wait a minute, this seems to be repeated. I think we can compute the value in the create_Ord itself.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -463,6 +464,94 @@ namespace Abs {
...
+        body.reserve(al, 1);
+        SetChar dep;
+        dep.reserve(al, 1);
+        /*
+         if (c is character) then
+            r = ord(c)
+        */
+        if (is_character(*arg_types[0])) {
+            ASR::expr_t *res;
+            res = EXPR(ASR::make_IntegerConstant_t(al, loc, int64_t(args[0]), return_type));
```
Thirumalai-Shaktivel, line 543: This doesn't seem to work as expected.
I think, the compile-time value cannot be evaluated in the instantiate_ord.

Thirumalai-Shaktivel:
@faze-geek Can you please rebase or merge main into this branch?
I added some changes to LPython from LFortran today morning.

Also, after thinking about only compile_time support for ord as you mentioned above is not possible. We have to implement both compile and runtime for `InstrinsicFunction`. I have left comments; if you have any doubts, just ping me.

src/libasr/pass/intrinsic_function_registry.h
Thirumalai-Shaktivel: You can check, how partition is handled in this PR: https://github.com/lcompilers/lpython/pull/1673

faze-geek:
Thanks @Thirumalai-Shaktivel for the set of reviews, I will go through them soon. First I'll go through https://github.com/lcompilers/lpython/pull/1673 and test it out once.

src/libasr/pass/intrinsic_function_registry.h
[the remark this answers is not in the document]
faze-geek: Removed the if statement. `IntrinsicAbs()` also has an unnecessary check, I'll modify it later.

faze-geek: Yess I followed the norm here, all possible error checking in `create_Ord`  abd then the compile time calculation in `eval_Ord`.

Issue #1728, opened 2023-04-21
sizeof(x) doesn't work for external symbol `x`


certik:
```
semantic error: Symbol 5 is not yet supported in sizeof.
   --> a.py:313:21
    |
313 |             ctx_id, sizeof(common_dev_host),
    |                     ^^^^^^^^^^^^^^^^^^^^^^^ 
```

certik:
This is partially fixed by https://github.com/lcompilers/lpython/pull/1729.

PR #1729, opened 2023-04-21
Make sizeof() work for external symbols

certik:
Fixes #1728.

PR #1730, opened 2023-04-21
Support imported ``StructType`` in ``sizeof``

czgdp1807:
Fixes https://github.com/lcompilers/lpython/issues/1728
https://github.com/lcompilers/lpython/pull/1729#pullrequestreview-1395080556

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -6567,13 +6567,19 @@ public:
                 ASR::ttype_t* arg_type = nullptr;
                 if( ASR::is_a<ASR::Var_t>(*args[0].m_value) ) {
                     ASR::Var_t* arg_Var = ASR::down_cast<ASR::Var_t>(args[0].m_value);
-                    if( ASR::is_a<ASR::Variable_t>(*arg_Var->m_v) ) {
-                        arg_type = ASR::down_cast<ASR::Variable_t>(arg_Var->m_v)->m_type;
-                    } else if( ASR::is_a<ASR::StructType_t>(*arg_Var->m_v) ) {
+                    ASR::symbol_t* arg_Var_m_v = ASRUtils::symbol_get_past_external(arg_Var->m_v);
+                    if( ASR::is_a<ASR::Variable_t>(*arg_Var_m_v) ) {
+                        // TODO: Import the underlying struct if arg_type is of Struct type
+                        // Ideally if a variable of struct type is being imported then its underlying type
+                        // should also be imported automatically. However, the naming of the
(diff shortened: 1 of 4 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -6567,13 +6567,19 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
-                        arg_type = ASR::down_cast<ASR::Variable_t>(arg_Var->m_v)->m_type;
-                    } else if( ASR::is_a<ASR::StructType_t>(*arg_Var->m_v) ) {
+                    ASR::symbol_t* arg_Var_m_v = ASRUtils::symbol_get_past_external(arg_Var->m_v);
+                    if( ASR::is_a<ASR::Variable_t>(*arg_Var_m_v) ) {
+                        // TODO: Import the underlying struct if arg_type is of Struct type
+                        // Ideally if a variable of struct type is being imported then its underlying type
+                        // should also be imported automatically. However, the naming of the
+                        // underlying struct type might lead to collisions, so importing the type
+                        // here seems like a better choice. Should be done later when the case arises.
+                        arg_type = ASR::down_cast<ASR::Variable_t>(arg_Var_m_v)->m_type;
```
certik, line 6577: This is related to https://docs.lfortran.org/en/asr/asr_nodes/expression_nodes/Var/. We already allow Var to point to a Function (symbol node), so that we can assign the function to a function pointer. This is almost identical: allow the structure itself to be passed as an argument to sizeof(). So Var could also point to Struct (=what we currently call StructType). We'll add StructType as a type, just like FunctionType.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -6567,13 +6567,19 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                    } else if( ASR::is_a<ASR::StructType_t>(*arg_Var_m_v) ) {
```
certik, line 6578: Based on my previous comment, this shouldn't even pass verify(). I guess it does in LPython, but will not once we sync ASR with LFortran, where I improved the pass. But we'll simply allow Struct as well in verify.

czgdp1807: Could you please share the PR where you improved the ASR verify pass? I will apply those changes and test right away.

certik: Hm, another question: what type is the Var, or the sizeof argument, if it points to Struct (=StructType currently)? The type is the "struct class itself, not an instance". After refactor, for the instance of a Struct has a type StructType. But what type does Struct itself have?

Issue #1613, opened 2023-03-23
ccall: allow to use a header file


certik:
Right now, every `@ccall` always generates a C forward declaration (in the C backend). In addition, we would also like to have a mode, something like:
```py
@ccall(header="some_header.h")
def f(x: i32) -> f64:
    pass

# later
e = f(5)
```
Which would generate the following code:
```c
#include "some_header.h"

...
e = f(5)
```
and it would *not* forward declare `f`. So in particular, if `f` is actually a C macro declared in the `some_header.h` file, it would also just work.

Alternative syntax:
```py
from ltypes import headers

h = headers("some_header.h")
@ccall(h)
def dummy():
    pass

@ccall
def f(x: i32) -> f64:
    pass

# later
e = f(5)
```

We might need to figure out the best design, but we need this feature.

certik:
To simplify, one idea might be:
```
with ltypes.header("some_header.h"):
    def f(x: i32) -> f64:
        pass

    def g(x: i32) -> f64:
        pass
```

PR #1731, opened 2023-04-21
C: Support header files in `ccall`

Smit-create:


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3425,6 +3425,35 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                        if (name == "ccall") {
+                            current_procedure_abi_type = ASR::abiType::BindC;
+                            current_procedure_interface = true;
+                            if (call_d->n_keywords > 0) {
+                                for (size_t i=0; i < call_d->n_keywords; i++) {
+                                    if (std::string(call_d->m_keywords[i].m_arg) == "header") {
+                                        if (AST::is_a<AST::ConstantStr_t>(*call_d->m_keywords[i].m_value)) {
+                                            std::string header_name = AST::down_cast<AST::ConstantStr_t>(
+                                                        call_d->m_keywords[i].m_value)->m_value;
+                                            bindc_name = s2c(al, header_name);
```
Smit-create, line 3441: I'm using `bindc_name` to test the prototype and use it as a header file name which in LLVM is a function name to be called from C. Once this works, do we want to add a new argument in the function type to check which header file it is contained in?

integration_tests/bindc_05b.h
```diff
@@ -1,2 +1,7 @@
+#ifndef BINDC_05BH
+#define BINDC_05BH
```
czgdp1807, line 2: What if the user doesn't define these guard macros in their header file? Probably then using `@ccall(header="header_name")` twice will give errors. I feel https://github.com/lcompilers/lpython/issues/1613#issuecomment-1481903888 syntax is better.

certik: We need to ensure a header file is included exactly once.

src/lpython/semantics/python_ast_to_asr.cpp
certik: I think we do. Probably not in FunctionType but in Function.

integration_tests/bindc_05b.h
Smit-create: > What if the user doesn't define these guard macros in their header file?

C/CPP uses a `set` to store the headers, so that won't be an issue.https://github.com/lcompilers/lpython/blob/b28a44046c9f375ce6b8f39b00226cdb36e1ccfe/src/libasr/codegen/asr_to_c_cpp.h#L114

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: I don't think Function has a mechanism to store the c header file. It just has bindc_name for its interop with C.
```
| Function(symbol_table symtab, identifier name, ttype function_signature,
        identifier* dependencies, expr* args, stmt* body, expr? return_var,
        access access, bool deterministic, bool side_effect_free)
```

Should we go with this?
```diff
diff --git a/src/libasr/ASR.asdl b/src/libasr/ASR.asdl
index 34af5be3c..3f4b29b4a 100644
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -363,7 +363,7 @@ ttype
     | CPtr()
     | TypeParameter(identifier param, dimension* dims)
     | FunctionType(ttype* arg_types, ttype? return_var_type,
-        abi abi, deftype deftype, string? bindc_name, bool elemental,
+        abi abi, deftype deftype, string? bindc_name, string? c_header, bool elemental,
         bool pure, bool module, bool inline, bool static, ttype* type_params,
         symbol* restrictions, bool is_restriction)
```

certik:
Thanks a lot @Smit-create for working on this! It looks like I found a workaround with https://github.com/lcompilers/lpython/pull/1732 and https://github.com/lcompilers/lpython/pull/1734 so that I am not blocked today anymore. This buys us time to design this well. We definitely need this feature.

src/libasr/ASR.asdl
```diff
@@ -363,7 +363,7 @@ ttype
     | CPtr()
     | TypeParameter(identifier param, dimension* dims)
     | FunctionType(ttype* arg_types, ttype? return_var_type,
-        abi abi, deftype deftype, string? bindc_name, bool elemental,
+        abi abi, deftype deftype, string? bindc_name, string? c_header, bool elemental,
```
certik, line 366: `c_header` I think should go into `Function`, not be part of the type (neither logical nor physical: https://github.com/lfortran/lfortran/issues/1601).

integration_tests/bindc_05b.h
certik: There is no issue.

certik (approved):
With those two changes this should be fine.

tests/reference/c-expr7-bb2692a.stdout
```diff
@@ -1,6 +1,6 @@
-#include <complex.h>
-#include <inttypes.h>
-#include <math.h>
+#include "complex.h"
+#include "inttypes.h"
+#include "math.h"
```
certik, line 3: I think this change is a problem, it should use `<`, not `"`.

certik (commented):
There is one more thing that we should fix.

src/libasr/codegen/asr_to_c.cpp
```diff
@@ -713,8 +713,8 @@ R"(
             }
         }
         std::string to_include = "";
-        for (auto s: headers) {
-            to_include += "#include <" + s + ".h>\n";
+        for (auto &s: headers) {
+            to_include += "#include \"" + s + "\"\n";
```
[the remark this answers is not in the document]
Smit-create: Done, thanks!

tests/reference/c-expr7-bb2692a.stdout
Smit-create: Yep, thanks. Reverted that.

certik (approved):
I think this is good enough now.

I noticed there are some default header files and macros always included and we should eventually handle them in the same way (put the macros into a header file and only include it if needed), but that's for later.

I think this is good to go in. Thanks for this @Smit-create, very important new feature.

Issue #1727, opened 2023-04-21
Function dependencies not respected with --disable-main


certik:
import_order_01.py:
```python
from import_order_01b import f

@ccallback
def main1():
    a: i32 = f()
    print(a)
```

import_order_01b.py:
```python
from lpython import i32

def f() -> i32:
    return 42
```
Compile using:
```console
$ lpython -I. --show-c --disable-main import_order_01.py
...

// Implementations
void main1()
{
    int32_t a;
    a = f();
    printf("%d\n", a);
}

int32_t f()
{
    int32_t _lpython_return_variable;
    _lpython_return_variable = 42;
    return _lpython_return_variable;
}
```
This then fails to compile due to:
```console
$ clang x.c                
x.c:41:9: error: implicit declaration of function 'f' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    a = f();
        ^
1 error generated.
```
The issue is that `main1` depends on `f`, but `f` gets generated below the function. It only happens when `--disable-main` is used, but we need to use it.

certik:
I am not sure how to best test this in integration tests, due to the `--disable-main`. The `@ccallback` is needed, otherwise the function gets treated as unused.

czgdp1807:
The `--disable-main` doesn't matter. Even without it, `f()` goes below `main1`,

```c
...

// Implementations
void main1()
{
    int32_t a;
    a = f();
    printf("%d\n", a);
}

int32_t f()
{
    int32_t _lpython_return_variable;
    _lpython_return_variable = 42;
    return _lpython_return_variable;
}

int main(int argc, char* argv[])
{
    return 0;
}
```

czgdp1807:
Its a problem with `pass_wrap_global_stmts_into_program`. It only shifts user defined functions into `_global_symbols` (generated module) only when they are called. In your case `main1` is never called so it stays in the global scope. Now C backend first visits all the functions in global scope and since `main1` lies in global scope so C code of it is generated first. Then `f` is generated when the modules are visited. If `main1` is shifted to `_global_symbols` then it depends on `import_order_01` and hence `f` will always be generated before `main1` (which is correct).

End moral - So, IMO global scope should only contain intrinsic functions generated by `libasr` in `intrinsic_function_registry.h`. **All user defined functions should be shifted to `_global_symbols` modules irrespective of the fact whether they are called by the user or not.**

PR #1732, opened 2023-04-21
Always insert global symbols to`` _global_symbols`` module

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1727
https://github.com/lcompilers/lpython/issues/1727#issuecomment-1517366001

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -203,6 +203,72 @@ macro(RUN)
     endif()
 endmacro(RUN)
 
+# only compiles till object file
+# to make sure that the generated code is syntactically correct
+# but we cannot generate an executable due to --disable-main option enabled.
+macro(COMPILE)
+    set(options FAIL)
+    set(oneValueArgs NAME IMPORT_PATH)
+    set(multiValueArgs LABELS EXTRAFILES)
+    cmake_parse_arguments(COMPILE "${options}" "${oneValueArgs}"
+                          "${multiValueArgs}" ${ARGN} )
+    set(name ${COMPILE_NAME})
+    set(import_path ${COMPILE_IMPORT_PATH})
+    if (NOT name)
+        message(FATAL_ERROR "Must specify the NAME argument")
+    endif()
+
+    if (${KIND} IN_LIST COMPILE_LABELS)
+        if (KIND STREQUAL "llvm")
+            if (import_path)
+                add_custom_command(
+                    OUTPUT ${name}.o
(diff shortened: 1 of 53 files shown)


Thirumalai-Shaktivel (approved):
LGTM! Thanks for fixing this
So, from now on, we will always have `two symbols` in the `TranslationUnit`. If the program uses an `IntrinsicFunction` then the `TranslationUnit` also includes that `function` in the `global scope`.

Empty program:
```Clojure
(TranslationUnit
    (SymbolTable
        1
        {
            _global_symbols:
                (Module
                    (SymbolTable
                        3
                        {
                            
                        })
                    _global_symbols
                    []
                    .false.
                    .false.
                ),
            main_program:
                (Program
                    (SymbolTable
                        2
                        {
                            
                        })
                    main_program
                    []
                    []
                )
        })
    []
)
```

czgdp1807:
Thanks for pointing out. I don't create `_global_symbols` now if it global scope doesn't have any user defined symbols. Now the behaviour will be same as `main` branch when the input file is empty.

certik (approved):
I think it looks good, thanks for fixing it!

certik:
Hm, this PR broke `--disable-main`, it now always inserts `main`:
```console
(lp) lpython/integration_tests(main) $ lpython --show-c --disable-main import_order_01.py -I.
#include <inttypes.h>

#include <stdlib.h>
#include <stdbool.h>
#include <stdio.h>
#include <string.h>

#define ASSERT(cond)                                                           \
    {                                                                          \
        if (!(cond)) {                                                         \
            printf("%s%s", "ASSERT failed: ", __FILE__);                       \
            printf("%s%s", "\nfunction ", __func__);                           \
            printf("%s%d%s", "(), line number ", __LINE__, " at \n");          \
            printf("%s%s", #cond, "\n");                                       \
    }
#define ASSERT_MSG(cond, msg)                                                  \
    {                                                                          \
        if (!(cond)) {                                                         \
            printf("%s%s", "ASSERT failed: ", __FILE__);                       \
            printf("%s%s", "\nfunction ", __func__);                           \
            printf("%s%d%s", "(), line number ", __LINE__, " at \n");          \
            printf("%s%s", #cond, "\n");                                       \
            printf("%s", "ERROR MESSAGE:\n");                                  \
            printf("%s%s", msg, "\n");                                         \
    }


struct dimension_descriptor
{
    int32_t lower_bound, length;
};

// Implementations
int32_t f()
{
    _lpython_return_variable = 42;
}

void main1()
{
    int32_t a;
    a = f();
    printf("%d\n", a);
}

{
    return 0;
}
```
Hopefully it can be fixed with a subsequent PR, otherwise we'll need to revert it.

czgdp1807:
Let me see. Probably a corner case.

PR #1734, opened 2023-04-21
Respect ``--disable-main`` in global_stmts pass

czgdp1807:
https://github.com/lcompilers/lpython/pull/1732#issuecomment-1518121655

PR #1735, opened 2023-04-21
Ensure the main is not present for --disable-main

certik:
This fixes a regression introduced by #1732.


certik:
The fix in https://github.com/lcompilers/lpython/pull/1734 is better, so this PR should not be merged, I am posting it here only because I tested it with our production code and it works.

certik:
Closing in favor of https://github.com/lcompilers/lpython/pull/1734.

PR #1738, opened 2023-04-21
Fix a test failure on linux

certik:
This fails in CPython, it was not able to achieve the prescribed accuracy. Lowering the accuracy fixed the problem.


certik:
Previously it was giving this error on Linux for me:
```console
$ PYTHONPATH=../src/runtime/lpython python elemental_01.py
  File "/home/tron1/ondrej/repos/lpython/integration_tests/elemental_01.py", line 166, in <module>
    elemental_mul()
  File "/home/tron1/ondrej/repos/lpython/integration_tests/elemental_01.py", line 89, in elemental_mul
    verify1d_mul(array_a, array_b, array_c, 100)
  File "/home/tron1/ondrej/repos/lpython/integration_tests/elemental_01.py", line 51, in verify1d_mul
    assert abs(array_a[i]**2.0 * 5.0*array_b[i]**3.0 - result[i]) <= eps
AssertionError
```
This error does not happen at the CI, but I am not sure what the issue is. I printed the numbers:
```console
0.0 71.0 76.0 11064390080.0
```
And they are getting large, so there might be some numerical change, such as using NumPy power vs Python power, or something like that.

Smit-create (commented):
It's fine. Thanks!

PR #1739, opened 2023-04-21
PIP: Support using packages installed via pip

Shaikh-Ubaid:
This PR adds initial support of using pip installed packages with lpython. The python packages uploaded to `pypi` are currently here https://github.com/Shaikh-Ubaid/lpython_packages.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -184,3 +184,63 @@ jobs:
             which node
             node -v
             node src/lpython/tests/test_lpython.js
+
+  test_pip_pkgs:
+    name: Test PIP Installable Packages
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - uses: mamba-org/provision-with-micromamba@v15
+        with:
+          environment-file: ci/environment.yml
+          extra-specs: |
+            bison=3.4
+
+      - uses: hendrikmuhs/ccache-action@main
+        with:
+          variant: sccache
+          key: ${{ github.job }}-${{ matrix.os }}
+
+      - name: Setup Platform (Linux)
+        shell: bash -l {0}
+        run: |
+            echo "LFORTRAN_CMAKE_GENERATOR=Unix Makefiles" >> $GITHUB_ENV
+            echo "WIN=0" >> $GITHUB_ENV
+            echo "MACOS=0" >> $GITHUB_ENV
+            echo "ENABLE_RUNTIME_STACKTRACE=yes" >> $GITHUB_ENV
+
+      - name: Build Linux
+        shell: bash -l {0}
+        run: |
+            xonsh ci/build.xsh
+
+      - name: PIP show version
+        shell: bash -l {0}
+        run: |
+            python -m pip -V
+
+      - name: PIP install required packages
+        shell: bash -l {0}
+        run: |
+            python -m pip install lpynn numpy
+
+      - name: PIP show installed packages
+        shell: bash -l {0}
+        run: |
+            python -m pip list
+
+      - name: Test PIP Packages with Python
+        shell: bash -l {0}
+        run: |
+            export PYTHONPATH=./src/runtime/lpython
+            python integration_tests/test_pip_import_01.py
+
+      - name: Test PIP Packages with LPython
+        shell: bash -l {0}
+        run: |
+            pip_pkg_path=$(python -c "import site; print(site.getsitepackages()[0])")
+            echo $pip_pkg_path
+            ./src/bin/lpython integration_tests/test_pip_import_01.py -I $pip_pkg_path
--- a/integration_tests/test_pip_import_01.py
+++ b/integration_tests/test_pip_import_01.py
@@ -1,0 +1,45 @@
+from lpynn.perceptron import init_perceptron, print_perceptron, normalize_input_vectors, Perceptron, train_dataset
+from lpython import i32, f64
+
+def main0():
+    p: Perceptron = Perceptron(0, [0.0], 0.0, 0, 0.0, 0.0, 0)
+    init_perceptron(p, 2, 0.05, 10000, 90.0)
+    print_perceptron(p)
+    print("=================================")
+
+    input_vectors: list[list[f64]] = [[-1.0, -1.0], [-1.0, 1.0], [1.0, -1.0], [1.0, 1.0]]
+    outputs: list[i32] = [1, 1, 1, -1]
+
+    normalize_input_vectors(input_vectors)
+    train_dataset(p, input_vectors, outputs)
+    print_perceptron(p)
+    print("=================================")
+
+    assert p.cur_accuracy > 50.0
+    assert p.epochs_cnt > 1
+    assert abs(p.weights[0] - (-0.1)) < 1e-12
+    assert abs(p.weights[1] - (-0.1)) < 1e-12
+    assert abs(p.weights[2] - (0.1)) < 1e-12
+
+def main1():
+    p: Perceptron = Perceptron(0, [0.0], 0.0, 0, 0.0, 0.0, 0)
+    init_perceptron(p, 2, 0.05, 10000, 90.0)
+    print_perceptron(p)
+    print("=================================")
+
+    input_vectors: list[list[f64]] = [[-1.0, -1.0], [-1.0, 1.0], [1.0, -1.0], [1.0, 1.0], [1.5, 1.0]]
+    outputs: list[i32] = [1, 1, -1, 1, -1]
+
+    normalize_input_vectors(input_vectors)
+    train_dataset(p, input_vectors, outputs)
+    print_perceptron(p)
+    print("=================================")
+
+    assert p.cur_accuracy > 50.0
+    assert p.epochs_cnt > 1
+    assert abs(p.weights[0] - (-0.22)) < 1e-12
+    assert abs(p.weights[1] - (0.1)) < 1e-12
+    assert abs(p.weights[2] - (0.1)) < 1e-12
+
+main0()
+main1()
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -3746,7 +3746,7 @@ public:
             in any directory other than src/runtime will also
             be ignored.
         */
-        if (mod_sym == "lpython") {
+        if (mod_sym == "lpython" || mod_sym == "numpy") {
             return ;
         }
 
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -3751,15 +3751,6 @@ public:
         }
 
         /*
-            If import_path is not empty then insert it
-            in the second priority. Top priority path
-            is runtime library path.
-        */
-        for( auto& path: import_paths ) {
-            paths.push_back(path);
-        }
-
-        /*
             Search all the paths in order and stop
             when the desired module is found.
         */
@@ -3820,7 +3811,13 @@ public:
         if (!t) {
             std::string rl_path = get_runtime_library_dir();
             SymbolTable *st = current_scope;
-            std::vector<std::string> paths = {rl_path, parent_dir};
+            std::vector<std::string> paths;
+            for (auto &path:import_paths) {
+                paths.push_back(path);
+            }
+            paths.push_back(rl_path);
+            paths.push_back(parent_dir);
+
             if (!main_module) {
                 st = st->parent;
             }
@@ -3870,7 +3867,12 @@ public:
     void visit_Import(const AST::Import_t &x) {
         ASR::symbol_t *t = nullptr;
         std::string rl_path = get_runtime_library_dir();
-        std::vector<std::string> paths = {rl_path, parent_dir};
+        std::vector<std::string> paths;
+        for (auto &path:import_paths) {
+            paths.push_back(path);
+        }
+        paths.push_back(rl_path);
+        paths.push_back(parent_dir);
         SymbolTable *st = current_scope;
         std::vector<std::string> mods;
         for (size_t i=0; i<x.n_names; i++) {


Shaikh-Ubaid:
This is currently a WIP (Work In Progress). Therefore, the CI tests might not pass currently.

Shaikh-Ubaid:
I would like to request a quick review to know if I am proceeding in the expected/desired direction.

src/lpython/utils.cpp
```diff
@@ -102,6 +102,68 @@ std::string get_runtime_library_header_dir()
...
+    }
+
+    // If no site-packages directory was found, return an empty string
+    if (paths.empty()) {
+        return "";
+    }
+
+    // Return the first site-packages directory found
+    return paths[0];
+}
```
certik, line 164: This seems very fragile.

Why not for now just require the user to supply the correct `-I` flags?

certik:
Let's first get everything working with the `-I` flags. Then we can support Conda, if it can be done easily on all platforms (via `$CONDA_PREFIX`). We can also support an environment variable `LPYTHON_PYTHONPATH` or something like that, and the environment that supports LPython can then set it.

I would not try to run cpython from lpython, that seems fragile and it will be very hard or impossible to ensure it always succeeds, and it will also be slow. Rather, the parent environment that runs lpython can set `LPYTHON_PYTHONPATH`, for example `conda activate` can set it. But we can search `$CONDA_PREFIX` either automatically, or with a command line flag like `--search-active-conda-env`.

Shaikh-Ubaid:
>Let's first get everything working with the -I flags

It seems the `-I` flag works for many/most cases. One of the concerns I have is with respect to the priority of paths to be searched for a particular package/module. The current priority as per the current main branch is as follows:

- runtime library path
- current directory of the file under consideration (called as `parent_dir` in code).
- import paths (provided by the user)

An uncommon example could be where the user defines his own package which has the same name as one of the packages defined in the `runtime library`. In such a case, there is no possibility for the user to import his package as his package will never be found as the package from runtime library will always be picked up. If we provide `import paths` more priority than the runtime library path then the user can specify location of his package using the `-I` option and his package would be picked up instead of the one in runtime library.

Shaikh-Ubaid:
Also should the current directory of the file (called as parent directory in code) have the most priority as compared to runtime library or import paths?

certik:
I think `-I` should take priority over a runtime library.

I think if you do `lpython a.py`, we need to decide if it should automatically find a package "b" in the same directory. I know some Python distribution allow it, and some don't. I think inside a Python package, say once it can find "b", then inside "b", such as "b/x.py", inside `x.py` I think it can automatically find `b/y.py`. I think that's how Python works.

Regarding priority, let's test how Python works with regards to PYTHONPATH.

Shaikh-Ubaid:
>Then we can support Conda, if it can be done easily on all platforms (via $CONDA_PREFIX). We can also support an environment variable LPYTHON_PYTHONPATH or something like that, and the environment that supports LPython can then set it.

Yes, I considered these before using the current approach. 

It seems if we use the `CONDA_PREFIX`, then `lpython` seems to depend on `conda` for execution. Similarly, with the current approach in this PR, `lpython` is dependent on `cpython` for execution, which I think and as you shared is not good as it could be slow, also seems fragile. (Although using `cpython` does not seem good, I implemented it so that we discuss and figure a better approach.)

Shaikh-Ubaid:
>We can also support an environment variable LPYTHON_PYTHONPATH

Do you mean to set this variable in `CMakeLists.txt` (from root directory)? For example, if we can set it to appropriate location of `site-packages` during build time of `lpython`, we can then just use `std::getenv("LPYTHON_PYTHONPATH")` in the code to get the location of `site-packages`.

certik:
To be safe, we can require to enable CONDA_PREFIX support via a compiler option like `--search-active-conda-env`, that way we are not depending on anything. 

Yes, it's hard to figure out the right approach here. It needs to be robust and fast, and as easy to use as possible.

Shaikh-Ubaid:
By the way, this is what I found when consulting with ChatGPT about paths to be searched:

**Prompt:** Can you just provide me with a list of paths that needs to be searched when importing a python package? What would be the list of paths to be searched on windows, linux, macos?

**Response:**

Sure! Here are the paths that are searched by default on Windows, Linux, and macOS when importing a Python package:

Windows

1. The directory of the script or the current directory.
2. The directories specified in the PYTHONPATH environment variable.
3. The installation-dependent default directory.

Linux and macOS

1. The directory of the script or the current directory.
2. The directories specified in the PYTHONPATH environment variable.
3. The installation-dependent default directory.
4. The user-specific site-packages directory (~/Library/Python/<version>/site-packages on macOS and ~/.local/lib/python<version>/site-packages on Linux).
5. The system-wide site-packages directory (/usr/local/lib/python<version>/site-packages and /usr/lib/python<version>/site-packages on Linux, and /Library/Python/<version>/site-packages and /System/Library/Frameworks/Python.framework/Versions/<version>/Extras/lib/python on macOS).

Note that the exact paths may vary depending on the specific installation and configuration of Python on your system. Also, this list only covers the default search paths - it's possible to modify sys.path at runtime to add additional directories to the search path.

certik:
> > We can also support an environment variable LPYTHON_PYTHONPATH
> 
> Do you mean to set this variable in `CMakeLists.txt` (from root directory)? For example, if we can set it to appropriate location of `site-packages` during build time of `lpython`, we can then just use `std::getenv("LPYTHON_PYTHONPATH")` in the code to get the location of `site-packages`.

We decided for now not to use any environment variables like this, to prevent a similar issue with Python, that a user environment can break things easily. But we can enable it with a command line option, that way it's opt-in, but not enabled by default. Regarding cmake, are you talking about a user project that uses cmake and calls lpython to build things? Yes, that's one option, but I think just supplying the correct `-I` is more robust.

Shaikh-Ubaid:
>We decided for now not to use any environment variables like this, to prevent a similar issue with Python, that a user environment can break things easily. But we can enable it with a command line option, that way it's opt-in, but not enabled by default.

Yes, got it.

Shaikh-Ubaid:
Upcoming:
- [ ] We need to get the location of `site-packages` directory or any other directory where `pip` installs packages.

Shaikh-Ubaid:
There is already package with name `lpython` on `pypi`, hence it is not available. For the name `ltypes`, `pypi` returns an error that `ltypes` is too similar to existing package. Hence, for uploading to pypi, I renamed the package to `lptypes`.

.github/workflows/CI.yml
```diff
@@ -184,3 +184,52 @@ jobs:
             which node
             node -v
             node src/lpython/tests/test_lpython.js
+
+  test_pip_pkgs:
```
Shaikh-Ubaid, line 188: I would like to possible have a separate test case at the CI for testing pip installable packages (for example, similar to how we have unique CI jobs for different backends at LFortran). So, that if there is any failure, it is clear from the jobs which failed if the failure is in pip installed package tests or in regular tests.

.github/workflows/CI.yml
```diff
@@ -184,3 +184,52 @@ jobs:
...
+        shell: bash -l {0}
+        run: |
+            python integration_tests/test_pip_import_01.py
+
+      - name: Test PIP Pacakges with LPython
+        shell: bash -l {0}
+        run: |
+            pip_pkg_path=$(python -c "import site; print(site.getsitepackages()[0])")
+            echo $pip_pkg_path
+            ./src/bin/lpython integration_tests/test_pip_import_01.py -I $pip_pkg_path
```
Shaikh-Ubaid, line 235: **Line 228 and 235:** I currently hardcoded the tests here. But this does not seem good/maintainable (enough). I would like to know how we could improve this.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3622,19 +3622,10 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
             in any directory other than src/runtime will also
             be ignored.
         */
-        if (mod_sym == "lpython") {
+        if (mod_sym == "lpython" || mod_sym == "lptypes" || mod_sym == "numpy") {
```
Shaikh-Ubaid, line 3625: I uploaded the `lpython.cpp` to `pypi` with the name `lptypes` (`lpython` is not available, also `ltypes` is not available). I would like to know if the name `lptypes` is fine or if we should use any other name.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3746,7 +3743,12 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
     void visit_Import(const AST::Import_t &x) {
         ASR::symbol_t *t = nullptr;
         std::string rl_path = get_runtime_library_dir();
-        std::vector<std::string> paths = {rl_path, parent_dir};
+        std::vector<std::string> paths;
+        for (auto &path:import_paths) {
+            paths.push_back(path);
+        }
+        paths.push_back(rl_path);
+        paths.push_back(parent_dir);
```
Shaikh-Ubaid, line 3751: This logic seems to repeat at two places, so we could probably wrap it into a function which returns a vector of strings. I would like to know if the current approach is fine or if we should wrap this code into a function.

integration_tests/test_pip_import_01.py
```diff
@@ -0,0 +1,39 @@
+from lpynn.perceptron import init_perceptron, print_perceptron, normalize_input_vectors, Perceptron, train_dataset
+from lptypes import i32, f64
```
Shaikh-Ubaid, line 2: The added test is such that both the packages it uses/needs are not available in the source code of `lpython`. (They are present in the `lpython` repository with the names `lnn` and `lpython`.) The above test makes sure that we are not using packages from any `lpython` project sub-directory and instead these are being used from the `pip` installed packages.

Shaikh-Ubaid:
>Let's first get everything working with the -I flags.

We are currently testing the `pip` installed packages using the `-I` flag in a separate `CI` job.

.github/workflows/CI.yml
```diff
@@ -184,3 +184,52 @@ jobs:
...
+        shell: bash -l {0}
+        run: |
+            xonsh ci/build.xsh
+
+      - name: PIP install required packages
+        shell: bash -l {0}
+        run: |
+            python -m pip install lptypes lpynn numpy
+
+      - name: Test PIP Pacakges with Python
```
certik, line 225:
```suggestion
      - name: Test PIP Packages with Python
```

.github/workflows/CI.yml
```diff
@@ -184,3 +184,52 @@ jobs:
...
+        shell: bash -l {0}
+        run: |
+            python integration_tests/test_pip_import_01.py
+
+      - name: Test PIP Pacakges with LPython
```
certik, line 230:
```suggestion
      - name: Test PIP Packages with LPython
```

.github/workflows/CI.yml
certik: We can worry about that later, I think it's fine for now.

integration_tests/test_pip_import_01.py
certik: The `lpynn` is fine. I would be careful with `lptypes`. I don't have an opinion right now, but I would be conservative and not allow it for now, and require that the special types that LPython understands are imported from "lpython" only, that is our own blessed version.

src/lpython/semantics/python_ast_to_asr.cpp
certik: I see, we need a CPython package, for compatibility. I would call the Python module lpython. For now I would distribute it with the "lpython" compiler, say the conda package. For pypi, maybe "lpython_emulation" pypi package would work? It would still install lpython.py.

certik (commented):
I left some comments. In general this looks great, thanks for working on this. This is very useful and once it works, it will allow us to have regular pure Python packages that just work with CPython via pip, but also just work with LPython without any modification.

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: Got it. 

One concern:
>For now I would distribute it with the "lpython" compiler, say the conda package

Every package that we develop uses the `i32/i64/...` types from `lpython`. Do you mean to package the `lpython.py` and `goto.py` (since `lpython.py` imports `with_goto` from `goto.py`) with every package that we upload to `pypi` (and `conda`)?

Since every package almost always uses types from `lpython`, I think it would not work if its dependencies are not included or the dependencies do not already exist on `pypi`.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -202,8 +202,8 @@ Result<std::string> get_full_path(const std::string &filename,
         return file_path;
     } else {
         // If this is `lpython`, do a special lookup
-        if (filename == "lpython.py") {
-            file_path = runtime_library_dir + "/lpython/" + filename;
+        if (filename == "lpython.py" || filename == "lptypes.py") {
+            file_path = runtime_library_dir + "/lpython/" + "lpython.py";
```
certik, line 206: Let's remove `lptypes.py`, let's only support lpython.py.

Shaikh-Ubaid: >Let's remove lptypes.py, let's only support lpython.py.

Got it.

Since we are currently unable to upload `lpython.py` as an `lpython` package, I will include it (and its dependencies) as module(s) inside every package.

Shaikh-Ubaid: >I will include it (and its dependencies) as module(s) inside every package.

Or maybe we need not include them. I just attempted uploading the `lpynn` package without including the `lpython.py` (and its dependency file) and it seems to upload.

Shaikh-Ubaid:
This is ready.

certik (approved):
Looks great!

certik:
Polish the history or squash it.

Shaikh-Ubaid:
>Polish the history or squash it before merging.

Done. Adding to auto-merge as it seems approved.

Issue #1737, opened 2023-04-21
If you create module a.sys, it fails


certik:
The `a.sys` user module collides with the standard library module "sys". Probably something to do with module name mangling, we have to prepend "a." to "sys".

PR #1743, opened 2023-04-22
Fix user import when it overlaps with runtime library modules

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1737.

### DIFF
--- a/integration_tests/test_import/sys.py
+++ b/integration_tests/test_import/sys.py
@@ -1,2 +1,5 @@
-def hi_from_user_sys():
+from lpython import i32
+
+def hi_from_user_sys() -> i32:
     print("hi from user sys!")
+    return -5
--- a/integration_tests/test_import/sys.py
+++ b/integration_tests/test_import/sys.py
@@ -1,0 +1,2 @@
+def hi_from_user_sys():
+    print("hi from user sys!")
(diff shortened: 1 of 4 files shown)


Shaikh-Ubaid:
There could be other approaches to fix https://github.com/lcompilers/lpython/issues/1737, for example, changing/reversing the priority of `runtime library` path and `parent_dir` or using name mangling. 

This PR fixes the issue by pruning the paths to be searched (as we need not re-search the package/module).

Shaikh-Ubaid:

Thirumalai-Shaktivel:
LGTM!

certik (approved):
I think it's fine. @czgdp1807 can you also please review?

czgdp1807 (approved):

integration_tests/test_import/sys.py
```diff
@@ -0,0 +1,2 @@
+def hi_from_user_sys():
+    print("hi from user sys!")
```
czgdp1807, line 2: Probably return an integer value from here.

integration_tests/test_import_05.py
```diff
@@ -0,0 +1,3 @@
+from test_import.sys import hi_from_user_sys
+
+hi_from_user_sys()
```
czgdp1807, line 3: And then assert here. So that we know that we are not having any hidden bugs.

Shaikh-Ubaid: Done! Thanks for the guidance!

Shaikh-Ubaid:
I am adding this to auto-merge as it seems approved.

czgdp1807 (approved):

PR #1747, opened 2023-04-24
Initial implementation of `String` Validation Methods

harshsingh-24:
- Adds onto `String` Methods by giving support of basic string `validation` methods.
- The following functions are implemented: 

1. `str.islower()` - 
Return `True` if all cased characters the string are lowercase and there is at least one cased character, `False` otherwise.
2. `str.isupper()` - 
Return `True` if all cased characters the string are uppercase and there is at least one cased character, `False` otherwise.
3. `str.isdecimal()` - 
Return `True` if all characters in the string are decimal characters and there is at least one character, `False` otherwise.
4. `str.isascii()` - 
Return `True` if the string is empty or all characters in the string are ASCII, `False` otherwise. ASCII characters have code points in the range `U+0000-U+007F`.

- Specification (mentioned above) for each of the functions has been obtained from [Python docs](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str).
- Appropriate test cases for each module has been added satisfying good coverage criteria

### DIFF
--- a/integration_tests/test_str_attributes.py
+++ b/integration_tests/test_str_attributes.py
@@ -182,6 +182,108 @@ def partition():
     seperator = "apple"
     assert s.partition(seperator) == ("rendezvous 5", "", "")
 
+def islower():
+    # Case 1: When constant string is present
+    assert "".islower() == False
+    assert "APPLE".islower() == False
+    assert "4432632479".islower() == False
+    assert "%#$#$#32a".islower() == True
+    assert "apple".islower() == True
+    assert "apple is a fruit".islower() == True
+
+    # Case 2: When variable string is present
+    s: str
+    s = ""
+    assert s.islower() == False
+    s = "APPLE"
+    assert s.islower() == False
+    s = "238734587"
+    assert s.islower() == False
+    s = "%#$#$#32a"
+    assert s.islower() == True
+    s = "apple"
+    assert s.islower() == True
+    s = "apple is a fruit"
+    assert s.islower() == True
+
+def isupper():
+        # Case 1: When constant string is present
+    assert "".isupper() == False
+    assert "apple".isupper() == False
+    assert "4432632479".isupper() == False
+    assert "%#$#$#32A".isupper() == True
+    assert "APPLE".isupper() == True
+    assert "APPLE IS A FRUIT".isupper() == True
+
+    # Case 2: When variable string is present
+    s: str
+    s = ""
+    assert s.isupper() == False
+    s = "apple"
+    assert s.isupper() == False
+    s = "238734587"
+    assert s.isupper() == False
+    s = "%#$#$#32A"
+    assert s.isupper() == True
+    s = "APPLE"
+    assert s.isupper() == True
+    s = "APPLE IS A FRUIT"
+    assert s.isupper() == True
+
+def isdecimal():
+    # Case 1: When constant string is present
+    assert "".isdecimal() == False
+    assert "apple".isdecimal() == False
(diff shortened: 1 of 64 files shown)


harshsingh-24:
Hey @certik @Thirumalai-Shaktivel !! I think the build is failing because `isupper`, `islower` and other functions are also present in C standard library and it throws an errror saying - 

```
expr2__tmp__generated__.c:126:6: warning: conflicting types for built-in function ‘islower’; expected ‘int(int)’ [-Wbuiltin-declaration-mismatch]
  126 | void islower()
      |      ^~~~~~~
expr2__tmp__generated__.c:9:1: note: ‘islower’ is declared in header ‘<ctype.h>’
    8 | #include <lfortran_intrinsics.h>
  +++ |+#include <ctype.h>
    9 |
expr2__tmp__generated__.c:149:6: warning: conflicting types for built-in function ‘isupper’; expected ‘int(int)’ [-Wbuiltin-declaration-mismatch]
  149 | void isupper()
      |      ^~~~~~~
expr2__tmp__generated__.c:149:6: note: ‘isupper’ is declared in header ‘<ctype.h>’
expr2__tmp__generated__.c:198:6: warning: conflicting types for built-in function ‘isascii’; expected ‘int(int)’ [-Wbuiltin-declaration-mismatch]
  198 | void isascii()
      |      ^~~~~~~
```

How do you think, I should fix it up?

certik:
Why does the C backend generate `void isascii()`?

harshsingh-24:
> Why does the C backend generate `void isascii()`?

@certik Even I am confused about that. I will have to look into it.

Smit-create:
`int islower( int arg );`, etc. are already declared under `<ctype.h>`.

harshsingh-24:
> `int islower( int arg );`, etc. are already declared under `<ctype.h>`.

@Smit-create I need to support `islower` for LPython as well. Is it not possible to implement it?

integration_tests/test_str_attributes.py
```diff
@@ -182,6 +182,108 @@ def partition():
     seperator = "apple"
     assert s.partition(seperator) == ("rendezvous 5", "", "")
 
+def islower():
```
Thirumalai-Shaktivel, line 185: I think the error might be caused by this?
Try using different name for the function.

Thirumalai-Shaktivel: You can use `lpython a.py --show-c` to see the C code.

harshsingh-24:
Can you please review @Thirumalai-Shaktivel @certik @Smit-create ? The build has succeeded.

integration_tests/test_str_attributes.py
harshsingh-24: Yep. That fixed it. Thanks

czgdp1807 (approved):

certik (approved):
We need to start using IntrinsicFunction for these things. I think it's fine to merge, but we will have to rework it.

certik:
@harshsingh-24 can you open up an issue for the name conflict? That's a bug that we need to fix.

harshsingh-24:
> We need to start using IntrinsicFunction for these things. I think it's fine to merge, but we will have to rework it.

Yes @certik . I was waiting for @Thirumalai-Shaktivel `str.partition` implementation with Intrinsic Functions. When he will be done with that, I will start refactoring.

harshsingh-24:
> @harshsingh-24 can you open up an issue for the name conflict? That's a bug that we need to fix.

Done @certik.

Issue #1232, opened 2022-10-27
Unify stmt and expr


certik:
Currently in Python we can have code as follows:
```python
def g():
    print("Something")
    return 3

def f():
    2 + g()
    return 42

f()
```
Which will print "Something".

This is currently represented in the compiler roughly as:
```
(= 
                        (Var 3 __lcompilers_dummy) 
                        (FunctionCall 
                            1 g 
                            () 
                            [] 
                            (Integer 4 []) 
                            () 
                            ()
                        ) 
                        ()
                    )
```

If we only have expressions, then the above will not be an assignment, but just the expression (`2+g()` or `g()`) and we will not use the result.

certik:
However, one downside is that in expressions like `2+g()` suddenly some Expr nodes (the former Stmt nodes) would not be allowed. That is not very consistent. Also, many times we accept an expression, say for dimensions, indexing, or loop begin/end, etc. We would have to add verify() checks for each case that it is not a former Stmt node, and handle the case if it is (or assert, etc.).

Here is an alternative solution. Let's add an `Expr(expr expression)` node into `stmt`. This is exactly what Python does at the AST level. Then a list of statements can contain an `Expr` node, and it tells the backend that one should drop the result. Currently we use the `Assignment()` node with a dummy variable for that. So an `Expr` node is an iterative improvement of this idea. It cleans things up, and either it will be good enough, or if not, we can still unify Expr and Stmt later.

PR #1758, opened 2023-05-04
Add `list.reverse` intrinsic function

virendrakabra14:
Currently gives an `Internal Compiler Error` at
```
File ".../lpython/src/libasr/asr_utils.h", line 39
    return ASR::down_cast<ASR::stmt_t>(f);
AssertFailed: is_a<T>(*f)
```

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1929,6 +1929,18 @@ public:
         tmp = list_api->index(plist, item, asr_el_type, *module);
     }
 
+    void generate_ListReverse(ASR::expr_t* m_arg) {
+        ASR::ttype_t* asr_el_type = ASRUtils::get_contained_type(ASRUtils::expr_type(m_arg));
+        int64_t ptr_loads_copy = ptr_loads;
+        ptr_loads = 0;
+        this->visit_expr(*m_arg);
+        llvm::Value* plist = tmp;
+
+        ptr_loads = !LLVM::is_llvm_struct(asr_el_type);
+        ptr_loads = ptr_loads_copy;
+        list_api->reverse(plist, *module);
+    }
+
     void visit_IntrinsicFunction(const ASR::IntrinsicFunction_t& x) {
         switch (static_cast<ASRUtils::IntrinsicFunctions>(x.m_intrinsic_id)) {
             case ASRUtils::IntrinsicFunctions::ListIndex: {
@@ -1945,6 +1957,10 @@ public:
                     }
                 }
                 break ;
+            }
+            case ASRUtils::IntrinsicFunctions::ListReverse: {
+                generate_ListReverse(x.m_args[0]);
+                break;
             }
             default: {
                 throw CodeGenError( ASRUtils::IntrinsicFunctionRegistry::
--- a/src/libasr/codegen/llvm_utils.cpp
+++ b/src/libasr/codegen/llvm_utils.cpp
@@ -2626,6 +2626,63 @@ namespace LCompilers {
         shift_end_point_by_one(list);
     }
 
+    void LLVMList::reverse(llvm::Value* list, llvm::Module& module) {
+
+        /* Equivalent in C++:
+         *
+         * int i = 0;
+         * int j = end_point - 1;
+         * 
+         * tmp;
+         *
+         * while(j > i) {
+         *      tmp = list[i];
+         *      list[i] = list[j];
+         *      list[j] = tmp;
+         *      i = i + 1;
+         *      j = j - 1;
+         *  }
+         */
+
+        llvm::Value* end_point = LLVM::CreateLoad(*builder,
+                                        get_pointer_to_current_end_point(list));
+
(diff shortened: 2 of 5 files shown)


certik:
I think this looks good.

virendrakabra14:
I think the issue is that `ASR.asdl` has `IntrinsicFunction` only as `expr`, whereas `list.reverse` is a statement.

czgdp1807:
I think the problem in the above comment and https://github.com/lcompilers/lpython/issues/1232#issuecomment-1293624014 is exactly same. We might need to create `Expr(expr expression)` where `expression` will store `IntrinsicFunction` and then we can do `make_Expr_t(...)` to store this `expression`. @certik What do you say? I think this stores the complete information to go from ASR to LLVM and if needed back from ASR to Python code.

Another approach is that we can add `IntrinsicFunctionStmt(int intrinsic_id, expr* args, int overload_id)` in `stmt` in `ASR.asdl`. This would help in representing those intrinsic which don't return anything and are present just as a statement in the user code.

I guess the first approach is better because it helps in resolving #1232.

certik:
I think adding a `Expr(expr expression)` statement node is the least disruptive approach for now and we can iterate on this later if this is not enough.

virendrakabra14:
I added an `Expr` node, but errors remain the same.

czgdp1807:
Well you are not using it for wrapping the intrinsic function node in “create_ListReverse”.

virendrakabra14:
> Well you are not using it for wrapping the intrinsic function node in “create_ListReverse”.

I tried that
```
return ASR::make_Expr_t(al, loc, ASRUtils::EXPR(ASR::make_IntrinsicFunction_t(...)));
```

But we also need to allow similar to `void` return types for expressions.
Currently the issue is at `visit_ttype`.

czgdp1807:
Yes. `list.reverse` won’t return anything. `IntrinsicFuntionStmt` seems tempting right now. Another way is to make presence of `ttype type` optional in `IntrinsicFunction` i.e., `ttype type` -> `ttype? type` and then in `ListReverse::verify_args` make sure that its nullptr. Another option is to introduce `Void` type in ASR. Best one seems to be `Void` but using `ttype? type` is also fine. @certik What do you say?

certik:
I would use IntrinsicFunction.
```
    | IntrinsicFunction(int intrinsic_id, expr* args, int overload_id,
            ttype type, expr? value)                               
```
The `type` is `FunctionType`, which already allows the return type to be nil, so that is already supported. Regarding `expr* args`, @czgdp1807 why do we need it? It seems all the information we need is already represented in `type`, isn't it?

czgdp1807:
> Ok, so the current design already allows to "return nil". I don't think any change is needed?

There is one needed. The following one,

```diff
diff --git a/src/libasr/ASR.asdl b/src/libasr/ASR.asdl
index 12ada08a1..bdaa0ea1d 100644
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -225,7 +225,7 @@ expr
     | FunctionCall(symbol name, symbol? original_name, call_arg* args,
             ttype type, expr? value, expr? dt)
     | IntrinsicFunction(int intrinsic_id, expr* args, int overload_id,
-            ttype type, expr? value)
+            ttype? type, expr? value)
     | StructTypeConstructor(symbol dt_sym, call_arg* args, ttype type, expr? value)
     | EnumTypeConstructor(symbol dt_sym, expr* args, ttype type, expr? value)
     | UnionTypeConstructor(symbol dt_sym, expr* args, ttype type, expr? value)
```

We do the same with `return_var_type` in `FunctionType` (see below),

https://github.com/lcompilers/lpython/blob/bf59fb4eec90ce9170da132e55206104ef23a0c6/src/libasr/ASR.asdl#L365

certik:
I don't think that's the right change. The `type` is a `FunctionType`, so you always require this to be present, both for Functions (that return something) and Subroutines (that don't return anything). Inside the `FunctionType` the "return" is optional.

czgdp1807 (changes requested):
Everything looks good but please resolve the conflicts (update `main` locally and then merge into the branch of this PR, git will ask you to resolve the conflicts, do it, add and commit the changes). Do a clean build afterwards, see if all the tests pass and then push.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1887,6 +1887,11 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                              dict_type->m_value_type, name2memidx);
     }
 
+    void visit_Expr(const ASR::Expr_t& x) {
+        // std::cout << "inside visit Expr" << std::endl;
```
czgdp1807, line 1891: Remove this line.

czgdp1807:
Seems like some syntax creeped in while resolving conflicts. @kabra1110 Do you need any help in fixing or would you able to deal with it on your own.

czgdp1807 (approved):

czgdp1807:
@certik Does this look good to you?

certik (approved):
I think that's fine.

Issue #1760, opened 2023-05-04
Bug: i32() does not seem to convert from float correctly


certik:
Encountered in https://github.com/lcompilers/lpython/pull/1751#discussion_r1185286130. We need to write a small test for this that fails in master (that is, we will use assert to check the result and it will fail either in CPython or LPython, showing different results) and fix it.

PR #1761, opened 2023-05-04
Make i32/i64 convert float to int in CPython

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1760.

Main Branch:
```bash
$ cat examples/expr2.py 
from lpython import i32, f64

def main0():
    a: f64
    a = 3.25
    b: i32 = i32(a) * 4
    print(b)
    assert b == 12

main0()
$ lpython examples/expr2.py --backend llvm
12
$ python examples/expr2.py               
13.0
  File "/Users/ubaid/Desktop/OpenSource/lpython/examples/expr2.py", line 10, in <module>
    main0()
  File "/Users/ubaid/Desktop/OpenSource/lpython/examples/expr2.py", line 8, in main0
    assert b == 12
AssertionError
```

PR Branch:
```
$ lpython examples/expr2.py --backend llvm
12
$ python examples/expr2.py                
12
```


Shaikh-Ubaid:
https://github.com/lcompilers/lpython/blob/d2bf0f19fbc5bc36636b99965ae7d73b0f393a11/integration_tests/test_str_to_int.py#L5

It seems that in `lpython`, we currently use `int` for converting strings to integer.

doubt: Does `i32` also needs to convert a string to an integer?

certik:
> Does i32 also needs to convert a string to an integer?

Good question. In CPython, you use `int` to convert string to integer. So doing `i32(int("123"))` vs `i64(int("123"))` I think is not a bad syntax. And in LPython we can understand it and do the right thing. And naked `int("123")` would not be allowed in LPython. Later, once LPython supports arbitrary integers, then it could return such an integer.

Shaikh-Ubaid:
https://github.com/lcompilers/lpython/blob/d2bf0f19fbc5bc36636b99965ae7d73b0f393a11/tests/complex1.py#L11-L17

I observed that for `complex` types as well we (need-to) use both `c32/64` and `complex` together. I think this is/was a feature we previously designed. Similarly, it seems using `i32(int())` is also a design feature (probably might not be a bug). 

Currently, I think the bug (or unexpected working) are the following:
- `i32` converts `i64` and `f32/f64`, but does not convert `str`.
-  just using `i32` converts a float to integer in `lpython`, but it does not do the same in case of `cpython`.
- there seems some confusion/overlap between the workings of `i32` and the intrinsic function `int()`.

Shaikh-Ubaid:
In this PR, which of the cases shared here https://github.com/lcompilers/lpython/pull/1761#issuecomment-1535533828 should we tackle?

certik:
I would just tackle those that are needed in your changed tests. The rest we can tackle in another PR and add tests.

Shaikh-Ubaid:
>I would just tackle those that are needed in your changed tests. The rest we can tackle in another PR and add tests.

Sure, got it. This PR makes the `i32/i64` convert `float` to `integer` in `cpython`. I think this is is ready.

src/runtime/lpython/lpython.py
```diff
@@ -13,15 +13,31 @@
...
+    "i64": int,
+    "f32": float,
+    "f64": float,
+    "c32": complex,
+    "c64": complex,
+    "c_ptr": lambda x: x,
+    "Const": lambda x: x,
+    "Callable": lambda x: x,
+    "Pointer": lambda x: x,
+}
```
Shaikh-Ubaid, line 29: For `cpython`, the types `f32/64`, `c32/64` would also support auto-conversions. (We did not update the integration_tests yet for these. I guess we can update the tests later progressively. I think we could experience the auto-conversion with `i32/64` for some time and based on it update the integration_tests for `f32/64`, `c32/64`.

certik: What is an auto-conversion?

certik (approved):
I think this looks good.

src/runtime/lpython/lpython.py
Shaikh-Ubaid: Sorry Sir, probably it should just be `conversion`. Previously, `i32/64` did not do conversion (and we have to explicitly use `int()`) and in this PR they support (implicit/auto) conversion to `integers`.

PR #1763, opened 2023-05-07
CI: Install and use lpython as a package

Shaikh-Ubaid:
Using `pip install lpython_emulation`, we can now install and import `lpython` as a package. This PR updates the CI to use `lpython` installed via `pip` (previously we used `lpython` from the runtime library).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -225,6 +225,8 @@ jobs:
       - name: PIP install required packages
         shell: bash -l {0}
         run: |
+            # Package lpynn has lpython_emulation as dependency
+            # Hence, it should by default install lpython_emulation
             python -m pip install lpynn numpy
 
       - name: PIP show installed packages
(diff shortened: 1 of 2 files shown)


czgdp1807 (approved):

Shaikh-Ubaid:
I updated the description of this PR. This is ready.

.github/workflows/CI.yml
```diff
@@ -225,6 +225,8 @@ jobs:
       - name: PIP install required packages
         shell: bash -l {0}
         run: |
+            # Package lpynn has lpython_emulation as dependency
+            # Hence, it should by default install lpython_emulation
             python -m pip install lpynn numpy
```
certik, line 230: I think that this is problematic. If we change anything in `lpython.py` this test will break, since we have to update it in `lpynn` as well.

At the very least, let's pin a specific version of the lpynn package, as well as numpy.

Shaikh-Ubaid: Done.

Shaikh-Ubaid:
We updated the `lnn` package in https://github.com/lcompilers/lpython/pull/1751. I missed to reflect/propagate the updates to `integration_tests/test_pip_import_01.py` and `lpynn`. I updated them in this PR.

certik (approved):
Awesome. I think for now this is good. Thanks!

Issue #1588, opened 2023-03-17
ASR: add UnsignedInteger


certik:
The UnsignedInteger, just like the default (signed) Integer should be checked in Debug and ReleaseSafe mode for overflow (from both sides). In Release mode they are not checked for performance.

We have to carefully check that this design fixes all the pitfalls at https://github.com/lcompilers/lpython/wiki/ASR-Design#unsigned-integers, but it looks like it might.

If wraparound is required, then the design at #1578 should be used.

Implementation:

* [x] Initial implementation: https://github.com/lcompilers/lpython/pull/1764
* [x] Add tests for all corner cases, harden the implementation
  * [x] https://github.com/lcompilers/lpython/pull/1769
  * [x] https://github.com/lcompilers/lpython/pull/1770
  * [x] https://github.com/lcompilers/lpython/pull/1775
* [x] Debug time wraparound check -> implement later: https://github.com/lcompilers/lpython/issues/1776

certik:
For example, for `unsigned int a = 1000; signed int b = -1; (a > b)`, in LPython it would look like:
```python
a: u16 = 1000
b: i16 = -1
a > u16(b)
```
We must decide how `u16(b)` where `b` is signed should be handled. It can either cast negative integers to positive (no bit changes), or it can give an error message.

rebcabin:
I believe C just re-uses the bit pattern: https://replit.com/join/jlxoloezhd-brianbeckman

certik:
Yes, C does. Rust, on the other hand, does not allow you to do this:
```rust
    let a: u8 = 100;
    let b: i8 = -1;
    let val: bool = a > b;                    
```
which gives:
```console
$ cargo run     
   Compiling xx v0.1.0 (/private/tmp/xx)
error[E0308]: mismatched types
 --> src/main.rs:5:25
  |
5 |     let val: bool = a > b;
  |                         ^ expected `u8`, found `i8`
  |
help: you can convert an `i8` to a `u8` and panic if the converted value doesn't fit
  |
5 |     let val: bool = a > b.try_into().unwrap();
  |                          ++++++++++++++++++++

For more information about this error, try `rustc --explain E0308`.
error: could not compile `xx` due to previous error
```
and this:
```rust
    let a: u8 = 100;
    let b: i8 = -1;
    let val: bool = a > b.try_into().unwrap();
```
gives:
```console
$ cargo run     
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
[...]

PR #1764, opened 2023-05-08
Add unsigned int types

Smit-create:
#1588


Smit-create:
@certik Do you think we should move forward with these changes? I wanted your approval before making changes as it will include changes backend as well as some ASR.

certik (approved):
Yes, I think this is the minimal design that will get us there.

This should allow a relatively clean implementation that is separate from the integer implementation and should be easy to maintain. It does add some code everywhere, but it is maintainable for now, so I think we should start with this.

Thanks @Smit-create!

certik:
This now prints 5:
```python
def main0():
    x: u32
    x = (u32(2)+u32(3))*u32(5)
    print(x)

main0()

#if __name__ == "__main__":
#    main()
```
The ASR is missing the arithmetic expression, it just has 5. But otherwise it seems to work. There are still many bugs and corner cases to get right.

certik:
Now it works, just constants not generated.

certik:
Constants work now. 

Thorough tests must be added.

certik:
The added test does not work yet.

certik:
The test works now.

We need to add many more tests, but the basics are working.

@czgdp1807 can you please review this? I'll polish the history tomorrow and merge.

czgdp1807 (approved):
Approving it but you might have noticed the type generation in LLVM. We need to clean it up and decouple it from ASRToLLVMVisitor. :-).

certik:
Yes, it's all over the place, it's a mess. Ideally this is done at just once place, and all decisions shifted to physical types to ASR.

certik (approved):
This is good enough now to merge, I polished the commits.

More PRs will be needed with more tests, and fixing corner cases and Debug time wraparound tests.

Smit-create:
Thanks, @certik for completing it!

PR #1765, opened 2023-05-08
Update array indexing to use strides and offsets

czgdp1807:
Should be merged along with https://github.com/lfortran/lfortran/pull/1648

### DIFF
--- a/src/libasr/asdl_cpp.py
+++ b/src/libasr/asdl_cpp.py
@@ -2458,10 +2458,15 @@ static inline ASR::ttype_t* expr_type0(const ASR::expr_t *f)
                 LCOMPILERS_ASSERT(e->m_external);
                 LCOMPILERS_ASSERT(!ASR::is_a<ASR::ExternalSymbol_t>(*e->m_external));
                 s = e->m_external;
-            } else if (s->type == ASR::symbolType::Function) {
-                return ASR::down_cast<ASR::Function_t>(s)->m_function_signature;
             }
-            return ASR::down_cast<ASR::Variable_t>(s)->m_type;
+            if (s->type == ASR::symbolType::Function) {
+                return ASR::down_cast<ASR::Function_t>(s)->m_function_signature;
+            } else if( s->type == ASR::symbolType::Variable ) {
+                return ASR::down_cast<ASR::Variable_t>(s)->m_type;
+            } else {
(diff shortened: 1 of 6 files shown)


czgdp1807:
> Then in our optimizer we determine which arrays are contiguous (at compile time), and then we can generate two versions of the function, and use the contiguous version with contiguous arrays.

This is being done in `pass_array_by_data_pass` ASR pass. Whenever its possible to pass array by just using its data pointer it does that. Its messy right now but physical types will help in cleaning that up. :-)

> One can consider adding more descriptors, such as our current one in master (in addition to the one you are creating in this PR)

To be noted that no new descriptor is being added here in this PR. I am just starting to use strides. Earlier they were being ignored and we used to calculate strides whenever we indexed an array. But now those strides are filled at the time of declaring the array and then loaded when indexed. So no re-calculations every time we index the array. To support `b => a(:, 1)` i.e., associating an array pointer variable to a slice we need to use strides and offsets. See https://github.com/lfortran/lfortran/pull/1630#issuecomment-1536264752.

> Much later we can consider a combination of the two.

It will be partially present once I complete this PR. In fact its already there partially in `main` as well. `pass_array_by_data` ASR pass helps with that. For completion we need more use cases which we will discover with time.

Cases where we can create arrays without descriptors,

1. Fixed sized arrays.
2. The array isn't sliced using variable indices such as `a(x:y, z:)`. Or in other words slice always comes out to be of fixed size.
3. Associations like, `b => a(:, 1)` aren't present.
4. Calls like `call f(a(:, 1))` aren't present.
5. and there might be more conditions for removing a descriptor.

certik:
I think `a(:, 1)` doesn't require a descriptor either, but `a(1, :)` does.

czgdp1807:
That’s a special case. In general, `b => a(:, x:y, z, t:, …)` will need a descriptor so that `b(i, j, k, …)` always refers to and modifies the correct element inside data of `a`.

certik:
I think so. I think these general array operations always require a runtime descriptor to represent these slices correctly.

Then our optimizer can in special cases get rid of it in ASR, possibly via inlining and other techniques. But in general we need a runtime descriptor.

czgdp1807:
Correct. That will happen with time.

certik (approved):

PR #1767, opened 2023-05-09
Allow using emulation with LPYTHON_PY_MOD_NAME

certik:


certik:
Looks like this is not needed after all, so I am going to close this PR. If we need it in the future, we can revive it.

Issue #1710, opened 2023-04-14
To add boolean keys in dictionaries.


faze-geek:
Since boolean values namely `True `and `False` are hashable they can be used as dictionary keys - 
```
def f():
    d: dict[bool,i32]
    d = {True: 1, False: 0}
    print(d)
f()
(lp) C:\Users\kunni\lpython>python try.py
{True: 1, False: 0}

(lp) C:\Users\kunni\lpython>src\bin\lpython try.py
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
LCompilersException: Hashing bool isn't implemented yet.
```
I am working on this. @czgdp1807 @certik is there a way / design that should be followed to implement the hashing ? Kindly let me know any reference prs / issues I should follow.

faze-geek:
The hashing logic is simple but I want to know how can I register Boolean in `key_asr_type` switch case. Can someone point me where it is done?

```
switch( key_asr_type->type ) {
            case ASR::ttypeType::Integer: {
            }
+            case ASR::ttypeType::Boolean: {
+                // Map true to 1 and false to 0
+                llvm::Value* bool_hash = builder->CreateZExt(
+                    key, llvm::Type::getInt32Ty(context)
+                );
+                return builder->CreateSRem(bool_hash, capacity);
+            }
```

AtmegaBuzz:
@faze-geek are you still working on this ?

faze-geek:
> @faze-geek are you still working on this ?

Hi @AtmegaBuzz . I am working on this.

czgdp1807:
`dict[bool, f64]` -> `llvm::ArrayType` of size 2. Mixed types `dict[i32, dict[str, dict[bool, f64]]]` can be dealt with later when we complete https://github.com/lcompilers/lpython/pull/1111. The reason being `dict[bool, f64]` will have a maximum size of 2 (a boolean can be either `True` or `False`, nothing more). So using the complicated `dict` algorithms when we already know the maximum size which is very small feels unncessary. Rest when @kabra1110 will implement, then we will deal with the drawbacks of this `llvm::ArrayType` approach.

faze-geek:
Closing as completed in #1771

PR #1771, opened 2023-05-10
Dict bool keys

kabra1110:
Work in progress. Towards #1710.


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1962,6 +1962,21 @@ namespace LCompilers {
...
+                
+                llvm_utils->create_if_else(is_key_one, [&]() {
+                    LLVM::CreateStore(*builder, llvm::ConstantInt::get(
+                                    context, llvm::APInt(32, 1)), hash_value);
+                }, [=]() {
+                    LLVM::CreateStore(*builder, llvm::ConstantInt::get(
+                                    context, llvm::APInt(32, 0)), hash_value);
+                });
+                return hash_value;
+            }
```
czgdp1807, line 1979: This logic is correct but the if-else check seems unnecessary. A simple cast from `i1` to `i32` should have done the job. We will see in our tomorrow's meeting how can we do that.

czgdp1807 (changes requested):
Add tests like looping and writing/reading again and again from a bool dictionary. Value can be of different types like, integer, float, string, etc.

integration_tests/test_dict_bool.py
czgdp1807, line 1: We are doing `d[True]` and `d[False]`. Let's **also** do something like,

```
d[i%2 == 0] += 1
```

And check the count of odd and even numbers for some range `[0, n)`. To make sure things work correctly for runtime boolean values.

tests/reference/asr-test_dict_bool-6f905fd.stdout
czgdp1807, line 1: Did you generate these by hand?

czgdp1807 (changes requested):

tests/reference/asr-test_dict_bool-6f905fd.stdout
kabra1110: No, I added (and later removed) a test in `./tests/tests.toml` with `asr` and `asr_json` options.

czgdp1807: Ah I see.Then you can remove these `.stdout` files.

integration_tests/test_dict_bool.py
kabra1110: Looks like `+=` isn't currently supported for `dict` yet. I have added this test with `=`.

czgdp1807: No worries.

> Looks like `+=` isn't currently supported for `dict` yet.

Let's support this in a separate PR, once you complete this and your other open PR.

tests/reference/asr_json-test_dict_bool-8307f87.stdout
czgdp1807, line 1: Delete this as well.

tests/reference/asr_json-test_dict_bool-8307f87.json
czgdp1807, line 1: This one too.

czgdp1807 (changes requested):
If `llvm` test is enabled in `integration_tests/CMakeLists.txt` then most of the time registering the same test in `.toml` file is not needed.

czgdp1807 (approved):

PR #1777, opened 2023-05-11
Support numpy size() and array.size

### LINKED ISSUE #1773: Implement numpy.size()
Currently it returns a placeholder (Integer 1234), but we now have an ArraySize ASR node, so we should just use it.

Shaikh-Ubaid:
fixes #1773


Shaikh-Ubaid:
This is ready.

certik (approved):
Thank you! This PR is good to merge.

Since we are doing it, in a new PR can you also please add a test like:
```
from numpy import size
a = empty((2, 3))
assert size(a) == 2*3
assert size(a, 0) == 2
assert size(a, 1) == 3
```
It will probably already work, but just in case. I usually use the `size` function like this, just like in Fortran.

Shaikh-Ubaid:
>Since we are doing it, in a new PR can you also please add a test like:

Sure, thank you for the guidance. It seems the backends expect dimension value to start from `1`, where as the `axis` attribute of `numpy size()` starts from `0`. I fixed it by adding a constant `1` to the user provided dimension and added the test case in https://github.com/lcompilers/lpython/pull/1780.

PR #1780, opened 2023-05-11
Fix numpy size() and add test

Shaikh-Ubaid:
towards https://github.com/lcompilers/lpython/pull/1777#pullrequestreview-1422615315.


Shaikh-Ubaid:
Ready.

certik:
Perfect. Since we are doing this, can you please also add in a test for some unsigned integer array, say `u32`? I implemented those recently, and there might still be some bugs.

Also let's rebase on top of the latest master, just to make sure.

Shaikh-Ubaid:
Ready.

certik (approved):

integration_tests/array_size_02.py
```diff
@@ -0,0 +1,90 @@
+from lpython import i32, f64, c32, c64, u32
+from numpy import empty, size
+
+def main0():
+    x: i32[4, 5, 2] = empty([4, 5, 2])
+    y: f64[24, 100, 2, 5] = empty([24, 100, 2, 5])
```
certik, line 6: I noticed one problem here, but I created a dedicated issue for this: https://github.com/lcompilers/lpython/issues/1790, as we need to test that LPython reports an error here.

PR #1794, opened 2023-05-12
Allow importing enums

### LINKED ISSUE #1783: Cannot import enum
lpython 0.13.0 will not allow enum imports.  The following works:

```python
# file: common.py
from enum import Enum, auto

from lpython import i32

from numpy import empty

class Constants(Enum):
    NUM_ELEMS: i32 = auto()

xs: i32[Constants.NUM_ELEMS] = empty(Constants.NUM_ELEMS)
```

but attempting to import the enum fails with the following error:

```python
# file: main.py
from common import Constants

from lpython import i32

from numpy import empty

xs: i32[Constants.NUM_ELEMS] = empty(Constants.NUM_ELEMS)
```

```
$ lpython --show-c main.py
semantic error: Only Subroutines, Functions, StructType, Variables and ExternalSymbol are currently supported in 'import'
 --> /var/tmp/main.py:1:1
  |
1 | from common import Constants
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^


```

Smit-create:
Fixes #1783 

cc @certik

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -484,6 +484,8 @@ RUN(NAME enum_04             LABELS cpython llvm c)
 RUN(NAME enum_05             LABELS llvm c
         EXTRAFILES enum_05b.c)
 RUN(NAME enum_06             LABELS cpython llvm c)
+RUN(NAME enum_07             IMPORT_PATH ..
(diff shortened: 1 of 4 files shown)


certik (approved):

Issue #1784, opened 2023-05-11
Disallow parameter (argument) redefinition


certik:
```python
def f(x: i32) -> i32:
    x = 2
    return x * x

print(f(3)) # => 4
```
Currently this is allowed:
```console
$ lpython a.py 
4
```
But we should give a compile time error message "input parameter `x` cannot be assigned to"

certik:
For a comparison, in Fortran this:
```fortran
integer function f(x) result(y)
integer, intent(in) :: x
x = 2
y = x*x
end function
```
is not allowed:
```console
$ gfortran a.f90 
a.f90:3:0:

    3 | x = 2
      | 
Error: Dummy argument 'x' with INTENT(IN) in variable definition context (assignment) at (1)
$ lfortran a.f90 
semantic error: Cannot assign to an intent(in) variable `x`
 --> a.f90:3:1
  |
3 | x = 2
  | ^ 
```

Shaikh-Ubaid:
It seems that the example
```python
from lpython import i32

def f(x: i32) -> i32:
    x = 2
    return x * x

print(f(3)) # => 4
```
works in `cpython`.
```bash
$ python examples/expr2.py
4
```

Also, I think assignment to function parameter is allowed in `C/CPP`, for example:
```cpp
$ cat examples/expr2.cpp 
#include <iostream>

int f(int x) {
    x = 2;
    return x * x;
}

int main() {
    std::cout << f(3) << std::endl;
}
$ g++ examples/expr2.cpp -o expr2cpp
$ ./expr2cpp 
4
```

Thus, it seems assignment to function parameter could be allowed in `lpython`.

Shaikh-Ubaid:
I shared a concern here https://github.com/lcompilers/lpython/pull/1803#discussion_r1192906762.

PR #1803, opened 2023-05-13
Error on assignment to function parameter

Shaikh-Ubaid:
towards https://github.com/lcompilers/lpython/issues/1784.

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -4844,6 +4844,13 @@ public:
                     std::string var_name = std::string(v->m_name);
                     throw SemanticError("Assignment to loop variable `" + std::string(to_lower(var_name)) +"` is not allowed", target->base.loc);
                 }
+                if (sym->type == ASR::symbolType::Variable) {
+                    ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(sym);
+                    if (v->m_intent == ASR::intentType::In) {
+                        throw SemanticError("Assignment to an input function parameter `"
+                            + std::string(v->m_name) + "` is not allowed", target->base.loc);
+                    }
+                }
             }
             tmp_vec.push_back(ASR::make_Assignment_t(al, x.base.base.loc, target, tmp_value,
                                     overloaded));
--- a/src/runtime/lpython_builtin.py
+++ b/src/runtime/lpython_builtin.py
@@ -198,18 +198,20 @@ def bin(n: i32) -> str:
         return '0b0'
     prep: str
     prep = '0b'
-    if n < 0:
-        n = -n
+    n_: i32
+    n_ = n
+    if n_ < 0:
+        n_ = -n_
         prep = '-0b'
     res: str
     res = ''
-    if (n - _lpython_floordiv(n, 2)*2) == 0:
+    if (n_ - _lpython_floordiv(n_, 2)*2) == 0:
         res += '0'
     else:
         res += '1'
-    while n > 1:
-        n = _lpython_floordiv(n, 2)
-        if (n - _lpython_floordiv(n, 2)*2) == 0:
+    while n_ > 1:
+        n_ = _lpython_floordiv(n_, 2)
+        if (n_ - _lpython_floordiv(n_, 2)*2) == 0:
             res += '0'
         else:
             res += '1'
@@ -227,16 +229,18 @@ def hex(n: i32) -> str:
         return '0x0'
     prep: str
     prep = '0x'
-    if n < 0:
+    n_: i32
+    n_ = n
+    if n_ < 0:
         prep = '-0x'
-        n = -n
+        n_ = -n_
     res: str
     res = ""
     remainder: i32
-    while n > 0:
-        remainder = n - _lpython_floordiv(n, 16)*16
-        n -= remainder
-        n = _lpython_floordiv(n, 16)
+    while n_ > 0:
+        remainder = n_ - _lpython_floordiv(n_, 16)*16
+        n_ -= remainder
+        n_ = _lpython_floordiv(n_, 16)
         res += hex_values[remainder]
     return prep + res[::-1]
 
@@ -252,16 +256,18 @@ def oct(n: i32) -> str:
         return '0o0'
     prep: str
     prep = '0o'
-    if n < 0:
+    n_: i32
+    n_ = n
+    if n_ < 0:
         prep = '-0o'
-        n = -n
+        n_ = -n_
     res: str
     res = ""
     remainder: i32
-    while n > 0:
-        remainder = n - _lpython_floordiv(n, 8)*8
-        n -= remainder
-        n = _lpython_floordiv(n, 8)
+    while n_ > 0:
+        remainder = n_ - _lpython_floordiv(n_, 8)*8
+        n_ -= remainder
+        n_ = _lpython_floordiv(n_, 8)
         res += _values[remainder]
     return prep + res[::-1]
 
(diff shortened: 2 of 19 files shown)


src/runtime/lpython_builtin.py
```diff
@@ -198,18 +198,19 @@ def bin(n: i32) -> str:
...
+    if (n_ - _lpython_floordiv(n_, 2)*2) == 0:
         res += '0'
     else:
         res += '1'
-    while n > 1:
-        n = _lpython_floordiv(n, 2)
-        if (n - _lpython_floordiv(n, 2)*2) == 0:
+    while n_ > 1:
+        n_ = _lpython_floordiv(n_, 2)
+        if (n_ - _lpython_floordiv(n_, 2)*2) == 0:
```
Shaikh-Ubaid, line 213: It seems that the runtime library assigns to function parameters at several places. In this PR, I updated three such places. With respect to https://github.com/lcompilers/lpython/issues/1784#issuecomment-1546507259, I would like to know if we should update all places in the runtime library to not allow assignment to function parameters or if we should keep allowing assignment to function parameters.

certik: We should not allow assigning to function parameters, unless the parameter is marked Out or InOut, but that's only allowed for arrays. Integers in Python can only be In, and we should not allow assigning to them.

certik:
Awesome, thanks for this!

src/runtime/lpython_builtin.py
Shaikh-Ubaid: Sure, got it.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4621,6 +4621,12 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     std::string var_name = std::string(v->m_name);
                     throw SemanticError("Assignment to loop variable `" + std::string(to_lower(var_name)) +"` is not allowed", target->base.loc);
                 }
+                if (sym->type == ASR::symbolType::Variable) {
+                    ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(sym);
+                    if (v->m_intent == ASR::intentType::In) {
+                        throw SemanticError("Assignment to function parameter `" + std::string(v->m_name) + "` is not allowed", target->base.loc);
+                    }
+                }
```
czgdp1807, line 4629: Well there is one case like,

```python
def f(x: ClassA) -> i32:
    x.Member1 = 2
    return x.Member1 * x.Member1
```

Should the above error or work?

Shaikh-Ubaid: I think/guess it should work. I think it is one of the ways to pass variables/values by reference in python. Objects are passed by reference in python. In the above example, we pass the object `x` and inside the function, a data member of `x` is updated. After the call to the function `f`, the changes to data members of `x` would be visible to the caller function.

For example, consider the following:
```python
$ cat examples/expr2.py   

@dataclass
class N:
    val: i32

def f1(n: i32):
    n = 2
    return

def f2(n: N):
    n.val = 2
    return

def main0():
    a: i32
    b: N = N(0)
    a = 4
    b.val = 4

    f1(a)
    f2(b)

    print(a)
    print(b.val)

main0()
$ python examples/expr2.py 
4
2
$ lpython examples/expr2.py
2
2
```

Shaikh-Ubaid:
For `integration_tests/generics_02.py` test case,

https://github.com/lcompilers/lpython/blob/b9850da2cff7be3d30f667aceee5c15f31252bfa/integration_tests/generics_02.py#L5-L13

Should it work or fail?

src/lpython/semantics/python_ast_to_asr.cpp
certik:
```
def f(x: ClassA) -> i32:
    x.Member1 = 2
    return x.Member1 * x.Member1
```
We have to decide what the default passing is, whether: `In`, `InOut`, `Out`. I would start with `In`, so the above is not allowed. To allow it, do:
```
def f(x: InOut[ClassA]) -> i32:
    x.Member1 = 2
    return x.Member1 * x.Member1
```
Then it is allowed.

certik:
This should fail, to allow it, do:
```
 def swap(x: InOut[T], y: InOut[T]): 
     temp: T 
     temp = x 
     x = y 
     y = temp 
     print(x) 
     print(y) 
```
Also note that Out and InOut in Python only really works for non-simple types, so not for ints or reals.

src/lpython/semantics/python_ast_to_asr.cpp
certik: See https://github.com/lcompilers/lpython/issues/1807.

Shaikh-Ubaid:
Ready.

integration_tests/CMakeLists.txt
```diff
@@ -509,7 +509,7 @@ RUN(NAME test_pkg_lnn_01        LABELS cpython llvm)
 RUN(NAME test_pkg_lnn_02        LABELS cpython llvm)
 
 RUN(NAME generics_01         LABELS cpython llvm c)
-RUN(NAME generics_02         LABELS cpython llvm c)
+# RUN(NAME generics_02         LABELS cpython llvm c)
```
Shaikh-Ubaid, line 512: I temporarily commented this out.

certik: I made the PR a draft until we resolve it. First let's get everything else working, then we'll fix this.

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: In the following example:
```python
def f(x: In[ClassA]) -> i32:
    x.Member1 = 2 # Assignment to Member1 which is a local variable of `x`
    x = ClassA(Member1=5, ...) # Assignment to x which is an intent In variable
    return x.Member1 * x.Member1
```
Do we print errors for both the above assignments or just for the second assignment (whose target variable has intent as `In`)?

Shaikh-Ubaid: If in case we print error for the first assignment `x.Member1 = 2 # Assignment to Member1 which is a local variable of x`, do we need to print error in `CPython` as well? (I think/guess by default cpython allows assignment to class member variables).

certik: > Do we print errors for both the above assignments or just for the second assignment (whose target variable has intent as In)?

I would print error for both. "In" means that "both the pointer and what it points to is constant". If you want to assign to any of it, you have to mark it "InOut".

Regarding CPython, I would not worry about it. The contract is: if LPython compiles it, it will run in CPython also. So if LPython enforces no modification, then CPython doesn't need to enforce it. In CPython we simply ignore type annotations.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4621,6 +4621,12 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     std::string var_name = std::string(v->m_name);
                     throw SemanticError("Assignment to loop variable `" + std::string(to_lower(var_name)) +"` is not allowed", target->base.loc);
                 }
+                if (sym->type == ASR::symbolType::Variable) {
+                    ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(sym);
+                    if (v->m_intent == ASR::intentType::In) {
+                        throw SemanticError("Assignment to function parameter `" + std::string(v->m_name) + "` is not allowed", target->base.loc);
```
certik, line 4627:
```suggestion
                        throw SemanticError("Assignment to intent In function parameter `" + std::string(v->m_name) + "` is not allowed", target->base.loc);
```

certik (approved):
I think that this is good.

Shaikh-Ubaid:
This PR prints errors on assignments to simple variables with intent `In`. I will handle other cases like assignments to struct local variables where the struct is passed with intent `In` in a new PR.

PR #1805, opened 2023-05-13
Support nullptr testing

### LINKED ISSUE #1781: Add a test for testing a pointer for nullptr
This might work: `p == empty_c_void_p()`, but we need to add a test for it, and if it doesn't work, we have to implement a solution.

Smit-create:
Fixes #1781

### DIFF
--- a/integration_tests/bindc_01.py
+++ b/integration_tests/bindc_01.py
@@ -8,6 +8,7 @@ print(queries, x)
 def test_issue_1781():
     p: CPtr = empty_c_void_p()
     assert p == empty_c_void_p()
+    assert not (p != empty_c_void_p())
 
 
 test_issue_1781()
--- a/integration_tests/bindc_01.py
+++ b/integration_tests/bindc_01.py
@@ -3,3 +3,11 @@ from lpython import c_p_pointer, CPtr, i16, Pointer, empty_c_void_p
 queries: CPtr = empty_c_void_p()
 x: Pointer[i16] = c_p_pointer(queries, i16)
 print(queries, x)
+
+
+def test_issue_1781():
+    p: CPtr = empty_c_void_p()
+    assert p == empty_c_void_p()
+
+
+test_issue_1781()
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -277,6 +277,8 @@ expr
     | StringOrd(expr arg, ttype type, expr? value)
     | StringChr(expr arg, ttype type, expr? value)
 
+    | CPtrCompare(expr left, cmpop op, expr right, ttype type, expr? value)
+
     | DictConstant(expr* keys, expr* values, ttype type)
     | DictLen(expr arg, ttype type, expr? value)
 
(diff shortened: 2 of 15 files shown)


integration_tests/bindc_01.py
```diff
@@ -1,5 +1,16 @@
 from lpython import c_p_pointer, CPtr, i16, Pointer, empty_c_void_p
 
-queries: CPtr = empty_c_void_p()
-x: Pointer[i16] = c_p_pointer(queries, i16)
-print(queries, x)
```
czgdp1807, line 5: Please don't edit this. Keep it as it is. Add your function in the lines that follow.

czgdp1807 (changes requested):

src/libasr/ASR.asdl
```diff
@@ -277,6 +277,8 @@ expr
     | StringOrd(expr arg, ttype type, expr? value)
     | StringChr(expr arg, ttype type, expr? value)
 
+    | CPtrCompare(expr left, cmpop op, expr right, ttype type, expr? value)
```
certik, line 280: The `cmpop` has various things like `>`. What does that mean for `CPtr`? I think we only want `==` and `!=`.

integration_tests/bindc_01.py
```diff
@@ -3,3 +3,11 @@
 queries: CPtr = empty_c_void_p()
 x: Pointer[i16] = c_p_pointer(queries, i16)
 print(queries, x)
+
+
+def test_issue_1781():
+    p: CPtr = empty_c_void_p()
+    assert p == empty_c_void_p()
```
certik, line 10: Test both `==` and `!=`.

src/libasr/ASR.asdl
Smit-create: C has pointer comparisons for different `cmpop`  as seen [here](https://stackoverflow.com/questions/9086372/how-to-compare-pointers).

certik: I see. We can allow it in ASR, since eventually we want to have a C frontend too. But in LPython I would mainly support `==` and `!=` for now for CPtr. The other ones we can, but don't have to right now.

Smit-create:
@certik The current diff does work fine for C and LLVM but fails in CPython because of two object comparisons.
```py
>>> import ctypes
>>> x = ctypes.c_void_p()
>>> y = ctypes.c_void_p()
>>> x == y
False
```

certik:
I see, we need to figure out a robust solution in CPython that will always work. If we have to do the `.value` trick, then maybe that's what is needed. Otherwise we can create our own wrapper for CPtr to handle `==` correctly.

czgdp1807:
Done in https://github.com/lcompilers/lpython/pull/1805/commits/1b162b940fbc6f7188fcfb5eb0d1349b5262fd22.

czgdp1807 (approved):

Smit-create:
Thanks @czgdp1807!

certik (approved):
Thanks!

Issue #1789, opened 2023-05-11
`array` of structures does not seem to work properly


certik:
```python
@dataclass
class Foo:
     x: i32
     y: i32

def main() -> None:
    foos: Foo[:] = array([Foo(1, 2), Foo(3, 4)])
    print("foos[0].x =", foos[0].x)
```
In C it generates:
```c
// Implementations
void _xx_lcompilers_changed_main_xx()
{
    struct xFoo foos_value;
    struct xFoo* foos = &foos_value;
    struct Foo *foos_data;
    foos->data = foos_data;
    foos->n_dims = 1;
    foos->dims[0].lower_bound = 0;
    foos->dims[0].length = 0;
    foos->x = 1;
    foos->y = 2;
    foos->x = 3;
    foos->y = 4;
    printf("%s%s%d\n", "foos[0].x =", " ", foos->data[(0 - foos->dims[0].lower_bound)].x);
}
```

certik:
Support array of structs and fix the above example, add a test.

PR #1811, opened 2023-05-15
Support array constant initialization for array of structs

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1789.


integration_tests/structs_22.py
```diff
@@ -0,0 +1,44 @@
...
+    foos: Foo[2] = array([Foo(1, 2), Foo(3, 4)])
+    print(foos[0].x, foos[0].y, foos[1].x, foos[1].y)
+
+    assert foos[0].x == 1
+    assert foos[0].y == 2
+    assert foos[1].x == 3
+    assert foos[1].y == 4
+
+def main1() -> None:
+    foos2: Foo2[3] = array([])
```
certik, line 25: The RHS is not correct, it's an array of length 0, assigning to array of length 3. So this should give either runtime or compile time error.

integration_tests/structs_22.py
```diff
@@ -0,0 +1,44 @@
...
+    foos2 = array([Foo2(-2.3, 42, u64(3)), Foo2(45.5, -3, u64(10001)), Foo2(1.0, -101, u64(100))])
```
certik, line 26: Just "merge" the previous two lines, then it should be ok.

certik:
Otherwise this looks good, thanks!

certik (approved):

Issue #1799, opened 2023-05-12
Need syntax for passing `CPtr`s to structs and arrays of structs


dylon:
The following works perfectly well in LPython and LPython generates valid C code from it, but the syntax is not supported by CPython:

```python
# file: main.py
from lpython import i32, dataclass

from numpy import empty

@dataclass
class Foo:
     x: i32
     y: i32

def init(foos: Foo[:]) -> None:
    foos[0] = Foo(1, 2)

def main() -> None:
    foos: Foo[1] = empty(1, dtype=Foo)
    init(foos)
    print("foos[0].x =", foos[0].x)

main()
```

```
$ lpython --show-c main.py > main.c

$ clang -o app main.c

$ ./app
foos[0].x = 1

$ lpython main.py
foos[0].x = 1

$ python main.py
Traceback (most recent call last):
  File "main.py", line 10, in <module>
    def init(foos: Foo[:]) -> None:
TypeError: 'type' object is not subscriptable
```

```c
// file: main.c
#include <complex.h>
#include <inttypes.h>

#include <stdlib.h>
#include <stdbool.h>
#include <stdio.h>
#include <string.h>
#include <lfortran_intrinsics.h>

#define ASSERT(cond)                                                           \
    {                                                                          \
        if (!(cond)) {                                                         \
            printf("%s%s", "ASSERT failed: ", __FILE__);                       \
            printf("%s%s", "\nfunction ", __func__);                           \
            printf("%s%d%s", "(), line number ", __LINE__, " at \n");          \
            printf("%s%s", #cond, "\n");                                       \
            exit(1);                                                           \
        }                                                                      \
    }
#define ASSERT_MSG(cond, msg)                                                  \
    {                                                                          \
        if (!(cond)) {                                                         \
            printf("%s%s", "ASSERT failed: ", __FILE__);                       \
            printf("%s%s", "\nfunction ", __func__);                           \
            printf("%s%d%s", "(), line number ", __LINE__, " at \n");          \
            printf("%s%s", #cond, "\n");                                       \
            printf("%s", "ERROR MESSAGE:\n");                                  \
            printf("%s%s", msg, "\n");                                         \
            exit(1);                                                           \
        }                                                                      \
    }


struct dimension_descriptor
{
    int32_t lower_bound, length;
};
struct Foo {
 int32_t x;
 int32_t y;
};


struct xFoo
{
    struct Foo *data;
    struct dimension_descriptor dims[32];
    int32_t n_dims;
[...]

PR #1814, opened 2023-05-16
Support syntax for passing struct in CPython

Shaikh-Ubaid:
Related - https://github.com/lcompilers/lpython/issues/1799.

```bash
$ cat examples/expr2.py 
# file: main.py

from numpy import empty

@dataclass
class Foo:
     x: i32
     y: i32

def init(foos: Foo[:]) -> None:
    foos[0] = Foo(1, 2)

def main() -> None:
    foos: Foo[1] = empty(1, dtype=Foo)
    init(foos)
    print("foos[0].x =", foos[0].x)

main()
$ lpython examples/expr2.py 
foos[0].x = 1
$ python examples/expr2.py 
foos[0].x = 1
```


Shaikh-Ubaid:
@czgdp1807 do you mean to add the example in https://github.com/lcompilers/lpython/issues/1799#issue-1708014713 as an integration_test?

czgdp1807:
Yes.

integration_tests/structs_25.py
```diff
@@ -0,0 +1,20 @@
...
+@dataclass
+class Foo:
+     x: i32
+     y: i32
+
+def init(foos: Foo[:]) -> None:
+    foos[0] = Foo(5, 21)
+
+def main0() -> None:
+    foos: Foo[1] = empty(1, dtype=Foo)
```
certik, line 13: I would just support the `Array[]` syntax for arrays of structs, like this:
```suggestion
    foos: Array[Foo, 1] = empty(1, dtype=Foo)
```
Then it should be possible to implement this cleanly in `lpython.py` as well as LPython.

certik (approved):
I think as a quick fix it is fine. I think it might break if `Foo` already defines `__class_getitem__` or if it supports the `[]`  like syntax. Right now LPython doesn't support it, so it's fine. But in the long run I think we will have to support the `Array[Foo]` syntax.

dylon:
This does not work in CPython, right now, because `__class_getitem__` returns `None`, which is not a valid type hint:

```
File "/home/dylon/Workspace/copperhead/demo-apps/lab4/emulation/gsi_emulation0.py", line 515, in gdl_context_desc_get
    c_p_pointer(ctx_desc_ptr, gdl_context_desc[:])
  File "/home/dylon/mambaforge/envs/lab4_emulation/lib/python3.10/site-packages/gsi_emulation/ltypes.py", line 488, in c_p_pointer
    targettype_ptr = ctypes.POINTER(convert_type_to_ctype(targettype))
  File "/home/dylon/mambaforge/envs/lab4_emulation/lib/python3.10/site-packages/gsi_emulation/ltypes.py", line 223, in convert_type_to_ctype
    raise NotImplementedError("Type cannot be None")
```

Shaikh-Ubaid:
>This does not work in CPython, right now, because __class_getitem__ returns None, which is not a valid type hint:

Please, could you share an example which fails?

dylon:
Sure, this reminds me of a bug I need to report that I worked around: https://github.com/lcompilers/lpython/issues/1817

With the workaround above, the following fails with the message above:

```python
from lpython import CPtr, c_p_pointer, p_c_pointer, dataclass, empty_c_void_p, pointer, Pointer

from numpy import empty

@dataclass
class Foo:
    pass

def bar(foos_ptr: CPtr) -> None:
    foos: Pointer[Foo[:]] = c_p_pointer(foos_ptr, Foo[:])

def main() -> None:
    foos: Foo[1] = empty(1, dtype=Foo)
    foos_ptr: CPtr = empty_c_void_p()
    p_c_pointer(pointer(foos), foos_ptr)
    bar(foos_ptr)

main()
```

```
$ PYTHONPATH=/home/dylon/Workspace/lpython/src/runtime/lpython python /var/tmp/main.py
  File "/var/tmp/main.py", line 18, in <module>
    main()
  File "/var/tmp/main.py", line 16, in main
    bar(foos_ptr)
  File "/var/tmp/main.py", line 10, in bar
    foos: Pointer[Foo[:]] = c_p_pointer(foos_ptr, Foo[:])
  File "/home/dylon/Workspace/lpython/src/runtime/lpython/lpython.py", line 524, in c_p_pointer
    targettype_ptr = convert_type_to_ctype(targettype)
  File "/home/dylon/Workspace/lpython/src/runtime/lpython/lpython.py", line 255, in convert_type_to_ctype
    raise NotImplementedError("Type cannot be None")
NotImplementedError: Type cannot be None
```

certik:
It's weird that it works at our CI. I wonder if it is Python version dependent. Or our CI is broken?

dylon:
This almost works, but leads to another bug (again, branching off my workaround from above): https://github.com/lcompilers/lpython/issues/1818

```python
def dataclass(arg):
    def __class_getitem__(idxs):
        return Array(arg, idxs)
    arg.__class_getitem__ = __class_getitem__
    return py_dataclass(arg)
```

certik:
I created a dedicated issue for all the bugs reported in the last few comments above: https://github.com/lcompilers/lpython/issues/1819, so that we don't forget about this. We'll investigate and fix it.

czgdp1807:
The code posted in https://github.com/lcompilers/lpython/pull/1814#issuecomment-1550200834 will not work with CPython because for dataclasses numpy just creates an array of Python objects. See below,

```python
from lpython import dataclass, i32

from numpy import empty

@dataclass
class Foo:
    x: i32
    y: i32

@dataclass
class Foe:
    x: i32
    y: i32
    z: i32

def main() -> None:
    foos = empty(2, dtype=Foo)
    print(foos.dtype)
    foos[0] = Foo(1, 2)
    foos[1] = Foe(3, 4, 5)
    print(foos[0], foos[1])

main()
```

```zsh
(lp) 15:11:37:~/lpython_project/lpython % python /Users/czgdp1807/lpython_project/debug.py
object
Foo(x=1, y=2) Foe(x=3, y=4, z=5)
```

So we won't be able to use `ctypes.Structure` here because `foos[0]`, `foos[1]` can easily store objects of different type. Also, `foos.dtype` is just a Python object. However I can make the following work,

```python
from lpython import CPtr, c_p_pointer, p_c_pointer, dataclass, empty_c_void_p, pointer, Pointer

from numpy import empty

@dataclass
class Foo:
    pass

def bar(foos_ptr: CPtr) -> None:
    foos: Pointer[Foo[:]] = c_p_pointer(foos_ptr, Foo[:])

def main() -> None:
    foos: Foo[1] = empty(1, dtype=Foo)
    foos_ptr: CPtr = empty_c_void_p()
    p_c_pointer(pointer(foos, Foo), foos_ptr)
    bar(foos_ptr)

main()
```

with the following diff,

```diff
diff --git a/src/runtime/lpython/lpython.py b/src/runtime/lpython/lpython.py
index c6c53cf31..a82a190f8 100644
--- a/src/runtime/lpython/lpython.py
+++ b/src/runtime/lpython/lpython.py
@@ -448,6 +448,8 @@ def pointer(x, type_=None):
         type_ = type(x)
     from numpy import ndarray
     if isinstance(x, ndarray):
+        if py_is_dataclass(type_):
+            return x
         return x.ctypes.data_as(ctypes.POINTER(convert_numpy_dtype_to_ctype(x.dtype)))
     else:
         if type_ == i8:
```

czgdp1807:
Also see https://github.com/lcompilers/lpython/pull/1822. Luck by chance I also ended up with https://github.com/lcompilers/lpython/pull/1814#issuecomment-1550430976.

Issue #1785, opened 2023-05-11
Support allocatable arrays


certik:
So that inside a function we can create an array of size `n`, where `n` is a local variable (not a function parameter):
```python
n: i32 = 5
A: Allocatable[f64[:]] = empty(n, ...)
```

Smit-create:
I'm working on this.

PR #1815, opened 2023-05-16
Support allocatable in lpython

Smit-create:
#1785


czgdp1807 (approved):

certik:
Looks good so far. Tests need to be added.

Smit-create:
Thanks. We need to figure out a syntax to support allocating the Allocatable arrays equivalent to `Allocate` in ASR. So do we want to support that using the `empty()` function itself? or do we want to have a subroutine just for allocating just like the `allocate` keyword in Fortran?

certik:
I think the syntax you chose is fine. This:
```python
    a: Allocatable[f64[:]] = empty((n,), dtype=float64)
```
is equivalent to:
```fortran
real(dp), allocatable :: a(:)
allocate(a(n))
```
And this:
```python
    b: Allocatable[i32[:]]
    n = 10
    b = empty((n,), dtype=int32)
```
is equivalent to:
```
integer, allocatable :: b(:)
n = 10
allocate(b(n))
```


I think the deallocation is also equivalent, when the array goes out of scope it gets deallocated (both in Python and Fortran).

The alternative design is to do something like:
```python
a: Allocatable[f64[:]] = allocate((n,), dtype=float64)
```
And somehow mark the array as "allocatable".

In the design of this PR, the array in CPython looks exactly like any other numpy array. In ASR it is "allocatable".

Let me ask this: do we currently have any way to do the following:
```python
n = 10
a: f64[n] = empty((n,), dtype=float64)
```
Here `f64[n]` is not allowed, since "n" is a local variable. If it was a function argument, then it is allowed and the array is a local stack array, not allocatable. So allocatable in ASR seems to be the only way to declare a local runtime array, correct? I think starting with explicit `Allocatable[f64[:]]` is the way to go for now. 

Another common use case is:
```fortran
integer, allocatable :: x(:)
do i = 1, 10
    allocate(x(i))
    x = ...
    print *, x
    deallocate(x)
end do
```
which in Python would become:
```python
for i in range(1, 11):
    x: Allocatable[i32[:]] = empty((i,), dtype=int32)
    x[:] = ...
    print(x)
```
When it goes out of scope, it gets deallocated, so we can allocate it again in the next iteration. Although here one issue is that for local variables in a loop there might be slightly different semantics. This however should be equivalent:
```fortran
integer, allocatable :: x(:)
allocate(x(10))
x = ...
print *, x
deallocate(x)
do i = 1, 10
    allocate(x(i))
    x = ...
    print *, x
    deallocate(x)
end do
```
which in Python would become:
```python
x: Allocatable[i32[:]] = empty((10,), dtype=int32)
x[:] = ...
print(x)
for i in range(1, 11):
    x = empty((i,), dtype=int32)
    x[:] = ...
    print(x)
```

It's a bit confusing to me, since in Fortran you need to explicitly deallocate if you want to allocate the same array again. It gets automatically deallocated when it goes out of scope. In Python I guess we would insert a deallocate before `x = empty((i,), dtype=int32)`.

certik (approved):
I think this looks good as a start.

Smit-create:
> Let me ask this: do we currently have any way to do the following:

No, we don't support that at present and will raise the following error:
```
semantic error: Only those local variables which can be reduced to compile time constant should be used in dimensions of an array.
 --> b.py:8:12
  |
8 |     a: f64[n] = empty((n,), dtype=float64)
  |   
```

> So allocatable in ASR seems to be the only way to declare a local runtime array, correct? I think starting with explicit Allocatable[f64[:]] is the way to go for now.

Yes, that's right.

Smit-create:
I'm adding this to auto-merge and will add the allocatable support for C backend in a new PR.

certik:
Thanks!

Issue #1807, opened 2023-05-13
Implement In, InOut, Out argument passing intents


certik:
Here `S` is any type (i32, f64, Struct, Array, etc.).

In:
```
def f(x: In[S]) -> None:
   x[5] = 5 # Assignment not allowed
```
InOut:
```
def f(x: InOut[S]) -> None:
   x[5] = 5 # Allowed
```
Out:
```
def f(x: Out[S]) -> None:
   x[5] = 5 # Allowed
```

When we write just:
```
def f(x: S) -> None:
   x[5] = 5 # ?
```

Then we have to decide what the default intent is. The three options that make the most sense are `In, InOut, Unspecified`, all of which are supported in ASR. The Unspecified intent mean that it can be any of `In`, `InOut`, `Out`, and we don't know which one. This is problematic, because the caller then doesn't know if you do `f(5)` if `5` should be passed by value (for In), or by reference (for InOut / Out). In Fortran we have to support it in ASR, but in Python I would not allow it, I would always specify intent (implicitly or explicitly), so ASR is always exactly one of In, InOut or Out. I think that is better for optimizations and checking.

A good backwards compatible choice is that default of `x: S` will be `x: In[S]`, so assignment is NOT allowed:
```
def f(x: S) -> None:
   x[5] = 5 # Assignment not allowed
```

Later, if we wanted, we can change the default to InOut, so assignment would be allowed and it would not break any existing code. However, given the fact that `int` can only be `In` (see below), for consistency defaulting everything to `In` seems like a good consistent choice.

I think being restrictive by default is good, and users opt-in for less restrictions. 

Finally, there is one Python specific issue: the Python semantics only allows InOut and Out for composite types like Class, Array, List. It is not allowed for primitive types like int, float, which are passed "by value" in CPython, so they can only be "In". As a result, `x: InOut[i32]` is a compile time error.

Note: Fortran also allows unspecified intent, and "value" attribute. I think this allows to actually assign to the argument, without modifying it in the caller. We CAN later support this, but I would start with being restrictive and not allow Unspecified intent. It does not close the door to later add it. I've never used this feature in Fortran, and I always specify intent, I think the compiler should even enforce it by default. So I recommend in LPython to start with the above minimal design, always require In, InOut, Out as the only options in ASR, and if it is not specified in the Python source code, the LPython frontend will default to `In` and set `In` in ASR. Later it can be changed / extended, without breaking anything.

Shaikh-Ubaid:
[...]

PR #1820, opened 2023-05-17
ASR: Support param access in, out, inout

Shaikh-Ubaid:
towards #1807.

### DIFF
--- a/integration_tests/intent_01.py
+++ b/integration_tests/intent_01.py
@@ -1,0 +1,26 @@
+from lpython import i32, u32, f64, dataclass, In, Out, InOut
+
+@dataclass
+class Foo:
+    p: i32
+
+def f(x: i32, y: In[f64], z: InOut[list[u32]], w: Out[Foo]):
+    assert (x == -12)
+    assert abs(y - (4.44)) <= 1e-12
+    z.append(u32(5))
+    w.p = 24
+
+
+def main0():
+    a: i32 = (-12)
+    b: f64 = 4.44
+    c: list[u32] = [u32(1), u32(2), u32(3), u32(4)]
+    d: Foo = Foo(25)
+
+    print(a, b, c, d.p)
+
+    f(a, b, c, d)
+    assert c[-1] == u32(5)
+    assert d.p == 24
+
+main0()
--- a/src/runtime/lpython/lpython.py
+++ b/src/runtime/lpython/lpython.py
@@ -84,6 +84,18 @@ Callable = Type("Callable")
 Union = ctypes.Union
 Pointer = PointerType("Pointer")
 
+
+class Intent:
+    def __init__(self, type):
+        self._type = type
+
+    def __getitem__(self, params):
+        return params
+
+In = Intent("In")
+Out = Intent("Out")
+InOut = Intent("InOut")
+
 # Generics
 
 class TypeVar():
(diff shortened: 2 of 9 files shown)


Shaikh-Ubaid:
I will update the commit history. I would like to request a quick review on this. We also have https://github.com/lcompilers/lpython/pull/1803 which prints error on assignment to intent in parameters.

examples/expr2.py
```diff
@@ -1,10 +1,2 @@
...
-    x = (2+3)*5
-    print(x)
-
-main0()
-
-# Not implemented yet in LPython:
-#if __name__ == "__main__":
-#    main()
+def f(x: In[list[i32]]) -> None:
+   x[5] = 5 # Assignment not allowed
```
certik, line 2: You should remove this change.

certik (approved):
Otherwise I think this looks good, thanks!

certik:
We should also add regular integration tests and CPython tests for this.

Shaikh-Ubaid:
>We should also add regular integration tests and CPython tests for this.

Done.

Shaikh-Ubaid:
I am adding this to auto-merge as it seems approved. I will work on further modications (if any) in another PR.

src/runtime/lpython/lpython.py
```diff
@@ -84,6 +84,18 @@ def __init__(self, type, dims):
...
+class Intent:
+    def __init__(self, type):
+        self._type = type
+
+    def __getitem__(self, params):
+        return params
+
+In = Intent("In")
+Out = Intent("Out")
+InOut = Intent("InOut")
```
Shaikh-Ubaid, line 97: I disabled the auto-merge. I think/guess this needs a review. I would like to know if this approach is fine for supporting `In`, `Out`, `InOut` in `CPython`.

Shaikh-Ubaid:
Ready.

integration_tests/intent_01.py
```diff
@@ -0,0 +1,26 @@
...
+
+def main0():
+    a: i32 = (-12)
+    b: f64 = 4.44
+    c: list[u32] = [u32(1), u32(2), u32(3), u32(4)]
+    d: Foo = Foo(25)
+
+    print(a, b, c, d.p)
+
+    f(a, b, c, d)
```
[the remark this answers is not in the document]
Shaikh-Ubaid:
```bash
$ cat examples/expr2.py 
from lpython import i32, u32, f64, dataclass, In, Out, InOut

@dataclass
class Foo:
    p: i32

def f(z: In[list[u32]], w: In[Foo]):
    print(z, w.p)


def main0():
    c: list[u32] = [u32(1), u32(2), u32(3), u32(4)]
    d: Foo = Foo(25)

    f(c, d)

    assert c[-1] == u32(4)
    assert d.p == 25

main0()
$ python examples/expr2.py 
[1, 2, 3, 4] 25
$ lpython examples/expr2.py
[1, 2, 3, 4] 25
```
It seems to work locally. Shall I add this as an integration_test?

Do you mean about printing errors when there is assignment/modification to `In` variables? I am hoping to handle printing errors in PR https://github.com/lcompilers/lpython/pull/1803 after completion/merging of this PR.

certik (approved):
I think this looks good, thanks @Shaikh-Ubaid !

Issue #1798, opened 2023-05-12
casting between struct array and CPtr causes lpython to segfault despite generating valid C code


dylon:
LPython 0.13.0 generates valid C code from the following snippet, but segfaults when it is executed directly:

```python
# file: main.py
from lpython import CPtr, i32, dataclass, c_p_pointer, Pointer, empty_c_void_p, p_c_pointer

from numpy import empty

@dataclass
class Foo:
     x: i32
     y: i32

def init(foos_ptr: CPtr) -> None:
    foos: Pointer[Foo[1]] = c_p_pointer(foos_ptr, Foo[1])
    foos[0] = Foo(1, 2)

def main() -> None:
    foos: Foo[1] = empty(1, dtype=Foo)
    foos_ptr: CPtr = empty_c_void_p()
    p_c_pointer(foos, foos_ptr)
    init(foos_ptr)
    print("foos[0].x =", foos[0].x)

main()
```

```
$ lpython --show-c main.py > main.c

$ clang -o app main.c

$ ./app
foos[0].x = 1

$ lpython main.py
Segmentation fault (core dumped)
```

```c
// file: main.c
#include <complex.h>
#include <inttypes.h>

#include <stdlib.h>
#include <stdbool.h>
#include <stdio.h>
#include <string.h>
#include <lfortran_intrinsics.h>

#define ASSERT(cond)                                                           \
    {                                                                          \
        if (!(cond)) {                                                         \
            printf("%s%s", "ASSERT failed: ", __FILE__);                       \
            printf("%s%s", "\nfunction ", __func__);                           \
            printf("%s%d%s", "(), line number ", __LINE__, " at \n");          \
            printf("%s%s", #cond, "\n");                                       \
            exit(1);                                                           \
        }                                                                      \
    }
#define ASSERT_MSG(cond, msg)                                                  \
    {                                                                          \
        if (!(cond)) {                                                         \
            printf("%s%s", "ASSERT failed: ", __FILE__);                       \
            printf("%s%s", "\nfunction ", __func__);                           \
            printf("%s%d%s", "(), line number ", __LINE__, " at \n");          \
            printf("%s%s", #cond, "\n");                                       \
            printf("%s", "ERROR MESSAGE:\n");                                  \
            printf("%s%s", msg, "\n");                                         \
            exit(1);                                                           \
        }                                                                      \
    }


struct dimension_descriptor
{
    int32_t lower_bound, length;
};
struct Foo {
 int32_t x;
[...]

PR #1822, opened 2023-05-17
Add ``lower_bounds`` in ``CPtrToPointer``

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1798


src/runtime/lpython/lpython.py
```diff
@@ -523,6 +528,8 @@ def __setattr__(self, name: str, value):
 def c_p_pointer(cptr, targettype):
     targettype_ptr = convert_type_to_ctype(targettype)
     if isinstance(targettype, Array):
+        if py_is_dataclass(targettype._type):
+            return ctypes.cast(cptr.value, ctypes.py_object).value
```
czgdp1807, line 532: `cptr.value` is nothing but the address of `foos: Foo[1] = empty(1, dtype=Foo)`. And `ctypes.cast(cptr.value, ctypes.py_object).value` just gives the original numpy array back, so effectively it becomes a pointer and you can easily make changes to it in other functions.

That's the only way forward if we want to keep using numpy arrays for arrays of dataclasses. If we can avoid using numpy arrays for dataclasses then we can define our own `empty` in `lpython.py` or in our own version of `numpy.py`. In our `empty` we can just create a `ctypes.Structure` array and return it via `empty`. For all the types like `i32`, etc we can keep calling `numpy.empty`. But I would not go for it unless extremely necessary.

certik (approved):

Issue #1818, opened 2023-05-16
c_p_pointer does not support pointers to structs with enum fields


dylon:
With the workarounds I mention in https://github.com/lcompilers/lpython/pull/1814, I've hit another bug which is that I cannot use `c_p_pointer` on pointers to structs with enum fields:

```python
from enum import Enum

from lpython import CPtr, c_p_pointer, p_c_pointer, dataclass, empty_c_void_p, pointer, Pointer, i32

from numpy import empty

class Value(Enum):
    TEN: i32 = 10

@dataclass
class Foo:
    value: Value

def bar(foo_ptr: CPtr) -> None:
    foo: Pointer[Foo] = c_p_pointer(foo_ptr, Foo)

def main() -> None:
    foo: Foo = Foo(Value.TEN)
    foo_ptr: CPtr = empty_c_void_p()
    p_c_pointer(pointer(foo), foo_ptr)
    bar(foo_ptr)

main()
```

```
$ python /var/tmp/main.py
Traceback (most recent call last):
  File "/var/tmp/main.py", line 23, in <module>
    main()
  File "/var/tmp/main.py", line 21, in main
    bar(foo_ptr)
  File "/var/tmp/main.py", line 15, in bar
    foo: Pointer[Foo] = c_p_pointer(foo_ptr, Foo)
  File "/home/dylon/Workspace/lpython/src/runtime/lpython/lpython.py", line 526, in c_p_pointer
    targettype_ptr = convert_type_to_ctype(targettype)
  File "/home/dylon/Workspace/lpython/src/runtime/lpython/lpython.py", line 262, in convert_type_to_ctype
    return convert_to_ctypes_Structure(arg)
  File "/home/dylon/Workspace/lpython/src/runtime/lpython/lpython.py", line 395, in convert_to_ctypes_Structure
    fields.append((name, convert_type_to_ctype(ltype_)))
  File "/home/dylon/Workspace/lpython/src/runtime/lpython/lpython.py", line 264, in convert_type_to_ctype
    raise NotImplementedError("Type %r not implemented" % arg)
NotImplementedError: Type <enum 'Value'> not implemented
```

PR #1823, opened 2023-05-17
Support structs with enum fields in ``c_p_pointer``

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/1818


integration_tests/structs_15.py
```diff
@@ -1,5 +1,6 @@
-from lpython import i32, i16, i8, i64, CPtr, dataclass, ccall, Pointer, c_p_pointer, sizeof
+from lpython import i32, i16, i8, CPtr, dataclass, ccall, Pointer, c_p_pointer, sizeof, ccallable
 
+@ccallable
```
czgdp1807, line 3: I feel that any ``dataclass`` which is being used in C code as well should be decorated with ``ccallable`` and ``lpython.py`` should raise an error in case we try to use a pure Python ``dataclass`` with C-APIs. What do you say @certik?

certik: Yes, I think so.

certik (approved):

certik:
I resolved the conflicts, I am going to keep the merge commit there.

PR #1835, opened 2023-05-18
RTL: fix an issue with abs()

### LINKED ISSUE #1825: _lpython_bit_length1 fails using conda-forge lpython
This function seems to inter infinite loop:
```c
LFORTRAN_API int32_t _lpython_bit_length1(int8_t num)
{
    int32_t res = 0;
    num = abs(num);
    for(; num; num >>= 1, res++);
    return res;
}
```
On Apple M1, using conda-forge. Steps to reproduce (with #1824):
```
mamba create -n t1 lpython python numpy
conda activate t1
cd integration_tests
./run_tests.py
```
It will hang in `test_bit_length`. This patch makes it pass:
```diff
--- a/integration_tests/test_bit_length.py
+++ b/integration_tests/test_bit_length.py
@@ -43,6 +43,6 @@ def ff5():
 ff()
 ff1()
 ff2()
-ff3()
+#ff3()
 ff4()
 ff5()
```
And a debugger says it hangs in `_lpython_bit_length1`.

certik:
Previously the abs() was somehow returning a negative number, because we called it with an int8_t argument, but it was expecting "int" argument.

The other bug was that "num" in the for loop stayed at -1 forever. We fixed that by using the condition "num > 0". We also switched to a while loop which make it more readable.

Fixes #1825.

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -794,24 +794,33 @@ LFORTRAN_API char* _lfortran_int_to_str8(int64_t num)
 LFORTRAN_API int32_t _lpython_bit_length1(int8_t num)
 {
     int32_t res = 0;
-    num = abs(num);
-    for(; num; num >>= 1, res++);
+    num = abs((int)num);
+    while (num > 0) {
+        num = num >> 1;
+        res++;
+    }
     return res;
 }
 
 LFORTRAN_API int32_t _lpython_bit_length2(int16_t num)
 {
     int32_t res = 0;
-    num = abs(num);
-    for(; num; num >>= 1, res++);
+    num = abs((int)num);
+    while (num > 0) {
+        num = num >> 1;
+        res++;
+    }
     return res;
 }
 
 LFORTRAN_API int32_t _lpython_bit_length4(int32_t num)
 {
     int32_t res = 0;
-    num = abs(num);
-    for(; num; num >>= 1, res++);
+    num = abs((int)num);
+    while (num > 0) {
+        num = num >> 1;
+        res++;
+    }
     return res;
 }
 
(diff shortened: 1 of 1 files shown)


Smit-create (approved):
Thanks, It's clean!

Shaikh-Ubaid:
Team, does this actually fix https://github.com/lcompilers/lpython/issues/1825?

On latest main, compiling using the following commands
```bash
./build0.sh
cmake -DWITH_FMT=yes -DCMAKE_CXX_FLAGS_RELEASE="-Wall -Wextra -O3 -funroll-loops -DNDEBUG" -DWITH_LLVM=yes .
cmake --build . -j16
```
I am experiencing the following:
```bash
$ python integration_tests/test_bit_length.py               
7
3
7
7
$ lpython integration_tests/test_bit_length.py               
AssertionError
```

Shaikh-Ubaid:
In https://github.com/lcompilers/lpython/pull/1835#issuecomment-1553425715, did I use the correct/expected flags for release mode compilation?

Also,
```bash
$ cat integration_tests/test_bit_length.py 
from math import floor, log2
from lpython import i8, i32, i16

def ff3():
    x: i16
    one: i16
    one = i16(1)
    x = -i16(one << i16(13))
    print("ff3(): x.bit_length() =", x.bit_length(), ", where x is", x)
    assert i32(x.bit_length()) == 14

ff3()
$ python integration_tests/test_bit_length.py 
ff3(): x.bit_length() = 14 , where x is -8192
$ lpython integration_tests/test_bit_length.py
ff3(): x.bit_length() = 0 , where x is -8192
AssertionError
```

certik:
@Shaikh-Ubaid can you open up a new issue? It looks like you are getting a failure:
```
$ lpython integration_tests/test_bit_length.py               
AssertionError
```
But it works on my laptop as well as the CI. Let's try to reproduce the problem that you are seeing and then we'll fix it.

Shaikh-Ubaid:
>@Shaikh-Ubaid can you open up a new issue? It looks like you are getting a failure:But it works on my laptop as well as the CI. Let's try to reproduce the problem that you are seeing and then we'll fix it.

Sure. Opened here https://github.com/lcompilers/lpython/issues/1836.

PR #1837, opened 2023-05-20
Print error on assign to input parameter

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1784, https://github.com/lcompilers/lpython/issues/1807.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -227,7 +227,7 @@ jobs:
         run: |
             # Package lpynn has lpython_emulation as dependency
             # Hence, it should by default install lpython_emulation
-            python -m pip install lpython_emulation==0.0.1.8 lpynn==0.0.1.3 numpy==1.24.3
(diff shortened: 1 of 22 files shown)


Shaikh-Ubaid:
I think the CI test "Test PIP Installable Packages" might fail. Because, it tests `integration_test/test_pip_import_01` which imports `lpynn` as well as `lpython` both of which are installed via `pip` and do not yet support the intent declaration syntax.

Shaikh-Ubaid:
Let's complete https://github.com/lcompilers/lpython/pull/1838 first. So that we can then update the `lpython_emulation` package to support intent (and also remove goto).

certik (approved):
I think this looks good. Can you please also add a test for the error? It could be in a new PR.

Shaikh-Ubaid:
>Can you please also add a test for the error? It could be in a new PR.

Sure, submitted here https://github.com/lcompilers/lpython/pull/1852.

Issue #1144, opened 2022-09-23
How to implement GoTo


certik:
The AST to ASR has to do something like the following.

* Maintain a hashtable/mapping of labels (strings) -> (unique IDs, bool declared); initialize it as empty when entering a new function
* Every time you encounger a "goto label", lookup the label in the mapping, two cases can happen:
   * if it is already there, use the ID from the mapping and create GoTo(ID)
   * if it is not there, add it to the mapping, and note that it has not been declared yet, and assign a unique ID to it; then use that unique ID in GoTo(ID)
* Every time you encounter "label label", lookup the label in the mapping, two cases can happen:
   * if it is there, change declared to yes, use the ID in GoToTarget(ID)
   * if it is not there, create a new label in the mapping, assign unique ID and declared as yes, then use that ID in GoToTarget(ID)
* at the end of the function, go over this mapping and check that each label has been declared. If there is one or more labels that has not been declared, raise a compiler error. This corresponds to the case of using a "goto label" without having the corresponding "label label" statement.

czgdp1807:
> Maintain a hashtable/mapping of labels (strings) -> (unique IDs, bool declared); initialize it as empty when entering a new function

Will we allow something like,

```python
@with_goto
def f():
   # body with goto and label
   @with_goto
   def g():
       # body with goto and label
```

If we are going to allow the above case, then will we generate hash maps for `f` and `g` separately? Will `goto` in `f` be allowed to jump to label defined inside in `g`?

certik:
> If we are going to allow the above case, then will we generate hash maps for f and g separately? 

I think so. We can handle this case later.

> Will goto in f be allowed to jump to label defined inside g?

I don't think so.

Issue #1826, opened 2023-05-17
Some CPython tests fail in integration_tests with conda-forge lpython


certik:
See #1825 for setup. These fail:
```
	156 - structs_04 (Failed)
	159 - structs_09 (Failed)
	160 - structs_10 (Failed)
	166 - structs_17 (Failed)
	204 - test_01_goto (Failed)
	225 - lpython_decorator_01 (Failed)
```
The actual failures are:
```
ld: library not found for -lpython3.10
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Traceback (most recent call last):
  File "/Users/ondrej/repos/lpython/integration_tests/lpython_decorator_01.py", line 4, in <module>
    @lpython
     ^^^^^^^
  File "/Users/ondrej/repos/lpython/src/runtime/lpython/lpython.py", line 784, in __init__
    assert r == 0, "Failed to create the shared library"
AssertionError: Failed to create the shared library
```
Here probably python3.11 is used. => must make the decorator more robust
```
Traceback (most recent call last):
  File "/Users/ondrej/repos/lpython/integration_tests/test_01_goto.py", line 35, in <module>
    test_goto()
  File "/Users/ondrej/repos/lpython/integration_tests/test_01_goto.py", line 29, in test_goto
    print(f())
          ^^^
  File "/Users/ondrej/repos/lpython/integration_tests/test_01_goto.py", line 8, in f
    goto .end
    ^^^^
NameError: name 'goto' is not defined
```
goto does not work somehow.

```
Traceback (most recent call last):
  File "/Users/ondrej/repos/lpython/integration_tests/structs_10.py", line 12, in <module>
    @dataclass
     ^^^^^^^^^
  File "/Users/ondrej/repos/lpython/src/runtime/lpython/lpython.py", line 47, in dataclass
    return py_dataclass(arg)
           ^^^^^^^^^^^^^^^^^
  File "/Users/ondrej/mambaforge/envs/t1/lib/python3.11/dataclasses.py", line 1221, in dataclass
    return wrap(cls)
           ^^^^^^^^^
  File "/Users/ondrej/mambaforge/envs/t1/lib/python3.11/dataclasses.py", line 1211, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1838, opened 2023-05-20
Remove goto

Shaikh-Ubaid:
towards https://github.com/lcompilers/lpython/issues/1826#issuecomment-1552149812

closes https://github.com/lcompilers/lpython/issues/1144


certik:
Thanks! I will get back to you on this, for now let's keep it open.

certik:
Ok, it looks like we can remove it. Let's do it after the ASR sync.

If in the future we nee to put this back in, it would not be complicated.

Issue #1804, opened 2023-05-13
Support ``+=`` with ``dict``


czgdp1807:
> Looks like `+=` isn't currently supported for `dict` yet.

```python
from lpython import i32

def test_dict_bool():
    d_int: dict[bool, i32] = {}
    i: i32

    for i in range(10, 99):
        d_int[i%2 == 0] += 1

test_dict_bool()
```

```zsh
(lp) 16:09:21:~/lpython_project/lpython % lpython integration_tests/test_dict_bool.py --show-llvm
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/czgdp1807/lpython_project/lpython/src/bin/lpython.cpp", line 1765
    return emit_llvm(arg_file, runtime_library_dir, lpython_pass_manager, compiler_options);
  File "/Users/czgdp1807/lpython_project/lpython/src/bin/lpython.cpp", line 720
    res = fe.get_llvm3(*asr, pass_manager, diagnostics, infile);
  File "/Users/czgdp1807/lpython_project/lpython/src/lpython/python_evaluator.cpp", line 58
    run_fn, infile);
  File "/Users/czgdp1807/lpython_project/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 8075
    v.visit_asr((ASR::asr_t&)asr);
  File "../libasr/asr.h", line 4681
  File "../libasr/asr.h", line 4657
  File "../libasr/asr.h", line 4682
  File "../libasr/asr.h", line 4407
  File "/Users/czgdp1807/lpython_project/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 1317
    ASR::symbol_t *mod = x.m_global_scope->get_symbol(item);
  File "../libasr/asr.h", line 4684
  File "../libasr/asr.h", line 4415
  File "/Users/czgdp1807/lpython_project/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 2703
    finish_module_init_function_prototype(x);
  File "/Users/czgdp1807/lpython_project/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 4248
    visit_Function(*s);
  File "/Users/czgdp1807/lpython_project/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 3862
    visit_procedures(x);
  File "/Users/czgdp1807/lpython_project/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 4199
[...]

PR #1843, opened 2023-05-21
Add dict increment

kabra1110:
Towards #1804.
Does not work with `str` keys. Looked around `llvm_utils`, but not sure what the issue is.


integration_tests/test_dict_increment.py
```diff
@@ -0,0 +1,55 @@
...
+    s1: str
+
+    d_int_int = {0: 1}
+    d_int_int[0] += 1000
+    assert d_int_int[0] == 1001
+
+    i2 = 1
+    d_int_int = {1: i2}
+    for i1 in range(10):
+        d_int_int[1] += i1
```
czgdp1807, line 20: Also test with runtime keys. Right now all the keys in your test are compile time constants.

czgdp1807 (changes requested):

czgdp1807:
> Looked around `llvm_utils`, but not sure what the issue is.

What's the error message you are receiving?

kabra1110:
> What's the error message you are receiving?

None. The value isn't incremented.
```
from lpython import i32

def test_dict_increment():
    d_str_int: dict[str, i32]
    d_str_int = {"1": 1}
    print(d_str_int["1"])
    d_str_int["1"] += 1
    print(d_str_int["1"])
    d_str_int["1"] = d_str_int["1"] + 1
    print(d_str_int["1"])

test_dict_increment()
```
```
1
1
2
```

integration_tests/test_dict_increment.py
kabra1110: Added

czgdp1807:
> ```python
> d_str_int["1"] += 1
> ```

This doesn't work.



> ```python
> d_str_int["1"] = d_str_int["1"] + 1
> ```

But this works. So why can't we implement `... += ...` as `... = ... + ...` in LLVM?

czgdp1807:
Also we will have to see how this operation looks like in ASR. We will see in our next meeting.

integration_tests/test_dict_increment.py
```diff
@@ -0,0 +1,65 @@
...
+        if j1 < 2.0:
+            j2 += 0.1
+        j1 += 0.1
+    assert d_bool_float[j1 < 2.0] == d_bool_float[j1 > 2.0]
+    assert d_bool_float[True] == j2
+
+    j1 = 2.0
+    d_str_float = {'key': j1}
+    while j1 < 4.0:
+        d_str_float['ke' + 'y'] += 0.1
```
czgdp1807, line 54:
```suggestion
    d_str_float = {'key': j1}
    s1 = "ke"
    while j1 < 4.0:
        d_str_float[s1 + 'y'] += 0.1
```

czgdp1807 (changes requested):

certik:
Looks like more work is needed. @kabra1110 go ahead and click on "Ready for review" once this is ready.

czgdp1807 (approved):

PR #1844, opened 2023-05-21
Introducing type() as an intrinsic function

anutosh491:
**My goal through this pr is to implement the inbuilt type function in python as an intrinsic function** (for 1 argument only ... It can take up 2 as well as 3 arguments but those cases can be addressed after this pr through overloading).

```
def main0():
    print(type('hi'))

main0()

#if __name__ == "__main__":
#    main()
<class 'str'>
```
My code doesn't give out the desired output as of now because I am not sure about a few things . I get the following 
```
1682743304
```

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1969,6 +1969,17 @@ public:
         list_api->reverse(plist, asr_el_type, *module);
     }
 
+    void generate_Type(ASR::expr_t* m_arg) {
+        this->visit_expr_wrapper(m_arg, true);
+        llvm::Value* item = tmp;
+
+        // Get the LLVM type of the value
+        llvm::Type* llvmType = item->getType();
+
+        // Create an LLVM expression for the type
+        tmp = builder->CreateBitCast(tmp, llvmType);
+    }
+
     void visit_IntrinsicFunction(const ASR::IntrinsicFunction_t& x) {
         switch (static_cast<ASRUtils::IntrinsicFunctions>(x.m_intrinsic_id)) {
             case ASRUtils::IntrinsicFunctions::ListIndex: {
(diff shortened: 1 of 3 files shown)


src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -1243,6 +1244,33 @@ namespace Partition {
...
+            const std::function<void (const std::string &, const Location &)> err) {
+
+        if (args.size() != 1) {
+            err("Intrinsic type function currently supports only 1 argument", loc);
+        }
+
+        // TODO - Replace this type , it shouldn't be integer .
+        ASR::ttype_t *type = ASRUtils::TYPE(ASR::make_Integer_t(al, loc,
+                            4, nullptr, 0));;
+
```
anutosh491, line 1267: Firstly here . Guessing what our IntrinsicFunction node should look like in the ASR 
```
                                        [(IntrinsicFunction
                                            Type
                                            [(StringConstant
                                                "hi"
                                                (Character 1 2 () [])
                                            )]
                                            0
                                            // Return Type of expected output
                                            //  Compile time value 
                                        )]
```
1) Here the expected ouput would be a `type` variable so we need return type of `type` and I'm not sure if we have a `ttype` for that so I've kept it as int as of now just to pass some checks .

2) Next we would get some compile time value from `eval_Type` and we would like to have an expr node to store this type , so I ended up introducing a `TypeConstant(ttype type)` in ASR.asdl to hold a constant type value .

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1969,6 +1969,17 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
 
+    void generate_Type(ASR::expr_t* m_arg) {
+        this->visit_expr_wrapper(m_arg, true);
+        llvm::Value* item = tmp;
+
+        // Get the LLVM type of the value
+        llvm::Type* llvmType = item->getType();
+
+        // Create an LLVM expression for the type
+        tmp = builder->CreateBitCast(tmp, llvmType);
```
anutosh491, line 1980: Also I am not too sure , if this is how generate_Type should have been introduced !? Maybe @czgdp1807 could help me out here .

certik:
We currently don't have a type as a type. In other words, we currently can't reason about types at runtime. We can implement a shortcut and only represent types as strings.

What is the use case here? You want to use this for debugging?

anutosh491:
> What is the use case here? You want to use this for debugging?

I thought this would be a good addition because SymPy's codebase use the `type()` and the `isinstance()` built function a good amount of times !

certik:
We'll have to tackle this, but I recommend to start with something easier. This requires us to figure out how to handle the types at runtime and I don't know the best design yet.

Keep this open, and we'll figure this out eventually. For now focus on symbolics, and later we'll get everything figured out.

kmr-srbh:
@Shaikh-Ubaid this can be closed because #2592 was merged.

anutosh491:
Perfect thanks @kmr-srbh for finishing this :)

PR #1845, opened 2023-05-21
Add `list.pop`

### LINKED ISSUE #1821: Issues with List.pop()
According to #941 , the pop method seems to be supported for lists . But we have the following errors on main
```
def main0():
    x: list[i32]
    x = [1, 2, 3, 4]
    print(x.pop(), x)

main0()

4 [1, 2, 3]
```
Through Lpython
```
def main0():
    x: list[i32]
    x = [1, 2, 3, 4]
    print(x.pop())

main0()

.......
  File "/home/anutosh491/lpython/lpython/src/libasr/../libasr/asr.h", line 4588, in ??
    case exprType::ListPop: { v.visit_ListPop((const ListPop_t &)x); return; }
LCompilersException: visit_ListPop() not implemented
```
```
def main0():
    x: list[i32]
    x = [1, 2, 3, 4]
    x.pop()
    print(x)

main0()

...........
  File "/home/anutosh491/lpython/lpython/src/libasr/asr_utils.h", line 39, in ??
    return ASR::down_cast<ASR::stmt_t>(f);
AssertFailed: is_a<T>(*f)
```

kabra1110:
Closes #1821.

---

**Issues**

1. Currently does not work with structures like `list[tuple[i32, i32]]`. Looks like the function is called twice (?)
```python
from lpython import i32

def list_pop():
    l: list[tuple[i32, i32]]
    l = [(1,2)]
    print(l.pop())
    print(l)

list_pop()
```
```
(1, IndexError: pop from empty list
```
This `IndexError` is output when trying to pop empty lists (as in CPython).

2. For now, supported only as expression *or* statement. The first commit supports statement `x.pop()` by wrapping `IntrinsicFunction` into `Expr`, while the second supports expression (so `print(x.pop())` works, for example).


czgdp1807 (changes requested):
We will investigate the bug you are facing on Saturday.

czgdp1807:
The problem is with `print_list_tuple` ASR pass. It transforms `print(l.pop())` into `print(l.pop()[0], l.pop()[1])` which calls `l.pop` twice leading to the error because in second call to `l.pop()` the list is empty. The correct transformation would be `t = l.pop(); print(t)`.

czgdp1807:
Use the following test case,

```python
from lpython import i32

def list_pop():
    l: list[tuple[i32, i32]]
    t: tuple[i32, i32]
    l = [(-4, 2)]
    t = l.pop()
    print(t)
    print(l)

list_pop()
```

czgdp1807:
For making the following work, just create the node for `list.pop` and then wrap it up inside `Expr` if its created inside `AST::Expr`.

```python
from lpython import i32

def list_pop():
    l: list[tuple[i32, i32]
    l = [(-4, 2)]
    l.pop()
    print(l)

list_pop()
```

src/lpython/semantics/python_attribute_eval.h
```diff
@@ -236,39 +250,39 @@ struct AttributeHandler {
...
+    //                 throw SemanticAbort();
+    //             }
+    //             idx = args[0];
+    //         } else {
+    //             // default is last index
+    //             idx = (ASR::expr_t*)ASR::make_IntegerConstant_t(al, loc, -1, int_type);
+    //         }
+
+    //         return make_ListPop_t(al, loc, s, idx, list_type, nullptr);
+    // }
```
czgdp1807, line 285:
```suggestion
```

tests/reference/asr-test_list3-5f4d2a8.stderr
```diff
@@ -1,5 +1,5 @@
-semantic error: Type mismatch in 'pop', List index should be of integer type
- --> tests/errors/test_list3.py:5:15
+semantic error: For now pop() takes no arguments
```
czgdp1807, line 1: These errors weren't there before. I would like to go back to the original errors. ``pop`` should take arguments and work with them.

czgdp1807 (changes requested):

tests/reference/asr-test_list3-5f4d2a8.stderr
kabra1110: Added a commit to allow `pop` with position argument. Does not seem to work correctly with nested iterables.
```python
from lpython import i32, f64

def test_list_pop():
    l1: list[tuple[i32, f64]]
    x: tuple[i32, f64]

    l1 = [(1, 1.0), (2, 4.0)]
    x = l1.pop(0)
    print(x)        # incorrect
    print(l1)       # correct

test_list_pop()
```
```
(2, 4.00000000000000000e+00)
[(2, 4.00000000000000000e+00)]
```

Also, errors correctly caught in `verify_args` lead to an `LCompilersException`. It does not seem related to this specific PR, though.

czgdp1807 (changes requested):
Insert the following as a test case,

```python
from lpython import i32, f64, InOut

def pop_wrapper(l: InOut[list[tuple[i32, f64]]], idx: i32) -> tuple[i32, f64]:
    return l.pop(idx)

def test_list_pop():
    l1: list[tuple[i32, f64]]
    x: tuple[i32, f64]
    i: i32

    l1 = [(1, 1.0), (2, 4.0), (3, 6.0)]

    x = pop_wrapper(l1, 0)
    print(x)
    assert x == (1, 1.0)

    l1.append((4, 8.0))
    assert x == (1, 1.0)

    print(l1)
    for i in range(len(l1)):
        assert l1[i] == (i + 2, 2.0*f64(i + 2))

test_list_pop()
```

integration_tests/CMakeLists.txt
```diff
@@ -421,6 +421,7 @@ RUN(NAME test_list_section   LABELS cpython llvm c NOFAST)
 RUN(NAME test_list_count     LABELS cpython llvm)
 RUN(NAME test_list_index     LABELS cpython llvm)
 RUN(NAME test_list_reverse   LABELS cpython llvm)
+RUN(NAME test_list_pop       LABELS cpython llvm NOFAST) # TODO: Remove NOFAST from here.
```
czgdp1807, line 424:
```suggestion
RUN(NAME test_list_pop       LABELS cpython llvm NOFAST) # TODO: Remove NOFAST from here.
RUN(NAME test_list_pop2       LABELS cpython llvm NOFAST) # TODO: Remove NOFAST from here.
```

czgdp1807 (changes requested):
Rest LGTM.

czgdp1807 (approved):

PR #1846, opened 2023-05-22
Intial Implementation for SymbolicExpression ttype 

anutosh491:
Through this Pr , I would like to solve a basic program involving symbolic expressions , which is 
```
from lpython import S
from sympy import Symbol

def main0():
   x: S = Symbol('x')
   print(x)

main0()
```

This essentially involves few steps 
1) Supporting imports from SymPy for CPython
2) introducing the `SymbolicExpression` ttype 
3) introducing `SymbolicSymbol` as an intrinsic function 
4) Framing a pass in intrinsic_function.cpp to make calls to SymEngine's C interface
5) getting the backend (preferrably LLVM) to work .

Currently I've tried to address the 2nd and 3rd step !


src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -1243,6 +1244,41 @@ namespace Partition {
     }
 } // namespace Partition
 
+namespace SymbolicSymbol {
+
+    static inline ASR::expr_t *eval_SymbolicSymbol(Allocator &al, const Location &loc,
+            Vec<ASR::expr_t*> &args) {
+        // TODO
+        return nullptr;
```
anutosh491, line 1252: My take on `eval_SymbolicSymbol` would be that we need to return a compile time value through SymEngine here . So something like this I guess 
```
    static inline ASR::expr_t *eval_SymbolicSymbol(Allocator &al, const Location &loc,
            Vec<ASR::expr_t*> &args) {
        if (args.size() == 1 && ASR::is_a<ASR::Character_t>(*args[0])) {
            ASR::Character_t* symbol_node = ASR::down_cast<ASR::Character_t>(args[0]);
            std::string symbol_name = symbol_node->m_name;
            SymEngine::Symbol symbol(symbol_name);
            // TODO: Use the symbol object for further evaluation
        }
        return nullptr;
    }
 ```
Then we might need an expr node to hold the SymEngine object !?

**Comment by certik** [comment]
See also https://github.com/lcompilers/lpython/pull/1591.

File: src/libasr/pass/intrinsic_function_registry.h (hunk 1 above)
Comment (anutosh491) [reply src/libasr/pass/intrinsic_function_registry.h:1252]: Or we could do whatever operations we want here through SymEngine and finally return the value as a string !?

**Comment by certik** [comment]
You must resolve conflicts so that the CI tests run.

**Comment by certik** [comment]
I think you have to update reference results with `./run_tests -u` and commit the changes.

**Comment by anutosh491** [comment]
Hello @certik . I've got the C code generated for a simple assignment program .  Like the following 
```
from sympy import Symbol
def main0():
    x: S = Symbol('x')

main0()
```

It looks like the following 
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --show-c examples/expr2.py 
#include <symengine/cwrapper.h>

#include <stdlib.h>
#include <stdbool.h>
#include <stdio.h>
#include <string.h>


struct dimension_descriptor
{
    int32_t lower_bound, length;
};

// Implementations
basic _lcompilers_symbolic_str(char * x)
{
    basic _lcompilers_symbolic_str;
    _lcompilers_symbolic_str = _lfortran_set_symbol(x);
    return _lcompilers_symbolic_str;
}

void main0()
{
    basic x;
    x = _lcompilers_symbolic_str("x");
}

void _lpython_main_program()
{
    main0();
}

{
    _lpython_set_argv(argc, argv);
    _lpython_main_program();
    return 0;
}
```
where `_lfortran_set_symbol` would be defined in `lfortran_intrinsic.c` and would be using the `symbol_set` function from `symengine/cwrapper.h` (somethine like `symbol_set(x, 'x')` ) . Does this approch look good to you ? 
I'll be ready with the C generated code for  print statement too !

**Comment by certik** [comment]
Yes, that looks good. As we discussed, I would "inline" the `_lcompilers_symbolic_str` function.

**Comment by anutosh491** [comment]
> Yes, that looks good. As we discussed, I would "inline" the `_lcompilers_symbolic_str` function.

Yes . I will address that in the following commit (just refining the code). Currently I have the following 
```
from sympy import Symbol
def main0():
    x: S = Symbol('x')
    y: S = Symbol('y')
    x = y
    z: S = x + y

main0()

(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --show-c examples/expr2.py 
#include <math.h>
#include <symengine/cwrapper.h>

#include <stdlib.h>
#include <stdbool.h>
#include <stdio.h>
#include <string.h>


struct dimension_descriptor
{
    int32_t lower_bound, length;
};

// Implementations
void main0()
{
    basic x;
    basic_new_stack(x);
    basic y;
    basic_new_stack(y);
    basic z;
    basic_new_stack(z);
    symbol_set(x, "x");
    symbol_set(y, "y");
    basic_assign(x, y);
    basic_add(z, x, y);
}

void _lpython_main_program()
{
    main0();
}

{
    _lpython_set_argv(argc, argv);
    _lpython_main_program();
    return 0;
}
```

File: tests/reference/asr-structs_02-f95782c.stderr [hunk 2]
```diff
@@ -1,5 +1,5 @@
-semantic error: struct S s must be initialized a value
- --> tests/errors/structs_02.py:8:5
+semantic error: S not supported yet in Attribute.
+ --> tests/errors/structs_02.py:9:5
   |
-8 |     s: S
-  |     ^^^^ 
+9 |     s.x = 2
```
certik, line 4: Good test --- when `S` is not `lpython.S`, but rather a user defined struct. We need to distinguish between the two cases.

certik:
Make sure you enable `-DWITH_RUNTIME_STACKTRACE=yes`.

certik:
Very good, excellent progress!

I think now we just need to print it, so that we can see some result when we compile this.

anutosh491:
> I think now we just need to print it, so that we can see some result when we compile this.

Yes, I have been generating C code , then taking the `main` function out of it , I generate a `c` file followed by an `o` and an `out` file linked with `symengine` for printing the output !


Also my latest commit adds functionality for supporting symbolic constants , for now I've added `pi` . So if we have the following program 
```
from sympy import Symbol, pi
def main0():
    x: S = Symbol('x')
    y: S = Symbol('y')
    x = pi
    z: S = x + y
    print(z)

main0()
```

The C code we generate for this looks like 
```
// Implementations
void main0()
{
    basic x;
    basic_new_stack(x);
    basic y;
    basic_new_stack(y);
    basic z;
    basic_new_stack(z);
    symbol_set(x, "x");
    symbol_set(y, "y");
    basic_const_pi(x);
    basic_add(z, x, y);
    printf("%s\n", basic_str(z));
}
```
It gives me `y + pi` when I print it out !

tests/reference/asr-structs_02-f95782c.stderr
anutosh491: Just to clarify , does this mean we should always have the import statement `from lpython import S` in any program we write ? Should using the `S` annotation without the import statement work ? 

Cause in python when we define a class with the same name as a built-in primitive data type, we essentially override the reference to that primitive type with our class definition.

This error essentially spawns from the following block in `get_type_from_var_annotation` . 
```
        } else if (var_annotation == "S") {
            type = ASRUtils::TYPE(ASR::make_SymbolicExpression_t(al, loc));
```

Here we could add a check if the import statement is present or not !

Thirumalai-Shaktivel: > Should using the S annotation without the import statement work ?

Yup, I think it might work, the import statement is used for CPython to compile.

Thirumalai-Shaktivel: Yup, that is one way, you can also do:
```diff
diff --git a/src/lpython/semantics/python_ast_to_asr.cpp b/src/lpython/semantics/python_ast_to_asr.cpp
index 849c157d3..c7a4b778d 100644
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -1040,8 +1040,6 @@ public:
             type = ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
         } else if (var_annotation == "CPtr") {
             type = ASRUtils::TYPE(ASR::make_CPtr_t(al, loc));
-        } else if (var_annotation == "S") {
-            type = ASRUtils::TYPE(ASR::make_SymbolicExpression_t(al, loc));
         } else if (var_annotation == "pointer") {
             LCOMPILERS_ASSERT(n_args == 1);
             AST::expr_t* underlying_type = m_args[0];
@@ -1073,6 +1071,8 @@ public:
                         }
                     }
                 }
+            } else if (var_annotation == "S") {
+                type = ASRUtils::TYPE(ASR::make_SymbolicExpression_t(al, loc));
             }
             if( raise_error ) {
                 throw SemanticError("Unsupported type annotation: " + var_annotation, loc);
```

Thirumalai-Shaktivel: It checks whether the S struct is defined, if not then assume it as a symbolic type.

tests/reference/run_dbg-test_assert_01-2f34744.stderr
```diff
@@ -1,5 +1,2 @@
-  File "tests/runtime_errors/test_assert_01.py", line 4
-    test()
-  File "tests/runtime_errors/test_assert_01.py", line 2
-    assert False
+warning: The `runtime stacktrace` is not enabled. To get the stacktraces, re-build LPython with `-DWITH_RUNTIME_STACKTRACE=yes`
```
Thirumalai-Shaktivel, line 1: Remove these warnings changes, 
re-build LPython with `-DWITH_RUNTIME_STACKTRACE=yes`
Do you use Windows? if so, do `python run_tests.py -u --skip-run-with-dbg` to update tests.

certik:
Here is the first test:
```python
from sympy import Symbol, pi
from lpython import S

def main0():
    x: S = Symbol('x')
    y: S = Symbol('y')
    x = pi
    z: S = x + y
    print(z)

main0()
```


TODO:
* [ ] Move your local `examples/expr2.py` to `integration_test/symbolics_01.py`
* [ ] Enable it using C and CPython and make the CI pass
  * [ ] Probably `lpython.py` needs adding `S` (`from lpython import S`)

tests/reference/asr-structs_02-f95782c.stderr
anutosh491: Yes , this should do the job !

tests/reference/run_dbg-test_assert_01-2f34744.stderr
anutosh491: Done .

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -2371,16 +2381,34 @@ R"(#include <stdio.h>
...
         headers.insert("math.h");
-        this->visit_expr(*x.m_args[0]);
-        out += "(" + src + ")";
-        src = out;
+        if (x.n_args == 0){
+            src = out;
+        } else if (x.n_args == 1) {
+            this->visit_expr(*x.m_args[0]);
+            if (x.m_intrinsic_id != static_cast<int64_t>(ASRUtils::IntrinsicFunctions::SymbolicSymbol)) {
+                out = "(" + src + ")";
```
Thirumalai-Shaktivel, line 2408:
```suggestion
                out += "(" + src + ")";
```

anutosh491: Thanks a lot ! Might have overlooked this !

anutosh491:
I have addressed all the TODO's except the github actions test for SymPy . I'm not sure what all should we have here ! Would this do for a start ?
```
test_sympy:
  name: Test Sympy
  runs-on: ubuntu-latest

  steps:
    - name: Checkout Repository
      uses: actions/checkout@v3
      with:
        fetch-depth: 0

    - name: Set up Conda Environment
      uses: mamba-org/setup-micromamba@v1
      with:
        environment-file: ci/environment.yml
        create-args: python=3.10

    - name: Install Sympy and Symengine
      run: conda install sympy symengine -c conda-forge
```

anutosh491:
The **Check Debug build** and **LPython CI / LPython CI (3.10, ubuntu-latest)** fail because importing `symengine` and `sympy` are failing . I tried adding them to the environment through `ci/environment.yml` but it didn't really help !

appveyor.yml
```diff
@@ -8,7 +8,7 @@ build_script:
 - call conda config --set always_yes yes --set changeps1 no
 - call conda info -a
 - call conda update -q conda
-- call conda install -c conda-forge python=3.7 re2c m2-bison xonsh llvmdev=11.1.0 jupyter xeus=1.0.1 xtl nlohmann_json cppzmq jupyter_kernel_test pytest
+- call conda install -c conda-forge python=3.7 re2c m2-bison xonsh llvmdev=11.1.0 jupyter xeus=1.0.1 xtl nlohmann_json cppzmq jupyter_kernel_test pytest symengine sympy
```
certik, line 11: I think we don't use appveyor anymore, so we can probably remove the whole file. In another PR.

ci/azure_install_macos.sh
```diff
@@ -5,7 +5,7 @@ set -ex
 conda config --set always_yes yes --set changeps1 no
 conda info -a
 conda update -q conda
-conda install -c conda-forge python=3.8 re2c bison=3.4 m4 xonsh llvmdev=11.0.1 toml cmake=3.17.0 jupyter pytest xeus=1.0.1 xtl nlohmann_json cppzmq jupyter_kernel_test
+conda install -c conda-forge python=3.8 re2c bison=3.4 m4 xonsh llvmdev=11.0.1 toml cmake=3.17.0 jupyter pytest xeus=1.0.1 xtl nlohmann_json cppzmq jupyter_kernel_test symengine sympy
```
certik, line 8: I would install specific versions of sympy and symengine.

certik:
Thanks @Thirumalai-Shaktivel for your help with this!

certik:
Great job! Indeed, everything is done, except the CI. Thanks @Thirumalai-Shaktivel for your help.

Ok, to add a CI, you need to add this to the github CI. And we need to not run this test by default, only in the CI job with sympy installed.

certik:
Here is some benchmark. SymPy:
```console
$ time PYTHONPATH=src/runtime/lpython python integration_tests/symbolics_01.py
y + pi
PYTHONPATH=src/runtime/lpython python integration_tests/symbolics_01.py  0.16s user 0.03s system 97% cpu 0.187 total
```
And LPython+SymEngine via the C backend:
```console
$ time lpython --backend=c --enable-symengine integration_tests/symbolics_01.py
y + pi
lpython --backend=c --enable-symengine integration_tests/symbolics_01.py  0.08s user 0.02s system 92% cpu 0.104 total
```
So exactly the same source code is faster to compile and run via LPython than it is to run via CPython+SymPy.  Very good!

If we exclude compilation and just run the binary:
```console
$ time ./symbolics_01.out
y + pi
./symbolics_01.out  0.00s user 0.01s system 74% cpu 0.013 total
```
This is solid!

CC @asmeurer, you might like these preliminary results.

ci/azure_install_macos.sh
Smit-create: I think we don't use this and so it would be good to remove this too.

src/bin/lpython.cpp
```diff
@@ -1562,6 +1568,7 @@ int main(int argc, char *argv[])
         app.add_flag("--get-rtl-dir", print_rtl_dir, "Print the path to the runtime library file");
         app.add_flag("--verbose", compiler_options.verbose, "Print debugging statements");
         app.add_flag("--enable-cpython", compiler_options.enable_cpython, "Enable CPython runtime");
+        app.add_flag("--enable-symengine", compiler_options.enable_symengine, "Enable Symengine runtime");
```
Thirumalai-Shaktivel, line 1571: I get it now, we haven't to specify `--enable-symengine` option to integration_tests

Thirumalai-Shaktivel: Similar to this: 
```console
$ lpython --backend=c integration_tests/symbolics_01.py       
symbolics_01__tmp__generated__.c:2:10: fatal error: 'symengine/cwrapper.h' file not found
#include <symengine/cwrapper.h>
...
```

Thirumalai-Shaktivel:
@anutosh491 I was able to add a Test to CI.
If you'd like to modify anything, please feel free to do so.

integration_tests/CMakeLists.txt
```diff
@@ -137,6 +137,30 @@ macro(RUN_UTIL RUN_FAIL RUN_NAME RUN_FILE_NAME RUN_IMPORT_PATH RUN_LABELS RUN_EN
...
+                    COMMAND ${LPYTHON} ${extra_args} --show-c
+                        ${CMAKE_CURRENT_SOURCE_DIR}/${file_name}.py > ${name}.c
+                    DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
+                    VERBATIM)
+            add_executable(${name} ${name}.c ${extra_files})
+            target_include_directories(${name} PRIVATE ${CMAKE_SOURCE_DIR}
+                "${Python_INCLUDE_DIRS}/..")
+            set_target_properties(${name} PROPERTIES LINKER_LANGUAGE C)
+            if (APPLE)
+                set(SYMPY_LIB "${Python_LIBRARY_DIRS}/libsymengine.dylib")
```
Thirumalai-Shaktivel, line 152: @certik Would this be okay?

anutosh491:
Thanks a lot @Thirumalai-Shaktivel and @certik for the help . I'll look into the changes introduces and suggested by y'all soon !

certik:
This starts to look really good. Don't rebase this PR, only push on top. From now on, we have one commit that passes all tests, so let's keep pushing on top, if we break the CI, we can push more changes to fix. And we always have a commit that works.

Now our goal is to polish this so that we can merge it.

.github/workflows/CI.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - name: Install Linux / macOS Conda Packages
         if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'macos')
         shell: bash -l {0}
-        run: conda install bison=3.4
+        run: conda install bison=3.4 sympy
```
certik, line 51: In here, let's not install sympy and only run all CPython (and LPython) tests that do not use sympy.

That way a regular LPython development does not depend on sympy in any way, you don't need to care, unless you use those features.

The sympy support will still get tested at the CI, so it won't get broken. But when I run tests locally, we don't depend on sympy.

integration_tests/CMakeLists.txt
```diff
@@ -546,6 +570,7 @@ RUN(NAME structs_22          LABELS cpython llvm c NOFAST)
 RUN(NAME structs_23          LABELS cpython llvm c NOFAST)
 RUN(NAME structs_24          LABELS cpython llvm c)
 RUN(NAME structs_25          LABELS cpython llvm c)
+RUN(NAME symbolics_01        LABELS cpython sympy)
```
certik, line 573: Let's do it like this:
```suggestion
RUN(NAME symbolics_01        LABELS cpython-sym c-sym)
```
Where `cpython-sym` is a cpython test that uses sympy (so it will only run at the CI with sympy included, otherwise it won't run). And `c-sym` is a C test with `symengine` included. Later we'll add `llvm-sym`, which will run an LLVM test with sympy included.

src/bin/lpython.cpp
```diff
@@ -1244,11 +1244,17 @@ int link_executable(const std::vector<std::string> &infiles,
...
+        }
         cmd += " -L" + base_path
             + " -Wl,-rpath," + base_path + " -l" + runtime_lib + " -lm";
+        if (compiler_options.enable_symengine) {
+            cmd += " -L$CONDA_PREFIX/lib -Wl,-rpath, $CONDA_PREFIX/lib -lsymengine";
+        }
         if (compiler_options.enable_cpython) {
             std::string py_version = "3.10";
-            std::string py_flags = R"(-I $CONDA_PREFIX/include/python)" + py_version + R"( -L$CONDA_PREFIX/lib -Wl,-rpath -Wl,$CONDA_PREFIX/lib -lpython)" + py_version + R"()";
+            std::string py_flags = R"(-I $CONDA_PREFIX/include/python)" + py_version + R"( -L$CONDA_PREFIX/lib -Wl,-rpath, $CONDA_PREFIX/lib -lpython)" + py_version + R"()";
```
certik, line 1257: I am not sure about this change. I thought the correct syntax is `-Wl,-rpath -Wl,$CONDA_PREFIX/lib`. Does it not work? I know it's slightly different on macOS and Linux, and one has to be careful here.

src/libasr/codegen/c_utils.h
```diff
@@ -527,6 +528,39 @@ class CCPPDSUtils {
...
+                    }
+                    default: {
+                        throw LCompilersException("IntrinsicFunction: `"
+                            + LCompilers::ASRUtils::get_intrinsic_name(intrinsic_id)
+                            + "` is not implemented");
+                    }
+                }
+            }
+            return result;
+        }
```
certik, line 562: Why not integrating this directly with `deep_copy` above?

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -2106,7 +2227,10 @@ namespace IntrinsicFunctionRegistry {
                 {"sum", {&Sum::create_Sum, &Sum::eval_Sum}},
                 {"list.index", {&ListIndex::create_ListIndex, &ListIndex::eval_list_index}},
                 {"list.reverse", {&ListReverse::create_ListReverse, &ListReverse::eval_list_reverse}},
-                {"list.pop", {&ListPop::create_ListPop, &ListPop::eval_list_pop}}
+                {"Symbol", {&SymbolicSymbol::create_SymbolicSymbol, &SymbolicSymbol::eval_SymbolicSymbol}},
+                {"list.pop", {&ListPop::create_ListPop, &ListPop::eval_list_pop}},
+                {"SymbolicAdd", {&SymbolicAdd::create_SymbolicAdd, &SymbolicAdd::eval_SymbolicAdd}},
+                {"pi", {&SymbolicPi::create_SymbolicPi, &SymbolicPi::eval_SymbolicPi}},
```
certik, line 2233: I would put `list.pop` at the top, right now it's between `Symbol` and `SymbolicAdd`.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -2211,7 +2336,10 @@ inline std::string get_intrinsic_name(int x) {
         INTRINSIC_NAME_CASE(ListIndex)
         INTRINSIC_NAME_CASE(Partition)
         INTRINSIC_NAME_CASE(ListReverse)
+        INTRINSIC_NAME_CASE(SymbolicSymbol)
         INTRINSIC_NAME_CASE(ListPop)
+        INTRINSIC_NAME_CASE(SymbolicAdd)
+        INTRINSIC_NAME_CASE(SymbolicPi)
```
certik, line 2342: Put `ListPop` above these.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3084,6 +3114,8 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
     void visit_Name(const AST::Name_t &x) {
         std::string name = x.m_id;
         ASR::symbol_t *s = current_scope->resolve_symbol(name);
+        std::set<std::string> not_cpython_builtin = {
+            "pi"};
```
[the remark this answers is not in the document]
certik: I see, it's similar to how other functions from numpy and sympy are handled. It might be ok for now, we can refactor this later.

certik:
I left some comments. Let's address those, then I'll do another round of review. After that hopefully we can merge.

I also merged the latest master, things seem to pass.

src/libasr/codegen/c_utils.h
anutosh491: Well the function definitons of both `get_deepcopy` and `get_deepcopy_symbolic` are different . 
```
std::string get_deepcopy(ASR::ttype_t *t, std::string value, std::string target)
std::string get_deepcopy_symbolic(ASR::expr_t *value_expr, std::string value, std::string target)
```
We call `get_deepcopy_symbolic` once we've confirmed the `ttype`
```
if (m_target_type->type == ASR::ttypeType::SymbolicExpression){
    ASR::expr_t* m_value_expr = x.m_value;
    src += alloc + indent + c_ds_api->get_deepcopy_symbolic(m_value_expr, value, target) + "\n";
}
```
But in this function we would have to pass the `m_value` too as we are essentially checking `m_intrinsic_id` and then mapping the corresponding symengine function with it .For eg
```
SymbolicSymbol : symbol_set
SymbolicAdd : basic_add
SymbolicPi : basic_const_pi
```
We can merge them both but we might also have to pass the `m_value` in `get_deepcopy`.

src/bin/lpython.cpp
anutosh491: Yeah even I thought the correct syntax involves having `-Wl` after `-rpath` . Not sure if this is affecting the CI !

Shaikh-Ubaid: @anutosh491 I think both the syntaxes (`-Wl,a,b,c` and `-Wl,a -Wl,b -Wl,c`) would/should work.

anutosh491:
I just resolved some merge conflicts as @Shaikh-Ubaid 's Pr introduced some , not sure why the CI is failing . Trying to figure it out !

Shaikh-Ubaid:
>I just resolved some merge conflicts as @Shaikh-Ubaid 's Pr introduced some , not sure why the CI is failing . Trying to figure it out !

It is the `fast` test that seems to fail. I think you were previously removing `sympy` from `FAST` in the below code.

https://github.com/lcompilers/lpython/blob/65bab540eeeb30791b686401519d2d7e5a68c6a3/integration_tests/CMakeLists.txt#L270

I guess you lost the change while resolving merge conflicts.

integration_tests/CMakeLists.txt
```diff
@@ -267,8 +291,8 @@ macro(RUN)
     if ((NOT DISABLE_FAST) AND (NOT RUN_NOFAST))
         set(RUN_EXTRA_ARGS ${RUN_EXTRA_ARGS} --fast)
         set(RUN_NAME "${RUN_NAME}_FAST")
-        list(REMOVE_ITEM RUN_LABELS cpython cpython_py) # remove cpython, cpython_py from --fast test
-        RUN_UTIL(RUN_FAIL RUN_NAME RUN_FILE_NAME RUN_LABELS RUN_ENABLE_CPYTHON RUN_EXTRAFILES RUN_EXTRA_ARGS)
+        list(REMOVE_ITEM RUN_LABELS cpython sympy cpython_py) # remove cpython, sympy, cpython_py from --fast test
+        RUN_UTIL(RUN_FAIL RUN_NAME RUN_FILE_NAME RUN_IMPORT_PATH RUN_LABELS RUN_ENABLE_CPYTHON RUN_EXTRAFILES RUN_EXTRA_ARGS)
```
Shaikh-Ubaid, line 295: I think we need not specify `RUN_IMPORT_PATH` here. It does not seem to used in `RUN_UTIL`.

integration_tests/CMakeLists.txt
```diff
@@ -551,6 +575,7 @@ RUN(NAME structs_23          LABELS cpython llvm c NOFAST)
 RUN(NAME structs_24          LABELS cpython llvm c)
 RUN(NAME structs_25          LABELS cpython llvm c)
 RUN(NAME structs_26          LABELS cpython llvm c)
+RUN(NAME symbolics_01        LABELS cpython sympy)
```
Shaikh-Ubaid, line 578: I think this might currently fail at the CI (in `LPython CI / LPython CI (3.10, ubuntu-latest) (pull_request)`). At the `CI`, I think we do not have `sympy` installed. You need to run this test only in the Job `LPython CI / Run SymPy tests (pull_request)`, where you have `sympy` installed.

Shaikh-Ubaid: I think this can be supported by following https://github.com/lcompilers/lpython/pull/1846#discussion_r1232530277.

anutosh491: Correct I guess this commit (https://github.com/lcompilers/lpython/commit/bd7e0d4593ddfac2aacecc8e03f46fe98558d831) did install sympy in the environment but it was taken out during the rebase !

anutosh491: And as suggested, sympy might not be a good addition to the environment due to reasons stated here https://github.com/lcompilers/lpython/pull/1846#discussion_r1232528749

integration_tests/CMakeLists.txt
Shaikh-Ubaid: The existing names use `_` (for example `wasm_x64`, `c_py`). So, maybe we can do `cpython_sym` and `c_sym`.

integration_tests/CMakeLists.txt
anutosh491: Checkout these comments too 
https://github.com/lcompilers/lpython/pull/1846#discussion_r1232527167
https://github.com/lcompilers/lpython/pull/1846#discussion_r1232530277

This is the direction we would like to go for but I am not fully sure on how to proceed .

integration_tests/CMakeLists.txt
anutosh491: So I'll have to introduce these in the backend right ? I guess here 
```
SUPPORTED_BACKENDS = ['llvm', 'c', 'wasm', 'cpython', 'x86', 'wasm_x86', 'wasm_x64', 'c_py', 'cpython_py', 'sympy']
```

integration_tests/CMakeLists.txt
Shaikh-Ubaid: Please checkout the commit [CI: Add separate job for CPython Interop tests](https://github.com/lcompilers/lpython/pull/1933/commits/56db034c8b54ba5d47d62c51dd2ec4ad6d193b58). I think it could guide you.

anutosh491: Sure ! Thanks , I'll go through it and get back !

integration_tests/CMakeLists.txt
Shaikh-Ubaid: Yes. Also, I suggest to use the names `cpython_sym` and `c_sym`. (These names suggest which backend it uses and also which library is uses. Also `cpython_sym` is distinct from `cpython`, hence tests using `sympy` would not be run locally when someone uses `./run_tests.py -b cpython`.)

anutosh491: Yes I do understand what You're trying to convey . Though I haven't really worked with CMake files before hence not fully sure on how I could introduce the blocks for `cpython_sym ` and `c_sym` as follows
```
elseif(KIND STREQUAL "cpython_sym ")
elseif(KIND STREQUAL "c_sym")
```
But yeah I realize that they would be closely related to how the cpython and c backends have been handled . Will try to figure it out . Any suggestions are welcome !

anutosh491: The block for `c_sym` which involves a C test including symengine , could look like the following
```
        elseif(KIND STREQUAL "c_sym")
            add_custom_command(
                OUTPUT ${name}.c
                COMMAND ${LPYTHON} ${extra_args} --show-c ${CMAKE_CURRENT_SOURCE_DIR}/${file_name}.py > ${name}.c
                DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${file_name}.py
                VERBATIM)
            add_executable(${name} ${name}.c ${extra_files})
            target_include_directories(${name} PRIVATE ${CMAKE_SOURCE_DIR}
                "${Python_INCLUDE_DIRS}/..")
            if (APPLE)
                set(SYMPY_LIB "${Python_LIBRARY_DIRS}/libsymengine.dylib")
            else()
                set(SYMPY_LIB "${Python_LIBRARY_DIRS}/libsymengine.so")
            endif()
            target_link_libraries(${name} lpython_rtlib ${SYMPY_LIB})
            add_test(${name} ${CMAKE_CURRENT_BINARY_DIR}/${name})
            if (labels)
                set_tests_properties(${name} PROPERTIES LABELS "${labels}")
            endif()
            if (${fail})
                set_tests_properties(${name} PROPERTIES WILL_FAIL TRUE)
            endif()
 ```

Comment (Shaikh-Ubaid) [reply integration_tests/CMakeLists.txt:578]: Yes, it looks good.

For `cpython_sym`, I think for now you can just do `elseif((KIND STREQUAL "cpython") OR (KIND STREQUAL "cpython_py") OR (KIND STREQUAL "cpython_sym"))`.

File: integration_tests/CMakeLists.txt [hunk 18]
```diff
@@ -267,7 +290,7 @@ macro(RUN)
     if ((NOT DISABLE_FAST) AND (NOT RUN_NOFAST))
         set(RUN_EXTRA_ARGS ${RUN_EXTRA_ARGS} --fast)
         set(RUN_NAME "${RUN_NAME}_FAST")
-        list(REMOVE_ITEM RUN_LABELS cpython cpython_py) # remove cpython, cpython_py from --fast test
+        list(REMOVE_ITEM RUN_LABELS cpython sympy cpython_py) # remove cpython, sympy, cpython_py from --fast test
```
Shaikh-Ubaid, line 293: I guess the above would be `cpython_sym` inplace of `sympy`?

anutosh491: Yes that's correct !

Shaikh-Ubaid: >Actually sympy essentially was replaced by c_sym, so maybe we could include that too ?

I think we should not include it. `cpython_sym` means the test runs using `CPython` and I think `CPython` does not support the `--fast` flag. Therefore tests for the backend `cpython_sym` should not be run with `--fast` flag. 

`c_sym` uses `lpython's` `c` backend. `lpython` supports `--fast` flag. Therefore if the given test file supports/compiles-with `--fast`, we should test both the versions non-fast and `--fast`.

anutosh491: Yes , that's a good argument . It does make sense .

certik:
@Shaikh-Ubaid can you please give it a thorough review? Is this consistent with the other tests that we recently added to LPython, regarding the Python interoperability?

certik:
Overall I think this PR looks pretty good. If @Shaikh-Ubaid is fine with it, I'll give it a thorough review and testing tomorrow and we can then merge it.

.github/workflows/CI.yml
```diff
@@ -333,3 +333,52 @@ jobs:
...
+              -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
+              -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+              -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+
+            cmake --build . -j16 --target install
+
+      - name: Debug failure
+        shell: bash -l {0}
+        run: |
+            ls /home/runner/micromamba/envs/lp/lib
```
Thirumalai-Shaktivel, line 378: We have to remove this, I used this for the debugging purpose.

.github/workflows/CI.yml
```diff
@@ -333,3 +333,52 @@ jobs:
...
+
```
Thirumalai-Shaktivel, line 379:
```suggestion
```
Please remove this. I used it for debugging purposes.

integration_tests/symbolics_01.py
```diff
@@ -0,0 +1,11 @@
+from sympy import Symbol, pi
+from lpython import S
+
+def main0():
+    x: S = Symbol('x')
+    y: S = Symbol('y')
+    x = pi
+    z: S = x + y
+    print(z)
```
Thirumalai-Shaktivel, line 9: I think, we need to use assert to test the output, something like `assert str(z)  == 'y + pi'`.
This can be done in the subsequent PRs.

src/libasr/asr_utils.h
```diff
@@ -1648,6 +1657,11 @@ inline int extract_dimensions_from_ttype(ASR::ttype_t *x,
             n_dims = extract_dimensions_from_ttype(ASR::down_cast<ASR::Const_t>(x)->m_type, m_dims);
             break;
         }
+        case ASR::ttypeType::SymbolicExpression: {
+            n_dims = 0;
+            m_dims = nullptr;
+            break;
+        }
```
Shaikh-Ubaid, line 1664: This can be merged with the below case. For example:
```suggestion
        case ASR::ttypeType::SymbolicExpression:
```

anutosh491:
I do see some improvements which should be introduced in `SymbolicAdd` etc
But I'll keep it for the next Pr and not disturb this one . As far as we are able to add a `symbol` to `pi` or something , I'll keep this as it is.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -676,6 +678,10 @@ R"(#include <stdio.h>
                             v->m_intent == ASRUtils::intent_return_var ) {
                             d += ";\n";
                         }
+                        if (ASR::is_a<ASR::SymbolicExpression_t>(*v->m_type)) {
+                            std::string v_m_name = v->m_name;
+                            d += indent + "basic_new_stack(" + v_m_name + ");\n";
```
Shaikh-Ubaid, line 683: @anutosh491 do variables allocated using `basic_new_stack()` need to be deallocated (probably using `basic_free_stack()`)? I think it need not be done in this PR, but I am curious to know.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1080,6 +1087,9 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
                         }
                     }
                 }
+            } else if (var_annotation == "S") {
+                type = ASRUtils::TYPE(ASR::make_SymbolicExpression_t(al, loc));
+                return type;
```
Shaikh-Ubaid, line 1092: Should this be one `if-else` level before, that is parallel to `if-else` for `pointer`, `bool`, `str`, etc.? Currently, it seems it is in the `if-else` for user-defined types like `Struct`, `Enum`.  Since `S` here seems to be an internal type, maybe it could make more sense to have this check parallel to the existing internal type checks.

PS: I guess this can be done in a separate PR.

src/libasr/codegen/asr_to_c_cpp.h
anutosh491: Yes , yes . I've had this discussion with @certik and we even went over how we could maybe solve it in the future . We shall be addressing this !

src/lpython/semantics/python_ast_to_asr.cpp
anutosh491: Ohh, this was something Thirumalai and I discussed over . Check this thread out https://github.com/lcompilers/lpython/pull/1846#discussion_r1228155048
We've kept it below `Struct` as `S` , the internal type can also be defined by the user and in that case we would be using it as a user defined type !

src/libasr/codegen/asr_to_c_cpp.h
anutosh491: Resolving .

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -1994,6 +1997,112 @@ namespace Partition {
...
+        return ASR::make_IntrinsicFunction_t(al, loc,
+                static_cast<int64_t>(ASRUtils::IntrinsicFunctions::SymbolicAdd),
+                args.p, args.size(), 0, to_type, compile_time_value);
+    }
+
+} // namespace SymbolicAdd
+
+namespace SymbolicPi {
+
+    static inline void verify_args(const ASR::IntrinsicFunction_t& /*x*/, diag::Diagnostics& /*diagnostics*/) {
```
Shaikh-Ubaid, line 2084: If it does not take arguments, I think we verify here that
```
ASRUtils::require_impl(x.n_args == 0, "SymbolicPi does not take arguments",
            x.base.base.loc, diagnostics);
```

PS: I think it can be done separately

anutosh491: Yes, this is a good catch , could be added to `verify_args`

anutosh491: Resolving as addressed .

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -2383,16 +2393,34 @@ R"(#include <stdio.h>
...
-        src = out;
+        if (x.n_args == 0){
+            src = out;
+        } else if (x.n_args == 1) {
+            this->visit_expr(*x.m_args[0]);
+            if (x.m_intrinsic_id != static_cast<int64_t>(ASRUtils::IntrinsicFunctions::SymbolicSymbol)) {
+                out += "(" + src + ")";
+                src = out;
+            }
+        }
```
Shaikh-Ubaid, line 2423: For Later: Instead of an `if-else` ladder, can we generalize this? For example:
```cpp
out = "(";
this->visit_expr(*x.m_args[0]);
out += src;
for (int i = 1; i < x.n_args; i++) {
    this->visit_expr(*x.m_args[i]);
    out += src;
}
out += ")";
```

anutosh491: Yes , we would be moving towards something like this once we start introducing intrinsic function that deal with multiple arguments like more binary operators like `SymbolicPow` etc and algorithms ( like `limits` which would be intaking a `function` , `symbol` and a `point` about which the limit needs to be calculated as arguments)

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: It might probably be fine for now. I think a better approach could be where we resolve the `type_annotation` at the start of the function `get_type_from_var_annotation` and handle appropriately. For example, instead of moving `else if (var_annotation == "S") ` inside, how about if we resolve the annotation first and bring `if (s) {` outside (probably make it the first `if`) parallel to other internal type checks. That way it also works for other predefined/internal types.

For example, diff with respect to main branch:
```bash
$ git diff
index 0c1783d1a..f529afa9a 100644
@@ -1005,7 +1005,31 @@ public:
         AST::expr_t** m_args=nullptr, size_t n_args=0,
         bool raise_error=true) {
         ASR::ttype_t* type = nullptr;
-        if (var_annotation == "i8") {
+        ASR::symbol_t *s = current_scope->resolve_symbol(var_annotation);
+        if (s) {
+            if (ASR::is_a<ASR::Variable_t>(*s)) {
+                ASR::Variable_t *var_sym = ASR::down_cast<ASR::Variable_t>(s);
+                if (var_sym->m_type->type == ASR::ttypeType::TypeParameter) {
+                    ASR::TypeParameter_t *type_param = ASR::down_cast<ASR::TypeParameter_t>(var_sym->m_type);
+                    type = ASRUtils::TYPE(ASR::make_TypeParameter_t(al, loc, type_param->m_param));
+                    return ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
+                }
+            } else {
+                ASR::symbol_t *der_sym = ASRUtils::symbol_get_past_external(s);
+                if( der_sym ) {
+                    if ( ASR::is_a<ASR::StructType_t>(*der_sym) ) {
+                        type = ASRUtils::TYPE(ASR::make_Struct_t(al, loc, s));
+                        return ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
+                    } else if( ASR::is_a<ASR::EnumType_t>(*der_sym) ) {
+                        type = ASRUtils::TYPE(ASR::make_Enum_t(al, loc, s));
+                        return ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
+                    } else if( ASR::is_a<ASR::UnionType_t>(*der_sym) ) {
+                        type = ASRUtils::TYPE(ASR::make_Union_t(al, loc, s));
+                        return ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
+                    }
+                }
+            }
+        } else if (var_annotation == "i8") {
             type = ASRUtils::TYPE(ASR::make_Integer_t(al, loc, 1));
             type = ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
         } else if (var_annotation == "i16") {
@@ -1055,32 +1079,10 @@ public:
             bool is_allocatable = false;
             type = ast_expr_to_asr_type(underlying_type->base.loc, *underlying_type, is_allocatable);
             type = ASRUtils::TYPE(ASR::make_Pointer_t(al, loc, type));
+        } else if (var_annotation == "S") {
+                type = ASRUtils::TYPE(ASR::make_SymbolicExpression_t(al, loc));
+                return type;
         } else {
-            ASR::symbol_t *s = current_scope->resolve_symbol(var_annotation);
-            if (s) {
-                if (ASR::is_a<ASR::Variable_t>(*s)) {
-                    ASR::Variable_t *var_sym = ASR::down_cast<ASR::Variable_t>(s);
-                    if (var_sym->m_type->type == ASR::ttypeType::TypeParameter) {
-                        ASR::TypeParameter_t *type_param = ASR::down_cast<ASR::TypeParameter_t>(var_sym-
>m_type);
-                        type = ASRUtils::TYPE(ASR::make_TypeParameter_t(al, loc, type_param->m_param));
-                        return ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
-                    }
-                } else {
-                    ASR::symbol_t *der_sym = ASRUtils::symbol_get_past_external(s);
-                    if( der_sym ) {
-                        if ( ASR::is_a<ASR::StructType_t>(*der_sym) ) {
-                            type = ASRUtils::TYPE(ASR::make_Struct_t(al, loc, s));
-                            return ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
-                        } else if( ASR::is_a<ASR::EnumType_t>(*der_sym) ) {
-                            type = ASRUtils::TYPE(ASR::make_Enum_t(al, loc, s));
-                            return ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
-                        } else if( ASR::is_a<ASR::UnionType_t>(*der_sym) ) {
-                            type = ASRUtils::TYPE(ASR::make_Union_t(al, loc, s));
-                            return ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
-                        }
-                    }
-                }
-            }
             if( raise_error ) {
                 throw SemanticError("Unsupported type annotation: " + var_annotation, loc);
             }
(END)
```

Shaikh-Ubaid:
@anutosh491 Please make a note of changes we discussed to be tackled later.

Shaikh-Ubaid (approved):
It looks good to me. Thank you so much for this @anutosh491 !

anutosh491:
> @anutosh491 Please make a note of changes we discussed to be tackled later.

Yes , I've made a note of the things that have been pointed out by you and @Thirumalai-Shaktivel . I'll address them in the upcoming Prs.

Shaikh-Ubaid:
Also another minor point, it seems `S` is a common variable name. For example, it could be used as
```
class S:
...

s: S
```
or 
```bash
S: Student // where Student is some class defined
```
It seems to (easily) conflict with the `LPython` annotation for `Symbolic` type. Is it possible/better to use another name for `LPython` `Symbolic` type?

certik:
We need to eventually allow renaming, like `from lpython import S as Symbolic` or `import lpython; lpython.S`, which will resolve any ambiguities. I think `S` is fine, since it's short and compatible with SymPy. The compiler should give an error message about any conflicts like this, so I think it's not a big problem.

I think `S` won't be used in non-symbolic code, so it's not a problem. In symbolic code like inside SymPy I can imagine `S` as a type being used a lot, so I think it helps if it is short.

certik (approved):
I reviewed by hand as much as I could. I think everything works!

certik:
Great job @anutosh491, this was the hardest. This provides the infrastructure and from here things should go a lot easier. Thank you @Shaikh-Ubaid for your review and help. Thanks @Thirumalai-Shaktivel for your great help with this.

@anutosh491 I would say from here, the next step is to add more symbolic features. Just add new tests and let's start supporting all the arithmetic operations, functions like sin/cos, operations like limit, integrate, diff, series, etc. Just use the C backend and the existing CI setup. As you start adding things, we'll have to iterate on the ASR design as needed. We'll have to eventually figure out how to free things correctly (`basic_free` in symengine).

After we have a reasonable subset working, then we'll tackle the LLVM backend and get all the symbolic tests working with LLVM also.

anutosh491:
Thanks a lot everyone who guided me on the PR . I'll send the second PR by today !

PR #1848, opened 2023-05-22
Support CPtr <-> u64 casting

### LINKED ISSUE #1736: Casting: Implement cptr_to_i64 and i64_to_cptr
We need some native LPython way to do the following casts:
```python
@ccall
def cptr_to_i64(p: CPtr) -> i64:
    pass

@ccall
def i64_to_cptr(p: i64) -> CPtr:
    pass

def increment_p_i16(p: CPtr, n: i32) -> CPtr:
    q: i64 = cptr_to_i64(p)
    q += i64(n * 2)
    return i64_to_cptr(q)
```
Currently they are implemented using:
```c
uint64_t cptr_to_i64(void *p) {
    return (uint64_t)p;
}

void* i64_to_cptr(uint64_t p) {
    return (void*)p;
}
```

Rather, I would like to just do:
```python
from lpython import cptr_to_i64, i64_to_cptr

def increment_p_i16(p: CPtr, n: i32) -> CPtr:
    q: i64 = cptr_to_i64(p)
    q += i64(n * 2)
    return i64_to_cptr(q)
```
And LPython will insert appropriate casting nodes in ASR.

Smit-create:
Fixes #1736


certik (approved):
I think this looks good, thanks for doing that!

certik (approved):

Issue #1849, opened 2023-05-23
Give an error message if function parameter is declared twice


certik:
This should give an error message that `m` is declared twice:
```python
def init_X(m:i32, n: i32, b: CPtr, m: i32) -> None:
    B: Pointer[i16[m*n]] = c_p_pointer(b, i16[m*n])
    i: i32
    j: i32
    for i in range(m):
        for j in range(n):
            B[i * n + j] = i16((i + j) % m)
```

Clang gives the following error message on the generated code:
```console
/Users/ondrej/repos/gsi/copperhead/benchmarks/matmul/build/mat_vec-host-generated.c:224:52: error: redefinition of parameter 'm'
void init_X(int32_t m, int32_t n, void* b, int32_t m)
                                                   ^
/Users/ondrej/repos/gsi/copperhead/benchmarks/matmul/build/mat_vec-host-generated.c:224:21: note: previous declaration is here
void init_X(int32_t m, int32_t n, void* b, int32_t m)
                    ^
```
We should give a similar error message, underlying both declarations of `m`.

PR #1854, opened 2023-05-23
Give an error message if function parameter is declared twice

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1849


certik:
Thanks!

PR #1861, opened 2023-05-26
@ccallable: Emit a header file with given name

### LINKED ISSUE #1797: @ccallable: emit a header file
This feature is the opposite of https://github.com/lcompilers/lpython/issues/1613. When creating some function that is exposed to C, such as:
```python
@ccallable(header="some_header.h")
def f(x: i32) -> f64:
    "Some documentation."
    return x + 1
```
it will not only create the function `f` that is callable from C, but it would also emit / create a C header file `some_header.h` that would contain the C signature, together with the docstring as documentation (using C comments).

We want to be able to specify different `@ccallable` in separate files and they would all contribute to the same C headerfile. Probably one could configure this using a command line argument, but one solution is to create the header file if it does not exist, and then keep appending to it (specifically, inserting into it, since there are typically the ifdef macros at the beginning and end). Another solution is to compile all the separate files into one ASR, and then create the headerfile at once.

We can also only emit the header file if requested at the command line.

Smit-create:
Fixes #1797

### DIFF
--- a/integration_tests/bindc_03.py
+++ b/integration_tests/bindc_03.py
@@ -42,6 +42,14 @@ def h(q_void: CPtr) -> None:
         print(el)
         assert el == i * i + i%2
 
+
+@ccallable(header="_test_bindc_03_my_header.h")
+def test_emit_header_ccallable() -> i32:
+    i: i32 = 5
+    assert i == 5
+    i = i*5
+    return i + 10
+
 def run():
     a: CPtr
     array_wrapped: ArrayWrapped = ArrayWrapped(a)
@@ -59,5 +67,6 @@ def run():
     f(array_wrapped.array)
     array_wrapped1 = array_wrapped
     h(array_wrapped1.array)
+    assert test_emit_header_ccallable() == 35
 
 run()
--- a/src/libasr/codegen/asr_to_c.cpp
+++ b/src/libasr/codegen/asr_to_c.cpp
@@ -656,34 +656,6 @@ R"(
 #include <string.h>
 #include <lfortran_intrinsics.h>
 
-#ifndef ASSERT
-#define ASSERT(cond)                                                           \
-    {                                                                          \
-        if (!(cond)) {                                                         \
-            printf("%s%s", "ASSERT failed: ", __FILE__);                       \
-            printf("%s%s", "\nfunction ", __func__);                           \
-            printf("%s%d%s", "(), line number ", __LINE__, " at \n");          \
-            printf("%s%s", #cond, "\n");                                       \
-            exit(1);                                                           \
-        }                                                                      \
-    }
-#endif
-
-#ifndef ASSERT_MSG
-#define ASSERT_MSG(cond, msg)                                                  \
-    {                                                                          \
-        if (!(cond)) {                                                         \
-            printf("%s%s", "ASSERT failed: ", __FILE__);                       \
-            printf("%s%s", "\nfunction ", __func__);                           \
-            printf("%s%d%s", "(), line number ", __LINE__, " at \n");          \
-            printf("%s%s", #cond, "\n");                                       \
-            printf("%s", "ERROR MESSAGE:\n");                                  \
-            printf("%s%s", msg, "\n");                                         \
-            exit(1);                                                           \
-        }                                                                      \
-    }
-#endif
-
 )";
 
         std::string indent(indentation_level * indentation_spaces, ' ');
--- a/src/libasr/codegen/asr_to_c.cpp
+++ b/src/libasr/codegen/asr_to_c.cpp
@@ -1,4 +1,5 @@
 #include <iostream>
+#include <fstream>
 #include <memory>
 
 #include <libasr/asr.h>
@@ -655,6 +656,7 @@ R"(
 #include <string.h>
 #include <lfortran_intrinsics.h>
 
+#ifndef ASSERT
 #define ASSERT(cond)                                                           \
     {                                                                          \
         if (!(cond)) {                                                         \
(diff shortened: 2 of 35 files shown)


src/libasr/codegen/asr_to_c.cpp
```diff
@@ -677,6 +682,7 @@ R"(
             exit(1);                                                           \
         }                                                                      \
     }
+#endif
```
certik, line 685: I think we should move this into `lfortran_intrinsics.h` that we require anyway.

certik:
Let's add a test for this?

certik (approved):
I think this looks great.

We should add a test that calls this from C and includes the header file, to ensure that everything actually works. We can do that in a next PR.

Smit-create:
Thanks @certik. The test is added and it works.

integration_tests/bindc_03.py
```diff
@@ -42,6 +42,14 @@ def h(q_void: CPtr) -> None:
         print(el)
         assert el == i * i + i%2
 
+
+@ccallable(header="_test_bindc_03_my_header.h")
+def test_emit_header_ccallable() -> i32:
+    i: i32 = 5
+    assert i == 5
+    i = i*5
+    return i + 10
```
Smit-create, line 51: See this test. The header file will be created and will be used in the original file.

certik: Ah, the generated file contains the header file?

I think we still need a handwritten file that includes it, to be sure that all the arguments are compatible.

Smit-create: % lpython --show-c integration_tests/bindc_03.py > b.c
Will generate two file.
1. main file:

```cpp
#include <complex.h>
#include <inttypes.h>
#include "_test_bindc_03_my_header.h"
#include "bindc_03b.h"

#include <stdlib.h>
#include <stdbool.h>
#include <stdio.h>
#include <string.h>


struct dimension_descriptor
{
    int32_t lower_bound, length;
};
struct ArrayWrapped {
 void* array;
};


struct i32
{
    int32_t *data;
    struct dimension_descriptor dims[32];
    int32_t n_dims;
    bool is_allocated;
};


inline void struct_deepcopy_ArrayWrapped(struct ArrayWrapped* src, struct ArrayWrapped* dest);


// Implementations
int32_t __lpython_overloaded_4___lpython_floordiv(int32_t a, int32_t b)
{
    double r;
    int32_t result;
    r = (double)(a)/(double)(b);
    result = (int32_t)(r);
    if (r >=   0.00000000000000000e+00 || (double)(result) == r) {
    }
    _lpython_return_variable = result - 1;
}

int32_t __lpython_overloaded_2___mod(int32_t a, int32_t b)
{
    _lpython_return_variable = a - __lpython_overloaded_4___lpython_floordiv(a, b)*b;
}

float _lfortran_caimag(float complex x);

double _lfortran_zaimag(double complex x);

void gpy(void* a, int32_t value, bool offset_value)
{
    g(a, value, offset_value);
}

void f(void* q_void)
{
    int32_t el;
    int32_t i;
    struct i32 q_value;
    struct i32* q = &q_value;
    q->n_dims = 1;
    q->dims[0].lower_bound = 0;
    q->dims[0].length = 0;
    q->data = (int32_t*) q_void;
    for (i=0; i<=10 - 1; i++) {
        {
            void* q2;
            q2 = (void*) &q->data[(i - q->dims[0].lower_bound)];
            gpy(q2, i*i, (bool)(__lpython_overloaded_2___mod(i, 2)));
            el = q->data[(i - q->dims[0].lower_bound)];
            printf("%d\n", el);
            ASSERT(el == i*i + __lpython_overloaded_2___mod(i, 2));
        }
    }
}

void h(void* q_void)
{
    int32_t el;
    int32_t i;
    struct i32 q_value;
    struct i32* q = &q_value;
    q->n_dims = 1;
    q->dims[0].lower_bound = 0;
    q->dims[0].length = 0;
    q->data = (int32_t*) q_void;
    for (i=0; i<=10 - 1; i++) {
        {
            el = q->data[(i - q->dims[0].lower_bound)];
            printf("%d\n", el);
            ASSERT(el == i*i + __lpython_overloaded_2___mod(i, 2));
        }
    }
}

void run()
{
    void* a;
    struct ArrayWrapped array_wrapped_value;
    struct ArrayWrapped* array_wrapped = &array_wrapped_value;
    struct ArrayWrapped array_wrapped1_value;
    struct ArrayWrapped* array_wrapped1 = &array_wrapped1_value;
    uint64_t q;
    int32_t size;
    void* x;
    array_wrapped->array = a;
    size = 10;
    a = get_array(size);
    ASSERT(a != NULL);
    array_wrapped->array = a;
    f(array_wrapped->array);
    q = (uint64_t)(a);
    x = (void*)(q);
    array_wrapped->array = x;
    f(array_wrapped->array);
    struct_deepcopy_ArrayWrapped(array_wrapped, array_wrapped1);
    h(array_wrapped1->array);
    ASSERT(test_emit_header_ccallable() == 35);
}

void _lpython_main_program()
{
    run();
}

{
    _lpython_set_argv(argc, argv);
    _lpython_main_program();
    return 0;
}

void struct_deepcopy_ArrayWrapped(struct ArrayWrapped* src, struct ArrayWrapped* dest) {
    dest->array =  src->array;
}
```

2. header file:

```cpp
#ifndef _TEST_BINDC_03_MY_HEADER_H
#define _TEST_BINDC_03_MY_HEADER_H

#include <complex.h>
#include <inttypes.h>

#include <stdlib.h>
#include <stdbool.h>
#include <stdio.h>
#include <string.h>


int32_t test_emit_header_ccallable()
{
    int32_t i;
    i = 5;
    ASSERT(i == 5);
    i = i*5;
    _lpython_return_variable = i + 10;
}



#endif
```

Smit-create: Yeah, you're right. `ccallable` needs to be updated in the LLVM backend so that it can call from C.

certik: The `test_emit_header_ccallable` must be in a C file, not a header file, otherwise the function will be generated twice by the C compiler and you will get linking errors.

PR #1863, opened 2023-05-27
Initial support for @pythoncall decorator

Shaikh-Ubaid:
towards https://github.com/lcompilers/lpython/issues/703.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -36,6 +36,15 @@ set_property(TARGET lpython_rtlib PROPERTY INTERFACE_LINK_LIBRARIES
     ${LPYTHON_RTLIB_LIBRARY})
 target_link_libraries(lpython_rtlib INTERFACE m)
 
+find_package(Python COMPONENTS Development)
+message("\n")
+message("System has the Python development artifacts: ${Python_Development_FOUND}")
+message("The Python include directories: ${Python_INCLUDE_DIRS}")
+message("The Python libraries: ${Python_LIBRARIES}")
+message("The Python library directories: ${Python_LIBRARY_DIRS}")
+message("The Python runtime library directories: ${Python_RUNTIME_LIBRARY_DIRS}")
+message("Python version: ${Python_VERSION}")
+
 enable_testing()
 
 message("\n")
@@ -57,7 +66,7 @@ message("LPYTHON_RTLIB_LIBRARY: ${LPYTHON_RTLIB_LIBRARY}")
 
 
 macro(RUN)
-    set(options FAIL)
+    set(options FAIL ENABLE_CPYTHON)
     set(oneValueArgs NAME IMPORT_PATH)
     set(multiValueArgs LABELS EXTRAFILES)
     cmake_parse_arguments(RUN "${options}" "${oneValueArgs}"
@@ -66,6 +75,11 @@ macro(RUN)
     set(import_path ${RUN_IMPORT_PATH})
     if (NOT name)
         message(FATAL_ERROR "Must specify the NAME argument")
+    endif()
+
+    set(enable_cpython "")
+    if (${RUN_ENABLE_CPYTHON})
+        set(enable_cpython "--enable-cpython")
     endif()
 
     if (${KIND} IN_LIST RUN_LABELS)
@@ -97,13 +111,13 @@ macro(RUN)
             if (import_path)
                 add_custom_command(
                     OUTPUT ${name}.c
-                    COMMAND ${LPYTHON} -I ${CMAKE_CURRENT_SOURCE_DIR}/${import_path} --show-c ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py > ${name}.c
+                    COMMAND ${LPYTHON} ${enable_cpython} -I ${CMAKE_CURRENT_SOURCE_DIR}/${import_path} --show-c ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py > ${name}.c
                     DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
                     VERBATIM)
             else ()
                 add_custom_command(
                     OUTPUT ${name}.c
-                    COMMAND ${LPYTHON} --show-c ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py > ${name}.c
+                    COMMAND ${LPYTHON} ${enable_cpython} --show-c ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py > ${name}.c
                     DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
                     VERBATIM)
             endif()
@@ -112,6 +126,12 @@ macro(RUN)
             target_include_directories(${name} PRIVATE ${CMAKE_SOURCE_DIR})
             set_target_properties(${name} PROPERTIES LINKER_LANGUAGE C)
             target_link_libraries(${name} lpython_rtlib)
+            if (${RUN_ENABLE_CPYTHON})
+                target_link_libraries(${name} Python::Python)
+                if (RUN_EXTRAFILES)
+                    file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/${RUN_EXTRAFILES} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+                endif()
+            endif()
             add_test(${name} ${CMAKE_CURRENT_BINARY_DIR}/${name})
             if (RUN_LABELS)
                 set_tests_properties(${name} PROPERTIES LABELS "${RUN_LABELS}")
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -476,6 +476,7 @@ RUN(NAME bindc_05   LABELS llvm c
         EXTRAFILES bindc_05b.c)
 RUN(NAME bindc_06   LABELS llvm c
         EXTRAFILES bindc_06b.c)
+RUN(NAME bindpy_01           LABELS cpython c ENABLE_CPYTHON EXTRAFILES bindpy_01_module.py)
 RUN(NAME test_generics_01    LABELS cpython llvm c)
 RUN(NAME test_cmath          LABELS cpython llvm c)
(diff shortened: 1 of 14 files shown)


Shaikh-Ubaid:
Example:
```bash
$ cat examples/expr3.py 
def main0():
    n: i32 = 5
    m: i32 = fun_from_py_module(n)
    print("hi")
    print("m recieved is: ", m)
    print("this time value received is:", fun_from_py_module(-22))
    print("And this time value received is:", fun_from_py_module(101))

    print("cpython_average(2, 3): ", cpython_average(2, 3))
    print("cpython_average(-100, -1): ", cpython_average(-100, -1))

    cpython_print_str("hi how are you?")
    pqr: str = "abcdefgh"
    cpython_print_str("hi how are you?: " + pqr)

    abc: str = cpython_return_random_str()
    print("abc: ", abc)

    print(cpython_str_concat(abc, pqr))

@pythoncall(module="py_module")
def cpython_str_concat(a: str, b: str)->str:
    pass

@pythoncall(module="py_module")
def fun_from_py_module(n: i32) -> i32:
    pass

@pythoncall(module="py_module")
def cpython_average(n: i32, m: i32) -> f64:
    pass

@pythoncall(module="py_module")
def cpython_print_str(x: str):
    pass

@pythoncall(module="py_module")
def cpython_return_random_str() -> str:
    pass

main0()
$ cat py_module.py 
def fun_from_py_module(n):
    import os
    print(os.getcwd())
    print("hi from cpython", n)
    return n + 2

def cpython_average(n, m):
    return (m + n) / 3

def cpython_print_str(x: str):
    print("I received the string:", x)

def cpython_return_random_str() -> str:
    return "this is some string"

def cpython_str_concat(a: str, b: str)->str:
    return "concatenated: " + a + " " + b
$ lpython examples/expr3.py --backend c
/Users/ubaid/Desktop/OpenSource/lpython
hi from cpython 5
hi
m recieved is:  7
/Users/ubaid/Desktop/OpenSource/lpython
hi from cpython -22
this time value received is: -20
/Users/ubaid/Desktop/OpenSource/lpython
hi from cpython 101
And this time value received is: 103
cpython_average(2, 3):  1.666667
cpython_average(-100, -1):  -33.666667
I received the string: hi how are you?
I received the string: hi how are you?: abcdefgh
abc:  this is some string
concatenated: this is some string abcdefgh
```
Note: `py_module.py` is in the root directory of the `lpython` project.

Shaikh-Ubaid:
WIP: CI Tests may/might not pass.

Shaikh-Ubaid:
Example from https://github.com/lcompilers/lpython/issues/703#issuecomment-1561538852,

```bash
$ cat examples/expr2.py 

@pythoncall(module="py_module")
def f(n: i32) -> str:
    pass

def main0():
    p: str
    p = f(3)
    print(p)

main0()

$ cat py_module.py 
def f(n) -> str:
    import sympy
    sympy.var("x")
    e = ((x+5)**n).expand()
    return str(e)
$ lpython examples/expr2.py --backend c         
x**3 + 15*x**2 + 75*x + 125
```

src/bin/lpython.cpp
```diff
@@ -1246,6 +1246,9 @@ int link_executable(const std::vector<std::string> &infiles,
         cmd += " -I " + rtlib_header_dir;
         cmd += " -L" + base_path
             + " -Wl,-rpath," + base_path + " -l" + runtime_lib + " -lm";
+        std::string py_version = "3.10";
+        std::string py_flags = R"(-I $CONDA_PREFIX/include/python)" + py_version + R"( -L$CONDA_PREFIX/lib -Wl,-rpath -Wl,$CONDA_PREFIX/lib -lpython)" + py_version + R"()";
+        cmd += " " + py_flags;
```
Shaikh-Ubaid, line 1251: I think this should be optional. We need these flags when there is a `@pythoncall()` in the user source code. Shall we add a compiler flag `--enable-cpython-runtime`? We also need the python version here so that we can link to the `pythonlib`. How do we get the python version that is installed on the system?

Shaikh-Ubaid:
How do we add support for import path (`-I` option)? The import paths are available with `lpython`, but not available with the generated `C` code. It is the generated `C` code which tries to open/read the `module_file` and evaluate the function present in it. I think we would need to provide/embed all the import paths into the generated `C` code to support module_files from user supplied import paths. 

The examples in https://github.com/lcompilers/lpython/pull/1863#issuecomment-1565357424 and https://github.com/lcompilers/lpython/pull/1863#issuecomment-1565395212 do not need `import` path since the module `py_module.py` is present in the current execution directory which is by default accessible.

src/libasr/codegen/c_utils.h
```diff
@@ -330,6 +330,87 @@ namespace CUtils {
...
+                        throw CodeGenError("get_py_obj_type_conv_func: Unsupported kind in unsigned int type");
+                }
+                break;
+            }
+            case ASR::ttypeType::Real: {
+                type_src = "PyFloat_AsDouble";
+                break;
+            }
+            case ASR::ttypeType::Character: {
+                type_src = "(char*)PyUnicode_AsUTF8";
```
Shaikh-Ubaid, line 405: `PyUnicode_AsUTF8()` returns a `const char *`. We are casting it here to `char *`.

src/bin/lpython.cpp
certik: Maybe just `--enable-cpython` and the Python version can be either specified at the command line. Alternatively we can specify the include path and the library name and path directly.

integration_tests/bindc_03.py
```diff
@@ -7,14 +7,14 @@
 class ArrayWrapped:
     array: CPtr
 
-@ccall(header="bindc_03b.h")
+@ccall(module="bindc_03b.h")
```
certik, line 10: I would call it a `header`, I think it's more intuitive what it means.

certik: The `@pythoncall(module="py_module")` usage is fine.

certik:
This looks great, thanks @Shaikh-Ubaid for working on this. Let's figure out how to test this, probably we need to add some new "label" into the integration_tests/CMakeLists.txt for this.

integration_tests/bindc_03.py
Shaikh-Ubaid: Got it. I reverted it back to using `header`. I added support both keywords `header` and `module`.

src/libasr/codegen/c_utils.h
Shaikh-Ubaid: I think we should create a copy of the `const char*` string. I think/guess we can support it in another PR.

src/bin/lpython.cpp
Shaikh-Ubaid: I added the flag `--enable-cpython`. Currently, I hardcoded the python_version and the include paths for python library.

integration_tests/CMakeLists.txt
```diff
@@ -112,6 +126,12 @@ macro(RUN)
             target_include_directories(${name} PRIVATE ${CMAKE_SOURCE_DIR})
             set_target_properties(${name} PROPERTIES LINKER_LANGUAGE C)
             target_link_libraries(${name} lpython_rtlib)
+            if (${RUN_ENABLE_CPYTHON})
+                target_link_libraries(${name} Python::Python)
+                if (RUN_EXTRAFILES)
+                    file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/${RUN_EXTRAFILES} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
Shaikh-Ubaid, line 132: With respect to https://github.com/lcompilers/lpython/pull/1863#issuecomment-1565401214, we do not yet support the `-I` option for importing/including the cpython module file. The Python module file can be accessed if it is in the same directory as of the `lpython` generated `--show-c` output file. Currently, to make the integration_tests to work we are copying the CPython module file to the binary directory so that it is present in the same directory as of the `lpython` generated `--show-c` output file.

Shaikh-Ubaid:
Ready.

integration_tests/CMakeLists.txt
```diff
@@ -456,6 +476,7 @@ RUN(NAME bindc_05   LABELS llvm c
         EXTRAFILES bindc_05b.c)
 RUN(NAME bindc_06   LABELS llvm c
         EXTRAFILES bindc_06b.c)
+RUN(NAME bindpy_01           LABELS c ENABLE_CPYTHON EXTRAFILES bindpy_01_module.py)
```
certik, line 479: Does this work with `cpython`? We should try to implement it, to make sure all the syntax works.

integration_tests/CMakeLists.txt
certik: That's fine, thanks!

integration_tests/bindpy_01.py
```diff
@@ -0,0 +1,77 @@
+from lpython import i32, i64, u32, u64, f32, f64
```
certik, line 1: we have to import `pythoncall` from here, and implement it in cpython mode.

src/bin/lpython.cpp
```diff
@@ -1246,6 +1246,11 @@ int link_executable(const std::vector<std::string> &infiles,
         cmd += " -I " + rtlib_header_dir;
         cmd += " -L" + base_path
             + " -Wl,-rpath," + base_path + " -l" + runtime_lib + " -lm";
+        if (compiler_options.enable_cpython) {
+            std::string py_version = "3.10";
+            std::string py_flags = R"(-I $CONDA_PREFIX/include/python)" + py_version + R"( -L$CONDA_PREFIX/lib -Wl,-rpath -Wl,$CONDA_PREFIX/lib -lpython)" + py_version + R"()";
```
certik, line 1251: I think that's fine for now. Conda is currently required.

certik:
I think this looks good. I would still get cpython working. That way the test will be `cpython` and `c`, which is fine. We can add `llvm` later.

Regarding how thing are handled at the ASR level, @czgdp1807 can you please review this? I think we discussed that we would insert explicit type conversions in ASR between our integer and BindPython integer. However I think we can also do that in later PRs.

certik (approved):
I think this looks good now, good enough to merge.

We can then iteratively improve this.

PR #1873, opened 2023-06-03
C: Support more intrinsic functions

Smit-create:
https://github.com/lfortran/lfortran/pull/1747


src/libasr/codegen/asr_to_c.cpp
```diff
@@ -806,6 +818,11 @@ R"(
     }
 
     void visit_Module(const ASR::Module_t &x) {
+        if (startswith(x.m_name, "lfortran_intrinsic_")) {
+            intrinsic_module = true;
+        } else {
+            intrinsic_module = false;
+        }
```
Shaikh-Ubaid, line 825: https://github.com/lfortran/lfortran/pull/1747#discussion_r1215537517

certik: We can fix this in a separate PR at all places (this is not the only occurrence).

certik:
These are good changes. It would be good to figure out how to test them, and possibly we can simply test it in LFortran. Let's keep this open and merge it in LFortran first.

certik:
I'll get the Fortran version in shape first.

Smit-create:
@certik The change is reflected in LPython after the sync. Just added a test so that we can merge this.

Smit-create:
Oh, the tests fail because of infinite recursion caused due to:
````c
// in main.c
double log(double x)
{
    double _lpython_return_variable;
    _lpython_return_variable = _lfortran_dlog(x);
}

// in lfortran_intrinsics.h
LFORTRAN_API double _lfortran_dlog(double x)
{
    return log(x);
}

````

czgdp1807:
Yes. I have experienced this too. I don't know how to fix this.

certik:
That seems like a bug in the runtime library.

Thirumalai-Shaktivel:
I think mangling the log name here would fix the issue: https://github.com/lcompilers/lpython/blob/d78559ba5eb15ba45af7218cdedc703c60a9ad06/src/runtime/math.py#L579-L580

PR #1877, opened 2023-06-04
Use add_or_overwrite_symbol

### LINKED ISSUE #1875: Tests fail locally
See https://lfortran.zulipchat.com/#narrow/stream/311866-LPython/topic/Integration.20Tests

certik:
Fixes #1875.

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -395,7 +395,7 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
             a_mod->m_intrinsic = intrinsic;
             mod2 = a_mod;
         }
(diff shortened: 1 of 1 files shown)


certik:
However often times using this function means there is a bug somewhere. Is replacing the older symbol intended here? We would need to investigate.

certik:
The other question is why this didn't trigger at the CI. Either it only happens on M1, or the assertions don't trigger at the CI.

certik:
I think we never test in Debug mode.

Shaikh-Ubaid:
>The other question is why this didn't trigger at the CI. Either it only happens on M1, or the assertions don't trigger at the CI.
>I think we never test in Debug mode.

Yes, we test release mode at the CI. In release mode, assertions are disabled and I think hence, this passes at the CI.

PR #1878, opened 2023-06-04
CI: Test LPython in Debug mode

certik:


certik:
Once this PR works, it should fail due to #1875. Then I want to merge #1877 and then check that this PR passes. Then we can merge it.

certik:
It now fails correctly with:
```
LCOMPILERS_ASSERT failed: /home/runner/work/lpython/lpython/src/libasr/asr_scopes.h
function add_symbol(), line number 76 at 
scope.find(name) == scope.end()
make[2]: *** [CMakeFiles/elemental_04.dir/build.make:74: elemental_04.o] Aborted (core dumped)
make[1]: *** [CMakeFiles/Makefile2:6130: CMakeFiles/elemental_04.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

certik:
Now the PR should pass.

certik:
The new test now passes. So I think this confirms that it works as expected.

PR #1879, opened 2023-06-04
Fix zero slice step

### LINKED ISSUE #1833: Edge case in slice step of Lists
Slice step 0 not checked.

```
def g():
    lst:list[i32] = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
    print(lst[0:10:0])
g()

(lp) faze-geek@DESKTOP-497JDCU:~/lpython/lpython$ python try.py
    print(lst[0:10:0])
          ~~~^^^^^^^^
ValueError: slice step cannot be zero
(lp) faze-geek@DESKTOP-497JDCU:~/lpython/lpython$ ./src/bin/lpython try.py
[]
```

kabra1110:
Fixes #1833.
I have added the check for any slice step. This holds for strings and lists.
Are there places where zero step is allowed? (I don't think so.)

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -3469,7 +3469,8 @@ public:
                     throw SemanticError("slice indices must be integers or None", tmp->loc);
                 }
                 ASR::expr_t* val = ASRUtils::expr_value(ASRUtils::EXPR(tmp_step));
-                if (val != nullptr && ASR::down_cast<ASR::IntegerConstant_t>(val)->m_n == 0) {
+                int64_t const_value = 1;
+                if (ASRUtils::is_value_constant(val, const_value) && const_value == 0) {
                     throw SemanticError("slice step cannot be zero", tmp_step->loc);
                 }
                 ai.m_step = ASRUtils::EXPR(tmp);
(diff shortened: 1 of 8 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3455,9 +3455,14 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
             }
             if (sl->m_step != nullptr) {
                 this->visit_expr(*sl->m_step);
+                ASR::asr_t *tmp_step = tmp;
                 if (!ASRUtils::is_integer(*ASRUtils::expr_type(ASRUtils::EXPR(tmp)))) {
                     throw SemanticError("slice indices must be integers or None", tmp->loc);
                 }
+                ASR::expr_t* val = ASRUtils::expr_value(ASRUtils::EXPR(tmp_step));
+                if (val != nullptr && ASR::down_cast<ASR::IntegerConstant_t>(val)->m_n == 0) {
```
Thirumalai-Shaktivel, line 3463: You can also use `is_value_constant`, but this is also fine, I think.
LGTM!

czgdp1807 (changes requested):
Address https://github.com/lcompilers/lpython/pull/1879/files#r1217472487 and then it should be good.

czgdp1807 (approved):

PR #1884, opened 2023-06-08
Accept shape in ``c_p_pointer`` if target is an array

czgdp1807:
Also, all types under `Pointer` should be of deferred shape. The same rule as Fortran.


certik:
I will investigate this later today how Fortran does it. We definitely want this feature, but we might also want to allow setting the size using a function argument (as it was before), so let me investigate first and I'll post here later.

certik (approved):
I think this looks great.

I didn't realize before that you just shifted the dimension/shape from the type to an argument, which is better.

Thanks!

PR #1885, opened 2023-06-08
Support Escaping/Unescaping strings in tokenizer

Shaikh-Ubaid:


Shaikh-Ubaid:
The step `Run mamba-org/provision-with-micromamba@main` seems to fail for three jobs at the CI.

Shaikh-Ubaid:
```bash
(lp) ubaid@ubaids-MacBook-Pro lpython % cat examples/expr2.py                     
r"""
Text
123\n\t
"""
(lp) ubaid@ubaids-MacBook-Pro lpython % python -m tokenize examples/expr2.py      
0,0-0,0:            ENCODING       'utf-8'        
1,0-4,3:            STRING         'r"""\nText\n123\\n\\t\n"""'
4,3-4,4:            NEWLINE        '\n'           
5,0-5,0:            ENDMARKER      ''             
(lp) ubaid@ubaids-MacBook-Pro lpython % lpython examples/expr2.py --show-ast        
(Module
    [(Expr
        (ConstantStr
            "\nText\n123\n\t\n"
            ()
        )
    )]
    []
)
```

In the above example, we have a raw string. For raw strings, escape sequences are treated as regular characters. We see that the current output of `lpython` is not as expected. Previously, the unescaping of string was done at the parser level, where we did not unescape strings which have `PREFIX_STRING` (see [semantics.h](https://github.com/lcompilers/lpython/blob/main/src/lpython/parser/semantics.h#L809)) of `r`.

certik:
Good point. I see two ways forward:

* unescape at the parser level
* add a special token for raw strings of the type `r"xxx"`, and then handle unescaping correctly. Modify the parser accordingly.

Shaikh-Ubaid:
>add a special token for raw strings of the type r"xxx", and then handle unescaping correctly. Modify the parser accordingly.

It seems that adding token for strings is possible, but is leading to many new tokens being added (raw_strings, bytes, raw_bytes, formatted strings, raw_formatted strings, etc.)

```
raw_str1 = ("r" | "R") (string1 | string2);
raw_str2 = ("r" | "R") (string3 | string4);
bytes1 = ("b" | "B") (string1 | string2);
bytes2 = ("b" | "B") (string3 | string4);
raw_bytes1 = ("rb" | "rB" | "Rb" | "RB" | "br" | "bR" | "Br" | "BR") (string1 | string2);
raw_bytes2 = ("rb" | "rB" | "Rb" | "RB" | "br" | "bR" | "Br" | "BR") (string3 | string4);
...
```

For the moment it seems it is better to proceed with unescaping at the parser level.

certik:
I think if you use `'rb'` then it is case insensitive in re2c. But it's fine.

certik:
Looks good to me. @Shaikh-Ubaid let me know when this is ready for review.

tests/reference/ast_new-string1-96b90b3.stdout
```diff
@@ -499,20 +499,20 @@
...
         (ConstantBytes
-            "b'\n'"
+            "b'\\n'"
             ()
         )
     )
     (Expr
         (ConstantBytes
-            "b'\n\n\\n'"
+            "b'\n\\n\\\\n'"
```
Shaikh-Ubaid, line 515: In the example above, the given string is
```
b'''
\n\\n'''
```

With respect to this example, it seems that all strings (including raw strings, bytes, raw bytes, etc.) need unescaping at the tokenizer level. And then we need to escape them if they are raw strings/ bytes/ raw bytes, etc (I think at the parser level).

certik: Why do raw strings have to be escaped?

Shaikh-Ubaid: I think you are right, Sir. I will look into it.

src/libasr/asdl_cpp.py
```diff
@@ -1934,12 +1934,12 @@ def visitField(self, field, cons):
...
                 if field.opt:
                     self.emit("if (x.m_%s) {" % field.name, 2)
-                    self.emit(    's.append("\\"" + get_escaped_str(x.m_%s) + "\\"");' % field.name, 3)
+                    self.emit(    's.append("\\"" + str_escape_c(x.m_%s) + "\\"");' % field.name, 3)
                     self.emit("} else {", 2)
                     self.emit(    's.append("[]");', 3)
                     self.emit("}", 2)
                 else:
-                    self.emit('s.append("\\"" + get_escaped_str(x.m_%s) + "\\"");' % field.name, 2)
+                    self.emit('s.append("\\"" + str_escape_c(x.m_%s) + "\\"");' % field.name, 2)
```
certik, line 1942: I would keep this change.

tests/reference/wat-bool1-234bcd1.stdout
```diff
@@ -184,7 +184,7 @@
     (data (;0;) (i32.const 4) "\0c\00\00\00\01\00\00\00")
     (data (;1;) (i32.const 12) "    ")
     (data (;2;) (i32.const 16) "\18\00\00\00\01\00\00\00")
-    (data (;3;) (i32.const 24) "\0a   ")
+    (data (;3;) (i32.const 24) "\n   ")
```
certik, line 187: Let's keep this change.

Shaikh-Ubaid:
See https://github.com/lcompilers/lpython/pull/1902#issue-1755530251.

Shaikh-Ubaid:
Ready.

Shaikh-Ubaid:
> Good point. I see two ways forward:
> 
> * unescape at the parser level
> * add a special token for raw strings of the type `r"xxx"`, and then handle unescaping correctly. Modify the parser accordingly.

This PR currently unescapes at tokenizer level. (LFortran also currently unescapes at tokenizer level since it needs to know the quote used to create the string literal).

The concern in the current approach of unescaping at the tokenizer level in LPython is that we are needing to create several/many new tokens like raw strings, unicode strings, formatted strings, raw formatted strings, bytes, raw bytes (and these also need support at the parser level).

Shaikh-Ubaid:
In the current approach, we unescape all `strings`, `bytes` unless it is marked `raw` in which case we do not unescape (and therefore the escape sequences present (if any) in the raw string/bytes would be treated as character literals).

Shaikh-Ubaid:
Ready.

tests/reference/ast_new-string1-96b90b3.stdout
```diff
@@ -129,7 +129,7 @@
     (Expr
         (ConstantStr
             "Text"
-            ()
+            "u"
```
certik, line 132: This change fixes a bug?

certik:
Thanks for this. I think you might have found and fixed another bug (https://github.com/lcompilers/lpython/pull/1885#discussion_r1228837650).

That bug aside, it does seem this approach is more complicated than doing it in the parser. Let's keep this PR open for a while and think about it.

@czgdp1807 let us know your opinion on this approach as well.

tests/reference/ast_new-string1-96b90b3.stdout
Shaikh-Ubaid: It seems that the prefixes "u" and "U" were meaningful for Python 2. In Python 3, all string literals are Unicode by default. Thus, it seems the prefixes "u" and "U" do not add much/any value in Python 3 (Reference https://chat.openai.com/share/1270fc0a-fb93-4638-822d-3d1619488027).

With respect to the above, I think it does not as such fix a bug. Previously, we were passing the kind only for "u" (prefix). Currently, we also pass the kind value for "U" (capital `U`) (prefix).

czgdp1807 (commented):
What's the status of this? Do we need this?

Shaikh-Ubaid:
>What's the status of this? Do we need this?

We need your opinion on it. Please, let us know what we should do. 

We currently handle unescaping at the parser level for LPython (and tokenizer level for LFortran). This PR supports unescaping at tokenizer level for LPython. 

Any approach works for me. Please feel free to close/merge it as required.

czgdp1807:
If there is no significant benefit of bringing un-escaping to the tokenizer level in LPython, then I wouldn't work on it for the time being. There are higher priority things to be done (like bug fixing, performance improvements to perform better than LPython's competitors), so let's work on those.

PR #1891, opened 2023-06-12
Handle array as return variable using the lpython decorator

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
Sometimes I get:
```
python(12438,0x1e33a1e00) malloc: Heap corruption detected, free list is damaged at 0x600000e7cb00
*** Incorrect guard value: 4634274385308418048
python(12438,0x1e33a1e00) malloc: *** set a breakpoint in malloc_error_break to debug
```
we need to fix this!

integration_tests/lpython_decorator_02.py
```diff
@@ -0,0 +1,15 @@
+from numpy import array
+from lpython import i32, f64, lpython
+
+@lpython
+def multiply(n: i32, x: f64[:]) -> f64[:]:
+    i: i32
+    for i in range(n):
+        x[i] *= 5.0
+    return x
```
Thirumalai-Shaktivel, line 9: Here is the generated C code for the above function:
```C
void multiply(int32_t n, struct r64* x, struct r64* _lpython_return_variable)
{
    int32_t __1_t;
    int32_t __1_v;
    int32_t i;
    for (i=0; i<=n - 1; i++) {
        x->data[(i - x->dims[0].lower_bound)] = x->data[(i - x->dims[0].lower_bound)]*  5.00000000000000000e+00;
    }
    __1_v = ((int32_t)x->dims[1-1].lower_bound);
    for (__1_t=((int32_t)_lpython_return_variable->dims[1-1].lower_bound); __1_t<=((int32_t) _lpython_return_variable->dims[1-1].length + _lpython_return_variable->dims[1-1].lower_bound - 1); __1_t++) {
        _lpython_return_variable->data[(__1_t - _lpython_return_variable->dims[0].lower_bound)] = x->data[(__1_v - x->dims[0].lower_bound)];
        __1_v = __1_v + 1;
    }
    return;
}
```
Here, I think the second for loop iteration should access the `x` variable instead of `_lpython_return_variable`. I need to investigate more about this.

integration_tests/lpython_decorator_02.py
```diff
@@ -0,0 +1,15 @@
+from numpy import array
+from lpython import i32, f64, lpython
+
+@lpython
+def multiply(n: i32, x: f64[:]) -> f64[:]:
```
certik, line 5: This needs to be changed to:
```python
n = TypeVar("n")
def multiply(n: i32, x: f64[:]) -> f64[n]:
```
The return array must have a length that is determined using the function arguments.

Thirumalai-Shaktivel:
This: https://github.com/lcompilers/lpython/pull/1891#issuecomment-1587247957 is fixed!

integration_tests/lpython_decorator_02.py
certik: The second loop seems to copy `x` into `_lpython_return_variable`, which seems correct:
```
        _lpython_return_variable->data[(__1_t - _lpython_return_variable->dims[0].lower_bound)] = x->data[(__1_v - x->dims[0].lower_bound)];
```

certik (approved):
I think this looks good.

Issue #1893, opened 2023-06-12
Hide the "numpy" module unless `--with-intrinsic-mods` is used


certik:
The ASR shows:
```
            numpy:
                (Module
                    (SymbolTable
                        3
                        {
                            __lpython_overloaded_0__arccos:
                                (Function
                                    (SymbolTable
                                        41
                                        {
                                            _lpython_return_variable:
```
But this should be hidden unless `--with-intrinsic-mods` is used.

PR #1895, opened 2023-06-12
Pickle: Hide "numpy" module unless `--with-intrinsic-mods` is used

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1893


Shaikh-Ubaid:
It seems that this issue https://github.com/lfortran/lfortran/pull/1805#issuecomment-1588136251 is being experienced here at LPython as well. I guess there is some issue at GitHub's end?

Shaikh-Ubaid:
Ready.

certik (approved):
I think that looks good. Maybe later we can simply mark the "numpy" ASR Module as intrinsic as well, so the checks will simplify. For now this is good, thanks!

PR #1897, opened 2023-06-13
Print unsigned ints correctly

### LINKED ISSUE #1896: Fix printing of unsigned integers
```python
def main():
    i: u8 = u8(255)
    print(i)

main()
```
This prints:
```console
$ lpython a.py
-1
```
But should print 255.

certik:
There are no tests, but there were no tests before and at least this PR makes the printing work, so it is worth merging.

Fixes #1896.


Smit-create (approved):
Looks good. We can improve this later anytime. Thanks.

Issue #1899, opened 2023-06-13
ASR verify: check that only signed integers are allowed in dimensions


certik:
See https://github.com/lcompilers/lpython/issues/1896#issuecomment-1589361311.

PR #1903, opened 2023-06-13
ASRVerify: Require signed integer array dimensions

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1899


certik (approved):

certik:
> @certik How do we test this?

Right now just by hand. We should eventually test all of these ASR verify checks systematically, one way to do that is via the MASR tool that @rebcabin is developing. We can setup randomized tests and ensure that if the ASR is outside of the "spec" that verify() always fails, and if it is inside the "spec" that it always succeeds. 

We'll focus on this after we deliver LFortran/LPython to beta. Right now just best effort solid ASR verify() is all we can do.

PR #1904, opened 2023-06-14
Allow duplicate imports

certik:
Give a warning.


tests/reference/asr_json-modules_02-70a491a.stderr
```diff
@@ -0,0 +1,14 @@
+[0;33;1mwarning[0;0m[0;1m: The symbol 'f' imported from modules_02b will shadow the existing symbol 'f'[0;0m
+ [0;34;1m-->[0;0m tests/../integration_tests/modules_02.py:1:28
```
Shaikh-Ubaid, line 2: It seems to be printing warnings in coloured format. Is it expected? I guess that the stderr outputs of existing reference tests do not contain colour printing codes.

certik: I think it's the JSON output that does not turn off colors. Can you create an issue for that? We ca fix it in a subsequent PR.

PR #1913, opened 2023-06-15
lpython.py: Add missing slots

Shaikh-Ubaid:
In, Out, InOut and dataclass were added to __slots__.


certik:
I think we can remove `__slots__` completely, it doesn't seem to restrict imports, so it doesn't serve a purpose. I misunderstood the feature I think.

PR #1931, opened 2023-06-17
Pythoncall: Support arrays of int, real as args

Shaikh-Ubaid:
towards https://github.com/lcompilers/lpython/issues/703

This PR supports passing of arrays of types int, real to @ pythoncall decorated functions. Returning arrays of simple types is yet to be supported.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
 
 project(lpython_tests C)
 
@@ -37,7 +37,13 @@ set_property(TARGET lpython_rtlib PROPERTY INTERFACE_LINK_LIBRARIES
     ${LPYTHON_RTLIB_LIBRARY})
 target_link_libraries(lpython_rtlib INTERFACE m)
 
-find_package(Python COMPONENTS Development)
+find_package(Python COMPONENTS Interpreter Development)
+execute_process(
+  COMMAND "${Python_EXECUTABLE}"
+  -c "import numpy; print(numpy.get_include())"
+  OUTPUT_VARIABLE NUMPY_INCLUDE_DIR
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
 message("\n")
 message("System has the Python development artifacts: ${Python_Development_FOUND}")
 message("The Python include directories: ${Python_INCLUDE_DIRS}")
@@ -45,6 +51,7 @@ message("The Python libraries: ${Python_LIBRARIES}")
 message("The Python library directories: ${Python_LIBRARY_DIRS}")
 message("The Python runtime library directories: ${Python_RUNTIME_LIBRARY_DIRS}")
 message("Python version: ${Python_VERSION}")
+message("Numpy Include Directory: ${NUMPY_INCLUDE_DIR}")
 
 enable_testing()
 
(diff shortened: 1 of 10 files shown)


integration_tests/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
```
Shaikh-Ubaid, line 1: I updated the cmake minimum version here. `cmake` with version `3.5`  is unable to find `python` from the `conda` environment (It always picks the system level `python`). It seems `cmake` with version `15` and above supports finding python from `conda` environment (reference https://gitlab.kitware.com/cmake/cmake/-/issues/21322#note_845923). `Python` from conda environment is needed as the conda environment contains the installed numpy package.

Shaikh-Ubaid:
```bash
$ cat expr2.py
from lpython import pythoncall, f32

@pythoncall(module = "expr2_module")
def return_multi_dim_arr() -> f32[2, 3]:
    pass

def main0():
    x: f32[2, 3] = return_multi_dim_arr()
    print("x = ", x)

main0()
$ cat expr2_module.py 
import numpy as np

def return_multi_dim_arr():
    a = np.array([3.12, -2.14, 1., 4., 5., 6.]).reshape((2, 3))
    print("CPython: ", a)
    return a
```
In the above example, the function `def return_multi_dim_arr() -> f32[2, 3]` gets transformed to `def return_multi_dim_arr(struct r32* _lpython_return_variable)` (I think via the `create_subroutine_from_function()` and/or `array_op` pass). Since it is transformed, its signature does not match to that of its counter-part defined in the `expr2_module.py`.

How do we tackle this? Shall we skip updating/modifying `@pythoncall` decorated functions in the `create_subroutine_from_function()` and `array_op` pass?

src/bin/lpython.cpp
```diff
@@ -1562,6 +1565,7 @@ int main(int argc, char *argv[])
         app.add_flag("--get-rtl-dir", print_rtl_dir, "Print the path to the runtime library file");
         app.add_flag("--verbose", compiler_options.verbose, "Print debugging statements");
         app.add_flag("--enable-cpython", compiler_options.enable_cpython, "Enable CPython runtime");
+        app.add_flag("--enable-cnumpy", compiler_options.enable_cnumpy, "Enable C-Numpy runtime");
```
certik, line 1568: I would call it
```suggestion
        app.add_flag("--enable-numpy", compiler_options.enable_cnumpy, "Enable NumPy runtime");
```

certik:
> Shall we skip updating/modifying `@pythoncall` decorated functions in the `create_subroutine_from_function()` and `array_op` pass?

Yes. I think it skips BindC functions and we just need it to also skip BindPython functions too.

certik (approved):
I think this looks great!

certik:
I am going to merge this, as it looks good. In subsequent PRs:

* [ ] Use `--enable-numpy` (see my comment above)
* [ ] Create a dedicated CI and update our integration_tests, that tests everything except python, via C and LLVM backends and ensures we have no dependency on CPython; and add a CI that only tests the Python interoperability tests (for now only C, later LLVM as well)

Shaikh-Ubaid:
>Use --enable-numpy

We can support this. A minor concern here is that people/users might get confused as `numpy` is also package that can be imported (In this PR, `cnumpy` refers to `numpy-c` library/API). For example, users might think `--enable-numpy` is used to enable importing numpy, that is, `from numpy import empty, ...`. In actual `--enable-numpy` is only needed if you are using `numpy` along with support for `cpython` runtime.

certik:
I see. Maybe we can call it `--link-numpy` instead then.

PR #1936, opened 2023-06-18
Allow partial initialization of structs

### LINKED ISSUE #1928: Partial initialization of dataclass does not work
The following works in oridinary CPython

```python
@dataclass
class StringIO:
    _buf     : str
    _0cursor : int = 0
    _len     : int = 0

if __name__ == '__main__':
    integer_asr : str = '(Integer 4 [])'
    test_dude   : StringIO = StringIO(integer_asr)
    assert test_dude._buf == integer_asr
    assert test_dude._len == 0 # len(integer_asr)
```

but the equivalent does not work in LPython

```python
@dataclass
class StringIO:
    _buf     : str
    _0cursor : i32 = i32(0)
    _len     : i32 = i32(0)

if __name__ == '__main__':
    integer_asr : str = '(Integer 4 [])'
    test_dude   : StringIO = StringIO(integer_asr) # , 0, len(integer_asr))
    assert test_dude._buf == integer_asr
    assert test_dude._len == 0 # len(integer_asr)
```

produces

```
└─(15:05:17 on brian-lasr ✹ ✭)──> ~/CLionProjects/lpython/src/bin/lpython lasr_lexer.py                                                               1 ↵ ──(Fri,Jun16)─┘
semantic error: StructConstructor arguments do not match the number of struct members
  --> lasr_lexer.py:29:30
   |
29 |     test_dude   : StringIO = StringIO(integer_asr) # , 0, len(integer_asr))
   |                              ^^^^^^^^^^^^^^^^^^^^^ 
```

Smit-create:
Fixes #1928


certik (approved):

certik:
I think the code change is fine, but I would like @rebcabin to review this as well.

Do we want to allow partial initialization of structs? Does CPython support that?

rebcabin:
Yes, CPython supports partial initialization of dataclasses. I will add a
repro case to the bug report.

rebcabin:
repro case : https://github.com/rebcabin/lpython/tree/brian-lasr/lasr/lpython/Issue1936

Works in CPython:

```python
from dataclasses import dataclass
i32 = int
from typing import \
    Generic, TypeVar, Optional, Union, \
    Tuple, Any, List, Pattern
T_ = TypeVar('T_')
class InOut(Generic[T_]):
    pass


@dataclass
class StringIO:
    _buf     : str
    _0cursor : i32 = i32(0)
    _len     : i32 = i32(0)


if __name__ == '__main__':
    integer_asr : str = '(Integer 4 [])'
    fd   : StringIO = StringIO(integer_asr)
    assert fd._len == 0
    assert fd._0cursor == 0
```

LPython version (comment out the compatibility block, that's the only change):

```python
# from dataclasses import dataclass
# i32 = int
# from typing import \
#     Generic, TypeVar, Optional, Union, \
#     Tuple, Any, List, Pattern
# T_ = TypeVar('T_')
# class InOut(Generic[T_]):
#     pass


@dataclass
class StringIO:
    _buf     : str
    _0cursor : i32 = i32(0)
    _len     : i32 = i32(0)


if __name__ == '__main__':
    integer_asr : str = '(Integer 4 [])'
    fd   : StringIO = StringIO(integer_asr)
    assert fd._len == 0
    assert fd._0cursor == 0
```

Improper (but not incorrect!) error message:

```bash
└─(15:33:11 on brian-lasr ✹ ✭)──> ~/CLionProjects/lpython/src/bin/lpython -I. lasr_lexer.py && ./lasr_lexer.out                                           ──(Sun,Jun18)─┘
semantic error: StructConstructor arguments do not match the number of struct members
  --> lasr_lexer.py:20:23
   |
20 |     fd   : StringIO = StringIO(integer_asr)
   |                       ^^^^^^^^^^^^^^^^^^^^^ 
```

rebcabin (approved):

rebcabin:
Changes look good. Can you write a test for a constructor call that has TOO MANY arguments? `StringIO` expects 0, 1, 2, or 3 arguments. Your test covers the case of 1 argument. Perhaps a negative of test of 4 arguments and also new tests with 0, 2, and 3 arguments would be warranted.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1286,8 +1286,8 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
                                          args, StructType, loc);
             }
 
-            if (args.size() > 0 && args.size() !=  StructType->n_members) {
-                throw SemanticError("StructConstructor arguments do not match the number of struct members", loc);
+            if (args.size() > 0 && args.size() >  StructType->n_members) {
+                throw SemanticError("StructConstructor arguments are greater the number of struct members", loc);
```
certik, line 1290: I would say:
"Struct constructor has more arguments than the number of struct members"

certik (approved):
I have an improvement to the error message, which can be done in a subsequent PR.

PR #1947, opened 2023-06-19
Return err, if an error has occurred in creating the C file

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
Thanks for the review!

PR #1951, opened 2023-06-19
ASR: Remove unnecessary compare checks

### LINKED ISSUE #1949: Bug: U16 Comparision - C backend
Error when compering u16 to u16:
```
semantic error: Compare: only Integer, Real, Logical, or String can be on the LHS and RHS.If operator is Eq or NotEq then Complex type is also acceptable

   --> /home/eamiel/tmp_cp_release_01/copperhead/demo-apps/belex-direct-fragment/device/../mat_vec_device.py:130:13

    |

130 |         if (u16(l1_grp) >= u16(GSI_L1_NUM_GRPS)):

    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

 
```

Smit-create:
Fixes #1949

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -5968,22 +5968,6 @@ public:
             right_type = ASRUtils::get_contained_type(right_type);
         }
         ASR::expr_t *overloaded = nullptr;
-        if (((left_type->type != ASR::ttypeType::Real &&
-            left_type->type != ASR::ttypeType::Integer) &&
-            (right_type->type != ASR::ttypeType::Real &&
-            right_type->type != ASR::ttypeType::Integer) &&
-            ((left_type->type != ASR::ttypeType::Complex ||
-            right_type->type != ASR::ttypeType::Complex) &&
-            x.m_ops != AST::cmpopType::Eq && x.m_ops != AST::cmpopType::NotEq) &&
-            (left_type->type != ASR::ttypeType::Character ||
-            right_type->type != ASR::ttypeType::Character)) &&
-            (left_type->type != ASR::ttypeType::Logical ||
-            right_type->type != ASR::ttypeType::Logical)) {
-        throw SemanticError(
-            "Compare: only Integer, Real, Logical, or String can be on the LHS and RHS."
-            "If operator is Eq or NotEq then Complex type is also acceptable",
-            x.base.base.loc);
-        }
 
         if (!ASRUtils::is_logical(*left_type) || !ASRUtils::is_logical(*right_type)) {
             cast_helper(left, right, false);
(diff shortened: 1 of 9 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -5922,22 +5922,6 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
-            x.m_ops != AST::cmpopType::Eq && x.m_ops != AST::cmpopType::NotEq) &&
-            (left_type->type != ASR::ttypeType::Character ||
-            right_type->type != ASR::ttypeType::Character)) &&
-            (left_type->type != ASR::ttypeType::Logical ||
-            right_type->type != ASR::ttypeType::Logical)) {
-        throw SemanticError(
-            "Compare: only Integer, Real, Logical, or String can be on the LHS and RHS."
-            "If operator is Eq or NotEq then Complex type is also acceptable",
-            x.base.base.loc);
-        }
```
certik, line 5940: I don't think this is the right fix. I think we need to keep this and add tests for this error to ensure the error actually works.

Then to fix the above issue, I think just include UnsignedInteger as an allowed type. Only allow comparing unsigned integers with each other.

Smit-create: This if-else is actually quite old and hack-ish as it misses many other checks that do currently work on LPython like comparing List, Tuple, CPtr, etc. Moreover, there isn't any test for the same and the message is also misguiding. Also, if there's anything not supported, it will surely raise an error in their individual `*Compare_t` nodes.

certik: Ok. Do we have tests for giving errors for things like `i16(x) + f16(y)` ? If so, then this PR is ok.

certik (approved):
Approved if https://github.com/lcompilers/lpython/pull/1951/files#r1234697033.

src/lpython/semantics/python_ast_to_asr.cpp
Smit-create: Yep, see:
```py
def f():
    x: i32 = 10
    y: i32 = 13
    print(i16(x) + f32(y))

f()
```
Gives:
```
 --> a.py:4:11
  |
4 |     print(i16(x) + f32(y))
  |           ^^^^^^   ^^^^^^ type mismatch (i16 and f32)
```

Also, the following:
```py
def f():
    x: i32 = 10
    y: i32 = 13
    print(i16(x) > f32(y))

f()
```

Gives:
```
 --> a.py:4:11
  |
4 |     print(i16(x) > f32(y))
  |           ^^^^^^   ^^^^^^ type mismatch ('i16' and 'f32')
```

certik: ok! And we have tests for it?

Smit-create: For the 2nd test case, we don't. But for the first one, we do have. I'll add some more in this PR.

certik (approved):
This looks great, thanks!

PR #1960, opened 2023-06-20
CI: Consolidate cpython_py and cpython

certik:
The two tests seem to be the same. The `c_py` requires Python to be
installed which otherwise is not required. But the `cpython_py` just
requires Python to be installed, which it is anyway for the `cpython`
backend, so the two labels are effectively the same. So we make them the
same.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -288,8 +288,7 @@ jobs:
             ctest
             ./run_tests.py -s
             cd integration_tests
-            ./run_tests.py -b llvm
-            ./run_tests.py -b c
+            ./run_tests.py -b llvm c
 
   cpython_interop:
     name: Test CPython Interop (@pythoncall)
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -331,7 +331,7 @@ jobs:
         shell: bash -l {0}
         run: |
             cd integration_tests
-            ./run_tests.py -b cpython_py c_py
+            ./run_tests.py -b cpython c_py
 
   sympy:
     name: Run SymPy tests
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -164,7 +164,7 @@ macro(RUN_UTIL RUN_FAIL RUN_NAME RUN_FILE_NAME RUN_LABELS RUN_ENABLE_CPYTHON RUN
             if (${fail})
                 set_tests_properties(${name} PROPERTIES WILL_FAIL TRUE)
             endif()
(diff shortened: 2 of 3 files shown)


certik:
I am going to merge it, as I think less backends to worry about, the better. If we need this in the future, we can easily add it back in.

.github/workflows/CI.yml
```diff
@@ -332,7 +331,7 @@ jobs:
         shell: bash -l {0}
         run: |
             cd integration_tests
-            ./run_tests.py -b cpython_py c_py
+            ./run_tests.py -b cpython c_py
```
Shaikh-Ubaid, line 334: Minor concern: We already test `cpython` in `text.xsh` (which is called in `LPython CI / LPython CI (3.10, ubuntu-latest) (pull_request)`). Do we need to test `cpython` again here?

Shaikh-Ubaid (commented):
It looks good to me. Thank you so much for this!

Shaikh-Ubaid:
I submitted some more improvements here https://github.com/lcompilers/lpython/pull/1961.

.github/workflows/CI.yml
certik: We don't, but I think it doesn't hurt to ensure the extra packages (sympy in this case) don't break anything. The tests run quickly.

Also I was thinking if the Python interop tests fail, but other tests pass, that we can infer what went wrong, as opposed to (all) the Python interop tests failing.

We can add it back in later, if needed.

Shaikh-Ubaid: >I think it doesn't hurt to ensure the extra packages (sympy in this case) don't break anything.

We have `numpy` in this case, which is already/by-default available in the `lp` conda environment. For `sympy` there is a separate CI Job called `Run SymPy tests`.

```
Also I was thinking if the Python interop tests fail, but other tests pass, that we can infer what went wrong, as opposed to (all) the Python interop tests failing.

We can add it back in later, if needed.
```
Yes, we can add it back in later, if needed. Also as per the current `main` branch, it seems:
- If there is an issue in **non** `python-interop` test case, then I think three CI jobs might fail 
    - LPython CI (3.10, ubuntu-latest)
    - LPython CI (3.10, macos-latest)
    - Test CPython Interop (`@pythoncall`)
 - If there is an issue in `python-interop` tests, then I think the same three CI jobs might fail
     - LPython CI (3.10, ubuntu-latest)
    - LPython CI (3.10, macos-latest)
    - Test CPython Interop (`@pythoncall`)

certik: Yes, the Python interop in pure Python mode is always tested with the rest of the Python tests. While the Python interop in the C mode is tested in separate CI jobs. I think that's fine.

Issue #1956, opened 2023-06-19
C: redefined warnings in macros used for constants


certik:
In `bindpy_02`:
```
/Users/ondrej/repos/lpython/integration_tests/_lpython-tmp-test-c_py/bindpy_02.c:360:13: warning: 'n' macro redefined [-Wmacro-redefined]
    #define n 3
            ^
/Users/ondrej/repos/lpython/integration_tests/_lpython-tmp-test-c_py/bindpy_02.c:224:13: note: previous definition is here
    #define n 5
            ^
[ 77%] Built target bindpy_01
/Users/ondrej/repos/lpython/integration_tests/_lpython-tmp-test-c_py/bindpy_02_FAST.c:360:13: warning: 'n' macro redefined [-Wmacro-redefined]
    #define n 3
            ^
/Users/ondrej/repos/lpython/integration_tests/_lpython-tmp-test-c_py/bindpy_02_FAST.c:224:13: note: previous definition is here
    #define n 5
            ^
```
The issue is that there is a constant `n` defined in two functions and it gets translated to a `#define n 3` macro, and then redefined in the next function.

There are couple approaches to this:

* One can define `n` and then undefine it at the end of the function; one issue is that if `n` was somehow defined by the user above, then this will still give a warning and then undefine it.
* One can use unique macro name, such as `_lcompilers_const_n1`, `_lcompilers_const_n2`, etc.
* C I think now allows a `const` keyword. Why not to use that? It should have a similar scope that we need and might just work for what we need.

PR #1962, opened 2023-06-20
C: Use `const` in place of `#define` for declaring constants

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1956


certik:
I think it works!

Issue #1952, opened 2023-06-19
Bug: Assignment to an input function parameter - C backend


ronnuriel:
semantic error: Assignment to an input function parameter `x` is not allowed
 --> test:2:5
  |
2 |     x = u16(11)
  |   

```
def foo(x: u16):
    x = u16(11)

foo(u16(10))
```

Shaikh-Ubaid:
>semantic error: Simple Type u16 cannot be intent InOut/Out

This is expected. Simple types cannot be `InOut`. In `CPython` if a variable (of any Simple Type `i32`, `i64`, `u32`, `f32`, ...) is updated in the callee function then the updated value is not reflected back in the caller function. This happens because `CPython` passes arguments by (copy of) object reference. When an object is updated inside a callee function, the original reference/variable remains intact, I think.

Shaikh-Ubaid:
```bash
$ cat examples/expr2.py 
def foo(x: u16):
    x = u16(11)

foo(u16(10))
$ lpython examples/expr2.py
semantic error: Assignment to an input function parameter `x` is not allowed
 --> examples/expr2.py:2:5
  |
2 |     x = u16(11)
  |     ^ Use InOut[x] to allow assignment


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that must be fixed).
```
It seems the bug here is the guiding message to use `InOut`. In this case, using `InOut` for a `Simple Type` is not allowed.

certik:
Yes, for simple types we need to improve the error message not to suggest InOut.

@ronnuriel are you trying to return a value from the function using the argument, or are you trying to just create a new local variable that happens to have the same name as an argument?

For now we decided not to allow "shadowing" arguments, to prevent bugs. If you don't need the argument, just rename it (or rename the local variable).

ronnuriel:
@certik No, Eyal tried to cast the var but he couldn't..

PR #1966, opened 2023-06-20
Fix error messages

Smit-create:
https://github.com/lcompilers/lpython/pull/1936#discussion_r1234416502
#1952

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -1266,15 +1266,15 @@ static inline std::string type_to_str_python(const ASR::ttype_t *t,
         }
         case ASR::ttypeType::Struct: {
             ASR::Struct_t* d = ASR::down_cast<ASR::Struct_t>(t);
-            return "struct " + std::string(symbol_name(d->m_derived_type));
+            return std::string(symbol_name(d->m_derived_type));
         }
         case ASR::ttypeType::Enum: {
             ASR::Enum_t* d = ASR::down_cast<ASR::Enum_t>(t);
-            return "enum " + std::string(symbol_name(d->m_enum_type));
+            return std::string(symbol_name(d->m_enum_type));
         }
         case ASR::ttypeType::Union: {
             ASR::Union_t* d = ASR::down_cast<ASR::Union_t>(t);
-            return "union " + std::string(symbol_name(d->m_union_type));
+            return std::string(symbol_name(d->m_union_type));
         }
         case ASR::ttypeType::Pointer: {
             ASR::Pointer_t* p = ASR::down_cast<ASR::Pointer_t>(t);
--- a/tests/reference/asr-structs_02-f95782c.stderr
+++ b/tests/reference/asr-structs_02-f95782c.stderr
@@ -1,4 +1,4 @@
-semantic error: struct S s must be initialized a value
+semantic error: `s` must be initialized with an instance of S
  --> tests/errors/structs_02.py:8:5
   |
 8 |     s: S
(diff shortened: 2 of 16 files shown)


src/libasr/asr_utils.h
```diff
@@ -1266,15 +1266,15 @@ static inline std::string type_to_str_python(const ASR::ttype_t *t,
...
         }
         case ASR::ttypeType::Enum: {
             ASR::Enum_t* d = ASR::down_cast<ASR::Enum_t>(t);
-            return "enum " + std::string(symbol_name(d->m_enum_type));
+            return std::string(symbol_name(d->m_enum_type));
         }
         case ASR::ttypeType::Union: {
             ASR::Union_t* d = ASR::down_cast<ASR::Union_t>(t);
-            return "union " + std::string(symbol_name(d->m_union_type));
+            return std::string(symbol_name(d->m_union_type));
```
Shaikh-Ubaid, line 1277: I think this is not needed.

This was a feature added in https://github.com/lcompilers/lpython/commit/7f7dbde2f14a2ef244bd585f4dd69bc599efe8f5. `Struct`, `Enum`, `Union` are derived types. For them, printing just the symbol name does not seem sufficient (we also need their type information). 

For example, for `s: i32`, `type_to_str_python()` returns `i32` which makes clear that it is an integer/`i32`.
For `class S:`, `type_to_str_python()` returns just `S` which does not inform of the type of `S` (`Struct` or `Enum` or `Union`).

For other types, `type_to_str_python()` returns the type name, where as for derived types `type_to_str_python()` returns the symbol name (and not the type). Therefore, we had prepended the type information to derived types in https://github.com/lcompilers/lpython/commit/7f7dbde2f14a2ef244bd585f4dd69bc599efe8f5.

tests/reference/asr-structs_02-f95782c.stderr
```diff
@@ -1,4 +1,4 @@
-semantic error: struct S s must be initialized a value
+semantic error: S s must be initialized a value
```
certik, line 1:
```suggestion
semantic error: `s` must be initialized with an instance of S
```

tests/reference/asr-structs_02-f95782c.stderr
```diff
@@ -1,4 +1,4 @@
...
  --> tests/errors/structs_02.py:8:5
   |
 8 |     s: S
```
certik, line 4: And add a suggestion:
```
^
s: S = S()
```

tests/reference/asr-test_assign9-51278b8.stderr
```diff
@@ -2,4 +2,4 @@ semantic error: Assignment to an input function parameter `x` is not allowed
  --> tests/errors/test_assign9.py:2:5
   |
 2 |     x = 2
-  |     ^ Use InOut[x] to allow assignment
+  |     ^ Assignment not supported for `x`
```
certik, line 5: For arrays, structs, etc the most common reason to get this error is when trying to modify the value in the caller. So the best suggestion is to use InOut (or Out).

For integers, people know that you cannot modify the variable in the caller that way, so I don't think that's the most common reason to get this error message.

I think there are two possibilities why somebody would assign to `x`:

* Trying to use the variable `x` locally in some algorithm and simply modifying it in the process, knowing that it doesn't change in the caller;
* Not needing the argument anymore and wanting to use a new local variable that happens to have the same name
* By a mistake wanting to use a new local variable that happens to shadow the argument, causing a bug

```suggestion
  |     ^ create a new local variable with a different name
```

certik (approved):

Issue #1938, opened 2023-06-18
Fixing redefinition error causes `SIGSEGV`


rebcabin:
LPython (correctly) reports a redefinition error for the following code

```python
@dataclass
class StringIO:
    _buf     : str
    _0cursor : i32 = i32(0)
    _len     : i32 = i32(0)


## Issue #1930: InOut type


def __post_init__(this: InOut[StringIO]):
    this._len = len(this._buf)


def read(this: InOut[StringIO], chars : i32) -> str:
    result : str = ''  # EOF
    if this._0cursor == this._len:
        result = ''  #
    else:
        mac : i32 = min(this._len, this._0cursor + chars)
        assert mac <= this._len
        result : str = this._buf[this._0cursor, mac]  ########### ATTENTION #############
        this._0cursor = mac
    return result


if __name__ == '__main__':
    integer_asr : str = '(Integer 4 [])'
    fd   : StringIO = StringIO(integer_asr, 0, 0)
    __post_init__(fd)  # Issue #1929
    assert fd._buf == integer_asr
    assert fd._len == len(integer_asr)

    a_char : str = fd.read(1)
    assert a_char == '('
```

produces

```bash
(lp) ┌─(~/CLionProjects/lpython/lasr/lpython)────────────────────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s000)─┐
└─(06:23:01 on brian-lasr ✹ ✭)──> ~/CLionProjects/lpython/src/bin/lpython lasr_lexer.py                                                               1 ↵ ──(Sun,Jun18)─┘
semantic error: Symbol is already declared in the same scope
  --> lasr_lexer.py:24:5
   |
24 |     result : str = ''  # EOF
   |     ~~~~~~~~~~~~~~~~~ original declaration
   |
30 |         result : str = this._buf[this._0cursor, mac]
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ redeclaration

```

Fixing the error as follows:

```python
@dataclass
class StringIO:
    _buf     : str
    _0cursor : i32 = i32(0)
    _len     : i32 = i32(0)


## Issue #1930: InOut type


def __post_init__(this: InOut[StringIO]):
    this._len = len(this._buf)


def read(this: InOut[StringIO], chars : i32) -> str:
    result : str = ''  # EOF
    if this._0cursor == this._len:
        result = ''  #
    else:
[...]

PR #1967, opened 2023-06-20
Catch errors for tuple indexing in strings and call to undefined struct members

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1938


Shaikh-Ubaid:
Ready.

tests/reference/asr-structs_10-cb8a283.stderr
```diff
@@ -0,0 +1,19 @@
+warning: Struct member functions are yet to be supported
+ --> tests/errors/structs_10.py:7:5 - 8:24
+  |
+7 |        def print_len(self: StringIO):
+  |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
+...
+  |
+8 |            print(self._len)
+  | ...^^^^^^^^^^^^^^^^^^^^^^^^ Skipped defining function print_len
```
certik, line 9: What does it mean Struct member functions are yet to be supported? Shouldn't it rather give an error message?

Shaikh-Ubaid:
>Q: what about print(foo,)? I believe the trailing comma inside the argument list is legal Python. I haven't tried yet to see what LPython will do with it.

LPython works with trailing comma:

```bash
$ cat examples/expr2.py 
def main0():
    x: i32
    x = (2+3)*5
    print(x,)

main0()
$ python examples/expr2.py
25
$ lpython examples/expr2.py
25
```

tests/reference/asr-structs_10-cb8a283.stderr
Shaikh-Ubaid: >What does it mean Struct member functions are yet to be supported? Shouldn't it rather give an error message?

I thought a `struct` is like a class, so it could/may have member functions defined inside it. Currently, we do not yet support member functions. The message warns that there is a function provided by the user inside the class and that
function is being skipped/ignored since we do not yet support defining member functions. We can as well give an error message (instead of a warning). The idea behind the warning was that the rest of the code (if semantically correct) could be compiled (and executed).

doubt: Do we plan to support member functions in `structs` in LPython? Or would `structs` only contain data members?

tests/reference/asr-string_01-78629c4.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Type mismatch in index, Index should be of integer type
```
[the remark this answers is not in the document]
Shaikh-Ubaid: >Or is this error for anytype

Yes, it is an error for anytype.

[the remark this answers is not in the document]
certik:
```
Type mismatch in index, expected a single integer
```

tests/reference/asr-structs_10-cb8a283.stderr
certik: Struct member functions are not supported

certik: And let's make it an error message, so that we get nice error messages for the current feature set. If we later decide to implement a larger subset of LPython, then we'll remove this error message and implement the feature. If we decide we don't want to do that, then we keep the error message.

tests/reference/asr-string_01-78629c4.stderr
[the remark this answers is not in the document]
Shaikh-Ubaid: Updated.

Shaikh-Ubaid:
Ready.

certik (approved):
Thanks!

Issue #1970, opened 2023-06-20
Instantiating `Lexer` class causes SIGABRT


rebcabin:
repro: https://github.com/rebcabin/lpython/tree/brian-lasr/lasr/LP-pycharm/Issue1970

This code causes a SIGABRT:

```python
def lexer_test():
    print('LEXER TEST')
    lexer : InOut[LasrLexer] = LasrLexer()  ############ ATTENTION #############
    assert True
```

produces:

```bash
/Users/brian/CLionProjects/lpython/src/bin/python /Users/brian/CLionProjects/lpython/lasr/LP-pycharm/lasr_lexer.py 
LCOMPILERS_ASSERT failed: /Users/brian/CLionProjects/lpython/src/libasr/asr.h
function expr_type0(), line number 47387 at 
false
ERROR MESSAGE:
6
Traceback (most recent call last):
  File "/Users/brian/CLionProjects/lpython/src/bin/lpython.cpp", line 1844
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir,
  File "/Users/brian/CLionProjects/lpython/src/bin/lpython.cpp", line 783
    r1 = LCompilers::LPython::python_ast_to_asr(al, lm, *ast, diagnostics, compiler_options,
  File "/Users/brian/CLionProjects/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 7529
    auto res2 = body_visitor(al, lm, *ast_m, diagnostics, unit, main_module,
  File "/Users/brian/CLionProjects/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 7482
    b.visit_Module(ast);
  File "/Users/brian/CLionProjects/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 4585
    visit_stmt(*x.m_body[i]);
  File "/Users/brian/CLionProjects/lpython/src/lpython/python_ast.h", line 1883
    void visit_stmt(const stmt_t &b) { visit_stmt_t(b, self()); }
  File "/Users/brian/CLionProjects/lpython/src/lpython/python_ast.h", line 1751
    case stmtType::FunctionDef: { v.visit_FunctionDef((const FunctionDef_t &)x); return; }
  File "/Users/brian/CLionProjects/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 4692
    handle_fn(x, *f);
[...]

PR #1975, opened 2023-06-21
Error for incorrect type annotation

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1970

### DIFF
--- a/tests/reference/asr-func_08-c137415.stderr
+++ b/tests/reference/asr-func_08-c137415.stderr
@@ -1,0 +1,5 @@
+semantic error: Intent annotation 'In' cannot be used here
+ --> tests/errors/func_08.py:3:17
+  |
+3 | def main0(x: In[In[i32]]):
+  |                 ^^^^^^^ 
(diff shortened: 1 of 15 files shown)


tests/reference/asr-func_08-c137415.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Intent annotation 'In' cannot be used here
+ --> tests/errors/func_08.py:3:17
+  |
+3 | def main0(x: In[In[i32]]):
```
Shaikh-Ubaid, line 4: The error message here is shared with the error message for local variables with annotation `In`/ `InOut` / `Out`. See `tests/errors/test_annassign_02.py` in this PR. Therefore the error message is kept slightly general to cover both the cases.

tests/reference/asr-func_08-c137415.stderr
```diff
@@ -0,0 +1,5 @@
...
+  |                 ^^^^^^^
```
Shaikh-Ubaid, line 5: The error message here is shared with the error message for local variables with annotation `In`/ `InOut` / `Out`. See `tests/errors/test_annassign_03.py` in this PR. Therefore the error message is kept slightly general to cover both the cases.

certik (approved):

PR #1977, opened 2023-06-21
Support `List` annotation

### LINKED ISSUE #1944: Bogus error message on `List[str]`
repro : https://github.com/rebcabin/lpython/tree/brian-lasr/lasr/lpython/Issue1944

The following works in ordinary CPython (after uncommenting the compatibility block at the top of the file):

```python
WHITESPACE   : List[str] = [" ", ",", NEWLINE, "\r", "\t"]
```

The error message from LPython does not lead me to a fix:

```bash
└─(15:12:05 on brian-lasr ✹ ✭)──> ~/CLionProjects/lpython/src/bin/lpython -I. lasr_lexer.py && ./lasr_lexer.out                                       2 ↵ ──(Sun,Jun18)─┘
semantic error: Variable 'str' not declared
   --> lasr_lexer.py:133:21
    |
133 | WHITESPACE   : List[str] = [" ", ",", NEWLINE, "\r", "\t"]
    |                     ^^^ 
```

Smit-create:
Fixes #1944


certik:
I would not support deprecated features. According to https://docs.python.org/3/library/typing.html#typing.List, "Deprecated alias to [list](https://docs.python.org/3/library/stdtypes.html#list).".

Rather, let's give a nice error message to use `list`.

Also, one thing is to support the intrinsic typing module, but here we are adding it to the `lpython` module, which is our own module --- and I think in general I would strive for "just one way of doing it". Since we support "list", I would not support "List".

rebcabin:
testing now prior to closing

Smit-create:
> Since we support "list", I would not support "List".

Okay, then let's give a nice error message.

rebcabin:
verified that 'List' gives nice error message. Closing.

PR #1979, opened 2023-06-21
Support `Literal` annotation

### LINKED ISSUE #1939: Support `Literal` type hint
The following should not raise an Error in LPython:

```python
└─(10:11:14 on brian-lasr ✹ ✭)──> ~/CLionProjects/lpython/src/bin/lpython lasr_lexer.py                                                               2 ↵ ──(Sun,Jun18)─┘
semantic error: Unsupported type annotation: Literal
  --> lasr_lexer.py:42:12
   |
42 | SEEK_SET : Literal[i32(0)] = i32(0)
   |            ^^^^^^^^^^^^^^^ 
```

It's information to the compiler that the type of `SEEK_SET` is a literal constant. It's kind of an avanced dependent type, and a way to get some semblance of immutability. In any event, it's necessary for an implementation of `Seekable`. 

See https://docs.python.org/3.10/library/typing.html?highlight=literal#typing.Literal and https://docs.python.org/3.10/library/io.html#io.SEEK_SET .

Smit-create:
Fixes #1939


certik:
@rebcabin `Literal[i32(0)]` means that the actual runtime value of this variable can only be `i32(0)`?

This PR is a start on that, but we do not actually check this, so the implementation is not finished. Given that, should we give a warning that this is not enforced yet?

Smit-create:
@rebcabin Can you please review this? Thanks

certik:
I am a little worried about adding types that we don't actually support yet.

Issue #1599, opened 2023-03-21
Array assignment fails


certik:
```python
from numpy import empty, float32

def main():
    a: f32[10000] = empty(10000, dtype=float32)
    b: f32[10000]
    b[:] = a[:]

main()
```
gives:
```console
$ lpython a.py
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/ondrej/repos/lpython/src/bin/lpython.cpp", line 1685
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir,
  File "/Users/ondrej/repos/lpython/src/bin/lpython.cpp", line 703
    res = fe.get_llvm3(*asr, pass_manager, diagnostics, infile);
  File "/Users/ondrej/repos/lpython/src/lpython/python_evaluator.cpp", line 58
    run_fn, infile);
  File "/Users/ondrej/repos/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 7025
    pass_manager.apply_passes(al, &asr, pass_options, diagnostics);
  File "/Users/ondrej/repos/lpython/src/libasr/pass/pass_manager.h", line 213
    _apply_passes(al, asr, _passes, pass_options, diagnostics);
  File "/Users/ondrej/repos/lpython/src/libasr/pass/pass_manager.h", line 104
    _passes_db[passes[i]](al, *asr, pass_options);
  File "/Users/ondrej/repos/lpython/src/libasr/pass/arr_slice.cpp", line 220
    ArraySectionVisitor v(al);
  File "../libasr/asr.h", line 39209
  File "../libasr/asr.h", line 4480
  File "../libasr/asr.h", line 4219
  File "../libasr/asr.h", line 39227
  File "../libasr/asr.h", line 4480
  File "../libasr/asr.h", line 4220
  File "../libasr/asr.h", line 39247
  File "/Users/ondrej/repos/lpython/src/libasr/pass/arr_slice.cpp", line 195
    visit_stmt(*m_body[i]);
  File "../libasr/asr.h", line 4495
  File "../libasr/asr.h", line 4241
  File "/Users/ondrej/repos/lpython/src/libasr/pass/arr_slice.cpp", line 207
    current_expr = const_cast<ASR::expr_t**>(&(x.m_value));
  File "/Users/ondrej/repos/lpython/src/libasr/pass/arr_slice.cpp", line 187
    replacer.replace_expr(*current_expr);
  File "../libasr/asr.h", line 38554
[...]

PR #1985, opened 2023-06-21
fix: added null check & type checking

arteevraina:
- [x] Checks if `*x` is `nullptr`.

Refer: #1599

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -5138,11 +5138,6 @@ public:
         if( x == nullptr ) {
             throw CodeGenError("Internal error: x is nullptr");
         }
-
-        // Check if the type of *x is expr_t or not.
-        if( !ASR::is_a<ASR::expr_t>(*x) ) {
-            throw CodeGenError("Internal error: x is not of type expr_t");
-        }
         
         this->visit_expr(*x);
         if( x->type == ASR::exprType::ArrayItem ||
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5134,6 +5134,16 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
     inline void visit_expr_wrapper(ASR::expr_t* x, bool load_ref=false) {
+        // Check if *x is nullptr.
+        if( x == nullptr ) {
+            throw CodeGenError("Internal error: x is nullptr");
+        }
+
+        // Check if the type of *x is expr_t or not.
+        if( !ASR::is_a<ASR::expr_t>(*x) ) {
+            throw CodeGenError("Internal error: x is not of type expr_t");
+        }
```
certik, line 5145: I think `x` has to be an `expr_t`, since that's the type of the argument. Also the `is_a` check works for the end nodes like IntegerBinOp; it works for expr if `x` was of type `asr_t`. 

I would remove this last check.

certik:
Can you check that this PR now reports an error for that issue instead of a segfault?

arteevraina:
> Can you check that this PR now reports an error for that issue instead of a segfault?

Yes, It gives an error message now instead of segfault for #1599 

```
(lp) arteevraina@Arteevs-MacBook-Air lpython % src/bin/lpython src/a.py                                         
code generation error: Internal error: x is nullptr


```

src/libasr/codegen/asr_to_llvm.cpp
arteevraina: This has been resolved.

certik:
Very good, let's see if the tests pass.

certik:
Thanks!

PR #2012, opened 2023-06-23
PythonCall: Support array of simple types as return type

Shaikh-Ubaid:
towards #703


Shaikh-Ubaid:
Ready.

certik (approved):
I think that looks good.

PR #2013, opened 2023-06-23
Move visit_Call in Common Visitor

### LINKED ISSUE #1981: LPython bogusly changes attribute value
repro: https://github.com/rebcabin/lpython/tree/brian-lasr/lasr/LP-pycharm/Issue1981

This is related to Issue #1978; i'm digging in to find out what's wrong. I reduced the test case to the bare minimum that illustrates the problem. 

The problem is that when I initialize a StringIO object from inside a function, an attribute value gets bogusly changed to a random number.

Consider the following code in `Issue1981/lasr_lexer.py`

```python
from lpython import (dataclass, InOut, i32,)
from lasr_stringio import (StringIO, __post_init__,)

NEWLINE      : str       = "\n"
WHITESPACE   : list[str] = [" ", ",", NEWLINE, "\r", "\t"]
EOF_SIGNAL   : str       = ''

@dataclass
class LasrLexer:
    fd   : StringIO = StringIO('', 0, 0) # prefer Optional[StringIO]
    line : i32 = 1 # prefer Line
    col  : i32 = 1 # prefer Col
    type : str = '' # prefer LTType
    val  : str = '' # prefer LTVal

def lexer_test():
    print('LEXER TEST')
    lexer : LasrLexer = LasrLexer(StringIO('  foo  bar  '))
    __post_init__(lexer.fd)

if __name__ == '__main__':
    lexer_test()
```

When I run it with LPython, I get

```bash
/Users/brian/CLionProjects/lpython/src/bin/python /Users/brian/CLionProjects/lpython/lasr/LP-pycharm/Issue1981/lasr_lexer.py 
LEXER TEST
StringIO __post_init__ preconditions:
    _buf =   foo  bar  .
    _len = 1.
    _0cursor = 1830662080.  ################ ATTENTION #############
StringIO __post_init__ postconditions:
    _buf =   foo  bar  .
    _len = 12.
    _0cursor = 1830662080.
```

When I change the code to the following, just moving the initialization inside main and out of the function `lexer_test`, I get the correct value for `_0cursor`:

```python
from lpython import (dataclass, InOut, i32,)
from lasr_stringio import (StringIO, __post_init__,)


NEWLINE      : str       = "\n"
WHITESPACE   : list[str] = [" ", ",", NEWLINE, "\r", "\t"]
EOF_SIGNAL   : str       = ''


@dataclass
class LasrLexer:
    fd   : StringIO = StringIO('', 0, 0) # prefer Optional[StringIO]
    line : i32 = 1 # prefer Line
    col  : i32 = 1 # prefer Col
    type : str = '' # prefer LTType
    val  : str = '' # prefer LTVal


if __name__ == '__main__':
    print('LEXER TEST')
    lexer : LasrLexer = LasrLexer(StringIO('  foo  bar  '))
    __post_init__(lexer.fd)
```

produces: 

```bash
/Users/brian/CLionProjects/lpython/src/bin/python /Users/brian/CLionProjects/lpython/lasr/LP-pycharm/Issue1981/lasr_lexer.py 
LEXER TEST
StringIO __post_init__ preconditions:
    _buf =   foo  bar  .
    _len = 0.
    _0cursor = 0.  ################ ATTENTION ###############
StringIO __post_init__ postconditions:
    _buf =   foo  bar  .
    _len = 12.
    _0cursor = 0.
```

Smit-create:
Fixes #1981

### DIFF
--- a/integration_tests/structs_27.py
+++ b/integration_tests/structs_27.py
@@ -1,4 +1,4 @@
-from lpython import dataclass, i32
+from lpython import dataclass, i32, u16, f32
 
 
 @dataclass
@@ -6,6 +6,14 @@ class StringIO:
     _buf     : str = ''
     _0cursor : i32 = 10
     _len     : i32 = 1
+
+@dataclass
+class StringIONew:
+    _buf     : str
+    _0cursor : i32 = i32(142)
+    _len     : i32 = i32(2439)
+    _var1    : u16 = u16(23)
+    _var2    : f32 = f32(30.24)
 
 
 def test_issue_1928():
@@ -47,4 +55,15 @@ def test_issue_1928():
     assert test_dude4._0cursor == 31
 
 
+def test_issue_1981():
+    integer_asr : str = '(Integer 4 [])'
+    test_dude   : StringIONew = StringIONew(integer_asr)
+    assert test_dude._buf == integer_asr
+    assert test_dude._len == 2439
+    assert test_dude._0cursor == 142
+    assert test_dude._var1 == u16(23)
+    assert abs(test_dude._var2 - f32(30.24)) < f32(1e-5)
+
+
+test_issue_1981()
 test_issue_1928()
--- a/integration_tests/structs_10.py
+++ b/integration_tests/structs_10.py
@@ -11,8 +11,13 @@ class Vec:
 
 @dataclass
 class MatVec:
-    mat: Mat = Mat([f64(0.0), f64(0.0)])
-    vec: Vec = Vec([f64(0.0), f64(0.0)])
+    # TODO: this should ideally be the following:
+    # mat: Mat = Mat(empty((2, 2), dtype=float64))
+    # vec: Vec = Vec(empty(2, dtype=float64))
+    # But since lpython ignores empty, the following is just a workaround.
+    # Follow the discussions here: https://github.com/lcompilers/lpython/issues/1809
+    mat: Mat = Mat(2.0)
+    vec: Vec = Vec(2.0)
 
 def rotate(mat_vec: MatVec) -> f64[2]:
     rotated_vec: f64[2] = empty(2, dtype=float64)
--- a/src/libasr/codegen/asr_to_c.cpp
+++ b/src/libasr/codegen/asr_to_c.cpp
@@ -146,11 +146,12 @@ public:
     void allocate_array_members_of_struct(ASR::StructType_t* der_type_t, std::string& sub,
         std::string indent, std::string name) {
         for( auto itr: der_type_t->m_symtab->get_scope() ) {
-            if( ASR::is_a<ASR::UnionType_t>(*itr.second) ||
-                ASR::is_a<ASR::StructType_t>(*itr.second) ) {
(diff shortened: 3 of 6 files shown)


integration_tests/structs_27.py
```diff
@@ -7,6 +7,14 @@ class StringIO:
     _0cursor : i32 = 10
     _len     : i32 = 1
 
+@dataclass
+class StringIONew:
+    _buf     : str
+    _0cursor : i32 = i32(142)
+    _len     : i32 = i32(2439)
```
certik, line 14: This PR assigns this to the "value" member of Variable. I am not sure if our code assumes that this is a constant, so the backend doesn't even visit this variable and uses the value instead. If so, then this won't work, since any modifications to `_len` will be lost.

certik: Let's add a test for modifying `_len`.

certik: Here is documentation for Variable:
```
`symbolic_value` the optional symbolic expression to initialize the variable
(e.g. `2+3+4+x`), this value must be compile time, but it is not necessarily a
constant (e.g., can contain binary operations, other variables, etc.)

`value` the optional constant expression holding the compile time value
(e.g. `5`, or `5.5`), it is a compile time constant.
```
So `value` and `symbolic_value` must be compile time constants. So let's require that for now as well in LPython.

A separate question is if they are set, if the `Variable` itself has to be a constant or not. I would say if it has to be a constant, then it should have a type Const. Otherwise it can change, which means `value` is the initial value, but the runtime value can change.

certik:
TODO:
* [ ] document that value in Variable is an initializer, but the runtime value can change.

certik:
TODO:
* [ ] inside a function, a statement `x: i32 = i32(0)` is now visited twice, which I think will cause problems.

certik:
I think `x: i32 = i32(0)` inside a function: symbol table visitor can declare the "x" variable, but it should not process the initializer. The body visitor processes the initializer and creates a regular assignment.

The `x: i32 = i32(0)` inside a class (or dataclass) will get fully processed by the symbol table visitor, and the RHS will become the "value", i.e., the ASR initializer.

certik:
Otherwise I think this is fine.

integration_tests/structs_27.py
Smit-create: Yeah, so I tested, and if the variable is having a value then it doesn't get changed after the assignment, so basically, it should only be used with the const variables.

Now, this problem gets trickier as Structs don't have a body, and, so `AnnAssign` which is an Assignment to the struct variables isn't possible with the current design.

Smit-create: One thing that we may do is add a field in the Struct to contain initializers for all it's variable members like the following:
```diff
     | StructType(symbol_table symtab, identifier name, identifier* dependencies,
         identifier* members, abi abi, access access, bool is_packed, bool is_abstract,
-        expr? alignment, symbol? parent)
+        expr* initializers, expr? alignment, symbol? parent)
```

Smit-create:
This is not required as #2056 is merged.

Issue #2011, opened 2023-06-23
BLOCKER: LPython can't find dict key, but CPython can find it


rebcabin:
repro: https://github.com/rebcabin/lpython/tree/brian-lasr/lasr/LP-pycharm/Issue2011

offending code:

```python
I4C: dict[str, i32] = {
    '0': 0, '1': 1, '2': 2, '3': 3,
    '4': 4, '5': 5, '6': 6, '7': 7,
    '8': 8, '9': 9,
    'a': 10, 'b': 11, 'c': 12, 'd': 13,
    'A': 10, 'B': 11, 'C': 12, 'D': 13,
    'e': 14, 'f': 15,
    'E': 14, 'F': 15}


def cnvi(s : str, base : i32=10) -> i32:
    """Assume input has been through 'match_integer'."""
    assert base == 10 or base == 8 or base == 16 or base == 2
    result : i32 = 0
    c : str
    t : str = strip_numerical_noise(s)
    pow: i32 = base ** (len(t) - 1)
    for c in t:
        incr : i32 = pow * I4C[c]
        result += incr
        pow = (pow // base)
    return result


if __name__ == '__main__':
    assert cnvi('0b0', base=2) == 0
    assert cnvi('0b1', base=2) == 1
    assert cnvi('0b10', base=2) == 2
    assert cnvi('0b11', base=2) == 3
    assert cnvi('0b1111_0100_111', base=2) == 1959
    assert cnvi('0b7a7', base=16) == 1959  ############# ATTENTION ##############
```

runs:

```bash
(lp) ┌─(~/CLionProjects/lpython/lasr/LP-pycharm/Issue2011)───────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s012)─┐
└─(17:51:22 on brian-lasr ✹ ✭)──> PYTHONPATH='../../../src/runtime/lpython' python lasr_lexer.py                                                          ──(Thu,Jun22)─┘
(lp) ┌─(~/CLionProjects/lpython/lasr/LP-pycharm/Issue2011)───────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s012)─┐
└─(17:51:30 on brian-lasr ✹ ✭)──> ~/CLionProjects/lpython/src/bin/lpython -I. lasr_lexer.py                                                               ──(Thu,Jun22)─┘
KeyError: The dict does not contain the specified key
```

certik:
Here is a minimal reproducible example:
```python
[...]

PR #2014, opened 2023-06-23
Fix ``LLVMDictSeparateChaining::dict_deepcopy``

czgdp1807:
Closes https://github.com/lcompilers/lpython/issues/2011


certik (approved):
Perfect. The fix is simple, thanks for doing this @czgdp1807, it would take me forever to figure out.

I have two questions:

* What was the problem?
* Is there any performance regression with this fix?

I am going to merge it, as it fixes the bug.

PR #2015, opened 2023-06-23
PythonCall: Code Refactor

Shaikh-Ubaid:
This PRs does some code refactoring related to `@pythoncall` support in C/C_CPP files. No new tests were added.

PR #2017, opened 2023-06-23
PythonCall: Support arrays of other simple types

Shaikh-Ubaid:
This PR adds support (and tests) for arrays of unsigned integers, bool, complex types for the `@pythoncall` decorator.

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -1076,7 +1076,7 @@ public:
         } else if (var_annotation == "str") {
             type = ASRUtils::TYPE(ASR::make_Character_t(al, loc, 1, -2, nullptr));
             type = ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
-        } else if (var_annotation == "bool") {
+        } else if (var_annotation == "bool" || var_annotation == "i1") {
             type = ASRUtils::TYPE(ASR::make_Logical_t(al, loc, 4));
             type = ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
         } else if (var_annotation == "CPtr") {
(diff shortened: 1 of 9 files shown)


Shaikh-Ubaid:
Only the latest `4` commits are part of this PR. The rest of the commits are part of the PR https://github.com/lcompilers/lpython/pull/2015. I will rebase this PR after merging/completion of https://github.com/lcompilers/lpython/pull/2015.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1076,7 +1076,7 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
         } else if (var_annotation == "str") {
             type = ASRUtils::TYPE(ASR::make_Character_t(al, loc, 1, -2, nullptr));
             type = ASRUtils::make_Array_t_util(al, loc, type, dims.p, dims.size());
-        } else if (var_annotation == "bool") {
+        } else if (var_annotation == "bool" || var_annotation == "i1") {
```
Shaikh-Ubaid, line 1079: It seems that `bool` is a predefined type in `CPython`. In `CPython`, we declare arrays of `bool` as `a: bool[5]`. Here, when declaring arrays of `bool` as a parameter to a function, we receive `TypeError: 'type' object is not subscriptable`;

Example:
```bash
$ cat examples/expr2.py 
from numpy import empty

def my_func(a: bool[:]):
    print(a)

def main0():
    x: bool[5] = empty([5], dtype=bool)
    my_func(x)

main0()
$ python examples/expr2.py
  File "/Users/ubaid/Desktop/OpenSource/lpython/examples/expr2.py", line 3, in <module>
    def my_func(a: bool[:]):
```

One of the ways to fix it could be to overload the default/inbuilt `bool` to support subscripting. This does not look great as it involves manipulating the inbuilt `bool`, which could (later) possibly lead to unexpected outcomes/experiences.

Other approach could be to define a `bool` type in `lpython.py` which supports subscripting and import it. This seems good but if used, the original/inbuilt `bool` gets hidded/shadowed. So, we might not be able to use `bool()` (I think we handle this by implementing `__call__()` magic method for our custom `bool`.

Another approach could be to define a custom type (for now, I named it `i1`) in `lpython.py` which supports subscripting.

I followed the third approach in this PR. Please share if it is fine.

Shaikh-Ubaid:
Ready.

Shaikh-Ubaid:
This PR needs a rebase after merging/completion of https://github.com/lcompilers/lpython/pull/2015. I am temporarily marking this as draft.

Shaikh-Ubaid:
Ready.

src/lpython/semantics/python_ast_to_asr.cpp
certik: Yes, 3rd option is fine for now. This is the issue with all intrinsic types, that they don't support our shortuct `i32[:]` notation. For those type, we can support the "full form" notation `Array[bool, :]` or something like that.

certik (approved):
I think that this looks good, thank you for this!

PR #1932, opened 2023-06-17
Support nested tuples

kabra1110:
Fixes #1886.

PR #2018, opened 2023-06-24
Remove `new` and `delete` from `TupleConcat`

kabra1110:
Fixes https://github.com/lcompilers/lpython/pull/1932#discussion_r1235964970


certik (approved):
Thanks! That should work.

PR #2019, opened 2023-06-24
Support character iteration over strings

faze-geek:
```
def f():
    s: str
    for s in "abcd":
        print(s)
f()

```
On main:
```
semantic error: Only function call `range(..)` supported as for loop iteration for now
 --> try.py:5:5 - 6:16
  |
5 |        for s in "abcd":
  |        ^^^^^^^^^^^^^^^^...
...
  |
6 |            print(s)
```

On branch :
```
a
b
c
d
```


certik (approved):
Awesome, thanks!

PR #2022, opened 2023-06-25
Handle the case where start is greater than end in `list.index()`

faze-geek:
A rather unlikely edge case -
```
def f():
    lst :list[i32] = [1, 2, 3]
    print(lst.index(2,2,1))

f()
```
On master :

```
  File "C:\Users\kunni\lpython\try.py", line 5, in <module>
    f()
  File "C:\Users\kunni\lpython\try.py", line 3, in f
    print(lst.index(2,2,1))
ValueError: 2 is not in list

2
```
On branch :
```
ValueError: The list does not contain the element: 2
```


certik (approved):

PR #2023, opened 2023-06-25
Add `dict.keys`

kabra1110:
Issue: Some extra zeros in list.
```python
from lpython import i32

def test_dict_keys_values():
    d: dict[i32, i32] = {1001:2002, 3003:4004}
    x: list[i32]
    x = d.keys()
    print(x)

test_dict_keys_values()
```
```
[0, 1001, 0, 3003, 0]
```


certik (approved):
I think this looks good. @kabra1110 is this ready?

certik:
Ah I see, there are some extra zeros.

Make sure to test the result and make it work in both LPython and CPython.

czgdp1807 (approved):

PR #2025, opened 2023-06-25
Add `list` comparison

### LINKED ISSUE #1832: Support List Compare like String Compare (index by index)
```
print([3,4,5]<[1,6,7])             # Expected output is false

semantic error: Compare: only Integer, Real, Logical, or String can be on the LHS and RHS.If operator is Eq or NotEq then Complex type is also acceptable
 --> try.py:2:11
  |
2 |     print([3,4,5]<[1,6,7])
```

kabra1110:
~WIP.~
Fixes #1832.
~This PR implements only less than operation.~


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -3408,17 +3429,73 @@ namespace LCompilers {
...
+            cnd = builder->CreateAnd(cnd, LLVM::CreateLoad(*builder, equality_holds));
+            builder->CreateCondBr(cnd, loopbody, loopend);
+        }
+
+        // body
+        llvm_utils->start_new_block(loopbody);
+        {
+            llvm::Value* i = LLVM::CreateLoad(*builder, idx);
+            // llvm::Value* t1i = llvm_utils->tuple_api->read_item(t1, i, LLVM::is_llvm_struct(
+            //                                 tuple_type->m_type[i]));
```
kabra1110, line 3479: How do we index `tuple_type->m_type` with run time value `i`?

czgdp1807: We can't. :-). Tuples should only be indexed with fixed indices available at compile time so that type of the indexed item can be figured out at compile time itself. Its the pattern as C++ `std::tuple`.

kabra1110: I see. Here we need to break out of the loop when equality does not hold at the first time. Run time values seem necessary for this.

czgdp1807 (commented):
Some abstract APIs to reduce code duplication - https://llvm.org/doxygen/classllvm_1_1IRBuilderBase.html#a843d43d070f0b1c6a133403edce488ef.

Issue #2006, opened 2023-06-22
Internal function `def` causes SEGSIGV


rebcabin:
Should perhaps produce a `NotImplemented` error message:

```bash
(lp) ┌─(~/CLionProjects/lpython/lasr/LP-pycharm)─────────────────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s012)─┐
└─(15:35:42 on brian-lasr ✹ ✭)──> cat > expr45.py                                                                                                         ──(Thu,Jun22)─┘
def foo():
    def bar():
        pass
    pass
(lp) ┌─(~/CLionProjects/lpython/lasr/LP-pycharm)─────────────────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s012)─┐
└─(15:36:41 on brian-lasr ✹ ✭)──> PYTHONPATH='../../src/runtime/lpython' python expr45.py                                                                 ──(Thu,Jun22)─┘
(lp) ┌─(~/CLionProjects/lpython/lasr/LP-pycharm)─────────────────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s012)─┐
└─(15:36:50 on brian-lasr ✹ ✭)──> ../../src/bin/lpython -I. ./expr45.py                                                                                   ──(Thu,Jun22)─┘
Traceback (most recent call last):
  File "/Users/brian/CLionProjects/lpython/src/bin/lpython.cpp", line 1844
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir,
  File "/Users/brian/CLionProjects/lpython/src/bin/lpython.cpp", line 783
    r1 = LCompilers::LPython::python_ast_to_asr(al, lm, *ast, diagnostics, compiler_options,
  File "/Users/brian/CLionProjects/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 7532
    auto res2 = body_visitor(al, lm, *ast_m, diagnostics, unit, main_module,
  File "/Users/brian/CLionProjects/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 7485
    b.visit_Module(ast);
  File "/Users/brian/CLionProjects/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 4620
    visit_stmt(*x.m_body[i]);
  File "/Users/brian/CLionProjects/lpython/src/lpython/python_ast.h", line 1883
[...]

PR #2035, opened 2023-06-26
fix: internal function def errors

arteevraina:
Towards: #2006

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -4761,6 +4761,12 @@ public:
     void visit_FunctionDef(const AST::FunctionDef_t &x) {
         SymbolTable *old_scope = current_scope;
         ASR::symbol_t *t = current_scope->get_symbol(x.m_name);
+
+        if (t == nullptr) {
+            // Throw Not implemented error.
+            throw SemanticError("Internal FunctionDef: Not implemented", x.base.base.loc);
+        }        
+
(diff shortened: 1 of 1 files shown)


certik:
Does get_symbol return a nullptr if it can't find the symbol?

arteevraina:
> Does get_symbol return a nullptr if it can't find the symbol?

Yes, it returns ``nullptr`` when it can't find the symbol [here](https://github.com/lcompilers/lpython/blob/main/src/libasr/asr_scopes.h#L62).

The problem here is ``add_symbol`` function is only called for outer function and not for internal function definition. This I have verified by debugging in my local environment.

Also, for outer function definition `add_symbol` is being called after calling `make_Function_t_util` [here](https://github.com/lcompilers/lpython/blob/a18690b44b5b9ef78037d3eca3f69d3397fdc9d5/src/lpython/semantics/python_ast_to_asr.cpp#L4232-L4251). I am not sure why it is not doing the same for internal function definition as well.

I tried doing something like this :

```
if (t == nullptr) {
            old_scope->add_symbol(x.m_name, (ASR::symbol_t *) &x);
            visit_FunctionDef(x);
}
```

But, in the above case, `x->type` is getting as ``Program`` instead of ``Function``. That's why I think it's necessary to call ``make_Function_t_util`` for internal function definition as well.

certik (approved):
Ok, I think we can merge it, it's a better error message for sure.

certik:
Thanks!

PR #2055, opened 2023-06-28
Support While Loop in Global scope

faze-geek:
```
i :i32 = 0
while(i<5):
    print(i)
    i=i+1
```

On main:
```
LCompilersException: visit_While() not implemented
```

On branch:
```
0
1
2
3
4
```


faze-geek:
Not sure why the tests fail, will need to investigate it.

faze-geek:
@Shaikh-Ubaid Can you have a look at the failing tests. Thanks!

integration_tests/loop_06.py
```diff
@@ -0,0 +1,8 @@
+from lpython import i32
+
+i: i32 = 0
+j: i32
+
+while(i<5):
+    j = i
+    print("j: ", j)
```
Shaikh-Ubaid, line 8:
```suggestion
    print("j: ", j)
    i = i + 1

assert j == 4

```
It seems the variable `i` was not incremented. Thus it leads to an infinite loop. The above change should fix it (Note: I also added an assert).

faze-geek: Oops. Didn't realize this :)
Will make the changes quick.

faze-geek:
This is ready !

Shaikh-Ubaid:
I would rebase and arrange the changes into two commits:
- Skip while in SymbolTableVisitor
- TEST: Add test for global while

czgdp1807 (commented):
@Thirumalai-Shaktivel This should be working now with the global statements pass you implemented?

Thirumalai-Shaktivel:
Yup, this works in the main!
```py
i :i32 = 0
while(i<5):
    print(i)
    i=i+1
```
```console
$ lpython examples/expr2.py
0
1
2
3
4
```

czgdp1807:
Let's close this then.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4520,6 +4520,10 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
         // We skip this in the SymbolTable visitor, but visit it in the BodyVisitor
     }
 
+    void visit_While(const AST::While_t &/*x*/) {
+        // We skip this in the SymbolTable visitor, but visit it in the BodyVisitor
+    }
```
Shaikh-Ubaid, line 4525: I think this failed in main because previously there was no `visit_While()` in `SymbolTableVisitor`. It seems the `visit_While()` for `SymbolTableVisitor` was recently added in https://github.com/lcompilers/lpython/commit/47bd6d539bb062d8133e222e8d08b5610db4a466. It appears that there was no test case added for global while. I guess we could add a test case, even a minor one like in this PR should be good. I think this would help in ensuring that the global while does not break with future changes.

PR #2056, opened 2023-06-29
Store initializer exprs for Structs in ASR

Smit-create:


Smit-create:
@certik This is an alternative approach to #2013 and I think this is more robust and doesn't create any confusion.

certik:
I think that this is fine. Isn't it equivalent to store an iniitalizer in the Variable (inside a struct) or in the Struct?

Smit-create:
> I think that this is fine. Isn't it equivalent to store an iniitalizer in the Variable (inside a struct) or in the Struct?

Yeah, that works. But it would require a lot of methods to be moved in CommonVisitor and may create problems of visiting them again in BodyVisitor which may affect the performance.

Smit-create:
@certik This PR fixes all the issues related to #1981, and #2013. Moreover, also fixes the test in #2058.

czgdp1807 (approved):

PR #2057, opened 2023-06-29
Added support for assert through SymbolicCompare

anutosh491:
The Pr adds support for the following 
1) Freeing the basic variables declared throughout the C generated code . For eg
```
from sympy import Symbol, pi
from lpython import S

def main0():
    x: S = Symbol('x')
    y: S = Symbol('y')
    print(pi + S(100))

main0()

// Implementations
void main0()
{
    basic x;
    basic_new_stack(x);
    basic y;
    basic_new_stack(y);
    symbol_set(x, "x");
    symbol_set(y, "y");
    basic queue2;
    basic_new_stack(queue2);
    basic queue3;
    basic_new_stack(queue3);
    basic_const_pi(queue3);
    basic queue4;
    basic_new_stack(queue4);
    integer_set_si(queue4, 100);
    basic_add(queue2, queue3, queue4);
    printf("%s\n", basic_str(queue2));
    basic_free_stack(queue4);
    basic_free_stack(queue3);
    basic_free_stack(queue2);
    basic_free_stack(y);
    basic_free_stack(x);
}
```

2) Support `equal to` and `not equal to` operators through `assert`
```
from sympy import Symbol, pi
from lpython import S

def test_operator_chaining():
    w: S = S(2)
    x: S = Symbol('x')
    y: S = Symbol('y')
    z: S = Symbol('z')

    a: S = x * w
    b: S = a + pi
    c: S = b / z
    d: S = c ** w

    assert(a == S(2)*x)
    assert(b == pi + S(2)*x)
    assert(c == (pi + S(2)*x)/z)
    assert(d == (pi + S(2)*x)**S(2)/z**S(2))
```
All test files have been updated to use assert

### DIFF
--- a/integration_tests/symbolics_01.py
+++ b/integration_tests/symbolics_01.py
@@ -6,6 +6,7 @@ def main0():
     y: S = Symbol('y')
     x = pi
     z: S = x + y
+    print(z)
     assert(z == pi + y)
     assert(z != S(2)*pi + y)
 
--- a/integration_tests/symbolics_01.py
+++ b/integration_tests/symbolics_01.py
@@ -6,6 +6,7 @@ def main0():
     y: S = Symbol('y')
     x = pi
     z: S = x + y
-    print(z)
+    assert(z == pi + y)
+    assert(z != S(2)*pi + y)
 
 main0()
--- a/integration_tests/symbolics_02.py
+++ b/integration_tests/symbolics_02.py
@@ -8,27 +8,33 @@ def test_symbolic_operations():
     # Addition
     z: S = x + y
     assert(z == x + y)
+    print(z)
     
     # Subtraction
     w: S = x - y
     assert(w == x - y)
+    print(w)
     
     # Multiplication
     u: S = x * y
     assert(u == x * y)
-    
+    print(u)
+
     # Division
     v: S = x / y
     assert(v == x / y)
-    
+    print(v)
+
     # Power
     p: S = x ** y
     assert(p == x ** y)
-    
+    print(p)
+
     # Casting
     a: S = S(100)
     b: S = S(-100)
     c: S = a + b
     assert(c == S(0))
+    print(c)
 
 test_symbolic_operations()
--- a/integration_tests/symbolics_02.py
+++ b/integration_tests/symbolics_02.py
@@ -7,28 +7,28 @@ def test_symbolic_operations():
     
     # Addition
     z: S = x + y
-    print(z)  # Expected: x + y
+    assert(z == x + y)
     
     # Subtraction
     w: S = x - y
-    print(w)  # Expected: x - y
+    assert(w == x - y)
     
     # Multiplication
     u: S = x * y
-    print(u)  # Expected: x*y
+    assert(u == x * y)
     
     # Division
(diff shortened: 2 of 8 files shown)


anutosh491:
@certik 
There are essentially two things I would like to point out here 

1) Firstly for comparison operators ( `==`, `!=` etc) , we can have two procedures I think 
i) We could have define something like `SymbolicCompare`( just like `StringCompare`, `ListCompare` etc) whenever we are comparing two symbolic expressions. Here would have to resolve stuff through `visit_SymbolicCompare`
ii) I think we can also end up introducing `SymbolicEqualTo` , `SymbolicNotEqualTo` as intrinsic function nodes (just like `SymbolicAdd`, `SymbolicMul`) . Here we would have to resolve stuff through `visit_IntrinsicFunction` through the switch case where we can implement logic for `SymbolicEqualTo` etc

I've chosen the first one as of now . I didn't see any particular advantages to the second approach and the first approach was doable easily.

2) The main concern any approach we choose could be the following . So the file for symengine's C wrapper (https://github.com/symengine/symengine/blob/2b575b9be9bb499d866dc3e411e6368ca0d1bb42/symengine/cwrapper.h#L256) essentially has function only to deal with `==` and `!=`

Once I implemented `SymbolicCompare`, I thought I might as well handle other comparison operators too like (`>`, `<` , `>=` etc) but then I realized that there are no functions to handle these . I hope I've not overlooked anything , you could check once .
So we have `basic_eq` and `basic_neq` for `==` and `!=` but no functions for others.
So we can do the following as of now
```
>>> x = Symbol('x')
>>> y = Symbol('y')
>>> x == y
False
>>> x != y
True
```
But not
```
>>> x > y
x > y
>>> x <= y
x <= y
```
Well it does serve our purpose as of now which is using `assert` instead of `print` while framing tests but we might want to consider this at some point !

anutosh491:
> Can you please separate this PR into two PRs, one for freeing variables, the other one for adding support for assert/compare?
Well most of the changes here are to support the assert statement . The code for freeing variables is actually quite minimal if you see.
1) Defining a set to store released variables 
2) Adding `queue[queue_front]` to the set whenever we move `queue_front` by 1
3) Finally a for loop to get elements out of the set , and add it to the main function through `basic_free_stack(var)` 
So it's around 7-8 lines overall , might not need a new PR. 

> Keep the print statements in the test, they test that printing works. And the assert tests that the result what we print is correct.

Yes, this should be done !

certik:
It's hard to review multiple independent changes. Please separate the PRs, that way we can ensure that each change is clean and we can merge it.

anutosh491:
Sure I'll do that right now !

anutosh491:
I've removed the changes related to freeing of variables , hope that makes it easier for you to review . Nothing changes, the tests should still pass as everything is seems except we are not freeing the stack space at the end . I'll raise a pr for that as soon as this goes in.

certik (approved):
I think this looks great!

certik:
@anutosh491 go ahead and submit a PR for the freeing of variables.

anutosh491:
I'll do it right away !Give me 5 minutes.

Issue #2054, opened 2023-06-28
C backend bug: ~ (not) operation


ronnuriel:
Hey, not op seems to be generated wrong:
```
def foo(grp: u16) -> u16:
    i: u16 = ~(u16(grp))

    return i


def foo2() -> u16:
    i: u16 = ~(u16(0xffff))

    return i

def foo3() -> u16:
    i: u16 = ~(u16(0xffff))

    return ~i

assert foo(u16(0)) == u16(0xffff), foo(u16(0))
assert foo(u16(0xffff)) == u16(0), foo(u16(0xffff))
assert foo2() == u16(0), foo2()
assert foo3() == u16(0xffff), foo3()
```

output:
```
#include <inttypes.h>

#include <stdlib.h>
#include <stdbool.h>
#include <stdio.h>
#include <string.h>
#include <lfortran_intrinsics.h>


// Implementations
uint16_t foo(uint16_t grp)
{
    uint16_t _lpython_return_variable;
    uint16_t i;
    i = grp;
    _lpython_return_variable = i;
    return _lpython_return_variable;
}

uint16_t foo2()
{
    uint16_t _lpython_return_variable;
    uint16_t i;
    i = (uint16_t)(65535);
    _lpython_return_variable = i;
    return _lpython_return_variable;
}

uint16_t foo3()
{
    uint16_t _lpython_return_variable;
    uint16_t i;
    i = (uint16_t)(65535);
    _lpython_return_variable = i;
    return _lpython_return_variable;
}

void _lpython_main_program()
{
    ASSERT_MSG(foo((uint16_t)(0)) == (uint16_t)(65535), foo((uint16_t)(0)));
    ASSERT_MSG(foo((uint16_t)(65535)) == (uint16_t)(0), foo((uint16_t)(65535)));
    ASSERT_MSG(foo2() == (uint16_t)(0), foo2());
    ASSERT_MSG(foo3() == (uint16_t)(65535), foo3());
}

int main(int argc, char* argv[])
{
    _lpython_set_argv(argc, argv);
    _lpython_main_program();
    return 0;
}
```

Please add them to Lpython tests

certik:
The first three asserts fail in CPython also, so that should not work in LPython. The last one succeeds in CPython but fails in LPython, here is a minimal test:
```python
from lpython import u16

def foo(grp: u16) -> u16:
    i: u16 = ~(u16(grp))
    return i


def foo2() -> u16:
[...]

Issue #2050, opened 2023-06-27
Lift Python interoperability from C to ASR


ubaidsk:
The next step is to lift the Python interoperability from C to ASR (see #1996) for both `@pythoncall` and `@pythoncallable`, this will make it work in all our backends. It should transform BindC functions into directly calls to Python C/API, by probably having some module with the C/API interfaces and then calling them appropriately, all in ASR.

_Originally posted by @certik in https://github.com/lcompilers/lpython/issues/1704#issuecomment-1609814856_

Shaikh-Ubaid:
```bash
PyAPI_DATA(PyObject *) PyExc_BaseException;
PyAPI_DATA(PyObject *) PyExc_Exception;
#if !defined(Py_LIMITED_API) || Py_LIMITED_API+0 >= 0x03050000
PyAPI_DATA(PyObject *) PyExc_StopAsyncIteration;
#endif
PyAPI_DATA(PyObject *) PyExc_StopIteration;
PyAPI_DATA(PyObject *) PyExc_GeneratorExit;
PyAPI_DATA(PyObject *) PyExc_ArithmeticError;
PyAPI_DATA(PyObject *) PyExc_LookupError;

PyAPI_DATA(PyObject *) PyExc_AssertionError;
PyAPI_DATA(PyObject *) PyExc_AttributeError;
PyAPI_DATA(PyObject *) PyExc_BufferError;
PyAPI_DATA(PyObject *) PyExc_EOFError;
PyAPI_DATA(PyObject *) PyExc_FloatingPointError;
PyAPI_DATA(PyObject *) PyExc_OSError;
PyAPI_DATA(PyObject *) PyExc_ImportError;
```

It seems there are several predefined exception codes which get defined when including `Python.h`. We currently use the exception `PyExc_ImportError` as follows:

For example:
```cpp
    if (_import_array() < 0) {
        PyErr_Print();
        PyErr_SetString(PyExc_ImportError, "numpy.core.multiarray failed to import");
        fprintf(stderr, "Failed to import numpy Python module(s)\n");
        return -1;
    }
```

How do we support the above exception codes at the ASR Level?

Shaikh-Ubaid:
The `C` backend also needs to include the following:
```cpp
#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
#include <Python.h>
#include <numpy/arrayobject.h>
```

Do we have a mechanism in the ASR to support these?

certik:
[...]

PR #2061, opened 2023-06-29
Fix unsigned integer bitnot

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2054


Shaikh-Ubaid:
In the commit [lpython.py: Use numpy functions](https://github.com/lcompilers/lpython/pull/2061/commits/465bb814cd4ad76b5c24ad1829800ce43825a018), I replaced custom functions with `numpy` functions. This way we need not implement every operator for custom Unsigned Integer class. A drawback here seems to be that everytime `lpython` is imported, it loads `numpy` (and thus could affect performance of `CPython`).

Please share if it is fine.

Shaikh-Ubaid:
Ready.

certik:
I would only depend on CPython for now by default. Of course if you use NumPy then you need NumPy, but I would not require it if you don't use NumPy.

certik:
I think this looks great. Do you mind splitting the assert improvement into a separate PR? It will be much easier to review.

Shaikh-Ubaid:
Ready.

integration_tests/bindpy_01.py
```diff
@@ -1,11 +1,11 @@
 from lpython import i32, i64, u32, u64, f32, f64, pythoncall
 
 @pythoncall(module = "bindpy_01_module")
-def add_ints(a: i32, b: i64, c: u32, d: u64) -> i64:
+def add_ints(a: u32, b: u32, c: u32, d: u32) -> u64:
     pass
 
 @pythoncall(module = "bindpy_01_module")
-def multiply_ints(a: i32, b: i64, c: u32, d: u64) -> i64:
+def multiply_ints(a: u32, b: u32, c: u32, d: u32) -> u64:
```
certik, line 8: Are the signed integers tested somewhere? It seems we are now only testing unsigned integers, but we need to test both signed and unsigned.

certik:
I think this PR is great now, except that it seems the signed integer tests were removed --- if you can test both signed and unsigned, then we can merge it.

integration_tests/bindpy_01.py
```diff
@@ -1,11 +1,19 @@
 from lpython import i32, i64, u32, u64, f32, f64, pythoncall
 
 @pythoncall(module = "bindpy_01_module")
-def add_ints(a: i32, b: i64, c: u32, d: u64) -> i64:
+def add_ints(a: i32, b: i32, c: i32, d: i32) -> i64:
```
certik, line 4: Is i64 not tested now?

integration_tests/bindpy_01.py
```diff
@@ -1,11 +1,19 @@
...
     pass
 
 @pythoncall(module = "bindpy_01_module")
-def multiply_ints(a: i32, b: i64, c: u32, d: u64) -> i64:
+def multiply_ints(a: i32, b: i32, c: i32, d: i32) -> i64:
+    pass
+
+@pythoncall(module = "bindpy_01_module")
+def add_unsigned_ints(a: u32, b: u32, c: u32, d: u32) -> u64:
```
certik, line 12: Let's test u64 too.

certik:
I would say in this PR make sure all previous tests are left, and just add some unsigned tests.

Then in the next PR, let's test rigorously i8, i16, i32, i64 and u8, u16, u32, u64 to make sure they all work.

Shaikh-Ubaid:
>Is i64 not tested now?
>Let's test u64 too.
>I would say in this PR make sure all previous tests are left, and just add some unsigned tests.

Yes, `i64` is not tested as a direct parameter. It might be tested as an array of `i64` (in some other test). 

These tests were added to test initial support of `cpython` interop. I would not make the test big. I would add it as a separate-test / new-test-file (if needed). I think simple and smaller tests are much simpler to debug and figure the failure. Testing all the types into one file could lead to tricky debugging, I think.

These types will be tested thoroughly as we add more support for CPython-Interop (For example the pass at https://github.com/lcompilers/lpython/issues/2050). When I added this test initially, I willingly kept the test case slightly less robust (and even slightly odd, for example multiplying/adding variables of different types together) so that we can have some freedom (or thinking/discussion opportunities) in the design/structure of CPython Interop. 

We are needing to update this test `bindpy_01.py` since the UnsignedInteger class defined in `lpython.py` (for `CPython`) supports operations only its own instances. This was needed because after any operation we need the unsigned integer value to be in its valid range (as per its kind). 

Currently, the updated test tests integers (only `i32`) and unsigned integers (only `u32`) which seems good. We can add more tests in a separate file, I think.

Please share what we should do.

certik (approved):

certik:
I think this is fine. We should now make sure all combinations are tested.

certik:
We made a mistake here. The UnsignedIntegerBitNot wraps around, and thus does NOT correspond do the CPython's `~` operator, which does not wrap around. A solution is here:

https://github.com/lcompilers/lpython/issues/2173

We need to create a new intrinsic function `bitnot(u)` that maps to `UnsignedIntegerBitNot`, and then we must forbid `~u` applied to unsigned integers. Then we remove the UnsignedInteger class in emulation mode. That brings us back to be compatible with CPython.

Issue #2060, opened 2023-06-29
warning: comparison of different enumeration types


Shaikh-Ubaid:
```bash
[ 82%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/utils.cpp.o
[ 83%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/python_serialization.cpp.o
[ 86%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/python_evaluator.cpp.o
[ 86%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/semantics/python_ast_to_asr.cpp.o
parser.tab.cc:2929:9: warning: variable 'yynerrs' set but not used [-Wunused-but-set-variable]
    int yynerrs;
        ^
1 warning generated.
In file included from /Users/ubaid/Desktop/OpenSource/lpython/src/lpython/semantics/python_ast_to_asr.cpp:14:
/Users/ubaid/Desktop/OpenSource/lpython/src/libasr/asr.h:32:26: warning: comparison of different enumeration types ('LCompilers::ASR::unitType' and 'LCompilers::ASR::asrType') [-Wenum-compare]
    return T::class_type == x.type;
           ~~~~~~~~~~~~~ ^  ~~~~~~
/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/semantics/python_ast_to_asr.cpp:3844:32: note: in instantiation of function template specialization 'LCompilers::ASR::is_a<LCompilers::ASR::TranslationUnit_t, LCompilers::ASR::asr_t>' requested here
        LCOMPILERS_ASSERT(ASR::is_a<ASR::TranslationUnit_t>(*tmp0));
                               ^
1 warning generated.
[ 87%] Linking CXX static library liblpython_lib.a
[ 87%] Built target lpython_lib
[ 88%] Building CXX object src/lpython/tests/CMakeFiles/test_lpython.dir/test_llvm.cpp.o
[ 89%] Building CXX object src/lpython/tests/CMakeFiles/test_lpython.dir/test_asm.cpp.o
[ 90%] Building CXX object src/lpython/tests/CMakeFiles/test_lpython.dir/test_stacktrace2.cpp.o
[ 91%] Building CXX object src/bin/CMakeFiles/lpython.dir/lpython.cpp.o
[ 93%] Building CXX object src/lpython/tests/CMakeFiles/test_stacktrace.dir/test_stacktrace.cpp.o
[...]

PR #2062, opened 2023-06-29
Build: Fix warning in debug mode

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2060


certik (approved):
In a separate PR let's setup the `-Werror` checks at the CI like we have for LFortran, that will prevent these warnings.

PR #2065, opened 2023-06-30
Support expression as message in assert

Shaikh-Ubaid:
towards https://github.com/lcompilers/lpython/pull/2061#issuecomment-1614069543, https://github.com/lcompilers/lpython/issues/2054

### DIFF
--- a/integration_tests/expr_17.py
+++ b/integration_tests/expr_17.py
@@ -1,0 +1,16 @@
+from lpython import i32, f64
+
+def main0():
+    a: i32 = 4
+    b: i32 = 3
+    c: i32 = 12
+    assert a * b == c, a * b
+
+    d: f64 = 0.4
+    e: f64 = 2.5
+    f: f64 = 1.0
+    assert abs((d * e) - f) <= 1e-6, abs((d * e) - f)
+
+    assert a == b + 1, "Failed: a == b + 1"
+
+main0()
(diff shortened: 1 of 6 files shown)


integration_tests/expr_17.py
```diff
@@ -0,0 +1,14 @@
...
+def main0():
+    a: i32 = 4
+    b: i32 = 3
+    c: i32 = 12
+    assert a * b == c, a * b
+
+    d: f64 = 0.4
+    e: f64 = 2.5
+    f: f64 = 1.0
+    assert abs((d * e) - f) <= 1e-6, abs((d * e) - f)
```
certik, line 12: Shouldn't the second argument be a string? According to the docs: https://docs.python.org/3/reference/simple_stmts.html#grammar-token-python-grammar-assert_stmt, it says that `assert expression1, expression2` is equivalent to:
```python
if __debug__:
    if not expression1: raise AssertionError(expression2)
```
And I thought you put in a string into an exception constructor.

In any case, using strings is the most common I think, so I would at least test that, to make sure strings work.

certik:

integration_tests/expr_17.py
Shaikh-Ubaid: It seems expressions are also supported (since the added tests in this PR with expressions as assert work with CPython).

Shaikh-Ubaid: I updated the test `expr_18` in [TEST: Add test for expr as assert message](https://github.com/lcompilers/lpython/pull/2065/commits/a42dd0c3256fa95d372ba516b1d170f73e2db5a3) to print a string on assert failure.

certik (approved):

PR #2070, opened 2023-06-30
LLVM: Fix visit_IfExp()

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2071

towards https://github.com/lfortran/lfortran/pull/1901#issuecomment-1615216892


certik:
Can you construct a test in https://github.com/lcompilers/lpython/blob/00b8d59e0601ffdb1d4cf5f729b6d236f420e071/integration_tests/test_ifexp.py that fails in master and is fixed in this PR?

Then let's test this thoroughly there.

certik (approved):
Awesome, thanks!

We should test it a lot more thoroughly, but this is good enough to merge.

integration_tests/test_ifexp_02.py
```diff
@@ -0,0 +1,17 @@
+from lpython import i32, f32
```
Shaikh-Ubaid, line 1: It seems that `f32` is unused here. So removing it.

PR #2078, opened 2023-07-03
Throw error for unsupported unary operand

Shaikh-Ubaid:
Example:
```python
$ cat examples/expr2.py   

@dataclass
class A:
    n: i32

def main0():
    a: A = A(10)
    print(-a)

main0()
```
Expected:
```bash
$ python examples/expr2.py    
  File "/Users/ubaid/Desktop/OpenSource/lpython/examples/expr2.py", line 11, in <module>
    main0()
  File "/Users/ubaid/Desktop/OpenSource/lpython/examples/expr2.py", line 9, in main0
    print(-a)
TypeError: bad operand type for unary -: 'A'
```
On main:
```bash
$ lpython examples/expr2.py
10
```
On branch:
```bash
$ lpython examples/expr2.py 
semantic error: Unary operator '-' not supported for type A
 --> examples/expr2.py:9:11
  |
9 |     print(-a)
  |           ^^ 


```

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -3443,6 +3443,9 @@ public:
             } else if (ASRUtils::is_complex(*operand_type)) {
                 throw SemanticError("Unary operator '~' not supported for complex type",
                     x.base.base.loc);
+            } else {
+                throw SemanticError("Unary operator '~' not supported for type " + ASRUtils::type_to_str_python(operand_type),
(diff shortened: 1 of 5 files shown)


Shaikh-Ubaid:
The changes in this PR make the compiler throw error instead of executing and producing incorrect output. I think ideally for structs, we need to check if there is `__neg__()` magic method defined (and call it if it exists). I think we can support it later in a separate PR.

certik (approved):

Issue #2086, opened 2023-07-03
Missing conversion from `u8` to `u32` ?


rebcabin:
MVE:

```python
from lpython import (i8, i32, i64, f32, f64,
                     u8, u32,
                     TypeVar, Const,
                     dataclass)
from numpy import empty, sqrt, float32, float64, int8


# Issue 2066
n = TypeVar("n")


HDC_DIM: Const[i32] = 8192


@dataclass
class LPBHV:
    # Issue 2083: can't say HDC_DIM
    dim : i32 = 8192
    a : u8[8192] = empty(8192, dtype=int8)


@dataclass
class LPBHV_small:
    dim : i32 = 4
    a : u8[4] = empty(4, dtype=int8)  ######### ATTENTION ###########


def as_hex(lphv_small : LPBHV_small) -> None:
    i : i32
    print("HEX")
    for i in range(4):
        elt : u32 = u32(lphv_small.a[i])  ############## ATTENTION ##############
        hx : str = hex(elt)
        print(hx, end='')
        print(' ' if i < 4 - 1 else '', end='')
    print('')
    print("BITS")
    for i in range(4):
        elt : u32 = u32(lphv_small.a[i])
        elt &= 0xff
        hx : str = hex(elt)
        print(hx, end='')
        print(' ' if i < 4 - 1 else '', end='')
    print('')


def g() -> None:
    lpbhv_small : LPBHV_small = LPBHV_small()
    print("DECIMAL")
    print(lpbhv_small.a)
    as_hex(lpbhv_small)


if __name__ == "__main__":
    print("Module HDC")
    g()
```

```bash
/Users/brian/CLionProjects/lpython/src/bin/python /Users/brian/CLionProjects/lpython/lasr/LP-pycharm/issue2086.py 
semantic error: Type mismatch in annotation-assignment, the types must be compatible
  --> /Users/brian/CLionProjects/lpython/lasr/LP-pycharm/issue2086.py:32:9
   |
32 |         elt : u32 = u32(lphv_small.a[i])
   |         ^^^             ^^^^^^^^^^^^^^^ type mismatch ('u32' and 'u8')
```

certik:
The line `a : u8[4] = empty(4, dtype=int8)` should be `a : u8[4] = empty(4, dtype=uint8)`, so the current line should give an error message.

The second line `elt : u32 = u32(lphv_small.a[i])` should work, i.e., `u32(x)` should work for `x: u8`. So that should not give an error message.

PR #2090, opened 2023-07-03
Support cast UnsignedIntegerToUnsignedInteger

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2086


certik (approved):
I think that this is fine, but let's check all combinations of casts:

u8, u16, u32, u64 and cast one to another. You can do that in a subsequent PR.

Issue #2129, opened 2023-07-09
math.pi causes linker errors


certik:
This global variables in math.py:
```python
from lpython import i8, i16, i32, f32, i64, f64, ccall, overload
                                     
          
pi: f64 = 3.141592653589793238462643383279502884197
e: f64 = 2.718281828459045235360287471352662497757
tau: f64 = 6.283185307179586        
```
Cause the following linker errors using the C backend if compiling and linking more than one file:
```console
duplicate symbol '_pi_64' in:
    CMakeFiles/a.dir/a.c.o
    b/b.mod.a(b.c.o)
duplicate symbol '_pi_32' in:
    CMakeFiles/a.dir/a.c.o
    b/b.mod.a(b.c.o)
ld: 2 duplicate symbols for architecture arm64
```

certik:
This was caused by the (correct) fix not to use C macros but global variables. But they now clash.

certik:
I think we need to make the names unique to a translation unit, similarly to what we did in LFortran recently. We need to port the solution from here: https://github.com/lfortran/lfortran/pull/1898 and use it for these global variables.

certik:
I think in the C backend we need to use a global map (using a unique hash) where we save the actual (unique) name, and look it up. Similarly how we handle similar things in the LLVM backend.

Smit-create:
Can you please add a small reproducible example? Is this something related to https://github.com/lcompilers/lpython/issues/2076?

Shaikh-Ubaid:
I am probably able to reproduce it:

```bash
$ cat a.py 
from lpython import f64
from math import pi

def area_of_circle(r: f64) -> f64:
    return pi * r * r
```
```
$ cat b.py
from lpython import f64
from math import pi

def main0():
    print("hi")

main0()
```
```
$ lpython --show-c a.py --disable-main > main1.c
$ lpython --show-c b.py  > main2.c              
$ gcc -c main1.c -I src/libasr/runtime                         
$ gcc -c main2.c -I src/libasr/runtime                         
$ gcc main1.o main2.o -o my_app -Lsrc/runtime -llpython_runtime
duplicate symbol '_tau' in:
    main1.o
    main2.o
duplicate symbol '_pi' in:
    main1.o
    main2.o
duplicate symbol '_e' in:
    main1.o
    main2.o
ld: 3 duplicate symbols for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

[...]

PR #2138, opened 2023-07-10
Implement lcompilers_unique_ID global variable

certik:
This is initialized to a unique number for each run.


certik:
To fix #2129, we can now use the `lcompilers_unique_ID` global variable, which is always initialized now with a unique ID. We just need to append it into the global variable name in the C backend, so that it doesn't clash.

Smit-create (approved):

PR #2142, opened 2023-07-10
C: Use unique names for variables

Smit-create:
#2129

### DIFF
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -1500,6 +1500,7 @@ int main(int argc, char *argv[])
         bool print_targets = false;
         bool print_rtl_header_dir = false;
         bool print_rtl_dir = false;
+        bool separate_compilation = false;
 
         std::string arg_fmt_file;
         // int arg_fmt_indent = 4;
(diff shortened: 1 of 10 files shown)


Smit-create:
@certik please let me know if this is the correct approach to move further. There are many tests that still fail and it would require fixes to complete this.

certik:
I think so. This seems what I was thinking.

certik:
To help with the tests, let's hide it behind a compiler option, that way the new name would only be used with the option is given, so current tests should pass.

certik:
Sure. Depends what the failures are --- the issue with unique names is that all the C tests then will fail, as the name is changing with every run (as it should!).

Smit-create:
I have added the fix, but it's kind of hack-ish as we need to keep in mind about unique_id  while emitting the symbol name which can fail easily. How about writing a pass if we have `--generate-object-code`, then it would replace all the symbol_names with unique names. This would also help with any such backend where there would be linking issues like C, C++, etc.

certik:
I think writing the ASR pass will be better, as it will fix this issue for all the backends. Here is what the pass could do:

* all module level symbols will receive proper name mangling: prepend module name, and optional unique hash (depending on a compiler option).

I think this might be as easy as just modifying the symbol table, since we don't reference things by name, except in ExternalSymbols, so those would have to be updated too.

Smit-create:
I'll keep this PR open as all the tests pass successfully. I'll work on writing a pass in a new PR tomorrow morning.

Smit-create:
Completed in #2149

PR #2143, opened 2023-07-10
Remove global variables

certik:
Due to #2129, the global variables do not always work well. Until that issue is fixed, a simple workaround is to move these global variables into local variables.

Reference tests updated.


Smit-create (approved):
Thanks, we can fix this as a follow up.

Issue #2139, opened 2023-07-10
New warning on macOS


certik:
```
/Users/ondrej/repos/lpython/src/libasr/codegen/llvm_utils.cpp:1487:21: warning: variable 'pred' is used uninitialized whenever switch default is taken [-Wsometimes-uninitialized]
                    default: {
                    ^~~~~~~
/Users/ondrej/repos/lpython/src/libasr/codegen/llvm_utils.cpp:1491:43: note: uninitialized use occurs here
                return builder->CreateCmp(pred, left, right);
                                          ^~~~
/Users/ondrej/repos/lpython/src/libasr/codegen/llvm_utils.cpp:1465:9: note: variable 'pred' is declared here
        llvm::CmpInst::Predicate pred;
        ^
```
I thought we test for this at the CI. Maybe we only do it for LFortran.

Shaikh-Ubaid:
>I thought we test for this at the CI. Maybe we only do it for LFortran.

Yes, we have to add `-Werror` at the CI. We previously discussed it here https://github.com/lcompilers/lpython/pull/2062#pullrequestreview-1506421777. I have a minor concern. When building `lpython` there is warning about unused variable `yynerrs` in `parser.tab.cc`. This file `parser.tab.cc` is a generated file, thus we are unable to use the variable or add `maybe unused`. We do not experience this in `lfortran` since it uses `%glr-parser`, where as (it seems) `lpython` uses `LR(1)` parser.

The warning of unused variable `yynerrs` in `parser.tab.cc` seems to be a bug in `bison`. This bug seems to be fixed in the latest `bison`, but they did not make a release of it post fixing. I had requested for a release here https://github.com/akimd/bison/issues/103.

certik:
There might be a way to turn off `-Werror` for the Bison file. Alternatively we can "post process" the file to remove "yynerrs" with a Python script.

Issue #1901, opened 2023-06-13
Build warnings on fresh clone


rebcabin:
```shell
[ 74%] Building CXX object src/libasr/CMakeFiles/asr.dir/codegen/llvm_utils.cpp.o
/Users/brian/CLionProjects/lpython/src/libasr/codegen/llvm_utils.cpp:2650:61: warning: unused parameter 'list_type' [-Wunused-parameter]
    void LLVMList::reverse(llvm::Value* list, ASR::ttype_t* list_type, llvm::Module& module) {
                                                            ^
/Users/brian/CLionProjects/lpython/src/libasr/codegen/llvm_utils.cpp:3169:42: warning: unused parameter 'module' [-Wunused-parameter]
                           llvm::Module& module) {
                                         ^
2 warnings generated.
[ 75%] Linking CXX static library libasr.a
[ 75%] Built target asr
[ 78%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/semantics/python_ast_to_asr.cpp.o
[ 78%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/parser/tokenizer.cpp.o
[ 80%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/parser/parser.tab.cc.o
[ 80%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/parser/parser.cpp.o
[ 82%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/utils.cpp.o
[ 83%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/python_serialization.cpp.o
[ 83%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/python_evaluator.cpp.o
[ 85%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/__/bin/tpl/whereami/whereami.cpp.o
[ 86%] Building CXX object src/lpython/CMakeFiles/lpython_lib.dir/pickle.cpp.o
parser.tab.cc:2929:9: warning: variable 'yynerrs' set but not used [-Wunused-but-set-variable]
    int yynerrs;
        ^
1 warning generated.
[ 87%] Linking CXX static library liblpython_lib.a
```

Smit-create:
This will be probably fixed by #1888

PR #2145, opened 2023-07-10
CI: Check for warnings

ubaidsk:
fixes https://github.com/lcompilers/lpython/issues/2139
fixes https://github.com/lcompilers/lpython/issues/1901


Shaikh-Ubaid:
The debug build fails with the error `‘void free(void*)’ called on unallocated object ‘yyssa’ [-Werror=free-nonheap-object]`. It works with `clang` on my system. It seems that this is a bug in gcc https://lists.gnu.org/archive/html/help-bison/2021-01/msg00025.html.

PR #2149, opened 2023-07-11
ASR pass for unique symbols

Smit-create:
#2129, #2142

### DIFF
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -243,6 +243,11 @@ int emit_asr(const std::string &infile,
     pass_options.always_run = true;
     pass_options.verbose = compiler_options.verbose;
     pass_options.pass_cumulative = compiler_options.pass_cumulative;
+    pass_options.all_symbols_mangling = compiler_options.all_symbols_mangling;
+    pass_options.module_name_mangling = compiler_options.module_name_mangling;
+    pass_options.global_symbols_mangling = compiler_options.global_symbols_mangling;
+    pass_options.intrinsic_symbols_mangling = compiler_options.intrinsic_symbols_mangling;
+
 
     pass_manager.apply_passes(al, asr, pass_options, diagnostics);
 
@@ -345,6 +350,10 @@ int emit_c(const std::string &infile,
     pass_options.run_fun = "f";
     pass_options.always_run = true;
     pass_options.verbose = compiler_options.verbose;
+    pass_options.all_symbols_mangling = compiler_options.all_symbols_mangling;
+    pass_options.module_name_mangling = compiler_options.module_name_mangling;
+    pass_options.global_symbols_mangling = compiler_options.global_symbols_mangling;
+    pass_options.intrinsic_symbols_mangling = compiler_options.intrinsic_symbols_mangling;
 
     pass_manager.apply_passes(al, asr, pass_options, diagnostics);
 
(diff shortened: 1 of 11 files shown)


Smit-create:
@certik This PR now passes all the integration tests with LLVM and C backend with every possible name-mangling. It just fails for one test and that has something to do with nested structs and External Symbol. Can you please test this PR against your use cases?

certik:
I would structure this ASR pass like this:

* First create a hashmap of unique symbol hash -> (new name) of all symbols that you want to rename and choose how to rename them. Make it configurable how this choice will be made, at the very least it looks like we will need the following options:
  * module name mangling on/off
  * only mangle global symbols, but not symbols in any modules (so that you can call a function from another module with the same name, using an "interface")
  * only global symbols and symbols in intrinsic modules
  * every symbol in all modules
* Then in the second pass go over and change the name, by changing the symbol, all dependencies to it and all external symbols that mention it. If there are other references elsewhere in ASR, those have to be changed as well.

We can then keep extending the logic in the first pass as needed, and it could be that there is no single way forward, and users will select the behavior on the command line that they need for their application / build system.

certik:
Register the pass in https://github.com/lcompilers/lpython/blob/main/src/libasr/gen_pass.py and run the script to generate the `.h` file. Keep the generated file committed.

src/bin/lpython.cpp
```diff
@@ -1614,7 +1614,7 @@ int main(int argc, char *argv[])
         app.require_subcommand(0, 1);
         CLI11_PARSE(app, argc, argv);
 
-        lcompilers_unique_ID = LCompilers::get_unique_ID();
+        lcompilers_unique_ID = "__" + LCompilers::get_unique_ID();
```
certik, line 1617: I would keep the `__` out of the unique ID and leave it to the use case to insert this properly.

src/libasr/pass/unique_symbols.cpp
```diff
@@ -0,0 +1,602 @@
...
+        }
+        for (auto &a : x.m_symtab->get_scope()) {
+            this->visit_symbol(*a.second);
+        }
+    }
+
+};
+
+
+class UniqueSymbolVisitor: public ASR::BaseWalkVisitor<UniqueSymbolVisitor> {
```
certik, line 211: Document here that this visitor does the renaming (and only renaming) using the `std::map`.

src/libasr/pass/unique_symbols.cpp
```diff
@@ -0,0 +1,602 @@
...
+    private:
+
+    Allocator& al;
+
+    public:
+    std::map<ASR::symbol_t*, std::string> sym_to_new_name;
+    std::map<std::string, ASR::symbol_t*> current_scope;
+
+    UniqueSymbolVisitor(Allocator& al_,
+    std::map<ASR::symbol_t*, std::string> sn) : al(al_), sym_to_new_name(sn){}
```
certik, line 221: a hashtable might be faster here.

src/libasr/pass/unique_symbols.cpp
Smit-create: Done

src/bin/lpython.cpp
Smit-create: Okay! done, thanks!

src/bin/lpython.cpp
```diff
@@ -1616,7 +1635,7 @@ int main(int argc, char *argv[])
         app.require_subcommand(0, 1);
         CLI11_PARSE(app, argc, argv);
 
-        lcompilers_unique_ID = LCompilers::get_unique_ID();
+        lcompilers_unique_ID = separate_compilation ? LCompilers::get_unique_ID(): "";
```
certik, line 1638: I would not do that here, but rather have `lcompilers_unique_ID` always have a unique ID. But have this logic of turning it on and off in asr_scopes. If asr_scopes doesn't have access to `separate_compilation` then let's create another variable `lcompilers_unique_ID_separate_compilation` and have that variable on and off as needed.

certik (approved):
I think that this is fine. Let's merge it and we can iterate on this as needed and fix any bugs as we encounter them.

Issue #2150, opened 2023-07-11
Add support for bitwise left and right shift operators in UnsignedInteger class


emoise-gsi:
In lpython v18.1, the class UnsignedInteger, defined in lpython.py doesn't support bitwise shift operators. Please add something like:

```
    def __lshift__(self, other):
        if isinstance(other, self.__class__):
            return UnsignedInteger(self.bit_width, self.value << other.value)
        else:
            raise TypeError("Unsupported operand type")

    def __rshift__(self, other):
        if isinstance(other, self.__class__):
            return UnsignedInteger(self.bit_width, self.value >> other.value)
        else:
            raise TypeError("Unsupported operand type")
```

emoise-gsi:
Not a blocker

PR #2152, opened 2023-07-11
Support unsigned bitshift operators

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2150

### DIFF
--- a/src/libasr/codegen/asr_to_c_cpp.h
+++ b/src/libasr/codegen/asr_to_c_cpp.h
@@ -2190,6 +2190,8 @@ PyMODINIT_FUNC PyInit_lpython_module_)" + fn_name + R"((void) {
 
     void visit_UnsignedIntegerBinOp(const ASR::UnsignedIntegerBinOp_t &x) {
         handle_BinOp(x);
+        int kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);
+        src = "(uint" + std::to_string(kind * 8) + "_t)(" + src + ")";
     }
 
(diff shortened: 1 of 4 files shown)


src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -2257,7 +2257,8 @@ PyMODINIT_FUNC PyInit_lpython_module_)" + fn_name + R"((void) {
             }
             default: throw CodeGenError("BinOp: " + std::to_string(x.m_op) + " operator not implemented yet");
         }
-        src = "";
+        src = "(" + CUtils::get_c_type_from_ttype_t(x.m_type) + ")";
```
Shaikh-Ubaid, line 2260: For unsigned ints, we are needing to explicitly cast them back to unsigned int. (C seems to promote unsigned ints to signed integers which leads to unexpected errors).

certik: I would only do this when it is needed. In particular, since shift returns signed in C, I would cast it to unsigned, but only for shifts. Not for any other operation, such as Add.

certik: To get started, for unsigned integers just cast the result to unsigned. Later we can refine it more.

Shaikh-Ubaid: Got it. Updated. Thank you for the guidance.

certik (approved):
I think that works. It's nice and clean. We can later further improve the C output, as needed. This should be correct.

Issue #2135, opened 2023-07-09
Equality test of `i8` arrays correctly fails, but with wrong error message


rebcabin:
The following code contains an error on the last line that CPython correctly identifies, but LPython produces a huge traceback and an incorrect error message:

```python
from lpython import (i8, i32, i64, f32, f64,
                     dataclass
                     )
from numpy import (empty,
                   int8,
                   )

r : i8 = i8(240)

@dataclass
class Foo:
    a : i8[4] = empty(4, dtype=int8)
    dim : i32 = 4

def trinary_majority(x : Foo, y : Foo, z : Foo) -> Foo:
    foo : Foo = Foo()

    # Issue 2132
    # assert foo.dim == x.dim == y.dim == z.dim

    assert foo.dim == x.dim
    assert x.dim == y.dim
    assert y.dim == z.dim

    i : i32
    for i in range(foo.dim):
        foo.a[i] = (x.a[i] & y.a[i]) | (y.a[i] & z.a[i]) | (z.a[i] & x.a[i])

    # Issue 2130
    # foo.a = x.a | y.a | z.a
    # foo.a = (x.a & y.a) | (y.a & z.a) | (z.a & x.a)
    return foo


t1 : Foo = Foo()
t1.a = empty(4, dtype=int8)

t2 : Foo = Foo()
t2.a = empty(4, dtype=int8)

t3 : Foo = Foo()
t3.a = empty(4, dtype=int8)

r1 : Foo = trinary_majority(t1, t2, t3)

assert r1.a == t1.a
```

Correct error message:

```bash
(lp) ┌─(~/CLionProjects/lpython/lasr/LP-pycharm)─────────────────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s012)─┐
└─(12:21:21 on brian-lasr ✹ ✭)──> PYTHONPATH='../../src/runtime/lpython' python issue2126.py                                                          3 ↵ ──(Sun,Jul09)─┘
Traceback (most recent call last):
  File "/Users/brian/CLionProjects/lpython/lasr/LP-pycharm/issue2126.py", line 46, in <module>
    assert r1.a == t1.a
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

LPython produces a huge traceback and an incorrect error message:

```bash
...

declare float @_lfortran_ssinh(float)

declare float @_lfortran_stan(float)

declare float @_lfortran_stanh(float)

[...]

PR #2157, opened 2023-07-13
Support array comparison

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2135


tests/reference/asr-arrays_02-da94458.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
certik, line 1: How about this:
```suggestion
semantic error: The truth value of an array with more than one element is ambiguous. Use any() or all()
```

tests/reference/asr-arrays_02-da94458.stderr
```diff
@@ -0,0 +1,5 @@
...
+  --> tests/errors/arrays_02.py:28:8
+   |
+28 | assert r1.a == t1.a
+   |        ^^^^^^^^^^^^
```
certik, line 5: And this:
```
help: use any(r1.a == t1.a) or all(r1.a == t1.a)
```

certik (approved):
I think that looks good.

You can merge this, and in the next PR you can try to improve the error message to be more helpful, I provided some ideas.

tests/reference/asr-arrays_02-da94458.stderr
Shaikh-Ubaid: >help: use any(r1.a == t1.a) or all(r1.a == t1.a)

Here, we do not have access to the string `r1.a == t1.a`. When printing diagnostics the input file is read (from disk) and the error line/string is obtained form the read source input.

Shall we do the following?
```
help: use any() or all()
```
Or shall we keep the message as it is currently in the main branch?

certik: We have access to AST and partial ASR, so we should be able to construct `t1`, `r1` and `a`. And we know it's a struct access and we know it's a comparison. So you use this information to create the text `any(r1.a == t1.a)`.

certik: I made this https://github.com/lcompilers/lpython/issues/2159, so that we don't forget.

PR #2158, opened 2023-07-13
Minor improvements

Shaikh-Ubaid:


examples/expr2.py
```diff
@@ -1,10 +1,8 @@
...
 def main0():
     x: i32
     x = (2+3)*5
     print(x)
 
 main0()
-
-# Not implemented yet in LPython:
-#if __name__ == "__main__":
-#    main()
```
certik, line 10: Why not use this, I think it is supported now?

certik (approved):

certik:
It can probably be called `main()` as well, as we have fixed that bug.

PR #2160, opened 2023-07-15
Refactor: Several Improvements

Shaikh-Ubaid:


czgdp1807 (commented):
For a compiler like ours which is in fast pace development mode, I wouldn’t remove code which never gets triggered by the tests. Instead I would prefer adding tests which trigger that code. Its a time consuming process but that’s better than removing unused code because they might be having some important logic we might have written but missed to add tests for it (< 100% code coverage is very normal).

Shaikh-Ubaid:
>For a compiler like ours which is in fast pace development mode, I wouldn’t remove code which never gets triggered by the tests.

Actually, it is not about the tests. The `current_module` value is always `nulltpr` (it is assigned this value during its initialization and its value is never updated. The code itself is dead code.

Shaikh-Ubaid:
Ready.

certik:
Can you send a PR to LFortran for this change in libasr as well? Let's see if it breaks anything there.

Shaikh-Ubaid:
>Can you send a PR to LFortran for this change in libasr as well? Let's see if it breaks anything there.

Sure, please see https://github.com/lfortran/lfortran/pull/1989.

certik (approved):
I think that this is fine, it works in LFortran.

Issue #2146, opened 2023-07-11
Declaring a u16 function breaks the % operation


certik:
Minimal example:
```python
from lpython import i16, i32

def f():
    i: i32 = 5
    print(i16(i % 1023))

def u16(x: i16) -> i32:
    if x >= i16(0):
        return i32(x)
    else:
        return i32(x) + 65536
```
This gives:
```console
$ lpython a.py 
warning: The module 'lpython_builtin' located in /Users/ondrej/repos/lpython/src/bin/../runtime/lpython_builtin.py cannot be loaded
 --> a.py:5:15
  |
5 |     print(i16(i % 1023))
  |               ^^^^^^^^ imported here

semantic error: Unsupported type annotation: u16
   --> /Users/ondrej/repos/lpython/src/bin/../runtime/lpython_builtin.py:504:26
    |
504 | def _lpython_floordiv(a: u16, b: u16) -> u16:
    |                          ^^^ 
```
The issue is likely here:
```c++
    ASR::ttype_t* get_type_from_var_annotation(std::string var_annotation,
        const Location& loc, Vec<ASR::dimension_t>& dims,
        AST::expr_t** m_args=nullptr, size_t n_args=0,
        bool raise_error=true, ASR::abiType abi=ASR::abiType::Source,
        bool is_argument=false) {
        ASR::ttype_t* type = nullptr;
        ASR::symbol_t *s = current_scope->resolve_symbol(var_annotation);
        if (s) {
            if (ASR::is_a<ASR::Variable_t>(*s)) {
...
```
Where it finds `s` as a "user" function. However this function was declared below and in the main module, not the built-in module, so unless the module is inlined, this shouldn't happen.

Finally, the error message should point to the user defined function and say that it can't be used for annotation.

Shaikh-Ubaid:
Since https://github.com/lcompilers/lpython/pull/2049 we were using the same `Translation_Unit_t`/`global_scope` across modules. The symbols defined in the `main_module` get added to the `Translation_Unit_t`/`global_scope`. Due to this,  other modules are able to access the symbols defined in the `main_module` (as the `global_scope` is shared with them). I think, therefore, the error described in this issue occurs.

I fixed this issue in https://github.com/lcompilers/lpython/pull/2161 by processing/forming the `main_module` as a `Module_t` (and embedding it inside the `Translation_Unit_t`. This way the symbols defined in the `main_module` are not added to the global scope (but instead to the module scope) and thus are not accessible by other modules.

[...]

PR #2161, opened 2023-07-15
Fix scoping issue

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2146

In this PR the `main_module` is also considered/treated as every other `module` and added into the ASR `Translation_Unit_t` as a `Module_t`.


Shaikh-Ubaid:
This PR is a continuation of https://github.com/lcompilers/lpython/pull/2160. (I was hoping to send them together but the PR was being huge.).

I will rebase this PR after merging/completion of https://github.com/lcompilers/lpython/pull/2160.

Shaikh-Ubaid:
Ready.

certik:
Thanks @Shaikh-Ubaid, very useful. @Thirumalai-Shaktivel, @czgdp1807 can you please review this and check if you see any issue with this approach?

This might be a simpler and more robust approach than what we were using so far.

Shaikh-Ubaid:
Just removed the function `move_symbols_from_global_scope()` which is unused now, since the pass that uses it was removed.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -7755,28 +7691,43 @@ Result<ASR::TranslationUnit_t*> python_ast_to_asr(Allocator &al, LocationManager
...
+        std::vector<ASR::asr_t*> tmp_vec;
+        get_calls_to_global_init_and_stmts(al, tu->base.base.loc, program_scope, mod, tmp_vec);
+
+        for (auto i:tmp_vec) {
+            prog_body.push_back(al, ASRUtils::STMT(i));
+        }
+
+        if (prog_body.size() > 0) {
+            prog_dep.push_back(al, s2c(al, mod_name));
+        }
```
Thirumalai-Shaktivel, line 7714: Could you move these lines inside the below if statement? So that, we don't need to fill prog body or dep when the `--disable-main` option is passed.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -7755,28 +7691,43 @@ Result<ASR::TranslationUnit_t*> python_ast_to_asr(Allocator &al, LocationManager
...
         // Note: we can modify this behavior for interactive mode later
-        LCompilers::PassOptions pass_options;
-        pass_options.disable_main = compiler_options.disable_main;
-        if (compiler_options.disable_main) {
-            if (tu->n_items > 0) {
-                diagnostics.add(diag::Diagnostic(
-                    "The script is invoked as the main module and it has code to execute,\n"
-                    "but `--disable-main` was passed so no code was generated for `main`.\n"
-                    "We are removing all global executable code from ASR.",
-                    diag::Level::Warning, diag::Stage::Semantic, {})
```
Thirumalai-Shaktivel, line 7768: I think we need to keep this warning?

src/libasr/pass/global_stmts.cpp
```diff
@@ -105,57 +105,31 @@ void pass_wrap_global_stmts(Allocator &al,
...
+        al, loc,
+        /* a_symtab */ fn_scope,
+        /* a_name */ fn_name,
+        nullptr, 0,
+        /* a_args */ nullptr,
+        /* n_args */ 0,
+        /* a_body */ body.p,
+        /* n_body */ body.size(),
+        /* a_return_var */ (return_var ? return_var_ref : nullptr),
+        (return_var ? ASR::abiType::BindC : ASR::abiType::Source),
```
Thirumalai-Shaktivel, line 122: This is fine, I think. If a tertiary operator doesn't suit our design, we can use some variable to store the values and pass them here.

tests/reference/llvm-expr_01-54467c1.stdout
```diff
@@ -5,16 +5,16 @@ source_filename = "LFortran"
 @1 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
 @2 = private unnamed_addr constant [5 x i8] c"%d%s\00", align 1
 
-define void @__module__global_symbols__lpython_main_program() {
+define void @__module___main_____main____global_statements() {
```
Thirumalai-Shaktivel, line 8: We need to handle this name mangling correctly.

Thirumalai-Shaktivel (approved):
Yup, this seems to be more robust, as we handle everything in the ASR itself.
This looks great! Thanks @Shaikh-Ubaid!

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: In the previous approach, we were removing the global executable code. I think therefore we were printing the warning that global code is being removed. In the current approach, we do not remove any global code (The global code is present inside the functions `global_init()` and `global_stmts()`.). When disable main is enabled, we just do not execute the global code (that is, it is present but not executing). Since, there is no loss or removal of code, I guess we need not print the warning.

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: Sure, I will update it.

tests/reference/llvm-expr_01-54467c1.stdout
Shaikh-Ubaid: Got it. I guess we can work on it in a separate PR.

Shaikh-Ubaid:
Thank you so much for the review @Thirumalai-Shaktivel! I appreciate it.

certik:
@Shaikh-Ubaid can you please merge https://github.com/lcompilers/lpython/pull/2160 and then update this branch against master, so that we can see just the changes? I want to see if we need to test it with LFortran as well.

certik (approved):
I think it's fine. Thanks for fixing this @Shaikh-Ubaid. If there is a problem, we'll fix it later.

PR #2164, opened 2023-07-16
LLVM: Support global unsigned integer variable

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2163

PR #2171, opened 2023-07-17
LLVM: Support Unsigned Int in visit_Variable

### LINKED ISSUE #2170: type casting between u16 and i32 needed
we expected to be able to convert 32768, which is i32, into u16. please help

```python
from lpython import u16, i32

i : i32
u : u16 = u16(32768)
x : i32

for i in range(i32(u)):
    x = i * 2
```

```bash
/Users/brian/CLionProjects/lpython/src/bin/python /Users/brian/CLionProjects/lpython/lasr/LP-pycharm/IssueCH152.py 
code generation error: Variable type not supported 1
 --> /Users/brian/CLionProjects/lpython/lasr/LP-pycharm/IssueCH152.py:4:1
  |
4 | u : u16 = u16(32768)
  | ^^^^^^^^^^^^^^^^^^^^ 


```

Smit-create:
Fixes #2170

### DIFF
--- a/integration_tests/test_unsigned_02.py
+++ b/integration_tests/test_unsigned_02.py
@@ -1,0 +1,10 @@
+from lpython import u16, i32
+
+# test issue 2170
+
+i : i32
+u : u16 = u16(32768)
+x : i32
+
+for i in range(i32(u)):
+    x = i * 2
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -571,6 +571,7 @@ RUN(NAME test_unary_op_04    LABELS cpython llvm c) # unary bitinvert
 RUN(NAME test_unary_op_05    LABELS cpython llvm c) # unsigned unary minus, plus
 RUN(NAME test_unary_op_06    LABELS cpython llvm c) # unsigned unary bitnot
 RUN(NAME test_unsigned_01    LABELS cpython llvm c) # unsigned bitshift left, right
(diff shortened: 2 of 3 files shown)


Shaikh-Ubaid:
Please see https://github.com/lcompilers/lpython/issues/2170#issuecomment-1638450450. I think the `else if` condition added in this PR would never be executed since the first `if` in the `visit_Variable()` would be executed for `Integers` as well as `UnsingedIntegers`. Thus, it seems the `else if` condition is more like a dead code.

https://github.com/lcompilers/lpython/blob/ea9715b2219838989597d8bedb0817d14df1c2cf/src/libasr/codegen/asr_to_llvm.cpp#L2356-L2378

Shaikh-Ubaid:
>This is fine, but after merging let's add tests for all unsigned kinds as well as checking that the value is present in the variable correctly.

Yes, I tested different unsigned kinds in https://github.com/lcompilers/lpython/pull/2164/files#diff-b6a08444f0697e1ea781c919bcdf3b29eba7b542bddef56ce0125f3c2997c973.

integration_tests/test_unsigned_02.py
```diff
@@ -0,0 +1,10 @@
+from lpython import u16, i32
+
+# test issue 2170
+
+i : i32
+u : u16 = u16(32768)
+x : i32
+
+for i in range(i32(u)):
+    x = i * 2
```
Shaikh-Ubaid, line 10: If in case we update the changes in this PR, I suggest we add an assert in this test case (and also possibly a print statement before the assert).

PR #2172, opened 2023-07-17
Fix a bug in cptr_to_u64 in emulation

certik:


certik:
Let's write a test for this as well, and investigate if the other function also needs this.

ronnuriel:
> Let's write a test for this as well, and investigate if the other function also needs this (update: it does, see commit below).

Hey I tested it and its not working,
    return ctypes.addressof(cptr)
TypeError: invalid type

certik:
@ronnuriel, awesome thanks. I submitted it here with you as an author: https://github.com/lcompilers/lpython/pull/2186. It would be nice to figure out an isolated test for this but it's more important to fix it, so if it passes, we'll merge it. Closing in favor of #2186.

PR #2176, opened 2023-07-18
Unsigned ints no wrap

Shaikh-Ubaid:


Shaikh-Ubaid:
This is a Work in progress (WIP). The tests might not pass.

Shaikh-Ubaid:
```
$ cat examples/expr2.py    
from lpython import u16, bitnot_u16, i32

u: u16 = u16(255)
u = bitnot_u16(u)
print(u)
print(~i32(u))
$ lpython examples/expr2.py
65280
255
$ python examples/expr2.py 
65280
-65281
```

It seems `bitnot()` and `~` could not be successively used.

Shaikh-Ubaid:
That [issue](https://github.com/lcompilers/lpython/pull/2176#issuecomment-1639549284) is fixed. The following is a concern:

```bash
$ cat examples/expr2.py 
from lpython import i16, u64, bitnot_u64

w: u64 = u64(123457889998923)
x: u64 = bitnot_u64(w)
print(w, i16(w), ~i16(w))
print(x, i16(x), ~i16(x))
$ python examples/expr2.py 
123457889998923 123457889998923 -123457889998924
18446620615819552692 18446620615819552692 -18446620615819552693
$ lpython examples/expr2.py
123457889998923 -26549 26548
18446620615819552692 26548 -26549
```

It seems `u8`, `u16`, `u32` and `u64` need to be some functions that could convert values appropriately in CPython.

Shaikh-Ubaid:
If the casts in CPython become `no-op`, the output value of CPython is not matching the output of LPython (Example: https://github.com/lcompilers/lpython/pull/2176#issuecomment-1640472819).

Shaikh-Ubaid:
For unsigned unary minus, shall we throw error in LPython similar to unsigned bit not?

certik:
The contract is that if LPython compiles and runs this without errors, then CPython produces the same value. The design of all i8-i64, u8-u64 is such that they are just no-op in CPython and they always produce exactly the same number as in LPython, since LPython only allows such casts that don't change the value. If the value is out of range, we give an error message. So this design should work.

Shaikh-Ubaid:
>only allows such casts that don't change the value. If the value is out of range, we give an error message.

Got it. Thank you so much for the clarification!

certik:
> For unsigned unary minus, shall we throw error in LPython similar to unsigned bit not?

Given that -0 = 0, the unary minus should be allowed for unsigned, but obviously it will only work for 0, for all other values it must give a runtime error.

Shaikh-Ubaid:
The CI might temporarily pass now. As a next step, we need to support 
- unsigned unary minus and uncomment the specific tests (and possibly update if the values are not in appropriate range)
- add error tests for `~` and `-`.

Shaikh-Ubaid:
Further steps could be to support compile time overflow checking and later runtime error checking for which we have a dedicated issue here https://github.com/lcompilers/lpython/issues/2183.

src/runtime/lpython/lpython.py
```diff
@@ -859,3 +707,11 @@ def __call__(self, *args, **kwargs):
...
             self.fn_name)
         return function(*args, **kwargs)
+
+def bitnot(x, bitsize):
+    return (~x) % (2 ** bitsize)
+
+bitnot_u8 = lambda x: bitnot(x, 8)
+bitnot_u16 = lambda x: bitnot(x, 16)
+bitnot_u32 = lambda x: bitnot(x, 32)
+bitnot_u64 = lambda x: bitnot(x, 64)
```
certik, line 717: Why not just calling it bitnot? And make the function work for all u8 - u64.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -7471,6 +7473,26 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                 ASR::ttype_t *type = ASRUtils::TYPE(ASR::make_Pointer_t(al, x.base.base.loc, type_));
                 tmp = ASR::make_GetPointer_t(al, x.base.base.loc, args[0].m_value, type, nullptr);
                 return ;
+            } else if( call_name.substr(0, 6) == "bitnot" ) {
+                parse_args(x, args);
+                if (args.size() < 1 or args.size() > 1) {
```
certik, line 7478: You cannot use `or` in C++, it fails with MSVC:
```suggestion
                if (args.size() < 1 || args.size() > 1) {
```

certik: Also why not just:
```suggestion
                if (args.size() != 1) {
```

certik:
Otherwise it looks good. After you fix the above, please do not merge it, I have to manually check and update our internal codes if needed before we can merge it.

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: Got it. Thank you for the guidance. I will update it.

src/runtime/lpython/lpython.py
certik: You are right, I didn't realize that. CPython puts quite a lot of limitations in what we can do, but I think it is worthy to just use the built-in `int` and `float` and build everything on top.

Let's use what you did then.

certik (approved):
I think this change is good. I'll mark it as draft, so that we don't accidentally merge it. I need to see if anything broke first.

certik:
Everything works for us, so I am merging this PR.

Thanks @Shaikh-Ubaid !

PR #2178, opened 2023-07-18
Fix unsigned to signed int cast

### LINKED ISSUE #2174: Bug in u16 to u32 conversion
```python
from lpython import u16, i32

u : u16 = u16(32768)

print(u, i32(u))
```
This prints:
```console
$ lpython a.py 
32768 -32768
```
But I think it should print:
```
32768 32768
```

Smit-create:
Fixes #2174


certik (approved):
Looks great, thanks!

Issue #2173, opened 2023-07-17
In emulation mode, u8-u64 should be represented by just int


certik:
Currently we created our own class UnsignedInteger and we provide a CPython incompatible behavior of some operators such as ~. Rather, if we need a CPython incompatible behavior, we should introduce some special functions, such as `from lpython import uneg`. The intrinsic operators should work just like in CPython.

Here is an example:
```python
print(u16(0xff), 0xff)
print(~u16(0xff), ~0xff)
```
This currently prints:
```
255 255
65280 -256
```
So the `~` is behaving differently for u16 and Python ints. It should not. The design of u16 and i16 is that they are strict value restricted subset of Python ints and should behave exactly identical on this subset, and LPython should give compile time or runtime error messages if any operation is performed outside of this subset. 

Since in CPython ~255 becomes -256, then the same behavior must happen on ~u16(255). In order to provide the other operation, we can make `uneg(255)` be 65280. Then the operation can be correctly implemented in emulation mode, and all unsigned integers can be just integers, which will speedup operation and remove many bugs that we currently have, when we are passing UnsignedInteger somewhere where a regular integer is expected.

I think the current behavior was introduced by trying to emulate C, but Python is not C. We need to be compatible with Python, not C. If C behavior is desired, it must be done using functions like `uneg`, not using intrinsic integer operators (which should rather behave like CPython does).

certik:
The "-" is two's complement, so
```
-(1) = -1
-(2) = -2
-(-1) = 1
-(-2) = 2
```
The "~" is one's complement, so:
```
~(1) = -2
~(2) = -3
~(-2) = 1
~(-3) = 2
```
And in Python it also does:
```
~(0) = -1
~(-1) = 0
```

The relation between "-" and "~" is:
```
~x = -x - 1
-x = ~x + 1
```

certik:
As a consequence, since `~x = -x-1`, this operation is by definition not supported for unsigned integers, because `~` applied to a non-negative number is negative, outside of range of unsigned. So to fix this:

* [ ] make LPython give a nice error message for `~` applied to unsigned, saying the result is out of range and providing suggestions how to obtain this operation (by properly casting to signed first, and then applying `~`)
* [ ] investigate if we need `uneg` intrinsic function, or if the workaround via casting to signed is good enough.

certik:
Example: this
```
u: u16 = u16(255)
u = ~u
```
will not compile. Must be rewritten to:
```
u: u16 = u16(255)
[...]

PR #2189, opened 2023-07-19
Unsigned integers no wrap II

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2173

towards https://github.com/lcompilers/lpython/pull/2176#issuecomment-1641364333

### DIFF
--- a/tests/reference/asr-unsigned_02-6563e58.stderr
+++ b/tests/reference/asr-unsigned_02-6563e58.stderr
@@ -1,0 +1,5 @@
+semantic error: The result of the unary minus `-` operation is negative, thus out of range for u16
+ --> tests/errors/unsigned_02.py:3:10
+  |
+3 | i: u16 = -u16(5)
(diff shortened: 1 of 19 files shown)


tests/reference/asr-unsigned_02-6563e58.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: The result of the unary minus `-` operation is negative, thus out of range for u16
+ --> tests/errors/unsigned_02.py:3:10
+  |
+3 | i: u16 = -u16(5)
+  |          ^^^^^^^ use -i32(u) for signed result
```
certik, line 5: It would be great if it could copy the actual expression inside:
```suggestion
  |          ^^^^^^^ use -i32(5) for signed result
```
But I guess for that we would have to have an ASR -> Python printer, which we plan to, but right now. So we can do that later.

certik (approved):
Looks great!

PR #2190, opened 2023-07-20
Added test based on basic_new_heap function

anutosh491:


anutosh491:
TODO for subsequent PRs:
* [x] Add a line under print to also test using assert: `assert basic_str(x) == "pi"`
* [ ] enable LLVM backend (we need to add `llvm_sym` into CMake and add this label)
* [ ] add a test for `basic_new_stack` (this should be a new test file)

anutosh491:
I've added the `assert` check in this pr itself as it was simple 1 line addition . I'll address the remaining TODO's through subsequent prs.

anutosh491:
I've added the `assert` check in this pr itself as it was simple 1 line addition . I'll address the remaining TODO's through subsequent prs.

anutosh491:
cc @certik All tests seems to pass . I think this could be merged .

certik (approved):
Yes, this is good. Thanks!

anutosh491:
Thanks . I'll create the new pull requests now!

PR #2193, opened 2023-07-20
C: Fix complex types

### LINKED ISSUE #2187: C generates _lfortran_caimag(float complex x) and it fails to compile
```
float _lfortran_caimag(float complex x);
```
It trips on the "float complex". It should be 
```
LFORTRAN_API float _lfortran_caimag(float_complex_t x);
```
As defined in `src/libasr/runtime/lfortran_intrinsics.h`.

Smit-create:
Fixes #2187


czgdp1807 (approved):

certik:
Thanks! I think this is the right fix.

PR #2196, opened 2023-07-20
Support ccall() for symengine and other libs

Shaikh-Ubaid:


Shaikh-Ubaid:
@anutosh491 It seems to work, but there is a catch.
```python
$ cat integration_tests/symbolics_07.py                    
from lpython import ccall, CPtr
import os

@ccall(header="symengine/cwrapper.h", py_mod="symengine", py_mod_path=f"{os.environ['CONDA_PREFIX']}/lib")
def basic_new_heap() -> CPtr:
    pass

@ccall(header="symengine/cwrapper.h", py_mod="symengine", py_mod_path=f"{os.environ['CONDA_PREFIX']}/lib")
def basic_const_pi(x: CPtr) -> None:
    pass

@ccall(header="symengine/cwrapper.h", py_mod="symengine", py_mod_path=f"{os.environ['CONDA_PREFIX']}/lib")
def basic_str(x: CPtr) -> str:
    pass

def main0():
    x: CPtr = basic_new_heap()
    basic_const_pi(x)
    s: str = basic_str(x)
    print(s)
    # assert s == "pi"

main0()
```
```bash
$ python integration_tests/symbolics_07.py                    
b'pi'
```
```bash
$ lpython integration_tests/symbolics_07.py --enable-symengine --backend c
pi 
```

The outputs by `lpython` and `cpython` do not match. Therefore the assert in the `integration_tests/symbolics_07.py` fails (I temporarily commented it out).

anutosh491:
I went through the example above and I think we get the b'pi' output through `CPython` while using `ccall` because `ctypes` library returns `strings` as `bytes` objects when calling C functions that return strings

integration_tests/symbolics_07.py
```diff
@@ -1,14 +1,15 @@
-from lpython import ccall
+from lpython import ccall, CPtr
+import os
 
-@ccall(header="symengine/cwrapper.h")
+@ccall(header="symengine/cwrapper.h", py_mod="symengine", py_mod_path=f"{os.environ['CONDA_PREFIX']}/lib")
```
certik, line 4: Is this the runtime library to use?

certik: If so, I would rename `py_mod` to `c_shared_library`

Shaikh-Ubaid: Got it. Thank you. I updated it.

anutosh491:
> Good catch @anutosh491! Thank you. I fixed it in [lpython.py: Decode ctypes strings before returning](https://github.com/lcompilers/lpython/pull/2196/commits/f6eab9cf62aad18cdb25f75d7014e78ed429fa1e).

Yes , this seems to be the correct way to address this . Thank you

certik (approved):
I think that this looks great.

Thank you.

This can also be used in the LLVM and C backends when linking. Right now we ignore it and one has to link by hand, which is fine.

PR #2201, opened 2023-07-22
Use optimisation flags for C compiler in lpython decorator

czgdp1807:
I think we should use all the optimisations flags by default in `lpython` decorator and generate as fast code as possible. It produces a significant difference when comparing with other JIT compilers.

### DIFF
--- a/integration_tests/lpython_decorator_01.py
+++ b/integration_tests/lpython_decorator_01.py
@@ -1,7 +1,7 @@
 from numpy import array
 from lpython import i32, f64, lpython
 
-@lpython
+@lpython("c", ["-ffast-math", "-funroll-loops", "-O3"])
 def fast_sum(n: i32, x: f64[:]) -> f64:
     s: f64 = 0.0
     i: i32
--- a/integration_tests/lpython_decorator_01.py
+++ b/integration_tests/lpython_decorator_01.py
@@ -1,7 +1,7 @@
 from numpy import array
 from lpython import i32, f64, lpython
 
-@lpython(optimisation_flags=["-ffast-math", "-funroll-loops", "-O3"])
+@lpython(backend="c", backend_optimisation_flags=["-ffast-math", "-funroll-loops", "-O3"])
 def fast_sum(n: i32, x: f64[:]) -> f64:
     s: f64 = 0.0
     i: i32
--- a/integration_tests/lpython_decorator_01.py
+++ b/integration_tests/lpython_decorator_01.py
@@ -1,7 +1,7 @@
 from numpy import array
 from lpython import i32, f64, lpython
 
-@lpython("c", ["-ffast-math", "-funroll-loops", "-O3"])
+@lpython(optimisation_flags=["-ffast-math", "-funroll-loops", "-O3"])
 def fast_sum(n: i32, x: f64[:]) -> f64:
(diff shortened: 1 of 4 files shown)


src/runtime/lpython/lpython.py
```diff
@@ -687,9 +687,9 @@ def get_rtlib_dir():
```
Thirumalai-Shaktivel, line 687: I'm not sure if the --fast` option works in the C backend. If it does, then we can use it in line 684?

czgdp1807: It doesn't I think.

czgdp1807:
`-funroll-loops` and `-ffast-math` are useful optimisations because most of the computations which lpython intends to handle involve loops and math, so enabling both of these will speed up the code. In fact, enabling `-O1`, `-O2` and `-O3` speeds up the code significantly when the `lpython` decorated function doesn’t return arrays. As a user I would like to pass these flags via `lpython` decorator but since its a class based decorator , its not so easy to pass these flags.

src/runtime/lpython/lpython.py
```diff
@@ -647,30 +648,28 @@ def ccallable(f):
...
 
-class Lpython:
-    """
-    The @lpython decorator compiles a given function using LPython.
+class LpythonJITCache:
 
-    The decorator should be used from CPython mode, i.e., when the module is
-    being run using CPython. When possible, it is recommended to use LPython
-    for the main program, and use the @cpython decorator from the LPython mode
-    to access CPython features that are not supported by LPython.
```
certik, line 657: This documentation should be somewhere.

integration_tests/lpython_decorator_01.py
```diff
@@ -1,7 +1,7 @@
 from numpy import array
 from lpython import i32, f64, lpython
 
-@lpython
+@lpython(optimisation_flags=["-ffast-math", "-funroll-loops", "-O3"])
```
certik, line 4: LPython does not support any of these flags.

Consequently, these are flags to be used with the C compiler that compiles the C code generated by our C backend.

Once we support LLVM, the flags we will support are different (we currently only support `--fast`).

So let's figure out how to do this. One option is:
```
@lpython(backend="c", backend_optimisation_flags=["-ffast-math", "-funroll-loops", "-O3"])
@lpython(backend="llvm", backend_optimisation_flags=["--fast"])
```
That might work.

certik (approved):
Awesome, this looks great! I left a comment.

integration_tests/lpython_decorator_01.py
czgdp1807: Makes sense. Right now the default backend is `c`, so `backend_optimisation_flags=["-ffast-math", "-funroll-loops", "-O3"]` makes sense in this example. Or do you want to make `backend` a compulsory argument (right now its optional and defaults to `c`?

certik: Let's make it compulsory if you use `backend_optimisation_flags`, otherwise things can break in the future once we make LLVM the default. Otherwise let's allow `@lpython` as is, and choose a default backend, currently C, but I think our default should actually be LLVM later on.

certik (approved):
Looks good to me!

PR #2205, opened 2023-07-24
Add ``get_builder0`` macro to declare auxiliary variables at entry point of functions

czgdp1807:
cc: @kabra1110


czgdp1807:
@certik I think we can merge this. Do you have any reviews that must be addressed before merging this?

certik:
Looks  good. The only issue is with using macros in header files. It will sooner or later cause a clash. One must use unique names, like LCOMPILERS_BUILDER_0 or something like that. That can be fixed later.

Issue #2162, opened 2023-07-15
CI: The Job LPython CI / Check Release build (pull_request) seems flaky


Shaikh-Ubaid:
It seems to falsely fail for me on first run. It seems to pass on second run attempt.

False Failures:
- https://github.com/lcompilers/lpython/actions/runs/5564532214/attempts/1
- https://github.com/lcompilers/lpython/actions/runs/5546663093/attempts/1
- https://github.com/lcompilers/lpython/actions/runs/5563369118/attempts/1
- https://github.com/lcompilers/lpython/actions/runs/5546663093/attempts/1

Smit-create:
Can you describe the exact issue? or point to failure? Rarely, it fails because of some issues on the GH actions side.

Shaikh-Ubaid:
>Can you describe the exact issue? or point to failure? Rarely, it fails because of some issues on the GH actions side.

I updated the description of this issue.

Shaikh-Ubaid:
The job [Check Release build](https://github.com/lcompilers/lpython/actions/runs/5608325831/jobs/10261274420#logs) passed in the third run attempt. Something does not seem right. I will explore more about this.

certik:
The failure seems different though:
```
make[2]: *** [CMakeFiles/modules_02_FAST.dir/build.make:74: modules_02_FAST.o] Illegal instruction (core dumped)
make[1]: *** [CMakeFiles/Makefile2:2838: CMakeFiles/modules_02_FAST.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```
and
```
Test project /home/runner/work/lpython/lpython
    Start 1: test_stacktrace
1/2 Test #1: test_stacktrace ..................   Passed    0.00 sec
    Start 2: test_lpython
2/2 Test #2: test_lpython .....................***Exception: Illegal  0.13 sec
```
and (3x):
```
make[2]: *** [CMakeFiles/array_03_decl_FAST.dir/build.make:74: array_03_decl_FAST.o] Illegal instruction (core dumped)
make[1]: *** [CMakeFiles/Makefile2:568: CMakeFiles/array_03_decl_FAST.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

Shaikh-Ubaid:
I think the original failure is 
```
2/2 Test #2: test_lpython .....................***Exception: Illegal  0.13 sec
```
[...]

PR #2212, opened 2023-07-25
CI: Provide release flags, exclude -march=native

Shaikh-Ubaid:
>I think we just need to lower the optimization options, perhaps keep -O3, but remove -march=native or something like that.

towards https://github.com/lcompilers/lpython/issues/2162#issuecomment-1644965229

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -335,6 +335,7 @@ jobs:
             export CXXFLAGS="-Werror"
             cmake . -GNinja \
               -DCMAKE_BUILD_TYPE=Release \
+              -DCMAKE_CXX_FLAGS_RELEASE="-Wall -Wextra -O3 -funroll-loops -DNDEBUG" \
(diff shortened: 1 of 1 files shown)


Shaikh-Ubaid:
I am unsure if the changes in this PR are the perfect fix. This seems to be more of an experimental PR. I think we can later try other steps/approaches if the failure in Release build is still experienced post merging of this PR.

czgdp1807 (commented):
So `-O3` optimisations weren't applied in the release build before this PR?

Shaikh-Ubaid:
>So -O3 optimisations weren't applied in the release build before this PR?

I think they were being applied. 

https://github.com/lcompilers/lpython/blob/634177d3ff5be102a71d64fe2d80670c85f93661/cmake/UserOverride.cmake#L12

In this PR, I just excluded the `-march=native` flag.

certik:
Yes, let's try this and see.

Issue #2221, opened 2023-07-29
Code Generation Errors for C / C++ source code


jusevitch:
The following code works when compiling to LLVM, but throws errors when I attempt to show the C and C++ source code.

I'm using LPython v0.19.0 on Ubuntu Linux 20.04.

Python code:

```python
from lpython import i32, f32

def add_list(x: list[f32]) -> f32:
    sum: f32 = f32(0.0)
    i: i32

    for i in range(len(x)):
        sum = sum + f32(x[i])
    return sum

def create_list(n: i32) -> list[f32]:
    x: list[f32]
    i: i32

    x = [f32(0.0)] * n
    for i in range(n):
        x[i] = f32(i)
    return x

def main0():
    x: list[f32] = create_list(i32(10))
    # x = [f32(1.0), f32(2.0), f32(3.0)]
    print(add_list(x))

if __name__ == "__main__":
    main0()
```

Output of `./src/bin/lpython --show-c examples/jamestest.py`:

```
BFD: DWARF error: section .debug_info is larger than its filesize! (0x93ef57 vs 0x530ea0)
Traceback (most recent call last):
  Binary file "/home/james/code/lpython_dir/lpython/src/bin/lpython", in _start()
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", in __libc_start_main()
  File "/home/james/code/lpython_dir/lpython/src/bin/lpython.cpp", line 1804, in ??
    return emit_c(arg_file, runtime_library_dir, lpython_pass_manager,
  File "/home/james/code/lpython_dir/lpython/src/bin/lpython.cpp", line 361, in ??
    auto res = LCompilers::asr_to_c(al, *asr, diagnostics, compiler_options, 0);
  File "/home/james/code/lpython_dir/lpython/src/libasr/codegen/asr_to_c.cpp", line 1418, in LCompilers::asr_to_c[abi:cxx11](Allocator&, LCompilers::ASR::TranslationUnit_t&, LCompilers::diag::Diagnostics&, LCompilers::CompilerOptions&, long)
    v.visit_asr((ASR::asr_t &)asr);
  File "/home/james/code/lpython_dir/lpython/src/libasr/../libasr/asr.h", line 4979, in LCompilers::ASR::BaseVisitor<LCompilers::ASRToCVisitor>::visit_asr(LCompilers::ASR::asr_t const&)
    void visit_asr(const asr_t &b) { visit_asr_t(b, self()); }
[...]

PR #2224, opened 2023-07-30
C: Fix ListConcat and ListSection

Shaikh-Ubaid:
fixes #2223

PR #2225, opened 2023-07-30
C: Support ListRepeat

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2221

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -504,7 +504,8 @@ RUN(NAME test_list_section2  LABELS cpython llvm c NOFAST)
 RUN(NAME test_list_count     LABELS cpython llvm)
 RUN(NAME test_list_index     LABELS cpython llvm)
 RUN(NAME test_list_index2    LABELS cpython llvm)
-RUN(NAME test_list_repeat    LABELS cpython llvm NOFAST)
+RUN(NAME test_list_repeat    LABELS cpython llvm c NOFAST)
+RUN(NAME test_list_repeat2   LABELS cpython llvm c NOFAST)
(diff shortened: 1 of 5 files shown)


Shaikh-Ubaid:
This PR needs to be rebased after merging/completion of https://github.com/lcompilers/lpython/pull/2224.

Shaikh-Ubaid:
Please note that the support for `ListRepeat` is only added for the `C` backend.

certik (approved):
I think this is fine.

Is the malloc part in C ever freed?

Shaikh-Ubaid:
>Is the malloc part in C ever freed?

I think it is not freed ever.

```bash
$ git grep "_lfortran_free"
src/libasr/codegen/asr_to_llvm.cpp:    inline void call_lfortran_free(llvm::Function* fn, llvm::Type* llvm_data_type) {
src/libasr/codegen/asr_to_llvm.cpp:        std::string func_name = "_lfortran_free";
src/libasr/codegen/asr_to_llvm.cpp:    inline void call_lfortran_free_string(llvm::Function* fn) {
src/libasr/codegen/asr_to_llvm.cpp:                    call_lfortran_free(free_fn, llvm_data_type);
src/libasr/codegen/llvm_utils.cpp:            std::string func_name = "_lfortran_free";
src/libasr/runtime/lfortran_intrinsics.c:LFORTRAN_API void _lfortran_free(char* ptr) {
src/libasr/runtime/lfortran_intrinsics.h:LFORTRAN_API void _lfortran_free(char* ptr);
```

We should fix it soon. I think it leads to memory leak.

certik:
Yes, it's time to fix all these.

PR #2226, opened 2023-07-30
Add a test for lpython emulation mode

### LINKED ISSUE #2222: Bug in LPython 0.19.0 emulation mode: 'function' object has no attribute 'restype'
It works with LPython 0.18.4, but breaks with LPython 0.19.0:
```
  File "/root/mambaforge/envs/ch_emulation/share/lpython/lib/lpython/lpython.py", line 398, in __call__
    if self.cf.restype == ctypes.c_char_p:
AttributeError: 'function' object has no attribute 'restype'
```
I still need to create a minimal example.

certik:
The new test fails due to #2222. We first verify that it fails at the CI and then we fix it with a commit.

Fixes #2222.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -763,3 +763,6 @@ RUN(NAME lpython_decorator_01    LABELS cpython)
 RUN(NAME lpython_decorator_02    LABELS cpython)
 
 COMPILE(NAME import_order_01     LABELS cpython llvm c) # any
+
+# LPython emulation mode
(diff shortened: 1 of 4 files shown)


certik:
It's the
```
@ccall
def f2(a: i64):
    pass
```
function that doesn't return anything and its pure Python implementation that caused the failure in #2222. We will fix it in the next commit.

certik:
Things correctly fail, so the test is a good one. Now we pushed in a commit that fixes it.

PR #2213, opened 2023-07-27
ASR: Support chained CompareOp

Smit-create:
Fixes #2131

PR #2229, opened 2023-07-30
Revert "ASR: Support chained CompareOp"

### LINKED ISSUE #2227: Possible regression: x > y == z should not be allowed
This used to correctly fail:
```
 --> /Users/ondrej/repos/lpython/integration_tests/expr_22.py:8:12
  |
8 |     assert x > y == z
  |            ^^^^^    ^ type mismatch ('bool' and 'i32')
```
But now it works. It seems this could should keep returning the same error, so we should add a test for this error message.

Smit-create:
Reverts lcompilers/lpython#2213

Fixes #2227


certik:
@Smit-create is this ready?

certik:
I am going to merge this, so that we can release a new version of LPython soon. It is important to ensure users don't start depending on this behavior, since it was incorrect and we would have to break it.

Smit-create:
Yes, thanks for merging. It was ready

PR #2235, opened 2023-07-31
Correction in `list.pop`

kabra1110:
Corrects implementation of `list.pop`. The `while` loop condition was incorrect.


czgdp1807 (changes requested):
Please add a test for which the `main` branch fails and your branch passes?

czgdp1807 (approved):

PR #2252, opened 2023-08-04
ASR: Fix Import as

### LINKED ISSUE #2153: `import` `as` does not work
```console
% cat a.py 
from lpython import i32

def f(x: i32) -> i32:
    return 2 *x
```

```console
% cat b.py
from lpython import f64
from a import f as fa

def main0():
    print(fa(3))

main0()
```
Error:
```bash
% lpython b.py -I .
semantic error: Function 'fa' is not declared and not intrinsic
 --> b.py:5:11
  |
5 |     print(fa(3))
  |           ^^^^^ 
```

CPython:
```
% PYTHONPATH=$PWD/src/runtime/lpython python b.py 
6
```

Smit-create:
Fixes #2153


certik (approved):
I think that's fine.

PR #2263, opened 2023-08-10
Refactoring the symbolic ASR pass

anutosh491:


anutosh491:
I've added a first level of refactoring here where I've brought down the repetetive code involved in all binary and unary operators/functions. cc @Thirumalai-Shaktivel 

The next step I see is unifying the switch case being used in all functions like `visit_Assignment` , `visit_Print` and ` visit_IntrinsicFunction`. We could just have a single function with the switch case I think

src/libasr/pass/replace_symbolic.cpp
```diff
@@ -294,6 +291,51 @@ class ReplaceSymbolicVisitor : public PassUtils::PassVisitor<ReplaceSymbolicVisi
...
+
+            if (ASR::is_a<ASR::IntrinsicFunction_t>(*value1) || ASR::is_a<ASR::Cast_t>(*value1)) {
+                if (ASR::is_a<ASR::IntrinsicFunction_t>(*value1)) {
+                    this->visit_IntrinsicFunction(*ASR::down_cast<ASR::IntrinsicFunction_t>(value1));
+                } else {
+                    this->visit_Cast(*ASR::down_cast<ASR::Cast_t>(value1));
+                }
+                ASR::symbol_t* var_sym1 = current_scope->get_symbol(symengine_stack.pop());
+                value1 = ASRUtils::EXPR(ASR::make_Var_t(al, loc, var_sym1));
+            }
```
Thirumalai-Shaktivel, line 307:
```suggestion
            if (ASR::is_a<ASR::IntrinsicFunction_t>(*value1)) {
                this->visit_IntrinsicFunction(*ASR::down_cast<ASR::IntrinsicFunction_t>(value1));
            } else if (ASR::is_a<ASR::Cast_t>(*value1)) {
                this->visit_Cast(*ASR::down_cast<ASR::Cast_t>(value1));
            } else {
                LCOMPILERS_ASSERT(false);
            }
            ASR::symbol_t* var_sym1 = current_scope->get_symbol(symengine_stack.pop());
            value1 = ASRUtils::EXPR(ASR::make_Var_t(al, loc, var_sym1));
```
Make sense?

anutosh491: wait so the logic here goes as follows , `value1` can possibly not be an `IntrinsicFunction` or a `Cast` and in those cases we don't need to edit `value1`. Only if `value1` is either one of those , then we need to edit value1 as done here . 
Hence through my code I just checked if it is one of them ... if yes I edit `value1` through this.
```
            ASR::symbol_t* var_sym1 = current_scope->get_symbol(symengine_stack.pop());
            value1 = ASRUtils::EXPR(ASR::make_Var_t(al, loc, var_sym1));
```
Are these the same ?

Thirumalai-Shaktivel: We have encountered only three cases (`Var_t`, `IntrinsicFunction_t` and `Cast_t`), right?
Any other cases?

Thirumalai-Shaktivel: Also, I think we should only allow CPtr type; for the rest, we have to throw an error or handle it by converting to CPtr type.

src/libasr/pass/replace_symbolic.cpp
```diff
@@ -294,6 +291,51 @@ class ReplaceSymbolicVisitor : public PassUtils::PassVisitor<ReplaceSymbolicVisi
...
+        const std::string& new_name, ASR::expr_t* target) {
+            ASR::expr_t* value1 = x->m_args[0];
+            ASR::expr_t* value2 = x->m_args[1];
+
+            if (ASR::is_a<ASR::IntrinsicFunction_t>(*value1) || ASR::is_a<ASR::Cast_t>(*value1)) {
+                if (ASR::is_a<ASR::IntrinsicFunction_t>(*value1)) {
+                    this->visit_IntrinsicFunction(*ASR::down_cast<ASR::IntrinsicFunction_t>(value1));
+                } else {
+                    this->visit_Cast(*ASR::down_cast<ASR::Cast_t>(value1));
+                }
```
Thirumalai-Shaktivel, line 304: This lines seems to be repeated as well, so we will create a new function say: `handle_arguments ` or something similar to handle it.
```c
ASR::expr_t* value1 = handle_arguments(x->m_args[0]);
```

Thirumalai-Shaktivel: Inside the function just check for `CPtr` type (for other `type` throw Exception), `Var_t`, `IntrinsicFunction_t` and `Cast_t` ( for other `expr_t` throw Exception).

anutosh491:
I've refactored the switch case through my latest commit . I guess I could address your reviews now and then the macros based refactoring could be done through the next PR for convinience.

src/libasr/pass/replace_symbolic.cpp
anutosh491: > Any other cases?

No , no other case .

anutosh491: I am a bit confused about your suggestion here , we only need to create `var_sym` and edit `value` if it is either of these 2 cases `IntrinsicFunction_t` and `Cast_t`. So I am not sure if something like this makes sense 
```
    void process_binary_operator(Allocator &al,  const Location &loc, ASR::IntrinsicFunction_t* x, SymbolTable* module_scope,
        const std::string& new_name, ASR::expr_t* target) {
            ASR::expr_t* value1 = x->m_args[0];
            ASR::expr_t* value2 = x->m_args[1];

            if (ASR::is_a<ASR::IntrinsicFunction_t>(*value1)) {
                this->visit_IntrinsicFunction(*ASR::down_cast<ASR::IntrinsicFunction_t>(value1));
            } else if (ASR::is_a<ASR::Cast_t>(*value1)) {
                this->visit_Cast(*ASR::down_cast<ASR::Cast_t>(value1));
            } else {
                LCOMPILERS_ASSERT(false);
            }
            ASR::symbol_t* var_sym1 = current_scope->get_symbol(symengine_stack.pop());
            value1 = ASRUtils::EXPR(ASR::make_Var_t(al, loc, var_sym1));

            if (ASR::is_a<ASR::IntrinsicFunction_t>(*value2)) {
                this->visit_IntrinsicFunction(*ASR::down_cast<ASR::IntrinsicFunction_t>(value2));
            } else if (ASR::is_a<ASR::Cast_t>(*value2)) {
                this->visit_Cast(*ASR::down_cast<ASR::Cast_t>(value2));
            } else {
                LCOMPILERS_ASSERT(false);
            }
            ASR::symbol_t* var_sym2 = current_scope->get_symbol(symengine_stack.pop());
            value2 = ASRUtils::EXPR(ASR::make_Var_t(al, loc, var_sym2));

        perform_symbolic_binary_operation(al, loc, module_scope, new_name, target, value1, value2);
    }
```

anutosh491: > Also, I think we should only allow CPtr type; for the rest, we have to throw an error or handle it by converting to CPtr type.

Also a bit confused about this . Could you show how this would help !

Thirumalai-Shaktivel: The arguments of IntrinisicFunction can be of any type, so we have to make sure we only pass `CPtr` to the `BindC` Function. If this is already handled while creating IntrinisicFunction, then no need to check for the type, I think.

Thirumalai-Shaktivel: Please ignore this review, I realised this: https://github.com/lcompilers/lpython/pull/2263#discussion_r1289655215 and provided a new suggestion: https://github.com/lcompilers/lpython/pull/2263#discussion_r1289625227

anutosh491: Actually if you look at the `visit_IntrinsicFunction` now, I think we are handling these things 
```
    void visit_IntrinsicFunction(const ASR::IntrinsicFunction_t &x) {
        if(x.m_type && x.m_type->type == ASR::ttypeType::SymbolicExpression) {
            SymbolTable* module_scope = current_scope->parent;

            ASR::ttype_t *type = ASRUtils::TYPE(ASR::make_SymbolicExpression_t(al, x.base.base.loc));
            std::string symengine_var = symengine_stack.push();
            ASR::symbol_t *arg = ASR::down_cast<ASR::symbol_t>(ASR::make_Variable_t(
                al, x.base.base.loc, current_scope, s2c(al, symengine_var), nullptr, 0, ASR::intentType::Local,
                nullptr, nullptr, ASR::storage_typeType::Default, type, nullptr,
                ASR::abiType::BindC, ASR::Public, ASR::presenceType::Required, false));
            current_scope->add_symbol(s2c(al, symengine_var), arg);
            for (auto &item : current_scope->get_scope()) {
                if (ASR::is_a<ASR::Variable_t>(*item.second)) {
                    ASR::Variable_t *s = ASR::down_cast<ASR::Variable_t>(item.second);
                    this->visit_Variable(*s);
                }
            }

            ASR::IntrinsicFunction_t &xx = const_cast<ASR::IntrinsicFunction_t&>(x);
            ASR::expr_t* target = ASRUtils::EXPR(ASR::make_Var_t(al, x.base.base.loc, arg));
            process_intrinsic_function(al, x.base.base.loc, &xx, module_scope, target);
        }
    }
```

What happens here is firstly we only want to edit those intrinsic function nodes which are of type `SymbolicExpression` . And also here we create a symbolic variables (`stack0` or something from the `symengine_stack`) then run it through `visit_Variable` so that 
1) The 4 statements which are the following are defined for the variable.
```
    _stack0: i64 = i64(0)
    stack0: CPtr = empty_c_void_p()
    p_c_pointer(pointer(_stack0, i64), stack0)
    basic_new_stack(stack0)
```

Thirumalai-Shaktivel: Yup, Let's check the type of `stack0` to be `CPtr`.

anutosh491: So I think how this goes is 
1) We only edit an intrinsic function of type SymbolicExpression
2) And while going from AST to ASR there are well defined rules as to what the arguments types can be. Like for binary operators
```
    static inline ASR::asr_t* create_##X(Allocator& al, const Location& loc,               \
            Vec<ASR::expr_t*>& args,                                                       \
            const std::function<void (const std::string &, const Location &)> err) {       \
        if (args.size() != 2) {                                                            \
            err("Intrinsic function `"#X"` accepts exactly 2 arguments", loc);             \
        }                                                                                  \
                                                                                           \
        for (size_t i = 0; i < args.size(); i++) {                                         \
            ASR::ttype_t* argtype = ASRUtils::expr_type(args[i]);                          \
            if(!ASR::is_a<ASR::SymbolicExpression_t>(*argtype)) {                          \
                err("Arguments of `"#X"` function must be of type SymbolicExpression",     \
                args[i]->base.loc);                                                        \
            }                                                                              \
        }  
```
Which means the arguments inside a symbolic type intrinsic function would also be symbolic and visit_IntrinsicFunction should be able to handle them.

anutosh491: Sure now that we have a proper framework in `intrinsic_registry.h` using the `create`, `eval` and `instantiate` functions , wherein we actually handle type check through create and also eval through some cases , I haven't disturbed that framework as of now.

anutosh491:
I think this is the almost all refactoring that could be achieved through functions . The functioncalls for `basic_new_stack` etc could be handled through macros like those used in `intrinsic_registry.h`

src/libasr/pass/replace_symbolic.cpp
```diff
@@ -294,405 +291,196 @@ class ReplaceSymbolicVisitor : public PassUtils::PassVisitor<ReplaceSymbolicVisi
...
+        if (ASR::is_a<ASR::IntrinsicFunction_t>(*arg) || ASR::is_a<ASR::Cast_t>(*arg)) {
+            if (ASR::is_a<ASR::IntrinsicFunction_t>(*arg)) {
+                this->visit_IntrinsicFunction(*ASR::down_cast<ASR::IntrinsicFunction_t>(arg));
+            } else if (ASR::is_a<ASR::Cast_t>(*arg)) {
+                this->visit_Cast(*ASR::down_cast<ASR::Cast_t>(arg));
+            }
+            ASR::symbol_t* var_sym = current_scope->get_symbol(symengine_stack.pop());
+            return ASRUtils::EXPR(ASR::make_Var_t(al, loc, var_sym));
+        }
+        return arg;
```
Thirumalai-Shaktivel, line 304:
```suggestion
        if (ASR::is_a<ASR::Var_t>(*arg)) {
            return arg;
        } else if (ASR::is_a<ASR::IntrinsicFunction_t>(*arg)) {
            this->visit_IntrinsicFunction(*ASR::down_cast<ASR::IntrinsicFunction_t>(arg));
        } else if (ASR::is_a<ASR::Cast_t>(*arg)) {
            this->visit_Cast(*ASR::down_cast<ASR::Cast_t>(arg));
        } else {
            LCOMPILERS_ASSERT(false);
        }
        ASR::symbol_t* var_sym = current_scope->get_symbol(symengine_stack.pop());
        return ASRUtils::EXPR(ASR::make_Var_t(al, loc, var_sym));
```
Otherwise, this PR looks great!

src/libasr/pass/replace_symbolic.cpp
```diff
@@ -294,405 +291,196 @@ class ReplaceSymbolicVisitor : public PassUtils::PassVisitor<ReplaceSymbolicVisi
...
-                        throw LCompilersException("IntrinsicFunction: `"
-                            + ASRUtils::get_intrinsic_name(intrinsic_id)
-                            + "` is not implemented");
-                    }
-                }
+    void visit_Assignment(const ASR::Assignment_t &x) {
+        SymbolTable* module_scope = current_scope->parent;
+        if (ASR::is_a<ASR::IntrinsicFunction_t>(*x.m_value)) {
+            ASR::IntrinsicFunction_t* intrinsic_func = ASR::down_cast<ASR::IntrinsicFunction_t>(x.m_value);
+            if (intrinsic_func->m_type->type == ASR::ttypeType::SymbolicExpression) {
```
Thirumalai-Shaktivel, line 482: Yup, I was talking about this in first review. As you did, we have to only allow the `SymbolicExpression` (which we later handle to CPtr). Otherwise, we might get Type mismatch for the BindC function (which might be difficult to debug).

certik:
Very good, you chopped off 1000 lines. Thanks @Thirumalai-Shaktivel for the review!

src/libasr/pass/replace_symbolic.cpp
anutosh491: Yes this has been taken care of in all `visit_Assignment` (line 483) , visit_Print (line 626) and visit_IntrinsicFunction (line 711) functions through a type check for SymbolicExpression.

anutosh491:
I think this is ready now.

Thirumalai-Shaktivel (approved):

certik (approved):

certik:
Great job @anutosh491 !

PR #2268, opened 2023-08-12
Added support for freeing variables in the ASR symbolic pass

anutosh491:
The framework here is simple
1) Anytime we realize that `basic_new_stack` would be required to allocate memory , that's enough information to confirm that the corresponding variable would also have to be freed at the end of the program. Hence I've declared `basic_free_stack` accordingly.
2) We run a loop through `symbolic_vars` which has all CPtr symbols that have been added to the function's symbol table.
3) We can loop through it and keep adding subroutine calls to the function body of `main0()`

### DIFF
--- a/src/libasr/pass/replace_symbolic.cpp
+++ b/src/libasr/pass/replace_symbolic.cpp
@@ -101,6 +101,7 @@ public:
 
         xx.n_body = func_body.size();
         xx.m_body = func_body.p;
+        symbolic_vars.clear();
     }
 
     void visit_Variable(const ASR::Variable_t& x) {
--- a/src/libasr/pass/replace_symbolic.cpp
+++ b/src/libasr/pass/replace_symbolic.cpp
@@ -81,11 +81,7 @@ public:
         std::string new_name = "basic_free_stack";
         ASR::symbol_t* basic_free_stack_sym = module_scope->get_symbol(new_name);
         Vec<ASR::stmt_t*> func_body;
-        func_body.n = 0;
-        func_body.reserve(al, 1);
-        for (size_t i = 0; i < xx.n_body; i++) {
-            func_body.push_back(al, xx.m_body[i]);
-        }
+        func_body.from_pointer_n_copy(al, xx.m_body, xx.n_body);
 
         for (ASR::symbol_t* symbol : symbolic_vars) {
             Vec<ASR::call_arg_t> call_args;
--- a/src/libasr/pass/replace_symbolic.cpp
+++ b/src/libasr/pass/replace_symbolic.cpp
@@ -54,6 +54,7 @@ public:
         ASR::Function_t &xx = const_cast<ASR::Function_t&>(x);
         SymbolTable* current_scope_copy = this->current_scope;
         this->current_scope = xx.m_symtab;
+        SymbolTable* module_scope = this->current_scope->parent;
         for (auto &item : x.m_symtab->get_scope()) {
             if (ASR::is_a<ASR::Variable_t>(*item.second)) {
                 ASR::Variable_t *s = ASR::down_cast<ASR::Variable_t>(item.second);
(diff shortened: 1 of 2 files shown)


anutosh491:
I have a doubt . I'm seeing some cases work, for example
```
def main():
   x: S = pi
```
Through the pass I can see this block 
```
                                    (SubroutineCall
                                        2 basic_free_stack
                                        2 basic_free_stack
                                        [((Var 3 x))]
                                        ()
                                    )
```

But when I have 
```
def main():
    print(Symbol("y"))
```
I get the following error 
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --show-asr --pass=symbolic examples/expr2.py 
ASR verify pass error: ASR verify: The symbol table was not found in the scope of `symtab`.

ASR verify pass error: ASR verify: Var::m_v `stack0` cannot point outside of its symbol table
  File "/home/anutosh491/lpython/lpython/src/bin/lpython.cpp", line 1797, in ??
    return emit_asr(arg_file, lpython_pass_manager, runtime_library_dir,
  File "/home/anutosh491/lpython/lpython/src/bin/lpython.cpp", line 252, in ??
    pass_manager.apply_passes(al, asr, pass_options, diagnostics);
  File "/home/anutosh491/lpython/lpython/src/libasr/pass/pass_manager.h", line 280, in LCompilers::PassManager::apply_passes(Allocator&, LCompilers::ASR::TranslationUnit_t*, LCompilers::PassOptions&, LCompilers::diag::Diagnostics&)
    apply_passes(al, asr, _user_defined_passes, pass_options,
LCompilersException: Verify failed in the pass: symbolic
```

cc @Thirumalai-Shaktivel

src/libasr/pass/replace_symbolic.cpp
```diff
@@ -54,6 +54,7 @@ class ReplaceSymbolicVisitor : public PassUtils::PassVisitor<ReplaceSymbolicVisi
         ASR::Function_t &xx = const_cast<ASR::Function_t&>(x);
         SymbolTable* current_scope_copy = this->current_scope;
         this->current_scope = xx.m_symtab;
+        SymbolTable* module_scope = this->current_scope->parent;
```
Thirumalai-Shaktivel, line 57: Clear `symbolic_vars` here, so that we will store only the symbolics that are present in the current scope.

Thirumalai-Shaktivel:
This was the cause for the ASR verify failure:

<details>
<summary>Details</summary>

```clojure
symbol_set:
   (Function
      (SymbolTable
            8
            {
               s:
                  (Variable
                        8
                        s
                        []
                        In
                        ()
                        ()
                        Default
                        (Character 1 -2 ())
                        ()
                        BindC
                        Public
                        Required
                        .true.
                  ),
               x:
                  (Variable
                        8
                        x
                        []
                        In
                        ()
                        ()
                        Default
                        (CPtr)
                        ()
                        BindC
                        Public
                        Required
                        .true.
                  )
            })
      symbol_set
      (FunctionType
            [(CPtr)
            (Character 1 -2 ())]
            ()
            BindC
            Interface
            "symbol_set"
            .false.
            .false.
            .false.
            .false.
            .false.
            []
            .false.
      )
      []
      [(Var 8 x)
      (Var 8 s)]
      [(SubroutineCall
            2 basic_free_stack
            2 basic_free_stack
            [((Var 3 x))]
            ()
      )]
      ()
      Public
      .false.
      .false.
      "symengine/cwrapper.h"
   )
````


</details>

File: src/libasr/pass/replace_symbolic.cpp (hunk 1 above)
Comment (Thirumalai-Shaktivel) [reply src/libasr/pass/replace_symbolic.cpp:57]: Or clear it at the end of the `visit_Function`.

Comment (anutosh491) [reply src/libasr/pass/replace_symbolic.cpp:57]: Ohh ok now I realize what the error  was . Thanks for pointing this out , I've tried to address this through the latest commit.

File: src/libasr/pass/replace_symbolic.cpp [hunk 2]
```diff
@@ -75,6 +76,32 @@ class ReplaceSymbolicVisitor : public PassUtils::PassVisitor<ReplaceSymbolicVisi
...
+
+        // freeing out variables
+        std::string new_name = "basic_free_stack";
+        ASR::symbol_t* basic_free_stack_sym = module_scope->get_symbol(new_name);
+        Vec<ASR::stmt_t*> func_body;
+        func_body.n = 0;
+        func_body.reserve(al, 1);
+        for (size_t i = 0; i < xx.n_body; i++) {
+            func_body.push_back(al, xx.m_body[i]);
+        }
```
Thirumalai-Shaktivel, line 88:
```suggestion
        from_pointer_n_copy(al, xx.m_body, xx.n_body);
```

Thirumalai-Shaktivel (approved):
I left one suggestion, otherwise this LGTM!

Thirumalai-Shaktivel (approved):
I left one suggestion, otherwise this LGTM!

anutosh491:
I've addressed the suggested change , should be ready now.

certik (approved):

certik:
Thanks! I think it looks good. I rebased it to make sure it still works and I set it to merge.

Issue #2270, opened 2023-08-12
Recursive call crashes program. Fibonacci simplest function


polkovnikov:
Under Windows `lpython` I tried to compile/run following `fib.py` file:

```py
def fib(n: i32) -> i32:
    return fib(n - 1) + fib(n - 2) if n >= 3 else 1
print(fib(30))
```

above program silently crashes. In regular CPython code above works well!

While same non-recursive variant

```py
def fib(n: i32) -> i32:
    a: i32 = 1
    b: i32 = 1
    i: i32 = 0
    for i in range(n - 2):
        a, b = b, a + b
    return b
print(fib(30))
```

runs in LPython without any problem and outputs correct answer `832040`

`================== UPDATE 1 ==================`

It appeared that only LLVM IR assembly is damaged by recursive crash, but C++ code doesn't.

I.e. if you do --show-llvm and compile .ll file manually through CLang then it crashes, but if you do --show-cpp and compile .cpp through CLang then it doesn't crash and produces correct output.

It appeared that most probably Python's default IF-ELSE inline statement is not interpreted correctly. Because following rewrite of code

```py
def fib(n: i32) -> i32:
    if n >= 3:
        return fib(n - 1) + fib(n - 2)
    else:
        return 1
print(fib(30))
```

works correctly and doesn't crash.

It means that `A if B else C` is converted to LLVM by LPython incorrectly. As you should know this statement should be lazy, meaning that you can't compute both A and C beforehand and then use one of results.

But if you look into LLVM IR assembly of the very first program of this post then you'll see something like this:

```
define i32 @__module___main___fib(i32* %n) {
  ...................................
  %1 = icmp sge i32 %0, 3
  ......... LOTS OF LINES computing both clauses of IF-ELSE beforehand without using %1 ...........
  %9 = select i1 %1, i32 %8, i32 1
  store i32 %9, i32* %_lpython_return_variable, align 4
  br label %return
  ....................
```

So you'll see that first it computes condition, then doesn't use this condition but computes both clauses of IF-ELSE without using result of condition, and at the very end it returns one of computed clause.

This is incorrect for Python's IF-ELSE and even for C++'s ternary `B ? A : C`, both Python and C++ SHOULD be lazy for this ternary operator.

Without LAZY evaluation of IF-ELSE of course fib() function will go into Infinite recursion and thus crash stack through overflow.

rebcabin:
possibly a stack overflow ? can you collect and report a stack trace?

[...]

PR #2271, opened 2023-08-12
LLVM: Fix visit_IfExp()

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2270

### DIFF
--- a/integration_tests/test_ifexp_03.py
+++ b/integration_tests/test_ifexp_03.py
@@ -1,0 +1,11 @@
+from lpython import i32
+
+def fib(n: i32) -> i32:
+    return fib(n - 1) + fib(n - 2) if n >= 3 else 1
+
+def main0():
+    res: i32 = fib(30)
+    print(res)
+    assert res == 832040
+
+main0()
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -626,6 +626,7 @@ RUN(NAME test_global         LABELS cpython llvm c)
 RUN(NAME test_global_decl    LABELS cpython llvm c)
 RUN(NAME test_ifexp_01          LABELS cpython llvm c)
 RUN(NAME test_ifexp_02          LABELS cpython llvm c)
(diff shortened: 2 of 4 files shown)


Smit-create (approved):
This looks fine to me. First, we evaluate the condition, and then the correct block works. This might have been missed while we refactored and used `create_if_else` wherever possible.

integration_tests/test_ifexp_03.py
```diff
@@ -0,0 +1,11 @@
+from lpython import i32
+
+def fib(n: i32) -> i32:
+    return fib(n - 1) + fib(n - 2) if n >= 3 else 1
+
+def main0():
+    res: i32 = fib(30)
```
polkovnikov, line 7: @Shaikh-Ubaid BTW, just in case if you don't know, you can use less than fib(30) value (smaller). This is to make test run faster. Even for fib(10) there will be a crash of recursion overflow in older (unfixed) code.

Shaikh-Ubaid: Thank you for sharing @polkovnikov! Yes, I agree `fib(10)` reproduces the error and could be faster that `fib(30)`. I used `fib(30)`, because:
- the test case with `fib(30)` seems to execute quickly on my system
```bash
$ time python integration_tests/test_ifexp_03.py
832040
python integration_tests/test_ifexp_03.py  0.13s user 0.01s system 98% cpu 0.147 total
$ time lpython integration_tests/test_ifexp_03.py
832040
lpython integration_tests/test_ifexp_03.py  0.10s user 0.02s system 102% cpu 0.117 total
```
- I think `fib(30)` might have much more recursion calls and therefore, huge stack space utilization due to it. Thus, I think the test also checks the working of the binary when there are many recursion calls and huge stack space utilization due to it. 


Also thank you so much for trying LPython and reporting the bug! We appreciate it.

certik:
Awesome, thanks @Shaikh-Ubaid !

Issue #2274, opened 2023-08-13
Simplify lpython.py


certik:
Right now the file is already quite small (less than 800 lines): https://github.com/lcompilers/lpython/blob/8df052e174adbd789fc7b03f5526e706195b3543/src/runtime/lpython/lpython.py, but every time it has some custom emulation layer, it almost always breaks.

In the past we have already removed two custom layers:

* We emulated UnsignedInteger and it was subtly broken in many cases. We removed it.
* `goto`: it was emulated and it broke in a new CPython version; we remove dit.

Now there is another breakage regarding pointers and packed structs using numpy arrays (https://github.com/lcompilers/lpython/issues/2267). 

We should try to remove as many custom emulation classes as possible, and only use native CPython classes (like we now use "int" to represent unsigned integer). We might need to remove some support for C pointers or pointers in general, because CPython does not have it, and our emulation is too fragile.

Ideally the file should only contain some types, and various conversion functions that can be implemented very robustly and that do not return any "emulated/wrapped" types.

This will make the file much more robust and resilient against future CPython versions.

rebcabin:
I think the only way to robustly support pointers is a "heavy" API interface, and it's probably not worth the effort. I think, design-wise, we should make beautiful array types and then discourage people from writing C-flavored Python.

certik:
I think that `lpython.py` should only contain:

* new type annotations, such as these: https://github.com/lcompilers/lpython/blob/8df052e174adbd789fc7b03f5526e706195b3543/src/runtime/lpython/lpython.py#L80-L98
* functions that implement some operation that is robustly and cleanly implemented using CPython and its own intrinsic types, an example is: https://github.com/lcompilers/lpython/blob/8df052e174adbd789fc7b03f5526e706195b3543/src/runtime/lpython/lpython.py#L760-L761
* decorators that are implemented in a clean and robust way, some examples are: `@overload`, `@pythoncall`, `@ccall`, `@lpython`.

What we do not want to do:

* depend on internals of ctypes and hoping they can "emulate" pointers, an example is https://github.com/lcompilers/lpython/blob/8df052e174adbd789fc7b03f5526e706195b3543/src/runtime/lpython/lpython.py#L594-L616
* Return wrapped entities, such as https://github.com/lcompilers/lpython/blob/8df052e174adbd789fc7b03f5526e706195b3543/src/runtime/lpython/lpython.py#L615.

[...]

PR #2275, opened 2023-08-14
Fix integration tests with Python 3.11

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2267

Python 3.11 has a breaking change for `@dataclass`. With python `3.11`, the default values for `@dataclass` members need to be `immutable` (or `hashable`). More info: https://docs.python.org/3.11/whatsnew/3.11.html#dataclasses, https://github.com/huggingface/datasets/issues/5230

The PR also adds a CI check for integrations tests with python `3.11.3`, so that it does break in future.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -685,6 +685,7 @@ RUN(NAME structs_31          LABELS cpython llvm c)
 RUN(NAME structs_32          LABELS cpython llvm c)
 RUN(NAME structs_33          LABELS cpython llvm c)
 RUN(NAME structs_34          LABELS cpython llvm c)
+RUN(NAME structs_35          LABELS cpython llvm)
 
 RUN(NAME symbolics_01        LABELS cpython_sym c_sym llvm_sym NOFAST)
 RUN(NAME symbolics_02        LABELS cpython_sym c_sym llvm_sym NOFAST)
--- a/src/runtime/lpython/lpython.py
+++ b/src/runtime/lpython/lpython.py
@@ -2,7 +2,7 @@ from inspect import getfullargspec, getcallargs, isclass, getsource
 import os
 import ctypes
 import platform
-from dataclasses import dataclass as py_dataclass, is_dataclass as py_is_dataclass
+from dataclasses import dataclass as py_dataclass, is_dataclass as py_is_dataclass, field
 import functools
 
 
@@ -11,7 +11,7 @@ __slots__ = ["i8", "i16", "i32", "i64", "u8", "u16", "u32", "u64", "f32", "f64",
         "overload", "ccall", "TypeVar", "pointer", "c_p_pointer", "Pointer",
         "p_c_pointer", "vectorize", "inline", "Union", "static",
         "packed", "Const", "sizeof", "ccallable", "ccallback", "Callable",
-        "Allocatable", "In", "Out", "InOut", "dataclass", "S"]
+        "Allocatable", "In", "Out", "InOut", "dataclass", "field", "S"]
 
 # data-types
 
--- a/src/runtime/lpython/lpython.py
+++ b/src/runtime/lpython/lpython.py
@@ -717,9 +717,9 @@ class LpythonJITCache:
         python_path = "-I" + get_python_inc() + " "
         numpy_path = "-I" + get_include() + " "
         rt_path_01 = "-I" + get_rtlib_dir() + "/../libasr/runtime "
-        rt_path_02 = "-L" + get_rtlib_dir() + " -Wl,-rpath " \
+        rt_path_02 = "-L" + get_rtlib_dir() + " -Wl,-rpath," \
             + get_rtlib_dir() + " -llpython_runtime "
-        python_lib = "-L" + get_python_lib() + "/../.. -lpython" + \
+        python_lib = "-L" + get_python_lib() + "/../.." + f" -Wl,-rpath,{get_python_lib()+'/../..'}" + " -lpython" + \
             get_python_version() + " -lm"
 
         # ----------------------------------------------------------------------
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -444,3 +444,58 @@ jobs:
             cd integration_tests
             ./run_tests.py -b c_sym cpython_sym llvm_sym
             ./run_tests.py -b c_sym cpython_sym llvm_sym -f
+
+  python_3_11:
+    name: Run Integration tests with Python 3.11
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - uses: mamba-org/setup-micromamba@v1
+        with:
+          environment-name: lp
+          condarc: |
+            channels:
+              - conda-forge
+          create-args: >-
+            llvmdev=11.1.0
+            bison=3.4
+            re2c
+            zlib
+            cmake
+            make
+            python=3.11.4
+            numpy
+
+      - uses: hendrikmuhs/ccache-action@main
+        with:
+          key: ${{ github.job }}-${{ matrix.os }}
+
+      - name: Show Python version
+        shell: bash -e -l {0}
+        run: python --version
+
+      - name: Build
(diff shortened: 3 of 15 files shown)


integration_tests/CMakeLists.txt
```diff
@@ -780,3 +780,4 @@ COMPILE(NAME import_order_01     LABELS cpython llvm c) # any
 
 # LPython emulation mode
 RUN(NAME lpython_emulation_01    LABELS cpython NOMOD)
+RUN(NAME lpython_emulation_02    LABELS cpython)
```
certik, line 783: Add `llvm` test as well. And rename the test, it should not be called emulation --- the emulation test is only for CPython only tests, which we have just one, to show how to hook your own implementations in CPython.

I think this is a "struct" test.

certik: Unless this test doesn't run in LPython yet, in which case keep this test cpython only and let's also add a second test that runs in LPython and still breaks in CPython 3.11.

src/runtime/lpython/lpython.py
```diff
@@ -53,6 +53,16 @@ def __class_getitem__(key):
...
 
+    class_mems = list(arg.__annotations__.keys())
+    class_mems_init = dict(list(filter(lambda x: x[0] in class_mems, arg.__dict__.items())))
+
+    for class_mem in class_mems_init.keys():
+        orig_val = class_mems_init[class_mem]
+        if isinstance(orig_val, (int, float, str, bool, tuple)):
+            setattr(arg, class_mem, py_field(default=orig_val))
+        else:
+            setattr(arg, class_mem, py_field(default_factory=lambda val=orig_val: val))
```
certik, line 64: What does this do exactly? I am trying to understand this code. 
```
class X:
    a: i32 = 123
    b: bool = True
    c: i32[:] = array([1, 2, 3])
    d: i32[:] = array([4, 5, 6])
```

Is `class_mems_init` equal to `{a: 123, b: True, c: array([1,2,3), d: array([4,5,6])}`?

What does this do: `py_field(default_factory=lambda val=orig_val: val)` ? I am trying to parse it. It is assigning to the `default_factory` argument, but it is assigning `lambda val=orig_val: val`, what does that mean? Is it a lambda function with a default argument?

In general, this seems to be adding a layer of indirection, away from the default CPython behavior, see here: https://github.com/lcompilers/lpython/issues/2274. 

What exactly is the problem with `dataclass` and numpy arrays? You can't store those there anymore?

Maybe there is no other way to fix this, but I am hoping we can just use CPython exactly as is, without adding more things on top. Even the behavior in the "main" branch seems modified by:
```
    def __class_getitem__(key):
        return Array(arg, key)
```
Why is that needed?

Shaikh-Ubaid: >What exactly is the problem with dataclass and numpy arrays? You can't store those there anymore?

Yes, We can't store them directory with Python 3.11. They made this change it https://docs.python.org/3.11/whatsnew/3.11.html#dataclasses (also shared in the description of this issue). Only types that are immutable (or atleast hashable) are allowed as default values for `@dataclass` members. Since, python 3.11, for immutable types, we need to use `field(default_factory=a_call_back_function_here_which_returns_the_default_value_for_the_parameter)` as the default value.

certik: I am completely confused. I thought you can store mutable things in dataclasses, I thought that is the whole point?

Shaikh-Ubaid: >What does this do: py_field(default_factory=lambda val=orig_val: val) ? I am trying to parse it. It is assigning to the default_factory argument, but it is assigning lambda val=orig_val: val, what does that mean? Is it a lambda function with a default argument?

Yes, it is a lambda function with a default argument. I initially tried several attempts using `py_field(default_factory=lambda: orig_val)` (returning the value of `orig_val` directly, without any default argument). But it does not do the correct initialization. For example, the following fails (also added as CPython emulation test):

```bash
$ cat examples/expr2.py 
from numpy import array

@dataclass
class X:
    c: i32[:] = array([1, 2, 3])
    d: i32[:] = array([4, 5, 6])

def main0():
    x: X = X()
    print(x)
    assert x.c.sum() == 6
    assert x.d.sum() == 15

main0()
$ python examples/expr2.py
X(c=array([4, 5, 6]), d=array([4, 5, 6]))
  File "/Users/ubaid/Desktop/OpenSource/lpython/examples/expr2.py", line 15, in <module>
    main0()
  File "/Users/ubaid/Desktop/OpenSource/lpython/examples/expr2.py", line 12, in main0
    assert x.c.sum() == 6
AssertionError
``` 
both `c` and `d` got assigned to the same value of `[4, 5, 6]` where as `c` was supposed to be assigned `[1, 2, 3]`. On investigating further, I realised that it happens because the return value of the lambda function is being evaluated when it is called and not when it is defined. To fix it we used the default argument which are evaluated at definition time.

rebcabin: Question: 

1. I am surprised that this change has not caused a firestorm of pushback from the Python community. Immutable objects (instances of classes) are exactly wrong for object-oriented programming (OOP). The entire point of OOP is mutable state :)
2. I would imagine this change in 3.11 is going to break lots and lots of released code in the field, no ?
3. I would imagine someone is going to "fix" this by introducing a legacy dataclass. Perhaps we need to do likewise: just implement our own legacy dataclass.

Observation:

Immutable objects are interesting and useful. Clojure, Haskell, other functional-programming languages (FPLs) have them. Efficient implementation is non-trivial (see everything that cites Chris Okasaki -- finger trees, Bloom filters, all kinds of sophisticated jazz).

rebcabin: Immutable objects are observably different from mutable objects. In Clojure, all mutable objects are syntactically different from immutable objects. Immutable objects are the default. Mutable objects are stored in atoms, agents, vars, a couple more, and must be accessed via AT-SIGN, kind of a thread-safe indirection operator (actually called Software Transactional Memory (STM)). Without STM, mutable objects must be manually synchronized -- a gigantic tarpit of problems for real-world programmers.

Maybe this is the reason 3.11 went immutable -- they're going multithreaded, no ? With multithreading, concurrent access is a must, and immutability makes it easy.

However, perhaps they (Python) forgot that mutable state is actually NECESSARY, and perhaps they forgot that they need language mechanisms for it.

integration_tests/CMakeLists.txt
Shaikh-Ubaid: >still breaks in CPython 3.11.

If we add a test that breaks, the CI might fail and we might not be able to merge.

src/runtime/lpython/lpython.py
certik: Regarding `A[:]`, you should not subscript `A` directly, but rather have a NumPy array of A, then everything works.

rebcabin: Clojure addressed this trifecta: {immutable-by-default, threadsafe-by-default, mutable-by-opt-in} in 2009, and it fixed the terrible problems introduced, on the one hand, by Haskell, which has NO mutability, and Java, on the other hand, which has no immutability but has threads. Java inflicted pain and bugs on a generation of programmers (see books by Brian Goetz). Haskell doomed itself to impracticality (except for finTech). Clojure found The Golden Mean and is, in my opinion, the best solution to Squaring the Circle of mutability and threads. I hope Python people are smart enough to learn from Clojure.

certik: I think I figured it out: https://github.com/lcompilers/lpython/issues/2276

certik:
@Shaikh-Ubaid I would say let's implement the solution from https://github.com/lcompilers/lpython/issues/2276, and not add any new layers into `lpython.py`, and later actually remove the custom wrapper of `dataclass` completely that we currently have there.

Shaikh-Ubaid:
This PR needs a rebase. The old `2` commits are part of https://github.com/lcompilers/lpython/pull/2280.

certik (approved):

Shaikh-Ubaid:
This PR does not seem to reflect the rebase. I merged it using the `merge with rebase` option at the web GUI.

PR #2279, opened 2023-08-16
Initial support for lambda functions

Shaikh-Ubaid:


certik (approved):
I think this is a good start. It names the lambda function based on the variable.

I think the next step is to assign it to a function pointer variable "x".

Issue #2276, opened 2023-08-14
Design of structs in LPython


certik:
We want to map our structs into ordinary CPython. For now we use dataclasses, but #2275 shows that Python 3.11 changed a few things. So let's use this opportunity to design it in such a way that we just use the `dataclass` decorator directly from CPython, we do not wrap it in anyway.

There are two ways to use dataclasses that works in all Python versions including 3.11.

### No Default Values

The first way is without initializing the members:
```python
from dataclasses import dataclass
from numpy import array, ndarray

@dataclass
class X:
    a: int
    b: bool
    c: list[int]
    d: ndarray

def main0():
    x: X = X(123, True, [1, 2, 3], array([4, 5, 6]))
    print(x)
    assert x.a == 123
    assert x.b == True
    assert x.c[0] == 1
    assert x.d[1] == 5
    x.c[0] = 3
    x.d[0] = 3
    print(x)
    assert x.c[0] == 3
    assert x.d[0] == 3

main0()
```
The fields are mutable, and there is no problem at all, it works with lists and numpy arrays also.

### Default Values

First of all, I don't think we need to provide default values. But if we do, here is how to do it:
```python
from dataclasses import dataclass, field
from numpy import array, ndarray

@dataclass
class X:
    a: int = 123
    b: bool = True
    c: list[int] = field(default_factory=lambda: [1, 2, 3])
    d: ndarray = field(default_factory=lambda: array([4, 5, 6]))

def main0():
    x: X = X()
    print(x)
    assert x.a == 123
    assert x.b == True
    assert x.c[0] == 1
    assert x.d[1] == 5
    x.c[0] = 3
    print(x)
    assert x.c[0] == 3

main0()
```
This works for all Python versions including 3.11. 

### TODO

* We need to figure out how to make our own types `i32[:,:]` work for numpy arrays.
* We need to figure out arrays of structs using numpy array

rebcabin:
Wow! Excellent!

certik:
The way this should be implemented in `lpython.py` is to just do `from dataclasses import dataclass, field` inside `lpython.py` and that's it. No custom wrappers. Later on, we will support `from dataclasses import dataclass, field` directly as an option.

We can do this step by step.

Shaikh-Ubaid:
>x: ndarray = array([X(), X(), X()])

I have the following doubt with respect to the above approach of arrays of structs. Please clarify.

[...]

PR #2280, opened 2023-08-16
Support field in dataclass member initialization

Shaikh-Ubaid:
towards #2276

### DIFF
--- a/integration_tests/structs_35.py
+++ b/integration_tests/structs_35.py
@@ -1,0 +1,26 @@
+from lpython import dataclass, field, i32
+from numpy import array
+
+@dataclass
+class X:
+    a: i32 = 123
+    b: bool = True
+    c: list[i32] = field(default_factory=lambda: [1, 2, 3])
+    d: i32[3] = field(default_factory=lambda: array([4, 5, 6]))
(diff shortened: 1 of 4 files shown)


integration_tests/structs_35.py
```diff
@@ -0,0 +1,26 @@
+from lpython import dataclass, field, i32
+from numpy import array
+
+@dataclass
+class X:
+    a: i32 = 123
+    b: bool = True
+    c: list[i32] = field(default_factory=lambda: [1, 2, 3])
+    d: i32[3] = field(default_factory=lambda: array([4, 5, 6]))
+    e: i32 = field(default=-5)
```
certik, line 10: Is this the same as:
```suggestion
    e: i32 = -5
```
?

certik (approved):
I think this looks good.

integration_tests/structs_35.py
Shaikh-Ubaid: Yes, I think they are equivalent.

certik: If they are equivalent, then I would almost not support the duplicate way of doing it ("just one way of doing things").

PR #2281, opened 2023-08-17
Run integration tests with the Latest Python version in the CI

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
What do you say about this, should we test the latest or specific version like 3.11?
@Shaikh-Ubaid @certik

CMakeLists.txt
```diff
@@ -4,10 +4,14 @@ set(CMAKE_USER_MAKE_RULES_OVERRIDE ${CMAKE_SOURCE_DIR}/cmake/UserOverride.cmake)
 
 # We don't execute this if we have a tarball
 if (LPYTHON_BUILD_ALL)
-  execute_process(COMMAND "build0.sh")
+    find_program(BASH_BIN bash)
+    execute_process(COMMAND "${BASH_BIN}" "-e" "build0.sh" WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} RESULT_VARIABLE _BUILD0_EXIT)
+    if (NOT _BUILD0_EXIT EQUAL 0)
+        message(FATAL_ERROR "Running build0.sh failed (see error above)")
+    endif ()
```
Thirumalai-Shaktivel, line 11: I will submit a new PR with these changes!

Shaikh-Ubaid:
>What do you say about this, should we test the latest or specific version like 3.11?

Ondrej Sir, always recommends to test specific version of dependencies. Also, it seems if we only test the latest version, on release of a new version, there is a possibility that the previous python version breaks (as it would not be tested at the CI). It seems ideally we might need to test several python versions like we do several LLVM versions at LFortran. I guess that can be done in future as more python versions are released. Currently, I think `Python 3.11.4` is the latest python version which we test at the CI.

Thirumalai-Shaktivel:
Yup, I had the same thought to test specific versions as we do for the LLVM (Later I thought all the tests which run in CPython 3.11 will surely run with the latest CPython, right? If this isn't the case then we will split it).

Why I say this because the new user might always use the latest CPython and it might fail sometimes because we didn't test the latest CPython.

certik:
Yes, definitely always test with specific versions. As @Shaikh-Ubaid said, we'll eventually test all Python versions that we support, often time in my experience just testing the recent version (3.11 now) and the oldest we support (say 3.8 or whatever it is) is enough. When 3.12 is released, very often it's enough to just upgrade to it, fix failures, and it will still work with 3.11 and all other versions in between. However, to be absolutely certain, the most robust way is to just test all versions that we support. From a practical perspective, if you look here: https://www.python.org/downloads/, there are only 5 "Active Python Releases", 3.8 - 3.12. So we should just test all of them.

czgdp1807:
Testing with 3.11 is okay for me. We should do it. But along side we should also test 3.10 separately.

Thirumalai-Shaktivel:
I updated the CI to test Python 3.12.

Shaikh-Ubaid (approved):

certik:
We now test `python-version: ["3.9", "3.10", "3.11", "3.12"]`, which is perfect. We can try to get 3.8 working, but it's lower priority and it becomes less and less important over time. From now on we simply add every new version to this list and ensure everything works.

PR #2277, opened 2023-08-15
Fix FMA pass

Smit-create:
Fixes #1671

PR #2282, opened 2023-08-17
Fix FMA pass to use IntrinsicFunction 

Smit-create:
Alternative version of #2277


certik (approved):
I think that looks good. @Smit-create this seems better than https://github.com/lcompilers/lpython/pull/2277 ? Let me know what you think.

Smit-create:
Yes, this looks concise, and design compatible.

Smit-create:
I'll also try to test this same with the LFortran to make sure we don't break anything there.

PR #2284, opened 2023-08-19
Syntactic sugar for Array annotation

Shaikh-Ubaid:
towards https://github.com/lcompilers/lpython/issues/2276#issuecomment-1679332772, https://github.com/lcompilers/lpython/pull/1814#discussion_r1195461845


certik (approved):
Perfect! Thanks for implementing this. I think that's exactly what we needed.

PR #2286, opened 2023-08-19
feat: :sparkles: add integer bit length in LPython's builtin methods

OmarMesqq:
I used @namannimmo10 suggestion to implement CPython's native [`bit_length()`](https://docs.python.org/3/library/stdtypes.html#int.bit_length) in LPython. I am new to compilers so please tell me if there's anything I should (or should not) change :) 

Reading through the [issue's thread](https://github.com/lcompilers/lpython/issues/281) I figured that the method is already available in  `python_ast_to_asr.cpp` . Do I have to write its implementation in C++ as well then?


certik:
@OmarMesqq awesome, thanks! Isn't `bit_length` already implemented? I see a test for it here: https://github.com/lcompilers/lpython/blob/8c42eaf54cea29d065acaa9f1be25fa2a3461136/integration_tests/test_bit_length.py. 

You can go to `integration_tests` and then do `git grep bit_length`. The same with other features, to see what is already implemented and tested.

Yes, we'll help you get up to speed. See if you can find some test case that currently fails, and then let's fix it.

OmarMesqq:
@certik Ok, thanks! I am studying the codebase and am updating the PR in the mean time. One question though: how is the CI implemented in this repo? Apparently my tests won't pass due to different hashes in the final build (which is expected since I am changing `src` files anyway).

certik:
You have to update the reference tests using `./run_tests.py -u`.

OmarMesqq:
Thanks a lot for your tips, @certik! Funny thing, when I first opened the PR whenever I'd run 
`(someInteger).bit_length()` I'd get `bit.length() isn't implemented yet`. I updated my branch from main and now the feature is there and all tests passing! Someone must have implemented it :) 

I'm closing this and focusing on other tasks

certik:
Awesome, I am glad it works @OmarMesqq. You probably had an older version of LPython.

Issue #1790, opened 2023-05-11
Enforce dtype for empty()


certik:
LPython should give an error message:
```python
    x: i32[4, 5, 2] = empty([4, 5, 2])
    y: f64[24, 100, 2, 5] = empty([24, 100, 2, 5])
```
Since in CPython the arrays would be both float (?) arrays of the default kind. Instead, it needs to enforce something like this:
```python
    from numpy import int32, float64
    x: i32[4, 5, 2] = empty([4, 5, 2], dtype=int32)
    y: f64[24, 100, 2, 5] = empty([24, 100, 2, 5], dtype=float64)
```

Shaikh-Ubaid:
I think we can enforce. In `CPython`, I think it is optional.

certik:
What I mean by that is that if dtype is not specified, then it will use the same dtype as CPython (I think it's float64).

Consequently, this `x: i32[4, 5, 2] = empty([4, 5, 2])` must fail, because the RHS is f64 and the LHS is i32.

certik:
I am assigning this to @czgdp1807, since he implemented #2118 for i8, so it just needs to be extended for all dtypes.

PR #2294, opened 2023-08-24
Support/Enforce dtype for empty()

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1790

### DIFF
--- a/integration_tests/bindpy_04.py
+++ b/integration_tests/bindpy_04.py
@@ -1,5 +1,5 @@
 from lpython import i1, i32, u32, f64, c64, pythoncall, Const, TypeVar
-from numpy import empty, uint32, complex64
+from numpy import empty, uint32, complex128
 
 n = TypeVar("n")
 m = TypeVar("m")
@@ -102,8 +102,8 @@ def test_2D_array_bools():
 # Complex
 def test_array_complexes():
     n: Const[i32] = 5
-    a: c64[n] = empty([n], dtype=complex64)
-    b: c64[n] = empty([n], dtype=complex64)
+    a: c64[n] = empty([n], dtype=complex128)
+    b: c64[n] = empty([n], dtype=complex128)
 
     i: i32
     for i in range(n):
@@ -122,8 +122,8 @@ def test_array_complexes():
 def test_2D_array_complexes():
     n: Const[i32] = 3
     m: Const[i32] = 4
-    a: c64[n, m] = empty([n, m], dtype=complex64)
-    b: c64[n, m] = empty([n, m], dtype=complex64)
+    a: c64[n, m] = empty([n, m], dtype=complex128)
+    b: c64[n, m] = empty([n, m], dtype=complex128)
 
     i: i32
     j: i32
--- a/integration_tests/generics_array_02.py
+++ b/integration_tests/generics_array_02.py
@@ -1,5 +1,5 @@
 from lpython import TypeVar, restriction, i32, f32
-from numpy import empty
+from numpy import empty, int32, float32
 
 n = TypeVar("n")
 T = TypeVar('T')
@@ -16,21 +16,21 @@ def add_float(x: f32, y: f32) -> f32:
 
 def g(n: i32, a: T[n], b: T[n], **kwargs):
   r: T[n]
-  r = empty(n)
+  r = empty(n, dtype=object)
   i: i32
   for i in range(n):
     r[i] = add(a[i], b[i])
   print(r[0])
 
 def main():
-    a_int: i32[1] = empty(1)
+    a_int: i32[1] = empty(1, dtype=int32)
     a_int[0] = 400
-    b_int: i32[1] = empty(1)
+    b_int: i32[1] = empty(1, dtype=int32)
     b_int[0] = 20
     g(1, a_int, b_int, add=add_integer)
-    a_float: f32[1] = empty(1)
+    a_float: f32[1] = empty(1, dtype=float32)
     a_float[0] = f32(400.0)
-    b_float: f32[1] = empty(1)
+    b_float: f32[1] = empty(1, dtype=float32)
     b_float[0] = f32(20.0)
     g(1, a_float, b_float, add=add_float)
 
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -1095,8 +1095,8 @@ static inline std::string type_encode_dims(size_t n_dims, ASR::dimension_t* m_di
     for( size_t i = 0; i < n_dims; i++ ) {
         ASR::dimension_t dim = m_dims[i];
         dims_str += "[";
-        dims_str += extract_dim_value(dim.m_start);
-        dims_str += ",";
+        // dims_str += extract_dim_value(dim.m_start);
+        // dims_str += ",";
         dims_str += extract_dim_value(dim.m_length);
         dims_str += "]";
     }
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -2430,21 +2430,13 @@ inline bool dimension_expr_equal(ASR::expr_t* dim_a, ASR::expr_t* dim_b) {
     if( !(dim_a && dim_b) ) {
         return true;
     }
-    ASR::expr_t* dim_a_fallback = nullptr;
-    ASR::expr_t* dim_b_fallback = nullptr;
-    if( ASR::is_a<ASR::Var_t>(*dim_a) &&
-        ASR::is_a<ASR::Variable_t>(
-            *ASR::down_cast<ASR::Var_t>(dim_a)->m_v) ) {
-        dim_a_fallback = ASRUtils::EXPR2VAR(dim_a)->m_symbolic_value;
+    int dim_a_int, dim_b_int;
+    if (ASRUtils::extract_value(ASRUtils::expr_value(dim_a), dim_a_int)
+        && ASRUtils::extract_value(ASRUtils::expr_value(dim_b), dim_b_int)) {
+        return dim_a_int == dim_b_int;
     }
-    if( ASR::is_a<ASR::Var_t>(*dim_b) &&
-        ASR::is_a<ASR::Variable_t>(
-            *ASR::down_cast<ASR::Var_t>(dim_b)->m_v) ) {
-        dim_b_fallback = ASRUtils::EXPR2VAR(dim_b)->m_symbolic_value;
-    }
-    if( !ASRUtils::expr_equal(dim_a, dim_b) &&
-        !(dim_a_fallback && ASRUtils::expr_equal(dim_a_fallback, dim_b)) &&
-        !(dim_b_fallback && ASRUtils::expr_equal(dim_a, dim_b_fallback)) ) {
+
+    if( !ASRUtils::expr_equal(dim_a, dim_b) ) {
         return false;
     }
     return true;
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -2492,6 +2492,13 @@ inline bool types_equal(ASR::ttype_t *a, ASR::ttype_t *b,
                             a2->m_dims, a2->n_dims,
                             b2->m_dims, b2->n_dims);
             }
+            case (ASR::ttypeType::TypeParameter) : {
+                ASR::TypeParameter_t* left_tp = ASR::down_cast<ASR::TypeParameter_t>(a);
+                ASR::TypeParameter_t* right_tp = ASR::down_cast<ASR::TypeParameter_t>(b);
+                std::string left_param = left_tp->m_param;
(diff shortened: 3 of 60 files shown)


integration_tests/array_03.py
```diff
@@ -1,20 +1,20 @@
...
         a[i] = f64(i+1)
     for i in range(n):
         assert abs(a[i] - f64(i + 1)) < 1e-12
-    b: Allocatable[i32[:]]
-    n = 10
-    b = empty((n,), dtype=int32)
-    for i in range(n):
+
+    m: Const[i32] = 10
+    b: Allocatable[i32[:]] = empty((m,), dtype=int32)
```
certik, line 14: Here `m` doesn't need to be constant. Instead this test is testing that it can be a runtime variable. So we need to preserve the test.

certik:
I think this is great. Thank you. I left an important comment to fix before we can merge.

tests/reference/asr-arrays_01-a617b64.stderr
```diff
@@ -2,7 +2,7 @@ semantic error: Type mismatch in procedure call; the types must be compatible
   --> tests/errors/arrays_01.py:15:9
    |
 15 |         [i8(214), i8(157), i8(3), i8(146)])
-   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type mismatch (passed argument type is list[i8] but required type is i8[4])
+   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type mismatch (passed argument type is list[i8] but required type is i8[0,4])
    |
  9 |     a : i8[4] = empty(4, dtype=int8)
-   |         ^^^^^ type mismatch (passed argument type is list[i8] but required type is i8[4])
+   |         ^^^^^ type mismatch (passed argument type is list[i8] but required type is i8[0,4])
```
Thirumalai-Shaktivel, line 8: This should be
```suggestion
   |         ^^^^^ type mismatch (passed argument type is list[i8] but required type is i8[4])
```
right?

src/libasr/asr_utils.h
```diff
@@ -2430,21 +2446,13 @@ inline bool dimension_expr_equal(ASR::expr_t* dim_a, ASR::expr_t* dim_b) {
...
-        dim_a_fallback = ASRUtils::EXPR2VAR(dim_a)->m_symbolic_value;
-    }
-    if( ASR::is_a<ASR::Var_t>(*dim_b) &&
-        ASR::is_a<ASR::Variable_t>(
-            *ASR::down_cast<ASR::Var_t>(dim_b)->m_v) ) {
-        dim_b_fallback = ASRUtils::EXPR2VAR(dim_b)->m_symbolic_value;
-    }
-    if( !ASRUtils::expr_equal(dim_a, dim_b) &&
-        !(dim_a_fallback && ASRUtils::expr_equal(dim_a_fallback, dim_b)) &&
-        !(dim_b_fallback && ASRUtils::expr_equal(dim_a, dim_b_fallback)) ) {
```
Thirumalai-Shaktivel, line 2447: @czgdp1807 is this okay to be removed?

src/libasr/asr_utils.h
```diff
@@ -2430,21 +2446,13 @@ inline bool dimension_expr_equal(ASR::expr_t* dim_a, ASR::expr_t* dim_b) {
...
+    int dim_a_int, dim_b_int;
+    if (ASRUtils::extract_value(ASRUtils::expr_value(dim_a), dim_a_int)
+        && ASRUtils::extract_value(ASRUtils::expr_value(dim_b), dim_b_int)) {
+        return dim_a_int == dim_b_int;
+    }
```
Thirumalai-Shaktivel, line 2453: I think expr_equal checks for integer constant or real constant. If any other cases are to be handled please add the cases to expr_equal.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3067,18 +3025,12 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
             process_variable_init_val(current_scope->get_symbol(var_name), x.base.base.loc, init_expr);
         }
 
-        if (is_allocatable && x.m_value && AST::is_a<AST::Call_t>(*x.m_value)) {
-            tmp = check_to_allocate_array(x.m_value, var_name, x.base.base.loc);
```
Thirumalai-Shaktivel, line 3071: We have to test these changes in LFortran as well.

tests/reference/asr-arrays_01-a617b64.stderr
Shaikh-Ubaid: We previously only supported encoding a single dimension to string. Now we support encoding multi-dimensions to string. We also encode the start index or lower bound which is more descriptive.

src/libasr/asr_utils.h
Shaikh-Ubaid: I think it is. Just below these deleted lines, we use `extract_value()` which I think already encompasses the above removed lines.

src/libasr/asr_utils.h
Shaikh-Ubaid: >I think expr_equal checks for integer constant or real constant

It seems to check for only `IntegerBinOp` and `Var`.

https://github.com/lcompilers/lpython/blob/621ffc09164e52945964bb59f65f491b5a8e024d/src/libasr/asr_utils.h#L2380-L2427

The idea here is to reuse the existing constructs. It might be possible to extend `expr_equal()`, but I think a better approach is to reuse the existing constructs as much as possible.

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: >We have to test these changes in LFortran as well.

I think we need not test this particular change. I think only the backend changes need to be tested. I submitted https://github.com/lfortran/lfortran/pull/2262 which includes the LibASR changes.

integration_tests/array_03.py
```diff
@@ -9,6 +9,7 @@ def f():
         a[i] = f64(i+1)
     for i in range(n):
         assert abs(a[i] - f64(i + 1)) < 1e-12
+
```
czgdp1807, line 12:
```suggestion
```

czgdp1807 (approved):

src/libasr/asr_utils.h
Thirumalai-Shaktivel: See LFortran

Shaikh-Ubaid: Let's look into the possibilities of improving this later. This PR is being lengthier.

tests/reference/asr-arrays_01-a617b64.stderr
Thirumalai-Shaktivel: If this is okay with @certik, Let's merge this PR.

integration_tests/bindpy_04.py
```diff
@@ -102,8 +102,8 @@ def test_2D_array_bools():
 # Complex
 def test_array_complexes():
     n: Const[i32] = 5
-    a: c64[n] = empty([n], dtype=complex64)
-    b: c64[n] = empty([n], dtype=complex64)
+    a: c64[n] = empty([n], dtype=complex128)
+    b: c64[n] = empty([n], dtype=complex128)
```
certik, line 106: Good catch. It's confusing, but I think it's less confusing to use c32 and c64 to mean 32bit and 64bit complex numbers (each component), so c32 corresponds to f32 and c64 to f64. NumPy uses the size of the whole complex number (both components together). Now LPython catches the mistake, so that's good.

integration_tests/generics_array_02.py
```diff
@@ -16,21 +16,21 @@ def add_float(x: f32, y: f32) -> f32:
 
 def g(n: i32, a: T[n], b: T[n], **kwargs):
   r: T[n]
-  r = empty(n)
+  r = empty(n, dtype=object)
```
certik, line 19: This is fine for now to get it to compile. Good catch. This we will have to fix later, since this should really be just:
```
r: T[n] = empty(n, dtype=T)
```
But that requires cooperation of numpy and generics. We are improving the generics in LFortran, once we are done and have a reasonably good design, we'll update LPython as well.

tests/reference/asr-test_annassign_type_mismatch-7dac7be.stderr
```diff
@@ -2,4 +2,4 @@ semantic error: Type mismatch in annotation-assignment, the types must be compat
  --> tests/errors/test_annassign_type_mismatch.py:4:5
   |
 4 |     x: i32[4] = [1, 2, 3, 4]
-  |     ^           ^^^^^^^^^^^^ type mismatch ('i32[4]' and 'list[i32]')
+  |     ^           ^^^^^^^^^^^^ type mismatch ('i32[0,4]' and 'list[i32]')
```
certik, line 5: This is a bug. The previous error message was correct.

tests/reference/asr-test_annassign_type_mismatch2-fc883f7.stderr
```diff
@@ -2,4 +2,4 @@ semantic error: Type mismatch in annotation-assignment, the types must be compat
  --> tests/errors/test_annassign_type_mismatch2.py:4:5
   |
 4 |     x: f64[5] = [1.0, 2.0, 3.0, 4.0, 5.0]
-  |     ^           ^^^^^^^^^^^^^^^^^^^^^^^^^ type mismatch ('f64[5]' and 'list[f64]')
+  |     ^           ^^^^^^^^^^^^^^^^^^^^^^^^^ type mismatch ('f64[0,5]' and 'list[f64]')
```
certik, line 5: Bug as well.

tests/reference/asr-test_dict2-4587f02.stderr
```diff
@@ -1,5 +1,5 @@
-semantic error: Type mismatch in dictionary key, the types must be compatible
+semantic error: Key type should be 'str' instead of 'i32'
  --> tests/errors/test_dict2.py:4:7
   |
 4 |     y[1] = -3
-  |       ^ type mismatch (found: 'i32', expected: 'str')
+  |       ^
```
certik, line 5: I liked the short label, showing what is wrong.

certik:
What is the default for "empty" in NumPy, is it f64? Perhaps later we can allow the default as well. For now it's fine.

I found a bug: https://github.com/lcompilers/lpython/pull/2294#discussion_r1307599731, so that should be fixed.

Other than that, the rest are minor comments. Great job! Thanks for implementing this. I'll mark it as Draft until the bug is fixed.

tests/reference/asr-test_annassign_type_mismatch-7dac7be.stderr
Shaikh-Ubaid: I agree. I was trying to also include the `start` index of the dimension in the error message as we now check for dimensions when we use `empty()` (previously we used to ignore/skip the call to empty including skipping the dimensions and the dtype). We now check both `start` and `length` of the dimension to match with that of the dimensions of the variable to which the value of `empty()` is being assigned to. The above mismatch error can be generated if there is a mismatch in either of the quantities `start` or `length` of the dimension.

Since, Python does not use lower_bounds, we can update it to only include the `length` value in the error message. I am slightly concerned about cases when the length dimensions are equal/match but there is a mismatch in the start dimension (due to some internal bug in the compiler). In such a case the error message would just print the length dimension and this would not make it clear where exactly the mismatch occurred.

Shaikh-Ubaid: `i32[0,4]` this seems confusing. We probably need a better approach/representation like `i32[0:4]`. I think we can look into this later. For now, I will update the code to only include the `length` dimension.

tests/reference/asr-test_annassign_type_mismatch2-fc883f7.stderr
Shaikh-Ubaid: Updated.

tests/reference/asr-test_annassign_type_mismatch-7dac7be.stderr
certik: Since this is in LPython where the lower bound is always zero, then don't print it at all. The syntax `i32[0,4]` means a 2D array, where the size of the first dimension is 0 (which I am not sure is even allowed).

certik (approved):
I think that looks good to merge. Let's add a test for an error message if the dtype mismatches the type annotation? That can be done in a subsequent PR.

Shaikh-Ubaid:
> Let's add a test for an error message if the dtype mismatches the type annotation? That can be done in a subsequent PR.

Yes, I have it planned.

Shaikh-Ubaid:
Thank you so much for the approval!

PR #2295, opened 2023-08-24
Refactor: Move test_modules folder to integration_tests

Shaikh-Ubaid:
The `test_modules` folder is used by `integration_tests/test_import_04.py`. I think the `test_modules` folder should not be in the root directory of the project. I think it should ideally reside in `integration_tests` directory.


certik (approved):
Yes, I think this is fine.

Issue #1827, opened 2023-05-17
Test lpython.py with multiple CPython versions


certik:
We need to test the lpython.py tests in integration_tests with multiple CPython versions. It looks like there are some bugs related to that.

rebcabin:
How about just picking CPython 3.11 and committing to it to reduce our
workload until LPython is 99% feature-complete?

Shaikh-Ubaid:
We currently test with `Python 3.10` and `Python 3.11.4` at the CI. Is this issue safe to close or shall we test other Python versions like `Python 3.9`, `Python 3.8` at the CI?

certik:
We should test with all the Python versions at: https://www.python.org/downloads/, so let's test with 3.8, 3.10, 3.11 and 3.12. To save time, we can probably skip 3.9. Once 3.8 is deprecated, we can test 3.9 instead.

Shaikh-Ubaid:
It seems `list` type is not subscriptable in Python `3.8`.

```
$ python --version
Python 3.8.17
```
```
$ cat examples/expr2.py 
from lpython import i32

def print_list(a: list[i32]):
    print(a)

def main0():
    x: list[i32] = [1, 2, 3]
    print_list(x)

main0()
```
```
$ python examples/expr2.py
Traceback (most recent call last):
  File "examples/expr2.py", line 3, in <module>
    def print_list(a: list[i32]):
TypeError: 'type' object is not subscriptable
```
```
$ lpython examples/expr2.py 
[1, 2, 3]
```

Please, could you share what we should do?

Shaikh-Ubaid:
It seems annotation for `list` is supported since `Python 3.9` (reference https://stackoverflow.com/a/64755075).

certik:
Ok, then Let's test Python 3.9, 3.10 and 3.11 explicitly, all three. Once 3.12 is released in October, we'll add it as well.

Shaikh-Ubaid:
For Python 3.8 and before, it seems one needs to use `from typing import List` (which I think is subscriptable).

Shaikh-Ubaid:
>Ok, then let's test Python 3.9, 3.10 and 3.11 explicitly, all three. Once 3.12 is released in October, we'll add it as well.

Sure, got it.

PR #2298, opened 2023-08-25
CI: Test integration_tests with Python 3.9, 3.10 and 3.11

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/1827#issuecomment-1692612951


Shaikh-Ubaid:
It seems `linear_regression`, `correlation`, `covariance` were added to the statistics module in `Python 3.10` (https://docs.python.org/3/library/statistics.html#statistics.correlation). I temporarily removed them in [tmp](https://github.com/lcompilers/lpython/pull/2298/commits/107355be4db1bf56321edeb5d7d6f8826f13d45b).

certik:
I have an idea. Why not to test `integration_tests/test_statistics.py` only with Python >= 3.10. That way we'll have a mechanism to test features in newer CPython versions. In LPython we can support the latest features, or we can even allow this with compiler options (but I would not bother for now, we can do that later).

Shaikh-Ubaid:
We currently test both `cpython` and `c_py` (as the test depends on CPython for execution). We will eventually support other backends like `llvm_py` and probably `c_sym`, `llvm_sym`.

integration_tests/CMakeLists.txt
```diff
@@ -623,7 +635,7 @@ RUN(NAME bindpy_01           LABELS cpython c_py ENABLE_CPYTHON NOFAST COPY_TO_B
 RUN(NAME bindpy_02           LABELS cpython c_py LINK_NUMPY COPY_TO_BIN bindpy_02_module.py)
 RUN(NAME bindpy_03           LABELS cpython c_py LINK_NUMPY NOFAST COPY_TO_BIN bindpy_03_module.py)
 RUN(NAME bindpy_04           LABELS cpython c_py LINK_NUMPY NOFAST COPY_TO_BIN bindpy_04_module.py)
-RUN(NAME bindpy_05           LABELS llvm_py c_py ENABLE_CPYTHON COPY_TO_BIN bindpy_05_module.py)
+RUN(NAME bindpy_05           LABELS llvm_py c_py ENABLE_CPYTHON COPY_TO_BIN bindpy_05_module.py REQ_PY_VER 3.10)
```
Shaikh-Ubaid, line 638: This test seems to fail with `Python 3.9`. For the moment, I marked it as `REQ_PY_VER 3.10`. I will tackle the failure with `Python 3.9` in a future PR.

Shaikh-Ubaid:
From https://github.com/lcompilers/lpython/issues/1827#issuecomment-1692612951,

>Ok, then let's test Python 3.9, 3.10 and 3.11 explicitly, all three

We now explicitly test all the three `3.9`, `3.10` and `3.11` Python versions with `integration_tests`.

integration_tests/CMakeLists.txt
```diff
@@ -334,6 +334,18 @@ macro(RUN)
         set(RUN_EXTRA_ARGS ${RUN_EXTRA_ARGS} -I${CMAKE_CURRENT_SOURCE_DIR}/${RUN_IMPORT_PATH})
     endif()
 
+    if (RUN_REQ_PY_VER)
+        string(REGEX REPLACE "([0-9]+)\\.([0-9]+)\\.([0-9]+)" "\\1" PY_MAJOR_VERSION "${Python_VERSION}")
+        string(REGEX REPLACE "([0-9]+)\\.([0-9]+)\\.([0-9]+)" "\\2" PY_MINOR_VERSION "${Python_VERSION}")
+        string(REGEX REPLACE "([0-9]+)\\.([0-9]+)" "\\1" REQ_PY_MAJOR_VERSION "${RUN_REQ_PY_VER}")
+        string(REGEX REPLACE "([0-9]+)\\.([0-9]+)" "\\2" REQ_PY_MINOR_VERSION "${RUN_REQ_PY_VER}")
+
+        if (PY_MINOR_VERSION LESS REQ_PY_MINOR_VERSION)
```
Shaikh-Ubaid, line 343: This compares only the Python minor versions for now. We might need to consider Python major versions when Python `4.0.0` is released, but I guess there might probably be some time for it.

Shaikh-Ubaid:
The CI is green. This is ready!

certik (approved):
I think this is good, thank you for fixing it!

PR #2300, opened 2023-08-26
TEST: Enable integration_tests/expr_16.py for LLVM

### LINKED ISSUE #1859: integration_tests/expr_16.py fails in LLVM
```console
$ lpython expr_16.py
...

declare void @_lcompilers_print_error(i8*, ...)

declare void @exit(i32)

define i32 @main(i32 %0, i8** %1) {
.entry:
  call void @_lpython_set_argv(i32 %0, i8** %1)
  call void @__module__global_symbols__lpython_main_program()
  ret i32 0
}

declare void @_lpython_set_argv(i32, i8**)
  %20 = sub i64 %19, i32 0
  %21 = add i64 %20, i32 1
  store i64 %21, i32* %17, align 4
 i32Both operands to a binary operator are not of the same type!
  %22 = mul i32 1, i64 %21



```

Shaikh-Ubaid:
fixes #1859 

The test case mentioned in the issue now works in the main branch. Enabling the specific test in the PR, thus hopefully closing the issue.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -487,7 +487,7 @@ RUN(NAME expr_13             LABELS llvm c
         EXTRAFILES expr_13b.c NOFAST)
 RUN(NAME expr_14             LABELS cpython llvm c)
 RUN(NAME expr_15             LABELS cpython llvm c)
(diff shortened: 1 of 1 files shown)


certik:
Thanks!

PR #2302, opened 2023-08-26
Refactor: ASR: Handle c_p_pointer() in visit_Call()

Shaikh-Ubaid:
This PR removes the use of
```cpp
is_c_pointer_call_copy = is_c_pointer_call;
visit_expr(...);
if (is_c_pointer_call) {
    ...
}
is_c_pointer_call = is_c_pointer_call_copy;
```
and handles the `c_p_pointer()` in `visit_Call()`,

thus making the code clean/tidy.

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -498,7 +498,6 @@ public:
     Vec<ASR::stmt_t*> *current_body;
     ASR::ttype_t* ann_assign_target_type;
     AST::expr_t* assign_ast_target;
-    bool is_c_p_pointer_call;
 
     std::map<std::string, int> generic_func_nums;
     std::map<std::string, std::map<std::string, ASR::ttype_t*>> generic_func_subs;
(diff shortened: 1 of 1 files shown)


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3012,11 +3022,12 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
             type = ASRUtils::TYPE(ASR::make_Allocatable_t(al, type->base.loc,
                 ASRUtils::type_get_past_pointer(type)));
         }
-        bool is_c_p_pointer_call_copy = is_c_p_pointer_call;
-        ASR::expr_t *value = nullptr;
+
+        create_add_variable_to_scope(var_name, type,
```
Shaikh-Ubaid, line 3026: The `c_p_pointer()` needs the variable to which it is being assigned. Therefore, in this PR, we first add the variable to scope, then visit the value/`c_p_pointer()` and later set the variables `init_val` in `process_variable_init_val()`.

certik (approved):
I think this is fine, thanks!

Issue #1396, opened 2023-01-03
Build0.sh fails when repo is forked in Github


rebcabin:
This morning, I created a fork of lcompilers/lpython so that I could submit pull requests according to github's recommendations. I got a cascade of build and test failures that I traced back to `./ci/version.sh`, which calls `git describe --tags --dirty`. This call fails because the fork doesn't have any tags. It works only if executed in a clone of the original repo. 

When executed in the fork, I get ...

```
(lp) ┌─(~/Documents/GitHub/lpython)──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s004)─┐
└─(11:57:38 on main ✭)──> ./ci/version.sh                                                                                                                 ──(Tue,Jan03)─┘
++ git describe --tags --dirty
fatal: No names found, cannot describe anything.
+ version=
```

Thirumalai-Shaktivel:
Fetching tags from origin might work.
```
git remote add lpython https://github.com/lcompilers/lpython.git
git fetch lpython main
```
and then Let's make a clean build:
```
git clean -dfx
./build0.sh
./build1.sh
```

certik:
This failure happens when there are missing tags. I noticed it can be fragile sometimes, so we should make the script more robust against this.

PR #2303, opened 2023-08-27
fix: :bug: create default version value for forked repos

OmarMesqq:
## [Related issue](https://github.com/lcompilers/lpython/issues/1396)

This change allows people who are trying to contribute to LPython to build the project locally without having to manually create a tag or fetch from origin. 

The local build will default to `0.0.0-0`


certik:
I would not do it by default like this, as I think we want to see the error message if something goes wrong with the tags.

Rather, why not creating a separate script for this?

OmarMesqq:
@certik good point. I assumed that, unlike the forked ones, the upstream repo will always have the configured tags but an error might indeed occur. 

I will undo changes in `version.sh` and propose a new shell script that handles the case for forked repos.

certik:
(Usually that is the case. But I've seen sometimes if the tags don't download for some reason, such as in your fork, then we want to see it break.)

ci/version.sh
```diff
@@ -11,8 +11,18 @@
...
 
-version=$(git describe --tags --dirty)
-version="${version:1}"
-echo $version > version
+set -ex 
+if [ $exit_code -ne 0 ]; then
+  echo "There were errors while getting the tags. Defaulting to 0.0.0-0"
+  version="0.0.0-0"
+  echo $version > version
+  exit 1
```
certik, line 22: Wouldn't this fail for you? It seems it would return a return code 1, which would exit the parent script too early?

OmarMesqq: I am sorry. PR is still WIP. I should find a final solution this weekend. But yes, it does exit the parent script. The reason I chose this approach is because of what you said about [if the tags fail, we want to see it break](https://github.com/lcompilers/lpython/pull/2303#issuecomment-1696502164). But I can change this

certik: Yes. I would leave the `version.sh` script alone. But I would implement a new script that is not called by our CI, but can be used by you. Then things will fail for us (as we want it), but they would work for you (as you want it).

OmarMesqq: Alright, `version.sh` checked out to main. Created the helper script in the root of the project. Please let me know if I should place it anywhere else. Maybe a `scripts` folder for this sort of stuff?

certik (approved):
I think that works!

certik:
Thanks!

abhi-glitchhg:
It would be helpful to mention this in the readme: https://github.com/lcompilers/lpython#linux-and-macos

PR #2306, opened 2023-08-27
Merge call in visit_Expr() with visit_Call()

### LINKED ISSUE #2305: User defined functions ignored when used as subroutine
https://github.com/lcompilers/lpython/blob/d24f131a153f6ae76a8287805875087ba0e367c9/src/lpython/semantics/python_ast_to_asr.cpp#L6684-L6686

Example:
```python
$ cat examples/expr2.py 
from lpython import i32, InOut

def reserve(a: InOut[list[i32]], b: i32):
    a.append(b)
    print("user defined reserve() called")

def main0():
    x: list[i32] = []
    reserve(x, 5)

    assert len(x) == 1
    assert x[0] == 5

main0()
```
``` bash
$ python examples/expr2.py      
user defined reserve() called
```
```bash
$ lpython_in_main examples/expr2.py
AssertionError
```

Shaikh-Ubaid:
fixes #2305


certik (approved):
I think that's fine. Thanks!

PR #2308, opened 2023-08-28
Improve dims error message format

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/pull/2307#discussion_r1308011096


certik (approved):

Issue #959, opened 2022-08-12
Implementing hyperbolic trigonometric, exponential and rounding functions in NumPy


czgdp1807:
Below I have provided some links to get started and then explained what you both are supposed to do,

**Links to get started**

1. https://github.com/lcompilers/lpython/blob/main/src/runtime/lpython_intrinsic_numpy.py - Contains definitions of `sin`, `cos`, `tan`, `log`, `log2`, `log10`, `sqrt` for `f32` and `f64` types. Some points to be noted - i) `@vectorize` decorator automatically vectorizes the function. Now by "vectorizes the function" I mean even if the functions are defined for single number input and single number outputs, you can call them upon are array inputs as well.
2. https://github.com/lcompilers/lpython/blob/main/integration_tests/elemental_01.py, https://github.com/lcompilers/lpython/blob/main/integration_tests/elemental_04.py - These are the test files for NumPy functions. You can see how `sin`, `cos`, `log` are called with array inputs even though they were defined for single input and single output. `@vectorize` does the job for you. You can also notice how the functions are tested. You can make the tests as natural/complicated as you want to. For example for testing `sin`, `cos` simultaneously, you can just verify `sin(array)**2 + cos(array)**2 == 1`. Similarly for trigonometric function, you can take a few trigonometric identities and try verifying them in the tests (with 1d, 2d, 5d, 6d, etc arrays as inputs). In fact you can also try finding the derivative of functions using arrays. For example for a function `f`, you can do `(f(array + harray) - f(array))/harray` and match with the formula you know for derivative of `f` you know. Bottomline is you can use your creativity to the full extent to write interesting test cases. In case you get errors don't work around the test case, instead let me know and I will fix the error for you.

Overall this task is easy-medium. Easy because extending `src/runtime/lpython_intrinsic_numpy.py` is fairly easy (just follow the already existing pattern) and medium because writing test cases requires some creativity and effort.

**What you have to do?**

- [x] `sinh` - https://github.com/lcompilers/lpython/pull/1002
- [x] `cosh` - https://github.com/lcompilers/lpython/pull/1002
- [x] `tanh` - https://github.com/lcompilers/lpython/pull/1064
- [x] `exp` - https://github.com/lcompilers/lpython/pull/1064
- [ ] `arcsinh`
- [ ] `arccosh`
- [ ] `arctanh`
- [ ] `fix`
- [ ] `floor`
- [ ] `ceil`
- [ ] `trunc`

[...]

PR #2316, opened 2023-09-07
[numpy] add trunc

khushi-411:
Hi,

The PR adds the `trunc` function to the lpython library. It follows the issue #959.

cc: @certik @czgdp1807

Looking forward to hearing your thoughts and feedback on this addition. Thank you for your time! 

One thing that's been on my mind is whether we have style checks like flake8, clang, or lintrunner in lpython. Do you think it would be a good idea to include them?

Thanks! :)


tests/reference/asr-array_01_decl-39cf894.json
```diff
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-array_01_decl-39cf894.stdout",
-    "stdout_hash": "137a0c427925ba7da2e7151f2cf52bfa9a64fede11fe8d2653f20b64",
+    "stdout_hash": "2aa47467473392c970bb1ddde961e3007d4c157bb0ea507b5e0db4a4",
     "stderr": null,
```
khushi-411, line 10: Just curious, why do we need these generated `.json` or `.stderr` files? Is there any place I could read on the importance of these additions? Thank you!

tests/reference/asr-array_01_decl-39cf894.stderr
```diff
@@ -0,0 +1,11 @@
...
+ --> tests/../integration_tests/array_01_decl.py:2:1
+  |
+2 | from numpy import empty, int16, int32, int64, float32, float64, complex64, complex128
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ imported here
+
+semantic error: Decorator: vertorize is not supported
+   --> $DIR/src/bin/../runtime/lpython_intrinsic_numpy.py:421:2
+    |
+421 | @vertorize
+    |  ^^^^^^^^^
```
Shaikh-Ubaid, line 11: I think this is not needed. This seems to be a reference output of an old version of the changes in this PR.

You can remove this by using:

(First ensure that your important changes (if any) are safe (either committed or in a new branch)):
```bash
rm -rf tests/reference/*
(After executing the above, you might need to press y here if there is any prompt saying "are you sure you want to delete all these files")
```
Now just generate the reference tests again.
```bash
./run_tests.py
```

tests/reference/asr-array_01_decl-39cf894.json
Shaikh-Ubaid: These are part of the reference tests. Reference tests help us to ensure that the output being generated is as expected. The `.stderr` reference output is generated when there is an error being observed when using the compiler.  This could be a syntax/parser error, a semantic error or I think even a code generation error (this is probably rare, we mostly handle all errors at the semantic level). The `.stderr` reference tests help us ensure that there is indeed an error being observed for different cases where there is an error expected (For example, we expect an error to be throw when there is a type mismatch). The error message is usually descriptive and guides the user towards the correct approach (If it is not then that might be a bug and we need to improve it.)

tests/reference/asr-array_01_decl-39cf894.stderr
khushi-411: Make sense. Committed in https://github.com/lcompilers/lpython/pull/2316/commits/e2bfbff7893496e45fe7eb793d90c6f68c2f7178. Thanks!

tests/reference/asr-array_01_decl-39cf894.json
khushi-411: Hi, @Shaikh-Ubaid!
Got it. Thanks for the clear description.

certik:
Intrinsic functions should be implemented using the IntrinsicFunction approach. See for example how sin/cos is implemented.

tests/reference/runtime-test_dict15-6f3af0d.stderr
```diff
@@ -1 +0,0 @@
-KeyError: The dict does not contain the specified key
```
Shaikh-Ubaid, line 1: Unrelated: These reference outputs got removed. They were added in https://github.com/lcompilers/lpython/pull/2261/files. In that PR, although we added the reference output, it seems we missed to enable those tests in `tests.toml`.

khushi-411: Ooo, I see. I've reverted the changes and enabled those tests in `tests.toml`. Please let me know if it sounds good to you. Thanks!

khushi-411:
> Intrinsic functions should be implemented using the IntrinsicFunction approach. See for example how sin/cos is implemented.

Hi, @certik! Thanks for looking into it. I've added `trunc` using the IntrinsicFunction approach. Could you please take another look and let me know if I could improve it somewhere?

src/runtime/lpython_intrinsic_numpy.py
```diff
@@ -410,3 +410,23 @@ def ceil(x: f32) -> f32:
...
+    return _lfortran_dtrunc(x)
+
+@ccall
+def _lfortran_strunc(x: f32) -> f32:
+    pass
+
+@overload
+@vectorize
+def trunc(x: f32) -> f32:
+    return _lfortran_strunc(x)
```
certik, line 432: This shouldn't be needed I think.

certik:
Excellent, this looks great. @Thirumalai-Shaktivel would you mind reviewing this please? Didn't we have some shorter way to define these common intrinsics? I don't see how right now, so this might be the way to do it.

Thirumalai-Shaktivel:
I think we can use the `create_exp_macro` macro. But, we have to update the macro to include the instantiate function as we do for `create_trig`.

The only difference between them is the type `create_trig` accepts both real and complex but `create_exp_macro` functions accept only the `real` type. 

I will add this to my to-do list, to redesign the handling.

Thirumalai-Shaktivel:
I see, we cannot include an instantiate function in the `create_exp_macro` macro.
I think let's keep the changes as is in this PR and Later while redesigned the handling, I will try to fix this.

src/runtime/lpython_intrinsic_numpy.py
Thirumalai-Shaktivel: Yup, let's remove this.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -2775,6 +2775,7 @@ PyMODINIT_FUNC PyInit_lpython_module_)" + fn_name + R"((void) {
             SET_INTRINSIC_NAME(Exp, "exp");
             SET_INTRINSIC_NAME(Exp2, "exp2");
             SET_INTRINSIC_NAME(Expm1, "expm1");
+	    SET_INTRINSIC_NAME(Trunc, "trunc");
```
Thirumalai-Shaktivel, line 2778: Please fix all the indents to use spaces.
We use 4 spaces as a tab.

src/libasr/codegen/asr_to_julia.cpp
```diff
@@ -1911,6 +1911,7 @@ class ASRToJuliaVisitor : public ASR::BaseVisitor<ASRToJuliaVisitor>
             SET_INTRINSIC_NAME(Exp, "exp");
             SET_INTRINSIC_NAME(Exp2, "exp2");
             SET_INTRINSIC_NAME(Expm1, "expm1");
+	    SET_INTRINSIC_NAME(Trunc, "trunc");
```
Thirumalai-Shaktivel, line 1914: ..

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -91,6 +92,7 @@ inline std::string get_intrinsic_name(int x) {
         INTRINSIC_NAME_CASE(Tanh)
         INTRINSIC_NAME_CASE(Gamma)
         INTRINSIC_NAME_CASE(LogGamma)
+	INTRINSIC_NAME_CASE(Trunc)
```
Thirumalai-Shaktivel, line 95: ..

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -2424,7 +2464,9 @@ namespace IntrinsicScalarFunctionRegistry {
                    verify_function>>& intrinsic_function_by_id_db = {
         {static_cast<int64_t>(IntrinsicScalarFunctions::LogGamma),
             {&LogGamma::instantiate_LogGamma, &UnaryIntrinsicFunction::verify_args}},
-        {static_cast<int64_t>(IntrinsicScalarFunctions::Sin),
+        {static_cast<int64_t>(IntrinsicScalarFunctions::Trunc),
+            {&Trunc::instantiate_Trunc, &UnaryIntrinsicFunction::verify_args}},
+	{static_cast<int64_t>(IntrinsicScalarFunctions::Sin),
```
Thirumalai-Shaktivel, line 2469: ..

tests/tests.toml
```diff
@@ -1000,6 +1000,14 @@ asr = true
...
 run = true
 
+[[test]]
+filename = "errors/test_dict15.py"
+run = true
+
+[[test]]
+filename = "errors/test_dict16.py"
+run = true
+
```
Thirumalai-Shaktivel, line 1010: Should this be removed?

src/libasr/codegen/asr_to_c_cpp.h
khushi-411: Sure, done. Thanks :-)

tests/tests.toml
khushi-411: Yes, the earlier PR missed to enable those tests. So, while updating references in this PR, it removed the required references. Hence, I enabled those tests so that we won't affect them.

Related discussion: https://github.com/lcompilers/lpython/pull/2316#discussion_r1319386819

src/runtime/lpython_intrinsic_numpy.py
certik: @khushi-411 can you please remove this change? I think it's not needed, and I think it will also bring down the amount of changes in reference tests, since the symbol table IDs should stay unchanged.

tests/tests.toml
certik: Yes, that's fine.

certik:
After you fix https://github.com/lcompilers/lpython/pull/2316#discussion_r1320173551, go ahead and click on "Ready for review", I'll give it one final review.

certik:
@Thirumalai-Shaktivel yes I agree, let's merge this PR as is, and we'll redesign later. We should probably create macros for the most common cases and just use them.

src/runtime/lpython_intrinsic_numpy.py
khushi-411: After removing this overload, the tests fail for the `fp32` case. The other case might be to obliterate it (but this is not even an option to be considered).

Therefore, I think we'll need this. Please correct me if I'm wrong; if lpython follows the numpy/array-api convention, `np.trunc` supports `fp32`, so I think we should add an overload for it. (And I guess for complex dtypes too...)

why do you think we should remove it?

certik: I would have to see the failure. Yes, we need to support all types and tests. The way it should work is that the frontend creates an IntrinsicFunction node, and then the LLVM backend calls an ASR->ASR pass where this node gets implemented, or it might even later implement it directly.

khushi-411: Interesting, will search about it more.

About the failures, I got:
```diff
semantic error: Arguments do not match for any generic procedure, trunc
  --> /home/khushi/Documents/lpython/integration_tests/elemental_13.py:54:16
   |
54 |     observed = trunc(arraynd)
   |                ^^^^^^^^^^^^^^ 
```
In this part of the code:
```python
    arraynd: f32[32, 16, 8, 4] = empty((32, 16, 8, 4), dtype=float32)

    for i in range(32):
        for j in range(16):
            for k in range(8):
                for l in range(4):
                    arraynd[i, j, k, l] = f32(f64((-1)**l) * sqrt(float(i + j + j + l)))

    observed: f32[32, 16, 8, 4] = empty((32, 16, 8, 4), dtype=float32)
    observed = trunc(arraynd)
```

certik: I think the code in the ast->asr has to be improved to take care of this, like it does for sin/cos.

khushi-411: I think `sin/cos`, too, uses `fp32` overloads as I did here; ref:
https://github.com/lcompilers/lpython/blob/main/src/runtime/lpython_intrinsic_numpy.py#L14-#L21

when I removed those overloads, I got the same errors as above. Please see:
```python
semantic error: Arguments do not match for any generic procedure, sin
  --> /home/khushi/Documents/lpython/integration_tests/elemental_01.py:10:24
   |
10 |         assert abs(sin(sin(array[i])) - result[i]) <= eps
   |                        ^^^^^^^^^^^^^ 
```

certik: I see, I didn't realize that. Ok then, we can leave it then, and fix it later.

certik (approved):
I think this is fine.

certik:
@czgdp1807 I think you probably want to merge this after the ASR sync is done, correct?

czgdp1807:
Correct.

khushi-411:
Great! Thanks very much for reviewing the PR @certik, @Thirumalai-Shaktivel, and @Shaikh-Ubaid!

certik:
Thanks @khushi-411 for the PR! We'll merge it once the ASR sync is complete.

PR #2318, opened 2023-09-10
Implement random.seed()

### LINKED ISSUE #2110: BLOCKER: How to seed randoms with entropy ?
This is a real BLOCKER but I hope easy to fix ? I need randoms that change from run-to-run. Our LPython random produces the same result every run.

```python
from random import random

print(random())  ########### PRODUCES SAME RESULT EVERY RUN ##########

# BUT, how to seed with time-of-day or internet noise
# or some other sort of variable entropy?

# NO WORK from random import seed
```

OmarMesqq:
Fixes #2110.

### DIFF
--- a/integration_tests/test_random.py
+++ b/integration_tests/test_random.py
@@ -52,6 +52,15 @@ def test_weibullvariate():
     r = random.weibullvariate(-5.6, 1.2)
     print(r)
 
+def test_seed():
+    random.seed(123)
+    t1: f64
+    t1 = random.random()
+    random.seed(321)
+    t2: f64
+    t2 = random.random()
+    assert t1 != t2
+
 def check():
     test_random()
     test_randrange()
@@ -60,5 +69,6 @@ def check():
     test_paretovariate()
     test_expovariate()
     test_weibullvariate()
+    test_seed()
 
 check()
--- a/integration_tests/test_random.py
+++ b/integration_tests/test_random.py
@@ -53,13 +53,31 @@ def test_weibullvariate():
     print(r)
 
 def test_seed():
+    random.seed()
+    t6: f64 = random.random()
     random.seed(123)
     t1: f64
     t1 = random.random()
     random.seed(321)
     t2: f64
     t2 = random.random()
+    random.seed(123)
+    t3: f64
+    t3 = random.random()
+    random.seed(0)
+    t4: f64
+    t4 = random.random()
+    random.seed(0)
+    t5: f64
+    t5 = random.random()
+    random.seed()
+    t7: f64 = random.random()
     assert t1 != t2
+    assert t1 == t3
+    assert t1 != t4
+    assert t1 != t5
+    assert t4 == t5
+    assert t6 != t7
 
 def check():
     test_random()
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -96,13 +96,14 @@ LFORTRAN_API void _lfortran_random_number(int n, double *v)
     }
 }
 
-LFORTRAN_API void _lfortran_init_random(unsigned seed) 
+LFORTRAN_API void _lfortran_init_random_seed(unsigned seed)
 {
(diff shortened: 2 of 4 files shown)


src/libasr/rng.cpp
```diff
@@ -0,0 +1,14 @@
...
+#include <random>
+
+extern "C" {
+    LFORTRAN_API double _lfortran_random() {
+        static std::random_device rd;
+        static std::mt19937 generator(rd());
+        static std::uniform_real_distribution<double> distribution(0.0, 1.0);
+
+        return distribution(generator);
+    };
```
certik, line 13: This must be implemented in the runtime library. However we use C there, not C++, so must be implemented in C.

OmarMesqq: In `libasr/runtime` or `runtime` ?

OmarMesqq:
@certik Can I use OS supplied PRNGs? I was thinking of writing the generator on my own, but it wouldn't be as cryptographically secure as `/dev/urandom` for instance

certik:
What issue is this trying to fix?

I would use the libc `rand()` and also allow to initialize it.

OmarMesqq:
@certik 
it fixes the [blocker](https://github.com/lcompilers/lpython/issues/2110) reported by @rebcabin 
With this final commit, running: 

```python 
from random import random 
print(random())
``` 
yields a different random number every run

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -96,8 +96,13 @@ LFORTRAN_API void _lfortran_random_number(int n, double *v)
...
 }
 
-LFORTRAN_API double _lfortran_random()
+LFORTRAN_API double _lfortran_random(double seed)
 {
+    if (seed == 0) {
+        srand((unsigned int)clock());
+    } else {
+        srand(seed);
+    }
```
certik, line 105: It seems the definition of `srand` is:
```
void srand( unsigned seed );
```
So `seed` must be of type `unsigned`.

This `srand` should be split into a dedicated function `_lfortran_init_random(unsigned seed)`. This `init_random()` function should only be called once, not for every `random` call. So we should expose it to the user via our runtime library somehow.

Ideally we would emulate what numpy or CPython already does. We would need to investigate.

certik (commented):
Perfect, thanks! I left some comments.

src/libasr/runtime/lfortran_intrinsics.c
OmarMesqq: Thanks for the feedback. I refactored the code using a logic similar to [CPython's](https://github.com/python/cpython/blob/80f958529bc5c466add532df4c8dd314995f689f/Lib/random.py#L135). They used classes, but that would make the code way more complex.

certik:
Thanks. I think this approach might work, I'll play with this soon.

certik:
We have to be 100% compatible with CPython. In CPython, `random.seed(0)` initializes the seed, so the following returns the same numbers in CPython:
```
    random.seed(0)
    print(random.random())
    random.seed(0)
    print(random.random())
```
But two different number in LPython (since 0 is used for clock initialization).

certik:
@OmarMesqq I fixed it up, let's see if this works.

OmarMesqq:
@certik Nice! Sorry, I didn't create more thorough tests. Already paying attention to the new `bug` issue you opened. Thanks  Ondrej!

certik:
Thank you!

integration_tests/test_random.py
```diff
@@ -52,6 +52,33 @@ def test_weibullvariate():
...
+    t5: f64
+    t5 = random.random()
+    random.seed()
+    t7: f64 = random.random()
+    assert t1 != t2
+    assert t1 == t3
+    assert t1 != t4
+    assert t1 != t5
+    assert t4 == t5
+    assert t6 != t7
```
Shaikh-Ubaid, line 80: We are comparing floating point values here. I think we need to use range comparison `abs(a - b) < 1e-5` here.

PR #2320, opened 2023-09-11
Further Refactoring for the ASR Symbolic Pass

anutosh491:
There were some repetetions while declaring functions like `basic_str` (repeated 3 times), `integer_set_si` (repeated 2 times). After this PR, there are no duplications while declaring any function in the module scope and every function is only declared once.


certik (approved):
I think that this is good, thanks!

PR #2323, opened 2023-09-13
[small] tidy up warnings

khushi-411:
Hi, the PR removes warnings during build time and tests. Please see:
```cpp
/home/khushi/Documents/lpython/src/libasr/runtime/lfortran_intrinsics.c: In function ‘_lfortran_read_int64’:
/home/khushi/Documents/lpython/src/libasr/runtime/lfortran_intrinsics.c: In function ‘_lfortran_read_int64’:
/home/khushi/Documents/lpython/src/libasr/runtime/lfortran_intrinsics.c:1994:19: warning: format ‘%lld’ expects argument of type ‘long long int *’, but argument 2 has type ‘int64_t *’ {aka ‘long int *’} [-Wformat=]
 1994 |         scanf("%lld", p);
      |                ~~~^   ~
      |                   |   |
      |                   |   int64_t * {aka long int *}
      |                   long long int *
      |                %ld
/home/khushi/Documents/lpython/src/libasr/runtime/lfortran_intrinsics.c:1994:19: warning: format ‘%lld’ expects argument of type ‘long long int *’, but argument 2 has type ‘int64_t *’ {aka ‘long int *’} [-Wformat=]
 1994 |         scanf("%lld", p);
      |                ~~~^   ~
      |                   |   |
      |                   |   int64_t * {aka long int *}
      |                   long long int *
      |                %ld
/home/khushi/Documents/lpython/src/libasr/runtime/lfortran_intrinsics.c:2008:27: warning: format ‘%lld’ expects argument of type ‘long long int *’, but argument 3 has type ‘int64_t *’ {aka ‘long int *’} [-Wformat=]
 2008 |         fscanf(filep, "%lld", p);
      |                        ~~~^   ~
      |                           |   |
      |                           |   int64_t * {aka long int *}
      |                           long long int *
      |                        %ld
/home/khushi/Documents/lpython/src/libasr/runtime/lfortran_intrinsics.c:2008:27: warning: format ‘%lld’ expects argument of type ‘long long int *’, but argument 3 has type ‘int64_t *’ {aka ‘long int *’} [-Wformat=]
 2008 |         fscanf(filep, "%lld", p);
      |                        ~~~^   ~
      |                           |   |
      |                           |   int64_t * {aka long int *}
      |                           long long int *
      |                        %ld
```
and
```python
warning: The symbol 'f' imported from modules_02b will shadow the existing symbol 'f'
 --> /home/khushi/Documents/lpython/integration_tests/modules_02.py:1:25
  |
1 | from modules_02b import f, f
  |                         ^ old symbol
  |
1 | from modules_02b import f, f
  |                            ^ new symbol
```
cc @certik @czgdp1807

### DIFF
--- a/integration_tests/modules_02.py
+++ b/integration_tests/modules_02.py
@@ -1,4 +1,4 @@
-from modules_02b import f, f
+from modules_02b import f
 from lpython import i32
 
 def main0():
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -1991,7 +1991,7 @@ LFORTRAN_API void _lfortran_read_int64(int64_t *p, int32_t unit_num)
 {
     if (unit_num == -1) {
         // Read from stdin
-        scanf("%lld", p);
+        scanf("%ld", p);
         return;
     }
 
@@ -2005,7 +2005,7 @@ LFORTRAN_API void _lfortran_read_int64(int64_t *p, int32_t unit_num)
     if (unit_file_bin) {
         fread(p, sizeof(*p), 1, filep);
     } else {
-        fscanf(filep, "%lld", p);
+        fscanf(filep, "%ld", p);
     }
 }
 
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -1965,7 +1965,7 @@ LFORTRAN_API void _lfortran_rewind(int32_t unit_num)
 }
 
 
-LFORTRAN_API void _lfortran_read_int32(int32_t *p, int32_t unit_num)
+LFORTRAN_API void _lfortran_read_int32([[maybe_unused]] int32_t *p, int32_t unit_num)
 {
     if (unit_num == -1) {
         // Read from stdin
@@ -1987,7 +1987,7 @@ LFORTRAN_API void _lfortran_read_int32(int32_t *p, int32_t unit_num)
     }
 }
 
-LFORTRAN_API void _lfortran_read_int64(int64_t *p, int32_t unit_num)
+LFORTRAN_API void _lfortran_read_int64([[maybe_unused]] int64_t *p, int32_t unit_num)
 {
     if (unit_num == -1) {
         // Read from stdin
@@ -2021,7 +2021,7 @@ LFORTRAN_API void _lfortran_read_int64(int64_t *p, int32_t unit_num)
     }
 }
 
-LFORTRAN_API void _lfortran_read_array_int8(int8_t *p, int array_size, int32_t unit_num)
+LFORTRAN_API void _lfortran_read_array_int8([[maybe_unused]] int8_t *p, int array_size, int32_t unit_num)
 {
     if (unit_num == -1) {
         // Read from stdin
@@ -2047,7 +2047,7 @@ LFORTRAN_API void _lfortran_read_array_int8(int8_t *p, int array_size, int32_t u
     }
 }
 
-LFORTRAN_API void _lfortran_read_array_int32(int32_t *p, int array_size, int32_t unit_num)
+LFORTRAN_API void _lfortran_read_array_int32([[maybe_unused]] int32_t *p, int array_size, int32_t unit_num)
 {
     if (unit_num == -1) {
         // Read from stdin
@@ -2073,7 +2073,7 @@ LFORTRAN_API void _lfortran_read_array_int32(int32_t *p, int array_size, int32_t
     }
 }
 
-LFORTRAN_API void _lfortran_read_char(char **p, int32_t unit_num)
+LFORTRAN_API void _lfortran_read_char([[maybe_unused]] char **p, int32_t unit_num)
 {
     if (unit_num == -1) {
         // Read from stdin
@@ -2098,7 +2098,7 @@ LFORTRAN_API void _lfortran_read_char(char **p, int32_t unit_num)
     }
 }
 
-LFORTRAN_API void _lfortran_read_float(float *p, int32_t unit_num)
+LFORTRAN_API void _lfortran_read_float([[maybe_unused]] float *p, int32_t unit_num)
 {
     if (unit_num == -1) {
         // Read from stdin
@@ -2120,7 +2120,7 @@ LFORTRAN_API void _lfortran_read_float(float *p, int32_t unit_num)
     }
 }
 
-LFORTRAN_API void _lfortran_read_array_float(float *p, int array_size, int32_t unit_num)
+LFORTRAN_API void _lfortran_read_array_float([[maybe_unused]] float *p, int array_size, int32_t unit_num)
 {
     if (unit_num == -1) {
         // Read from stdin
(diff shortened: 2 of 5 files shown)


Shaikh-Ubaid:
Please enable `-Werror` for this file.

Also I think, now this shows warnings in macos. For example, from the build step in https://github.com/lcompilers/lpython/actions/runs/6168547819/job/16741232055?pr=2323:
```bash
/Users/runner/work/lpython/lpython/lpython-0.19.0-220-g8e845514a/src/libasr/runtime/lfortran_intrinsics.c:1994:22: warning: format specifies type 'long *' but the argument has type 'int64_t *' (aka 'long long *') [-Wformat]
        scanf("%ld", p);
               ~~~   ^
               %lld
/Users/runner/work/lpython/lpython/lpython-0.19.0-220-g8e845514a/src/libasr/runtime/lfortran_intrinsics.c:2008:30: warning: format specifies type 'long *' but the argument has type 'int64_t *' (aka 'long long *') [-Wformat]
        fscanf(filep, "%ld", p);
                       ~~~   ^
                       %lld
/Users/runner/work/lpython/lpython/lpython-0.19.0-220-g8e845514a/src/libasr/runtime/lfortran_intrinsics.c:1994:22: warning: format specifies type 'long *' but the argument has type 'int64_t *' (aka 'long long *') [-Wformat]
        scanf("%ld", p);
               ~~~   ^
               %lld
/Users/runner/work/lpython/lpython/lpython-0.19.0-220-g8e845514a/src/libasr/runtime/lfortran_intrinsics.c:2008:30: warning: format specifies type 'long *' but the argument has type 'int64_t *' (aka 'long long *') [-Wformat]
        fscanf(filep, "%ld", p);
                       ~~~   ^
                       %lld
```

khushi-411:
Ah, I see. You are right; there are warning messages in macos but not in ubuntu. I think it's related to the compilers.

Just to confirm this, I checked my previous PR; there were warning messages in ubuntu but not in macos. Please see: https://github.com/lcompilers/lpython/actions/runs/6164474892/job/16730254578

I also checked on the Godbolt and found that `gcc` and `clang` work pretty well, but there are warnings in `msvc`. Link: https://godbolt.org/z/ba3vdTxeh.
.
I used `PRId64` in my latest commit (https://github.com/lcompilers/lpython/pull/2323/commits/de9bd28270e7048a574166fb5852d2a789f8eab7); let's see if all the build system are happy with it or not :-)

I mainly referred to this hack from [stack-overflow](https://stackoverflow.com/a/6299313/14473589).

Thanks!

Shaikh-Ubaid:
Please also enable `-Werror` check for this file. I think that would ensure checking for warnings at the CI and thus help warning free builds in future.

certik:
The runtime library must be build with `-Werror` and we need to ensure it fails on both macOS and Linux at the CI. That's a separate task.

Then we need to fix it. Yes, this `%lld` gives warnings on either macOS or Linux. We faced this before and I think I put some ifdefs, and use `%lld` on one, but `%ld` on the other, or something like that.

Thirumalai-Shaktivel:
See: https://github.com/lcompilers/lpython/blob/df414fb45d022917385e31ffe8a0309a43c32e04/src/libasr/runtime/lfortran_intrinsics.c#L2598-L2602

khushi-411:
Interesting! Thanks for sharing. I have updated it using Macros.

Thirumalai-Shaktivel:
I think we can try `if (scanf("%d", p)) { } // Suppress unused result warning`
But let's wait for other's opinions as well.

Shaikh-Ubaid:
I think we can also disable the warning related to the unused result using the flag `-Wno-unused-result` only for the specific `lfortran_intrinsics.c` file.

certik:
We can, but I think those are good warnings, and we should only disable them when we know the result value was meant to be ignored with the `[[unused-*]]` syntax.

khushi-411:
> I think we can try if (scanf("%d", p)) { } // Suppress unused result warning
But let's wait for other's opinions as well.

> I think we can also disable the warning related to the unused result using the flag -Wno-unused-result only for the specific lfortran_intrinsics.c file.

Thanks, @Thirumalai-Shaktivel and @Shaikh-Ubaid, for sharing two other ways to disable the warning message.

> We can, but I think those are good warnings, and we should only disable them when we know the result value was meant to be ignored with the [[maybe_unused]] syntax. So I would keep the warning on.

Got it! Thanks, @certik, for clarifying it. It sounds totally reasonable to me.

This PR is ready by my side. Please let me know if there's anything to be improved in this PR.

certik (approved):
I think this looks good. We should probably use a different ifdef for apple (I am not sure if this define can be off sometimes for apple), but for now this is good.

certik:
Thanks!

Shaikh-Ubaid:
Thank you so much @khushi-411. Great work! I appreciate it.

Shaikh-Ubaid:
>I used PRId64 in my latest commit (https://github.com/lcompilers/lpython/commit/de9bd28270e7048a574166fb5852d2a789f8eab7); let's see if all the build system are happy with it or not :-)

The above seems to be a correct approach. Please see https://chat.openai.com/share/51fe5b13-86a1-4ebe-809d-a71d7d284cab.

Shaikh-Ubaid:
For the unused result warnings, it seems we also have the possibility/approach of casting the result to `void` (when we do not need/use the result). For example,
```C
$ cat examples/main.c 
#include <stdio.h>

__attribute__((warn_unused_result))
int custom_function(int value) {
    return value * 2;
}

int main() {
    custom_function(5); // This will trigger a warning for unused result
    return 0;
}
```
```bash
$ clang -Wunused-result -o unused_result examples/main.c
examples/main.c:9:5: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
    custom_function(5); // This will trigger a warning for unused result
    ^~~~~~~~~~~~~~~ ~
1 warning generated.
```
Casting the result to `void` like
```C
int main() {
    (void) custom_function(5);
    return 0;
}
```
```bash
$ clang -Wunused-result -o unused_result examples/main.c
$ 
```

khushi-411:
> It seems for scanf() it is SCNd64 and for printf() it is PRId64.

Ah, right. My bad, I missed it. Thanks for pointing it out! :-) 
I think it might have worked earlier but would have shown some miscellaneous effect while dtype conversion. Happy we haven't used that.


> PS: It seems to work on mac. We might need to check on Linux.

Nice approach!

I tested it on Ubuntu, and it appears clang is happy, but gcc still throws the same warning msg. Please see:
```cpp
(lp) khushi@khushi:~/Documents/lpython$ clang -Wunused-result -o result main.c
(lp) khushi@khushi:~/Documents/lpython$
(lp) khushi@khushi:~/Documents/lpython$ gcc -Wunused-result -o result main.c
main.c: In function ‘main’:
main.c:9:12: warning: ignoring return value of ‘custom_function’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
    9 |     (void) custom_function(5); // This will trigger a warning for unused result
      |            ^~~~~~~~~~~~~~~~~~
```

PR #2328, opened 2023-09-17
Setup online lpython

Shaikh-Ubaid:
Demo: https://dev.lpython.org/

### DIFF
--- a/src/bin/CMakeLists.txt
+++ b/src/bin/CMakeLists.txt
@@ -48,29 +48,21 @@ set_target_properties(lpython PROPERTIES
 )
 
 if (HAVE_BUILD_TO_WASM)
-    # set(WASM_LINK_FLAGS
-    #     "-g0" # Store no debugging information in the generated wasm file. This helps reduce generated file size
-    #     "-Oz" # Optimize for size. With this code size ~ 2.4mb. Without this code size ~49mb
-    #     "-fexceptions" # Enable Cpp exception support
-    #     "--no-entry" # No start function to execute
-    #     "-s ASSERTIONS" # Compile with Assertions which (as per docs) are helpful to debug compilation process
-    #     "-s ALLOW_MEMORY_GROWTH" # Allow dynamic memory growth upto the maximum page size limit
-    #     "-s WASM_BIGINT" # Allow use of i64 integers. ASR is needing this option to be enabled.
-    #     "-s EXPORTED_RUNTIME_METHODS=['cwrap']" # Export cwarp. cwarp helps us to call our EMSCRIPTEN_KEEPALIVE functions
-    # )
-
-    # Some extra flags below that we may need in future. But these may/might increase the code size
-    # "--preload-file ./asset_dir"
-    # "-s SAFE_HEAP=1"
-    # "-s \"EXPORTED_RUNTIME_METHODS=['ccall']\""
-    # "-s EXPORTED_FUNCTIONS=\"['_free', '_malloc']\""
+    # "-g0": Store no debugging information in the generated wasm file. This helps reduce generated file size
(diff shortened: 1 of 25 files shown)


src/bin/CMakeLists.txt
```diff
@@ -48,29 +48,24 @@ set_target_properties(lpython PROPERTIES
...
+    # "-s EXPORTED_RUNTIME_METHODS=['cwrap']": Export cwarp. cwarp helps us to call our EMSCRIPTEN_KEEPALIVE functions
+    # "-fsanitize=undefined": Clang's Undefined Behaviour Sanitizer. The LPython parser segfaults.
+        # This option is for debugging, but currently helps avoid the segfault in the parser.
+    # "-s INITIAL_MEMORY=536870912": Start the wasm linear memory with sufficiently large size 512Mb.
 
     # Notes:
     # STANDALONE_WASM is disabling support for exceptions, so it is currently omitted
     # In build_to_wasm.sh, we need CMAKE_CXX_FLAGS_DEBUG="-Wall -Wextra -fexceptions" flags for exception support
-    set(WASM_COMPILE_FLAGS "-g0 -fexceptions")
+    set(WASM_COMPILE_FLAGS "-g0 -fexceptions -fsanitize=undefined")
```
Shaikh-Ubaid, line 66: LPython's parser segfaults in many cases when LPython is build to wasm. It seems LPython uses a different parser as compared to LFortran. I think LPython uses LR(1) parser where as LFortran uses GLR parser.

The above `-fsanitize` flag is actually a debugging flag, but it seems to make LPython's parser to work.

src/libasr/ASR.asdl
```diff
@@ -164,6 +168,7 @@ abi                   -- External     ABI
     | GFortranModule  --   Yes        GFortran
     | BindC           --   Yes        C
     | BindPython      --   Yes        Python
+    | BindJS          --   Yes        Javascript
```
Shaikh-Ubaid, line 171: I added a new decorator for calls to Javascript functions. We previously used the (BindC + implementation=interface) for recognizing the functions to be imported. But it seems some functions from the numpy intrinsic module use these exact parameters (BindC + interface) and thus we are unable to distinguish between these numpy intrinsic functions and the functions that call into Javascript.

tests/reference/asr-elemental_01-b58df26.stderr
```diff
@@ -1,5 +0,0 @@
-semantic error: Arguments do not match for any generic procedure, sin
-  --> tests/../integration_tests/elemental_01.py:10:24
-   |
-10 |         assert abs(sin(sin(array[i])) - result[i]) <= eps
-   |                        ^^^^^^^^^^^^^
```
Shaikh-Ubaid, line 5: These few files got removed. I think these were left over from some previous PR.

Shaikh-Ubaid:
After we merge this, we need to update the commit version at https://dev.lpython.org/.

src/bin/CMakeLists.txt
certik: Yes, LPython uses LALR(1), which should be an easier parser.

src/libasr/ASR.asdl
certik: I think that's fine. The same should be done in LFortran.

certik (approved):
I think that this is fine, thanks!

src/bin/CMakeLists.txt
Shaikh-Ubaid: With a previous version of emscripten, the LPython parser would abort with error in the lines of "large memory ~3.5Gbs requested which is more than the available ~2.4Gbs". With the latest version of emscripten, it says "aborted, runtime error". (I updated my installed `emscripten` to be sure the errors are not due to some old version. )

PR #2330, opened 2023-09-20
[numpy] add fix

khushi-411:
Follows #959


certik (approved):
I think that this looks good. We can later refactor how we handle all these rounding functions in ASR, including names (I didn't know what "Fix" should do, had to look it up: [numpy.fix](https://numpy.org/doc/stable/reference/generated/numpy.fix.html)), but for now I think this works.

PR #2331, opened 2023-09-20
Added support for functions to accept symbolic variables

anutosh491:
Now that we have started porting sympy functions to LPython, I realized that there are some errors while trying to have symbolic parameters. Consider the following program
```
from sympy import pi, Symbol

def printsym(x: S):
    print(x)

def main0():
    a : S = Symbol("x")
    printsym(a)
```
If we look at the C code generated for this.

```
void printsym(void* x)
{
    _x = 0;
    x = NULL;
    x = (void*) &_x;
    basic_new_stack(x);
    printf("%s\n", basic_str(x));
    basic_free_stack(x);
}

void main0()
{
    int64_t _a;
    void* a;
    _a = 0;
    a = NULL;
    a = (void*) &_a;
    basic_new_stack(a);
    symbol_set(a, "x");
    printsym(a);
    basic_free_stack(a);
}
```
But this would end up giving a segmentation fault, we don't need to apply the transformations which we apply on local variables (`a` in this case) to the parameterized variable (`x` in this case). Something as simple as the following would do the job.
```
void printsym(void* x)
{
    printf("%s\n", basic_str(x));
}
```
Hence the changes introduced are the following 
1) Only apply the transformation if a local variable is being considered. Taken care through an if check 
2) The parameter `x: S` is being changed to `x: CPtr` to replicate the void pointer (`printsym(void* x)`)
3) I don't think the parameterized variables need to be freed, only the local ones should be hence that is being taken care of .For eg
```
def printsym(x: S):
   a: S = pi
   print(a)
   print(x)
```
Here `a` should be freed but freeing of `x` has been taken care in `main0`

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -711,6 +711,7 @@ RUN(NAME symbolics_05        LABELS cpython_sym c_sym llvm_sym NOFAST)
 RUN(NAME symbolics_06        LABELS cpython_sym c_sym llvm_sym NOFAST)
 RUN(NAME symbolics_07        LABELS cpython_sym c_sym llvm_sym NOFAST)
 RUN(NAME symbolics_08        LABELS cpython_sym c_sym llvm_sym)
+RUN(NAME symbolics_09        LABELS cpython_sym c_sym llvm_sym NOFAST)
 
 RUN(NAME sizeof_01           LABELS llvm c
         EXTRAFILES sizeof_01b.c)
--- a/integration_tests/symbolics_09.py
+++ b/integration_tests/symbolics_09.py
@@ -1,0 +1,17 @@
+from sympy import Symbol, pi, S
+from lpython import S, i32
+
+def addInteger(x: S, y: S, z: S, i: i32):
+    _i: S = S(i)
+    print(x + y + z + _i)
+
+def call_addInteger():
+    a: S = Symbol("x")
+    b: S = Symbol("y")
+    c: S = pi
+    addInteger(a, b, c, 2)
+
+def main0():
+    call_addInteger()
+
+main0()
--- a/src/libasr/pass/replace_symbolic.cpp
+++ b/src/libasr/pass/replace_symbolic.cpp
@@ -640,13 +640,16 @@ public:
                 ASR::expr_t* cast_value = cast_t->m_value;
                 if (ASR::is_a<ASR::Var_t>(*cast_arg)) {
                     ASR::symbol_t* integer_set_sym = declare_integer_set_si_function(al, x.base.base.loc, module_scope);
+                    ASR::ttype_t* cast_type = ASRUtils::TYPE(ASR::make_Integer_t(al, x.base.base.loc, 8));
+                    ASR::expr_t* value = ASRUtils::EXPR(ASR::make_Cast_t(al, x.base.base.loc, cast_arg,
+                        (ASR::cast_kindType)ASR::cast_kindType::IntegerToInteger, cast_type, nullptr));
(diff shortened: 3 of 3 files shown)


anutosh491:
An example for reference 
```
from sympy import pi, Symbol

def printsym(x: S, y: S):
    z: S = Symbol("z")
    print(x + y + z + pi)

def main0():
    a : S = Symbol("x")
    b : S = Symbol("y")
    printsym(a, b)

if __name__ == "__main__":
   main0()
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=llvm  examples/expr2.py 
x + y + z + pi
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=c  examples/expr2.py 
x + y + z + pi
```

certik:
Can you add a test for this?

anutosh491:
I think so maybe we would need a `visit_SubRoutineCall` function in our pass to handle something like this 
```
from sympy import pi, Symbol
from lpython import S

def printsym(x: S):
    print(x)

def main0():
    printsym(Symbol("x"))

main0()
```
Where we are passing `Symbol("x")` directly rather than storing it in a symbolic variable first and then moving on from there.
I'll try to get this soon and add tests for the same .

certik:
Just add a test for the feature you implemented in this PR, and later once you implement `printsym(Symbol("x"))` you can test it too.

anutosh491:
Yes, I think so I can implement the case for `printsym(Symbol("x"))` ( through a `visit_SubroutineCall` function) in the subsequent PR.

anutosh491:
Also I am not sure why do we get an error through the `llvm` backend. Everything looks correct to me as in I don't see any obvious error . Consider this simple program 
```
from sympy import pi, Symbol

def main0():
    i: i32 = 2
    _i: S = S(i)
    print(_i)
```
The C backend works perfectly 
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --show-c examples/expr2.py 

void main0()
{
    int64_t __i;
    void* _i;
    int32_t i;
    __i = 0;
    _i = NULL;
    _i = (void*) &__i;
    basic_new_stack(_i);
    i = 2;
    integer_set_si(_i, i);
    printf("%s\n", basic_str(_i));
    basic_free_stack(_i);
}

(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=c examples/expr2.py 
2
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --show-llvm examples/expr2.py 

declare void @_lpython_set_argv(i32, i8**)
  %3 = load i32, i32* %i, align 4
 i64  call void @integer_set_si(void* %2, i32 %3)
```
@Thirumalai-Shaktivel would you mind helping me out here.

Thirumalai-Shaktivel:
I think you have to cast the second argument from i32 to i64. As you can see, `declare` accepts i64.
```llvm
declare void @integer_set_si(void*, i64)
...
  %3 = load i32, i32* %i, align 4
  call void @integer_set_si(void* %2, i32 %3)
...
```
In C, I think there might be an implicit cast for i32 to i64, so it works.

anutosh491:
> I think you have to cast the second argument from i32 to i64. As you can see, declare accepts i64.

Oops, sorry I think so I overlooked this. Thanks for the help !

anutosh491:
I have a few doubts regarding few corners, hence converting to draft.

anutosh491:
I think this is ready now.

certik (approved):
I think that this is good. Thanks!

PR #2334, opened 2023-09-22
Dftatom: Changes from lfortran LibASR

Shaikh-Ubaid:
related https://github.com/lfortran/lfortran/pull/2488

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -698,8 +698,7 @@ static inline ASR::symbol_t *get_asr_owner(const ASR::expr_t *expr) {
             return ASRUtils::get_asr_owner(ASR::down_cast<ASR::FunctionCall_t>(expr)->m_name);
         }
         default: {
-            throw LCompilersException("Cannot find the ASR owner of underlying symbol of expression "
-                                        + std::to_string(expr->type));
+            return nullptr;
         }
     }
     return nullptr;
--- a/src/libasr/pass/pass_utils.cpp
+++ b/src/libasr/pass/pass_utils.cpp
@@ -560,7 +560,25 @@ namespace LCompilers {
 
         ASR::expr_t* get_bound(ASR::expr_t* arr_expr, int dim, std::string bound,
                                 Allocator& al) {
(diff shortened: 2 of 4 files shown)


src/libasr/pass/pass_utils.cpp
```diff
@@ -560,7 +560,25 @@ namespace LCompilers {
 
         ASR::expr_t* get_bound(ASR::expr_t* arr_expr, int dim, std::string bound,
                                 Allocator& al) {
+            ASR::ttype_t* x_mv_type = ASRUtils::expr_type(arr_expr);
+            ASR::dimension_t* m_dims;
+            int n_dims = ASRUtils::extract_dimensions_from_ttype(x_mv_type, m_dims);
+            bool is_data_only_array = ASRUtils::is_fixed_size_array(m_dims, n_dims) && ASRUtils::get_asr_owner(arr_expr) &&
+                                    ASR::is_a<ASR::StructType_t>(*ASRUtils::get_asr_owner(arr_expr));
```
certik, line 567: Does "data only array" in this context mean "FixedSizeArray" inside a struct?

src/libasr/asr_utils.h
```diff
@@ -698,8 +698,7 @@ static inline ASR::symbol_t *get_asr_owner(const ASR::expr_t *expr) {
             return ASRUtils::get_asr_owner(ASR::down_cast<ASR::FunctionCall_t>(expr)->m_name);
         }
         default: {
-            throw LCompilersException("Cannot find the ASR owner of underlying symbol of expression "
-                                        + std::to_string(expr->type));
+            return nullptr;
```
certik, line 701: Note: this changes the API, now it will return `nullptr` if the node does not have an ASR owner.

certik (approved):

src/libasr/pass/pass_utils.cpp
Shaikh-Ubaid: Yes.

certik: Ok. I would maybe document this in a comment, or maybe name the variable `fixed_array_inside_struct`, or something like that.

PR #2337, opened 2023-09-25
Adding support for executing attribute/query calls without assigning to a prior variable

anutosh491:
I realize that we couldn't access attributes or query methods without assigning the expression to a variable beforehand.
So this works 
```
from sympy import pi, Symbol, sin, cos

def main0():
    x: S = Symbol("x")
    a: S = sin(x)
    print(a.diff(x))

main0()
```
But these cases don't 
```
from sympy import pi, Symbol, sin, cos

def main0():
    x: S = Symbol("x")
    print(sin(x).diff(x))
    print(sin(x).diff(x).diff(x))

main0()
```

### DIFF
--- a/integration_tests/symbolics_05.py
+++ b/integration_tests/symbolics_05.py
@@ -1,4 +1,4 @@
-from sympy import Symbol, expand, diff
+from sympy import Symbol, expand, diff, sin, cos, exp, pi
 from lpython import S
 
 def test_operations():
@@ -20,5 +20,17 @@ def test_operations():
     assert(diff(b, x) == S(3)*(x + y + z)**S(2))
     print(a.diff(x))
     print(diff(b, x))
+
+    # test diff 2
+    c:S = sin(x)
+    d:S = cos(x)
+    assert(sin(Symbol("x")).diff(x) == d)
+    assert(sin(x).diff(Symbol("x")) == d)
+    assert(sin(x).diff(x) == d)
+    assert(sin(x).diff(x).diff(x) == S(-1)*c)
+    assert(sin(x).expand().diff(x).diff(x) == S(-1)*c)
+    assert((sin(x) + cos(x)).diff(x) == S(-1)*c + d)
+    assert((sin(x) + cos(x) + exp(x) + pi).diff(x).expand().diff(x) == exp(x) + S(-1)*c + S(-1)*d)
+
 
 test_operations()
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -7241,6 +7241,36 @@ public:
             std::string res = n->m_value;
             handle_constant_string_attributes(res, args, at->m_attr, loc);
             return;
+        } else if (AST::is_a<AST::Call_t>(*at->m_value)) {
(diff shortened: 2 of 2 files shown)


anutosh491:
Converting to draft as few corner cases need to be checked and the `symbolics_05.py` test file needs to be updated with these cases.

anutosh491:
The has attribute can be added to the set of attributes once #2336 is merged. We can first merge this and then rebase that pr on top of this I think !

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -7220,6 +7241,32 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             ASR::expr_t* expr = ASR::down_cast<ASR::expr_t>(tmp);
             handle_builtin_attribute(expr, at->m_attr, loc, eles);
             return;
+        } else if (AST::is_a<AST::BinOp_t>(*at->m_value)) {
+            AST::BinOp_t* bop = AST::down_cast<AST::BinOp_t>(at->m_value);
+            std::set<std::string> symbolic_attributes = {
```
Thirumalai-Shaktivel, line 7246: We can declare this as a Global variable, but it's fine for now.

integration_tests/symbolics_05.py
```diff
@@ -21,4 +21,16 @@ def test_operations():
...
+    # test diff 2
+    c:S = sin(x)
+    d:S = cos(x)
+    assert(sin(Symbol("x")).diff(x) == d)
+    assert(sin(x).diff(Symbol("x")) == d)
+    assert(sin(x).diff(x) == d)
+    assert(sin(x).diff(x).diff(x) == S(-1)*c)
+    assert(sin(x).expand().diff(x).diff(x) == S(-1)*c)
+    assert((sin(x) + cos(x)).diff(x) == S(-1)*c + d)
+    assert((sin(x) + cos(x) + exp(x) + pi).diff(x).expand().diff(x) == exp(x) + S(-1)*c + S(-1)*d)
```
Thirumalai-Shaktivel, line 33: Why don't we test this in the CI?

anutosh491: I didn't catch you. Don't we execute `./run_tests.py -b=llvm_sym` and `./run_tests.py -b=c_sym` in `LPython CI / Run SymPy tests` ? I think so that checks this.

Thirumalai-Shaktivel: Okay, cool.

certik:
@czgdp1807, @Thirumalai-Shaktivel, @Shaikh-Ubaid any ideas how to best support this? We want to support custom SymPy methods, but do not want to pollute the frontend at the wrong places. I think this is similar to custom NumPy methods that we support (or will support).

Thirumalai-Shaktivel:
As Anutosh said, the constant like pi will be treated as variable and it is not declared in current scope, the same goes for function call as well.
So for that, I think:
1. We can add a new node (let's say SymbolicItems or SymbolicObjects) and handle this in the Symbolic pass.
2. We can add more symbols like SymbolicFunction, SymbolicConstant, SymbolicBinOp.

certik (approved):
I don't know how to improve it. We need this feature, so let's merge this for now. We might need to improve it later, once we know how.

Issue #946, opened 2022-08-09
Remove `const` from all visitors


certik:
Most visitors need to modify the ASR, and duplicating them just to be able to modify ASR will add more code, slower compilation, etc.

I think the AST->ASR does not modify AST, so having that visitor const is good to make sure we don't modify it accidentally.

The ASR->LLVM also does not modify ASR, so having that visitor `const` is also good, for code safety.

But almost all ASR passes modify ASR in place, so all those base visitors should not be `const`.

xmnlab:
hey @certik if you don't mind, I would love to contribute to this issue.
as it would be my first contribution to the project maybe it could take some time. 
do you have any deadline in mind for this issue?

Thirumalai-Shaktivel:
Thanks for your interest in contributing to LPython. We appreciate it! 
Go ahead, make the required changes, and submit the PR. 
If you have any doubts, please ask.

xmnlab:
thank you so much! Probably I will have questions about that :)
I started to take a look into that ... and tomorrow I have a meeting with @czgdp1807 
and I will post here any question I have :) 

thank you so much

khushi-411:
Hi,
@xmnlab, I wonder if you are still planning to work on this issue. If not, I'd like to pick this. :-)
I am looking forward to hearing from you. Thank you!

xmnlab:
Hi @khushi-411 !
Sorry, I've been very busy and I didn't have time to move this forward.
please go ahead! Thanks

khushi-411:
Thanks for confirming, @xmnlab! Happy to work on it

PR #2343, opened 2023-09-27
[cpp] remove const asr->cpp & asr->c

khushi-411:
Follows: #946


certik:
I think it has to be passed in as `&`, not by value, for performance.

PR #2364, opened 2023-10-06
Throw Exception for wrong argtype in join method

Agent-Hellboy:


src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -6669,6 +6668,12 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                 throw SemanticError("str.join() takes one argument",
                     loc);
             }
+            ASR::expr_t *arg_sub = args[0].m_value;
+            ASR::ttype_t *arg_sub_type = ASRUtils::expr_type(arg_sub);
+            if(!ASR::is_a<ASR::List_t>(*arg_sub_type)){
+                throw SemanticError("str.join() takes type list only",
+                     loc);
+            }
```
Shaikh-Ubaid, line 6676: This change seems good. You can add an error reference test for this. Here is an example of how to add an error reference test https://github.com/lcompilers/lpython/commit/7c2bbac2e99c368383509de975a937ee813857ef. You just add a test file in `tests/errors`. In the test file, you pass a type apart from list to `str.join()`. You enable this test by adding it to `tests/tests.toml`. Then you just need to update the reference tests using `./run_tests.py -u`. This will add a new reference `.stderr` and `.json` files.

Agent-Hellboy: Actually, I am still not checking the type of element inside the list which should be str, I checked grammar for asr but didn't get anything on how to iterate over list expression as it must be a list of str expressions. ast is getting generated perfectly but, I don't know how can I map ast nodes to asr nodes.

Thirumalai-Shaktivel: Try updating refs using `./run_tests.py -u` and stage all and commit the updated refs.

Thirumalai-Shaktivel: After implementing and committing the required changes, you can do the following to test everything on your machine:
```
./build0.sh
./build1.sh
./run_tests.py # if this fails, check for the failure and do ./run_tests.py -u to update refs.
cd integration_tests && ./run_tests.py # Here you can do `./run_tests.py -b c` to test c backend, use `-h` option to view all the available options
```
If everything works, then you are good to go.

tests/errors/string_01.py
```diff
@@ -2,4 +2,5 @@ def test_tuple_indicing_in_str():
     x: str = "abcdefgh"
     print(x[2, 5])
 
+
```
Shaikh-Ubaid, line 5:
```suggestion
```

Shaikh-Ubaid: @Agent-Hellboy please apply the above change and update the reference tests (`./run_tests.py -u`).

Shaikh-Ubaid: >Actually, I am still not checking the type of element inside the list which should be str.

Let's tackle this after the CI passes for the current changes.

Shaikh-Ubaid:
>please apply the above change and update the reference tests (./run_tests.py -u).

@Agent-Hellboy  After the change, please update the reference tests like I mentioned above. Ideally do the following:

```
rm -rf ./tests/reference/*
./run_tests.py -u

# now just commit all the reference test changes
```

tests/reference/asr-string_02-499c9ff.stdout
```diff
@@ -0,0 +1 @@
+(Var 3 p)
```
Shaikh-Ubaid, line 1: The CI tests seem to pass. I think we just need to remove this `stdout` output. For the `string_02` error test, as it results in an error, we would/should not have any `stdout` file generated for it.

You can simply remove this specific file and commit the changes.

Agent-Hellboy: okay, sure

integration_tests/test_str_01.py
```diff
@@ -59,6 +61,13 @@ def test_str_join2():
     res:str = a.join(p)
     assert res == "a**b"
 
+def test_str_join3():
+    a: str
+    a = "**-"
+    p:list[str] = ["a","b"]
+    res:str = a.join(p)
+    assert res == "a**-b"
```
Shaikh-Ubaid, line 69: We should probably add a new test for corner cases like when `a` here is an empty `string` or when `p` here is an empty `list`.

integration_tests/test_str_01.py
```diff
@@ -1,3 +1,5 @@
+from lpython import i32
```
Shaikh-Ubaid, line 1: @Agent-Hellboy Is the above import being used in the test case? If it is unused, I suggest we remove it.

integration_tests/test_str_01.py
Shaikh-Ubaid: This can be tackled in a separate/new PR.

Shaikh-Ubaid:
@Agent-Hellboy  I shared few comments above. This looks good to me overall.

integration_tests/test_str_01.py
Agent-Hellboy: I have added the test in the same PR.

integration_tests/test_str_01.py
Agent-Hellboy: Done

Shaikh-Ubaid (approved):
Amazing work @Agent-Hellboy ! Thank you so much for this! This looks good to me.

Issue #2332, opened 2023-09-20
Symbolics TODO


certik:
* [ ] get the mrv() function finished that doesn't need the SubsSet (see my `gruntz2` branch)
* [ ] Implement missing ASR features (important: focus only on a clean ASR design, for now in the frontend support method calls, and maybe later we'll design a clean interface):
  * [ ] `e.has(x)`
  * [ ] `e.is_Mul`, `e.is_Add`, `e.is_Pow`
  * [ ] `isinstance(e, log)`, `isinstance(e, exp)`
  * [ ] `e.exp.args[0]`
  * [ ] `exponent = int(e.exp)`
  * [ ] Set of symbolic expressions (`{exp(x)}`)

anutosh491:
Well if we plan on using `basic_get_args` for implemnting `e.exp` and `e.base` (instead of what I was trying to do in https://github.com/symengine/symengine/pull/1980), we would need a work around for the `CVecBasic` type just like we had one for the `basic` type.
If you remember, we introduced a workaround for declaring a basic variable as follows 
```
    _x: i64 = i64(0)
    x: CPtr = empty_c_void_p()
    p_c_pointer(pointer(_x, i64), x)
```

Now what `basic_get_args` does is, it takes in a pointer to `CVecBasic` and then it populates the value of `args` with the arguments.
```
//! Returns a CVecBasic of vec_basic given by get_args
CWRAPPER_OUTPUT_TYPE basic_get_args(const basic self, CVecBasic *args);
```
Hence we might need to think about a workaround for implementing `CVecBasic` in LPython.

certik:
You can also write an auxiliary C function that does this, if it simplifies things.

certik:
Regarding the `args`, let's first implement the following:
```
args: tuple[S] = (x**2).args
print(args[0], args[1])
```
Where we obtain the whole vector of arguments from SymEngine, and we convert it to a Python tuple. Then we just deal with a tuple. For that, we should first get tuples of type `S` working in LPython, that is, get this working first:
```
args: tuple[S] = (x, S(2))
print(args[0], args[1])
```
And then once that works, let's get `args: tuple[S] = (x**2).args` working. I think after that, the following will also automatically start working:
```
print((x**2).args[0], (x**2).args[1])
```

certik:
[...]

PR #2367, opened 2023-10-06
Added Support for Symbolic `func` attribute

anutosh491:
This Pr tries to add support for the func attribute. I've used a transform the following python program into this respective C code
```
def main0():
    x: S = pi
    y: S = Symbol("y")
    z: S = y + x
    y = y**x
    print(y.func)
```
```
def main0():
    _x: i64 = i64(0)
    x: CPtr = empty_c_void_p()
    basic_new_stack(x)
    basic_const_pi(x)
    _y: i64 = i64(0)
    y: CPtr = empty_c_void_p()
    basic_new_stack(y)
    symbol_set(y, "y")
    basic_pow(y, y, x)
    temp_int : i32 = basic_get_type(y)
    temp_str : str = ""
    if temp_int == 17:
        temp_str = "Pow"
    if temp_int == 15:
        temp_str = "Mul"
    if temp_int == 16:
        temp_str = "Add"
    print(temp_str)

main0()
```
So essentially we do the following 
1) Intorduce `temp_str`, `temp_int` variables and introduce the respective symbols in the symbol table
2) Frame if statements where we compare `temp_int` against integers and update `temp_str` accordingly

### DIFF
--- a/integration_tests/symbolics_02.py
+++ b/integration_tests/symbolics_02.py
@@ -1,35 +1,51 @@
-from sympy import Symbol, pi
+from sympy import Symbol, pi, Add, Mul, Pow
 from lpython import S
 
 def test_symbolic_operations():
     x: S = Symbol('x')
     y: S = Symbol('y')
-    p1: S = pi
-    p2: S = pi
+    pi1: S = pi
+    pi2: S = pi
     
     # Addition
     z: S = x + y
+    z1: bool = z.func == Add
+    z2: bool = z.func == Mul
     assert(z == x + y)
+    assert(z1 == True)
+    assert(z2 == False)
     print(z)
     
     # Subtraction
     w: S = x - y
+    w1: bool = w.func == Add
     assert(w == x - y)
+    assert(w1 == True)
     print(w)
     
     # Multiplication
     u: S = x * y
+    u1: bool = u.func == Mul
     assert(u == x * y)
+    assert(u1 == True)
     print(u)
 
     # Division
     v: S = x / y
+    v1: bool = v.func == Mul
     assert(v == x / y)
+    assert(v1 == True)
     print(v)
 
     # Power
     p: S = x ** y
+    p1: bool = p.func == Pow
+    p2: bool = p.func == Add
+    p3: bool = p.func == Mul
     assert(p == x ** y)
+    assert(p1 == True)
+    assert(p2 == False)
+    assert(p3 == False)
     print(p)
 
     # Casting
@@ -40,13 +56,13 @@ def test_symbolic_operations():
     print(c)
 
     # Comparison
-    b1: bool = p1 == p2
+    b1: bool = pi1 == pi2
     print(b1)
     assert(b1 == True)
-    b2: bool = p1 != pi
+    b2: bool = pi1 != pi
     print(b2)
     assert(b2 == False)
-    b3: bool = p1 != x
+    b3: bool = pi1 != x
     print(b3)
     assert(b3 == True)
     b4: bool = pi == Symbol("x")
--- a/integration_tests/symbolics_02.py
+++ b/integration_tests/symbolics_02.py
@@ -14,6 +14,10 @@ def test_symbolic_operations():
     assert(z == x + y)
     assert(z1 == True)
     assert(z2 == False)
+    if z.func == Add:
+        assert True
+    else:
+        assert False
     print(z)
     
     # Subtraction
@@ -21,6 +25,10 @@ def test_symbolic_operations():
     w1: bool = w.func == Add
     assert(w == x - y)
     assert(w1 == True)
+    if w.func == Add:
+        assert True
+    else:
+        assert False
     print(w)
     
     # Multiplication
@@ -28,6 +36,10 @@ def test_symbolic_operations():
     u1: bool = u.func == Mul
     assert(u == x * y)
     assert(u1 == True)
+    if u.func == Mul:
+        assert True
+    else:
+        assert False
     print(u)
 
     # Division
@@ -35,6 +47,10 @@ def test_symbolic_operations():
     v1: bool = v.func == Mul
     assert(v == x / y)
     assert(v1 == True)
+    if v.func == Mul:
+        assert True
+    else:
+        assert False
     print(v)
 
     # Power
@@ -46,6 +62,10 @@ def test_symbolic_operations():
     assert(p1 == True)
     assert(p2 == False)
     assert(p3 == False)
+    if p.func == Pow:
+        assert True
+    else:
+        assert False
     print(p)
 
     # Casting
--- a/src/libasr/pass/intrinsic_function_registry.h
+++ b/src/libasr/pass/intrinsic_function_registry.h
@@ -77,7 +77,9 @@ enum class IntrinsicScalarFunctions : int64_t {
     SymbolicExp,
     SymbolicAbs,
     SymbolicHasSymbolQ,
-    SymbolicFuncQ,
+    SymbolicAddQ,
+    SymbolicMulQ,
+    SymbolicPowQ,
     // ...
 };
 
@@ -138,7 +140,9 @@ inline std::string get_intrinsic_name(int x) {
         INTRINSIC_NAME_CASE(SymbolicExp)
         INTRINSIC_NAME_CASE(SymbolicAbs)
         INTRINSIC_NAME_CASE(SymbolicHasSymbolQ)
-        INTRINSIC_NAME_CASE(SymbolicFuncQ)
+        INTRINSIC_NAME_CASE(SymbolicAddQ)
+        INTRINSIC_NAME_CASE(SymbolicMulQ)
+        INTRINSIC_NAME_CASE(SymbolicPowQ)
         default : {
             throw LCompilersException("pickle: intrinsic_id not implemented");
         }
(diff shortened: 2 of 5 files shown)


anutosh491:
Hence something like the following works now 
```
from lpython import S
from sympy import Symbol, pi

def main0():
    x: S = pi
    y: S = Symbol("y")
    z: S = y + x
    y = y**x
    print(y.func)
    print(z.func)
    assert (y.func == "Pow")
    assert (z.func == "Add")

if __name__ == "__main__":
   main0()
Pow
Add
```

anutosh491:
The implementation is a surface level implementation right now, we can surely improve or optimize it here and there. 
cc @Thirumalai-Shaktivel if you see anything where I could improve upon or maybe a different implementation idea as a whole to address the func attribute you could let me know.

certik:
Instead of:
```
    temp_str : str = ""
    if temp_int == 17:
        temp_str = "Pow"
    if temp_int == 15:
        temp_str = "Mul"
    if temp_int == 16:
        temp_str = "Add"
```
Let's send a patch to SymEngine to add a function `symengine_type_to_str(17) -> "Pow"`, then you can just call it here.

certik:
I think the `.func` attribute can't be of a type string. Rather, it must be able to compare the `Pow` and `Add` symbols, and we can assume that we know the actual value of the RHS at compile time, so that we can always directly translate it to our ASR.

anutosh491:
Yeah returning a string was hackish here and not intended.
If only we had subtypes for in the ASR for SymbolicExpression (like `SymbolicPowExpression` etc) to further classify a `SymbolicExpression` it would have been great.
```
ttype
    = Integer(int kind)
    ....
    | SymbolicExpression()
    | FunctionType(ttype* arg_types, ttype? return_var_type,
        abi abi, deftype deftype, string? bindc_name, bool elemental,
        bool pure, bool module, bool inline, bool static,
        symbol* restrictions, bool is_restriction)
```
What are your thoughts here ? Which `ttype` should the `func` attribute return ?

anutosh491:
> Let's send a patch to SymEngine to add a function symengine_type_to_str(17) -> "Pow", then you can just call it here.

I think so we already have `basic_get_class_from_id`(https://github.com/symengine/symengine/blob/2b575b9be9bb499d866dc3e411e6368ca0d1bb42/symengine/cwrapper.h#L155) which can take in an Integer and return the class (as a string). Let me try using this !

anutosh491:
I've made the implementation much easier through the latest commit.
Now something like 
```
z: str = y.func
print(z)
```
Is transformed into the following through the pass
```    
z: str = basic_get_class_from_id(basic_get_type(y))
print(z)
```
The only question that remains is what should be the return type of the `func` attribute. I'm a bit confused here. Once I have some clarity on this, I would finish work on this !

certik:
I can see the following ways forward:

1. Adding types to ASR for symbolic expressions, such as `SymbolicPowExpression`. We might end up doing that, but I think it's a little bit early for this, I would do this later once we gain more experience. Adding types I think is a big change since now every place that accepts a type must learn how to handle the new symbolic types. First I would do some more minimal solution, and iterate. Once we actually use this in limits or other bigger codes, we will gain more experience how to design this well.
2. Add `SymbolicPowQ`, `SymbolicLogQ`, etc. nodes, that accept an expression and return true/false. Internally you implement them using SymEngine, one way or the other. I don't see any problem here, except that there might be a long list of these nodes.
3. Add `SymbolicIsAQ` node that accepts an expression and a SymEngine Type ID integer
4. Make `func` return an integer, the SymEngine Type ID integer.

Related issue: https://github.com/lcompilers/lpython/issues/2332

This requires a proper design, but for now let's get something going that unblocks us, and then we'll iteratively improve things. I suspect the fundamental issue here is that we are querying the type at runtime, so it's a runtime type. While in LPython so far all types are compile time.

I like the approach 3. the most. You can think of all those ASR IntrinsicFunctions like `SymbolicPowQ` as "runtime types", without introducing any integer ID, so this design does not close any door to a more sophisticated design. At the frontend side, you can easily map `(x**2).func == Pow` to `SymbolicPowQ(x**2)` in ASR, and in the backend, just implement `SymbolicPowQ` like you already did (I believe) by hardwiring the SymEngine ID number (for now), and later we'll initialize the Type ID numbers at the program start (at runtime), or even require to link against exactly the same symengine that was available at compile time, in which case we can simply hardwire the numbers for the given symengine version.

Do you see any problem with this approach?

anutosh491:
> I like the approach 3.

You mean approach 2 where we implementing `SymbolicPowQ` etc ?

anutosh491:
I'll just write down my understanding here so that anyone can correct me if I haven't understood things correctly.
Well if I go through the `def mrv` functions in `gruntz.py`, I see that there are type checks and some other queries like the following
```
1) e.is_Pow, e.is_Mul, e.is_Add
2) isinstance(e, log), isinstance(e, exp)
3) others such as e.has(x), e.base == S.Exp1, e.is_infinite
```
This makes me realize that we only need to focus on the query aspects of things here ( only having a bool output either `True` or `False`)
So essentially we don't need to think about cases like assigning or printing right now correct ?
```

**Comment by certik** [comment]
> You mean approach 2 (did you type 3 by mistake ?) where we implement `SymbolicPowQ` etc ?

Yes, I fixed it. I like approach 2 the most.

**Comment by certik** [comment]
Yes, what you wrote is correct. In the end, did you mean `SymbolicPowQ`? Here is how I would tackle each of the cases you wrote:
```
1) e.is_Pow, e.is_Mul, e.is_Add
```
I would use `SymbolicPowQ`, `SymbolicMulQ`, `SymbolicAddQ`.
```
2) isinstance(e, log), isinstance(e, exp)
```
I would use `SymbolicLogQ`, `SymbolicExpQ`.
```
3) others such as e.has(x)
```
I would use `SymbolicHasSymbolQ` (already implemented)
```
3) e.base == S.Exp1
```
I would use `SymbolicExp1Q(SymbolicArgs(e, 0))` (and 1 for "e.exp"). Here you can start with `SymbolicExp1Q`. Normally I would name it just `SymbolicEQ` but that looks like some kind of an "equation" or "equality", so better to use `SymbolicExp1Q` for now.
```
3) e.is_infinite
```
I would use `SymbolicInfinityQ`.

**Comment by Thirumalai-Shaktivel** [comment]
> 2. Add SymbolicPowQ, SymbolicLogQ, etc. nodes,

You meant to add in IntrinsicScalarFunction, right?
If so, then should we move symbolics items from IntrinisicScalarFunction to dedicated enum, say: Symbolics?

**Comment by certik** [comment]
> You meant to add in `IntrinsicScalarFunction`, right?
> If so, then should we move `symbolics items` from `IntrinisicScalarFunction` to dedicated enum, say: `Symbolics`?

Yes, those `SymbolicPowQ` etc. are part of `IntrinsicScalarFunction`. For now I would keep them there. Once we can compile the full `limit.py` and have more experience, we can figure out if a better design can be done. Right now I think it is not clear which approach is better, so we go with the "minimal design", which this one is, as it doesn't require to modify ASR at all.

**Comment by certik** [comment]
There are several ways to tackle it, but right now we have not figured out the most cleanest way.

To move forward, let's implement it like this:

* In AST->ASR, in `visit_Compare(AST::Compare_t ...)` we add a line of the kind: "is the LHS a `.func` called on a symbolic type"?
* If true: check that the right hand side is exactly "Pow" (and other types) and return `SymbolicPowQ`, otherwise give a nice error message
* If false: handle the compare node as usual

**Comment by anutosh491** [comment]
> You descent all the way to .func, and you set some internal flag (as well as some internal type) that the whole Compare node must be replaced

I've tried addressing the func attribute through this method in the latest commit. I think using an internal flag would be a suitable approach here. So now we can do something like the following
```

def main0():
    x: S = pi
    y: S = Symbol("y")
    y = y + x
    z1: bool = y.func == Pow
    z2: bool = y.func == Add
    print(z1)
    print(z2)

False
True
```
And for something that is yet to be implemented 
```
def main0():
    x: S = pi
    y: S = Symbol("y")
    y = y + x
    z1: bool = y.func == Log
    print(z1)
    
semantic error: Log symbolic type not supported yet
 --> examples/expr2.py:8:16
  |
8 |     z1: bool = y.func == Log
  |                ^^^^^^^^^^^^^ 
 ```

anutosh491:
The latest commit also adds some simple tests for the same in `symbolics_02.py`

integration_tests/symbolics_02.py
```diff
@@ -1,35 +1,51 @@
...
+    pi1: S = pi
+    pi2: S = pi
     
     # Addition
     z: S = x + y
+    z1: bool = z.func == Add
+    z2: bool = z.func == Mul
     assert(z == x + y)
+    assert(z1 == True)
+    assert(z2 == False)
```
certik, line 16: In addition to this, let's also add tests like this:
```python
if z.func == Add:
    assert True
else:
    assert False
```
as well as:
```python
assert z.func == Add
```
To ensure that the `z.func == Add` idiom works in all the situations:
* assigning to a bool variable (already tested)
* in `if`
* in `assert`

src/lpython/semantics/python_attribute_eval.h
```diff
@@ -46,7 +46,10 @@ struct AttributeHandler {
         symbolic_attribute_map = {
             {"diff", &eval_symbolic_diff},
             {"expand", &eval_symbolic_expand},
-            {"has", &eval_symbolic_has_symbol}
+            {"has", &eval_symbolic_has_symbol},
+            {"is_Add", &eval_symbolic_is_Add},
+            {"is_Mul", &eval_symbolic_is_Mul},
+            {"is_Pow", &eval_symbolic_is_Pow}
```
certik, line 52: I would not support `is_Add` at all.

certik:
Otherwise this is good.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -3412,6 +3472,9 @@ namespace IntrinsicScalarFunctionRegistry {
                 {"SymbolicExp", {&SymbolicExp::create_SymbolicExp, &SymbolicExp::eval_SymbolicExp}},
                 {"SymbolicAbs", {&SymbolicAbs::create_SymbolicAbs, &SymbolicAbs::eval_SymbolicAbs}},
                 {"has", {&SymbolicHasSymbolQ::create_SymbolicHasSymbolQ, &SymbolicHasSymbolQ::eval_SymbolicHasSymbolQ}},
+                {"is_Add", {&SymbolicAddQ::create_SymbolicAddQ, &SymbolicAddQ::eval_SymbolicAddQ}},
+                {"is_Mul", {&SymbolicMulQ::create_SymbolicMulQ, &SymbolicMulQ::eval_SymbolicMulQ}},
+                {"is_Pow", {&SymbolicPowQ::create_SymbolicPowQ, &SymbolicPowQ::eval_SymbolicPowQ}},
```
certik, line 3477: Remove these?

integration_tests/symbolics_02.py
certik: This should still be addressed.

src/libasr/pass/intrinsic_function_registry.h
anutosh491: I'll change the name of the string, I think we can relate better then !

certik: I see. Yes, change it, I think that will be clearer.

anutosh491:
I've addressed all changes and I think this is ready. We just need to remember than we are currently hardcoding values (TypeId's) for symbolic classes like `Pow (17)`, `Mul (15)` etc and we might want to replace these later using `basic_get_class_from_id` or something!

integration_tests/symbolics_02.py
certik: Can you add a test `assert z.func == Add` ?  I don't see it, unless I missed it.

certik (approved):
I think that this is fine. Do you want to merge as is, and send improvements in subsequent PRs?

anutosh491:
Sure . I will be addressing the `isinstance(e, Exp)` and `isinstance(e, Log)` cases in the next pr and I'll shall address anything that needs to added there.

integration_tests/symbolics_02.py
anutosh491: Will remember this and add it in a subsequent pr

PR #2372, opened 2023-10-07
Support floor division as intrinsic function

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2349


Shaikh-Ubaid:
Ready.

src/libasr/casting_utils.h
```diff
@@ -0,0 +1,21 @@
...
+    int get_type_priority(ASR::ttypeType type);
+
+    int get_src_dest(ASR::expr_t* left_expr, ASR::expr_t* right_expr,
+                      ASR::expr_t*& src_expr, ASR::expr_t*& dest_expr,
+                      ASR::ttype_t*& src_type, ASR::ttype_t*& dest_type,
+                      bool is_assign);
+
+    ASR::expr_t* perform_casting(ASR::expr_t* expr, ASR::ttype_t* src,
+                                 ASR::ttype_t* dest, Allocator& al,
+                                 const Location& loc);
```
certik, line 18: What is the idea behind these? I think we do not want to do any kind of automatic casting in ASR, nor be responsible for determining what proper casting must be done.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -1964,6 +1972,138 @@ namespace FlipSign {
...
+            r = float(a)/float(b)
+            tmp = i64(r)
+            if r < 0.0 and f64(tmp) != r:
+                tmp = tmp - 1
+            result = i32(tmp)
+            return result
+        */
+
+        ASR::expr_t *op1 = r64Div(CastingUtil::perform_casting(args[0], arg_types[0], real64, al, loc),
+            CastingUtil::perform_casting(args[1], arg_types[1], real64, al, loc));
```
certik, line 2092: I can see it is being used here. What kind of input arguments are allowed here? We need to be careful that we do not accept too many inputs, we should be explicit about exactly what types we accept in `_lpython_floordiv`.

Shaikh-Ubaid: We currently support `integers`, `unsigned integers`, `reals` and `logicals`.

certik:
In general the change is good I think, but I want to have good clarity on the casting situation. We should document this in comments, once I understand it.

src/libasr/casting_utils.h
Shaikh-Ubaid: >I think we do not want to do any kind of automatic casting in ASR

It is actually not for automatic casting. The floordiv intrinsic we defined in this PR is generalized so that it can work on any type (from the supported types of `int, unsigned int, real, logical`). The casting function is used to cast the intermediate temporary variable to the appropriate return type.

Shaikh-Ubaid:
>In general the change is good I think, but I want to have good clarity on the casting situation. We should document this in comments, once I understand it.

Sure, I clarified your queries above.

src/libasr/casting_utils.h
Shaikh-Ubaid: It is also used to cast the function arguments to the needed intermediate type.

certik: Is this a helper function to make `make_Cast_t_value` more convenient, but you still tell the initial and final type explicitly?

Shaikh-Ubaid:
Please let me know if you have further doubts in the implementation. I will be happy to clarify them.

certik (approved):
I think that this is fine. Would you mind submitting this PR against LFortran as well?

src/libasr/casting_utils.h
Shaikh-Ubaid: >Is this a helper function to make make_Cast_t_value more convenient, but you still tell the initial and final type explicitly?

Yes. The way it helps is we need not worry about what type it currently is and just specify what type we want it to be.

Shaikh-Ubaid:
Thank you so much for the review.

>Would you mind submitting this PR against LFortran as well?

Sure. Is it fine to merge this PR currently?

certik:
Yes, you can merge it.

PR #2374, opened 2023-10-08
Added support for comparing symbolic expressions

anutosh491:
While I was discussing comparsion of class types (Pow etc) with @Thirumalai-Shaktivel we realized that currently we aren't supporting comparing basic variables. Hence this is an implementation for that. Now we can do the following 
```
from lpython import S
from sympy import Symbol, pi, sin

def main0():
    x: S = pi
    y: S = pi
    print(x == pi)
    print(x == y)
    print(pi == y)
    print(pi == pi)
    print(x != sin(x))
    print(pi != sin(pi))
```
```
True
True
True
True
True
True
```

This should essentially help us with the following (https://github.com/sympy/sympy/blob/368eb0f483f2b4cd84dc0de3edc7abdc876d3d46/sympy/series/gruntz.py#L257)

### DIFF
--- a/integration_tests/symbolics_02.py
+++ b/integration_tests/symbolics_02.py
@@ -43,15 +43,15 @@ def test_symbolic_operations():
     b1: bool = p1 == p2
     print(b1)
     assert(b1 == True)
-    b2: bool = p1 == pi
+    b2: bool = p1 != pi
     print(b2)
-    assert(b2 == True)
+    assert(b2 == False)
     b3: bool = p1 != x
     print(b3)
     assert(b3 == True)
-    b4: bool = pi != Symbol("x")
+    b4: bool = pi == Symbol("x")
     print(b4)
-    assert(b4 == True)
+    assert(b4 == False)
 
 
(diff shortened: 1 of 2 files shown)


Thirumalai-Shaktivel (approved):
Perfect, let's add some tests.

anutosh491:
I've added some simple tests in the file where we are testing out binary operators. Hopefully this works.

integration_tests/symbolics_02.py
```diff
@@ -37,4 +39,19 @@ def test_symbolic_operations():
...
+    assert(b1 == True)
+    b2: bool = p1 == pi
+    print(b2)
+    assert(b2 == True)
+    b3: bool = p1 != x
+    print(b3)
+    assert(b3 == True)
+    b4: bool = pi != Symbol("x")
+    print(b4)
+    assert(b4 == True)
```
Thirumalai-Shaktivel, line 54: Add some false tests as well

certik (approved):

PR #2368, opened 2023-10-07
Adding Support for symbolics in the list data structure

anutosh491:
This is an attempt to allow the list data structure to handle symbolics.
So suppose we have something like 
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ cat examples/expr2.py 
from lpython import S
from sympy import Symbol, pi, sin

def main0():
    l1: list[S] = [pi]
    
main0()
```
Through the pass the following takes place
1) First we run through the list and then for each element (either a variable like `x` or an intrinsic function like `pi`, we do whatever's required like introducing the placeholder , changing type from `S` to `CPtr`)
2) Then we change the type of the list itself. 
So after the pass we have 
```
def main0():
    _stack0: i64 = i64(0)
    stack0: CPtr = empty_c_void_p()
    p_c_pointer(pointer(_stack0, i64), stack0)
    basic_new_stack(stack0)
    basic_const_pi(stack0)
    l1: list[CPtr] = [stack0]

main0()
```

PR #2378, opened 2023-10-09
Supporting assignment through `basic_assign`

anutosh491:
Addressing comment (https://github.com/lcompilers/lpython/pull/2368#issuecomment-1752019938).
Now all variable to variable assignments take place through `basic_assign` as recommended by symengine's documentation.

### DIFF
--- a/integration_tests/symbolics_01.py
+++ b/integration_tests/symbolics_01.py
@@ -6,7 +6,10 @@ def main0():
     y: S = Symbol('y')
     x = pi
     z: S = x + y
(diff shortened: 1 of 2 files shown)


anutosh491:
We won't really need some heavy testing here as we already had assignment working through the default `=` operator. I've still added a simple case in `symbolics_01.py`

Thirumalai-Shaktivel (approved):

certik (approved):
Very good, thanks!

PR #2384, opened 2023-10-18
Implemented query method for log class

anutosh491:
There's a TODO in this Pr which I need to address

So technically there is no `SYMENGINE_EXP` class which means to implement something like `e.func == exp`, we first need to check if `e.func == Pow and e.base == E`.

The first part here is easy but for the second part we need to use the `basic_get_args` which is defined as the following in sympy.
```
//! Returns a CVecBasic of vec_basic given by get_args
CWRAPPER_OUTPUT_TYPE basic_get_args(const basic self, CVecBasic *args);
```
It takes a pointer to an empty container I think and then populate it with the arguments. So basically we want the populated list and then the first element of that list would be our `base`.

Now the point is we shall pass a `CPtr` variable `e` as the first argument but I am not sure what would suffice for `CVecBasic *args`. I don't think a `pointer` to `List[CPtr]` works here so need to ideate on that.

### DIFF
--- a/src/libasr/pass/intrinsic_function_registry.h
+++ b/src/libasr/pass/intrinsic_function_registry.h
@@ -83,7 +83,6 @@ enum class IntrinsicScalarFunctions : int64_t {
     SymbolicMulQ,
     SymbolicPowQ,
     SymbolicLogQ,
-    SymbolicExpQ,
     // ...
 };
 
@@ -149,7 +148,6 @@ inline std::string get_intrinsic_name(int x) {
         INTRINSIC_NAME_CASE(SymbolicMulQ)
         INTRINSIC_NAME_CASE(SymbolicPowQ)
         INTRINSIC_NAME_CASE(SymbolicLogQ)
-        INTRINSIC_NAME_CASE(SymbolicExpQ)
         default : {
             throw LCompilersException("pickle: intrinsic_id not implemented");
         }
@@ -3151,7 +3149,6 @@ create_symbolic_query_macro(SymbolicAddQ)
 create_symbolic_query_macro(SymbolicMulQ)
 create_symbolic_query_macro(SymbolicPowQ)
 create_symbolic_query_macro(SymbolicLogQ)
-create_symbolic_query_macro(SymbolicExpQ)
 
 
 #define create_symbolic_unary_macro(X)                                                    \
@@ -3315,8 +3312,6 @@ namespace IntrinsicScalarFunctionRegistry {
             {nullptr, &SymbolicPowQ::verify_args}},
         {static_cast<int64_t>(IntrinsicScalarFunctions::SymbolicLogQ),
             {nullptr, &SymbolicLogQ::verify_args}},
-        {static_cast<int64_t>(IntrinsicScalarFunctions::SymbolicExpQ),
-            {nullptr, &SymbolicExpQ::verify_args}},
     };
 
     static const std::map<int64_t, std::string>& intrinsic_function_id_to_name = {
@@ -3429,8 +3424,6 @@ namespace IntrinsicScalarFunctionRegistry {
             "SymbolicPowQ"},
         {static_cast<int64_t>(IntrinsicScalarFunctions::SymbolicLogQ),
             "SymbolicLogQ"},
-        {static_cast<int64_t>(IntrinsicScalarFunctions::SymbolicExpQ),
-            "SymbolicExpQ"},
     };
 
 
@@ -3490,7 +3483,6 @@ namespace IntrinsicScalarFunctionRegistry {
                 {"MulQ", {&SymbolicMulQ::create_SymbolicMulQ, &SymbolicMulQ::eval_SymbolicMulQ}},
                 {"PowQ", {&SymbolicPowQ::create_SymbolicPowQ, &SymbolicPowQ::eval_SymbolicPowQ}},
                 {"LogQ", {&SymbolicLogQ::create_SymbolicLogQ, &SymbolicLogQ::eval_SymbolicLogQ}},
-                {"ExpQ", {&SymbolicExpQ::create_SymbolicExpQ, &SymbolicExpQ::eval_SymbolicExpQ}},
     };
 
     static inline bool is_intrinsic_function(const std::string& name) {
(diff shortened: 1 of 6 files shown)


anutosh491:
My initial thought process here was that rather than extracting args and then extracting the first element from the list, we can expose a `basic_get_base` function (https://github.com/symengine/symengine/pull/1980) but I don't think it was the best approach either.

certik:
I would not allow `.func == Exp`, because there is no such function. Rather, the user will have to do `e.func == Pow and `e.args == E` in the Python code. In other words, this can be implemented on top, but LPython itself would not have "Exp" as a type.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -82,6 +82,8 @@ enum class IntrinsicScalarFunctions : int64_t {
     SymbolicAddQ,
     SymbolicMulQ,
     SymbolicPowQ,
+    SymbolicLogQ,
+    SymbolicExpQ,
```
certik, line 86: I would remove `SymbolicExpQ`, since we do not represent Exp, only Pow.

certik:
Let's add a Python/SymPy test for "Log".

anutosh491:
Ohh wait but can't you do something like the following is sympy 
```
>>> from sympy import *
>>> x = Symbol("x")
>>> exp(x).func
exp
>>> exp(x).func == exp
True

**Comment by certik** [comment]
We will always support just a subset. So you can also do `exp(x).is_Log` too in SymPy, but in LPython you can't. So we won't allow `exp(x).func == exp`, you have to do `exp(x).func == Pow and exp(x).args[0] == E`, or you can write a function like `is_exp()` that is implemented internally using `exp(x).func == Pow and exp(x).args[0] == E`. LPython doesn't care, it just doesn't support it.

**Comment by anutosh491** [comment]
Ohh ok yes that makes sense to me and we only try implementing the approach you are talking about.
I'll design that in a subsequent pr, for now I'll add tests for `LogQ`.

I have also fixed `visit_Assert` to handle something like `assert(e.func == log)`.
I'll add tests for these too.

**Review by certik — APPROVED** [review]

**Comment by anutosh491** [comment]
> So we won't allow exp(x).func == exp, you have to do exp(x).func == Pow and exp(x).args[0] == E, or you can write a function like is_exp() that is implemented internally using exp(x).func == Pow and exp(x).args[0] == E

I think it would need to use `exp(x).base == E` rather than `exp(x).args[0] == E`. Because firstly if an expression is of type Pow, we can extract it's base
```
>>> (x*pi).base
AttributeError: 'Mul' object has no attribute 'base'
>>> (x**pi).base
x
```
Secondly the `args` method returns the arguments being used in the expression. So we have the following rather than getting `E` out of here.
```
>>> exp(x).args
(x,)
>>> exp(x).args[0]
x
```

**Comment by certik** [comment]
You can introduce `SymbolicPowBase` to return the base for Pow, otherwise give an error. However I think it is redundant.

We need to support `.args` anyway, don't we? So let's implement `args`, which in turn will fix this issue as well automatically.

**Comment by anutosh491** [comment]
Actually I am somewhat confused. We still plan to maintain consistency with sympy right ? (Atleast we've done that for all tests till now which means the tests through sympy or lpython return the same result). But consider the following.
If we don't go for `y.func == exp` and go for something like this instead.
```
from lpython import S
from sympy import pi, exp, Pow, Symbol

def main0():
    x: S = Symbol("x")
    y: S = exp(x)
    z: bool = y.func == Pow and y.base == E
    print(z)
```
We would be returning `True` for the following through `LPython` as discussed but we lose consistency with SymPy here 
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ PYTHONPATH="src/runtime/lpython" python examples/expr2.py 
False
```
Because through sympy we have the following 
```
>>> from sympy import *
>>> x = Symbol("x")
>>> y = exp(x)
>>> y.base
E
>>> y.func
exp
```
What comes in my mind here is that users could still use `y.func == exp` in the front end and once we catch hold the attribute `exp`, we could transform that in ASR to use `PowQ` and `BaseQ/ArgsQ`. Somewhere here 
https://github.com/lcompilers/lpython/blob/f8fbd8d3393ca1fab40a0bfcd08250be03b40d4a/src/lpython/semantics/python_ast_to_asr.cpp#L6040-L6046
But to begin with shouldn't we still start with `y.func == exp` to maintain consistency ?

**Comment by anutosh491** [comment]
But yeah, correct I think we might need indexing through args nonetheless as I can see at some places in `def mrv`. So yeah I'll try implementing args for now.

**Comment by certik** [comment]
The rule is that if LPython compiles it, then SymPy must as well. But not vice versa.

I didn't realize that `exp(x)` in sympy is indeed type "exp", not Pow. So we might need to do whatever it takes in the frontend to emulate this. In the backend (symengine) I would use the default, which his a Pow.

**Comment by anutosh491** [comment]
Yes correct. So to begin with we need to start with something like `e.func == exp` I suppose and then through the pass we might have to transform this such that we check the class to be Pow and the base to be `E`.

**Comment by certik** [comment]
I would get `y.func == Pow and y.args[0] == x` working first for things like `x**x`. That should work in both SymPy and SymEngine/LPython.

Then, we can add special handling for `y.func == exp` and rewrite it to `y.func == Pow and y.args[0] == E`.

**Comment by anutosh491** [comment]
> I would get y.func == Pow and y.args[0] == x

Actually I am not fully sure of how would can we represent this (`y.args[0] == x`) in ASR. 
`y.args[0]` looks like a query method for getting the 1st argument but if that's the case there can be infinite of such query methods which we would need to introduce.

What does make sense to me is something like this 
```
x : S = Symbol("x")
y : S = x**pi
args : list[S] = y.args
z: bool = y.func == Pow and args[0] == x
```
I think representing `y.args` makes more sense here as this can represent an IntrinsicFunction that holds the list of arguments overall.
But I am not sure how `y.args[0]` or `y.args[1]` would be represented. Can they be represented by the same intrinsic function (through the `overload id` I assume if we want to go forward with `args[]` as an intrinsic function as compared to `args` itself)

certik:
I would implement it using the design at https://github.com/lcompilers/lpython/issues/2393.

PR #2386, opened 2023-10-22
Implementing query method for exp class

anutosh491:
We start with something like the following in the frontend
```
from lpython import S
from sympy import pi, exp, Pow, Symbol

def main0():
    x: S = Symbol("x")
    y: S = exp(x)
    z: bool = y.func == exp
    print(z)
```
And I think we need to transform this into something like the following through the pass
```
def main0():
    _x: i64 = i64(0)
    x: CPtr = empty_c_void_p()
    basic_new_stack(x)
    symbol_set(x, "x")
    _y: i64 = i64(0)
    y: CPtr = empty_c_void_p()
    basic_new_stack(y)
    basic_exp(y, x)
    args: CPtr = vecbasic_new()
    basic_get_args(y, args)
    _base: i64 = i64(0)
    base: CPtr = empty_c_void_p()
    p_c_pointer(pointer(_base, i64), base)
    vecbasic_get(args, 0, base)
    _const: i64 = i64(0)
    const: CPtr = empty_c_void_p()
    p_c_pointer(pointer(_const, i64), const)
    basic_new_stack(const)
    basic_const_E(const)
    z: bool = basic_get_type(y) == 17 and basic_eq(base, const)
    print(z)
```

### DIFF
--- a/src/libasr/pass/intrinsic_function_registry.h
+++ b/src/libasr/pass/intrinsic_function_registry.h
@@ -83,6 +83,7 @@ enum class IntrinsicScalarFunctions : int64_t {
     SymbolicMulQ,
     SymbolicPowQ,
     SymbolicLogQ,
+    SymbolicExpQ,
     // ...
 };
 
@@ -148,6 +149,7 @@ inline std::string get_intrinsic_name(int x) {
         INTRINSIC_NAME_CASE(SymbolicMulQ)
         INTRINSIC_NAME_CASE(SymbolicPowQ)
         INTRINSIC_NAME_CASE(SymbolicLogQ)
+        INTRINSIC_NAME_CASE(SymbolicExpQ)
         default : {
             throw LCompilersException("pickle: intrinsic_id not implemented");
         }
@@ -3149,6 +3151,7 @@ create_symbolic_query_macro(SymbolicAddQ)
 create_symbolic_query_macro(SymbolicMulQ)
 create_symbolic_query_macro(SymbolicPowQ)
 create_symbolic_query_macro(SymbolicLogQ)
+create_symbolic_query_macro(SymbolicExpQ)
 
 
 #define create_symbolic_unary_macro(X)                                                    \
@@ -3312,6 +3315,8 @@ namespace IntrinsicScalarFunctionRegistry {
             {nullptr, &SymbolicPowQ::verify_args}},
         {static_cast<int64_t>(IntrinsicScalarFunctions::SymbolicLogQ),
             {nullptr, &SymbolicLogQ::verify_args}},
+        {static_cast<int64_t>(IntrinsicScalarFunctions::SymbolicExpQ),
+            {nullptr, &SymbolicExpQ::verify_args}},
     };
 
     static const std::map<int64_t, std::string>& intrinsic_function_id_to_name = {
@@ -3424,6 +3429,8 @@ namespace IntrinsicScalarFunctionRegistry {
             "SymbolicPowQ"},
         {static_cast<int64_t>(IntrinsicScalarFunctions::SymbolicLogQ),
             "SymbolicLogQ"},
+        {static_cast<int64_t>(IntrinsicScalarFunctions::SymbolicExpQ),
+            "SymbolicExpQ"},
     };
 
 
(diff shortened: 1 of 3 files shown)


anutosh491:
So basically after extracting the base we need to compare it against the constant `E` through `basic_eq` which tells me that along with `pi`, we would also need to introduce the constants `E` in Lpython which would be my next step.

anutosh491:
Okay I've made a separate PR (#2387) for addressing the above comment, so that the reviewing is easier on this PR and we could just focus on implementing `expQ` through this PR. @certik once you merge the above PR, I shall rebase this one !

anutosh491:
I was able to frame a patch for this and get an output 
```
from lpython import S
from sympy import pi, exp, Pow, Symbol, E

def main0():
    x: S = Symbol("x")
    y: S = exp(x)
    z: bool = y.func == exp
    print(z)

(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=llvm examples/expr2.py
True
```
But the code looks quite odd in comparison to other class query methods (because unfortunately symengine doesn't support an `exp` class)

So essentially just one statement in the frontend `e.func == exp` is replaced by the following 
```
    args: CPtr = vecbasic_new()
    basic_get_args(y, args)
    _base: i64 = i64(0)
    base: CPtr = empty_c_void_p()
    p_c_pointer(pointer(_base, i64), base)
    vecbasic_get(args, 0, base)
    _const: i64 = i64(0)
    const: CPtr = empty_c_void_p()
    p_c_pointer(pointer(_const, i64), const)
    basic_new_stack(const)
    basic_const_E(const)
    z: bool = basic_get_type(y) == 17 and basic_eq(base, const)
```
Which involves
1) Defining an argument vector
2) Extracting args from the expression
3) Defining a `base` to hold the 0th index of the argument vector (using `vecbasic_get(args, 0, base)`)
4) Defining a `const` to hold `E`
5) checking `type` as `Pow` and equating `base` against `const`

I'll paste the working patch below but just as I pointed out I am a bit doubtful if this is the best approach here. Because check the following case out . Through symengine we would get
```
from lpython import S
from sympy import pi, exp, Pow, Symbol, E

def main0():
    x: S = Symbol("x")
    y: S = exp(x)
    z: bool = y.func == Pow
    print(z)

(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=llvm examples/expr2.py
True
```
Whereas if through sympy we would get 
```
>>> exp(x).func == Pow
False
```
Hence we would have to tackle such cases too if we continue without introducing the `exp` class in SymEngine.

certik:
The question of introducing the exp class in SymEngine is a design question which has pros/cons, but I suspect we should not do it, it seems it does not speed things up, and Mathematica also doesn't do it, only SymPy. I opened up an issue https://github.com/symengine/symengine/issues/1984 to discuss that more.

Given that, and given that matching and querying also fails in Mathematica if you query for Exp, you have to query for Power, I would go with that design.

The problem is how to keep compatibility with SymPy. I think we can support `y.func == exp`, and internally check for `Pow(E,...)`, so that shouldn't be a problem. The problem is what to do about `y.func == Pow and y.args[0] == E`, which will work with LPython, but fail in SymPy. We have three options:

* Try to disallow it in LPython. However, what about just doing `exp(x).func == Pow`, which will pass in LPython but fail in SymPy. I don't see how we can disallow this, since the following must be allowed and passes in both SymPy and LPython: `(x**x).func == Pow`. So far I don't see a path forward here.
* Introduce `exp(x)` into SymEngine, but given https://github.com/symengine/symengine/issues/1984, I don't see this viable either.
* Allow `.func == Pow` in LPython, and yes, `exp(x)` will differ, which breaks the contract of "if it works in LPython, it must work in Python/SymPy`.

Any other option?

It seems the last option is the only way forward. This means there might not be any way we can enforce at compile time this difference. The solution might be that we might need to create a pure Python implementation of the Symbolic API, since SymPy itself differs slightly. This might be a good idea to do, and SymPy could later use it as well. For this pure Python implementation we will 100% guarantee that if it works in LPython, it will work with this pure Python implementation. SymPy will only work 99% or so. We will document that if you want to be compatible with sympy, to use `.func == exp`, then we can guarantee 100% compatibility. But if you use `y.func == Pow`, then if you want SymPy to work, you must ensure `y` is not `exp(x)`.

anutosh491:
As I have worked with SymPy's `gruntz.py`, I can say that differentiating `exp` and `Pow` in SymPy does play a role atleast in implementation of gruntz algorithm. I'm not sure of all advantages but this is actually a prominent one.

There would be cases where we would like to use `exp._eval_nseries` rather than the standard `Pow._eval_nseries` in gruntz. This is also pointed out by the note stating **Important** [here](https://github.com/sympy/sympy/blob/f751cd4a1da2c07a492b972302acab602dbcce52/sympy/functions/elementary/exponential.py#L23)

But yeah if we have don't see any other viable option, we can try not following the compilation rule (if works in Lpython, should work in SymPy) for this corner case and support only the `Pow` class and not the `exp` class.

anutosh491:
This can now be closed as we have decided not to go ahead with a query method for `exp` rather just use the `Pow` class and check if the `base` of the expression is `E`.

PR #2387, opened 2023-10-22
Implemented symbolic constant E

anutosh491:
I've made this pull request as a helping PR for #2386


anutosh491:
We've tested out the constant `pi` enough, E would behave just as same as `pi`. Not sure if I should add some test for it. All operations that could be done on `pi` can be replicated for `E`, so all tests that pass for expressions involving pi would also work for E

certik:
Definitely add a test for E. That's to ensure that everything actually works and that it will not get broken in the future.

anutosh491:
Done. I've added a minimalistic test function as of now which checks overall use of `E`.

anutosh491:
Not really sure why the `LPython CI / LPython CI (3.10, ubuntu-latest)` test fails . I am not able to replicate it locally !

certik (approved):

certik:
Thanks! What was the failure at the CI?

anutosh491:
I think so 1 test (`test_complex_02.py`) failed abruptly with the C backend with the fast flag. Thirumalai told me that he has seen the failure before and it occurs sometimes in the CI. He adviced me to re-run the tests!

certik:
It's a flaky test then. Can you report the error into an issue with a log of the failure? We need to keep track of this, so that we can fix it.

Issue #2385, opened 2023-10-19
current_scope's asr_owner is pointing to nullptr.


arteevraina:
In this [testcase](https://github.com/lcompilers/lpython/blob/main/integration_tests/lpconvexhull/lpconvexhull_main.py), `current_scope->asr_owner` is pointing to `nullptr` when `orientation` dependency is being added.

This simplest way to reproduce it is by adding this log at https://github.com/lcompilers/lpython/blob/main/src/lpython/semantics/python_ast_to_asr.cpp#L1190.

```
std::cout << ASRUtils::symbol_name(stemp) << " " << current_scope->asr_owner << std::endl;
```

It should log this :
```
orientation 0x0
```

certik:
Thanks! Let's create a minimal example out of this, and then fix it by adding an ASR verify check that `asr_owner` is present and points to the correct ASR node, as well as the actual fix.

arteevraina:
Here is the minimal example for this issue : 

```
from lpython import i32

def say_hello():
    print("Hello, World!")

def main():
    i: i32
    for i in range(10):
        say_hello()
        break
```

cc: @certik

arteevraina:
After investigating it more throughly, it's only the case when a function is called inside for loop is time when `current_scope->asr_owner` is pointing to `nullptr`.

Shaikh-Ubaid:
@arteevraina  I am yet to investigate the issue deeply but assuming that the `asr_owner` is not checked for `nullptr` in ASRVerify, I would take the following steps in the specified order to solve this:

- [ ] Add check for `asr_owner == nullptr` in the ASR Verify. If the `asr_owner` is nullptr then the ASRVerify fails.
- [ ] With the above check, the ASRVerify is now stronger/robust. Now, the example shared at https://github.com/lcompilers/lpython/issues/2385#issuecomment-1774024304 should fail with LPython (because of the asr_owner == nullptr check and assuming that the asr_owner in this case is nullptr, as you mentioned).
- [ ] Now, we have a test case that runs with CPython but fails for LPython in the ASRVerfiy stage. The last step would be to debug/support this in LPython.

PR #2390, opened 2023-10-23
Implemented asr verify checks to check for nullptr

arteevraina:
- [ ] Implemented asr verify checks to fail if functionCall / subroutineCall's current_symtab->asr_owner is equal to nullptr.

### DIFF
--- a/src/libasr/asr_verify.cpp
+++ b/src/libasr/asr_verify.cpp
@@ -855,6 +855,9 @@ public:
     void visit_SubroutineCall(const SubroutineCall_t &x) {
         require(symtab_in_scope(current_symtab, x.m_name),
             "SubroutineCall::m_name '" + std::string(symbol_name(x.m_name)) + "' cannot point outside of its symbol table");
+        require(current_symtab->asr_owner != nullptr, 
+            "SubroutineCall::m_name '" + std::string(symbol_name(x.m_name)) + "' must have a valid asr_owner");
+
         if (check_external) {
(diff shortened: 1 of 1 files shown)


certik (approved):
That seems like a no-brainer. `asr_owner` should never be null and it passes all the tests.

Shaikh-Ubaid:
>That seems like a no-brainer. asr_owner should never be null and it passes all the tests.

It passes all the tests, but I am concerned if it is a redundant check. It seems check for asr_owner is already present inside the respective symbol tab owners. I think we should now add this check to entities where we do not currently check it (if there exist any such entites where it is not checked).

Shaikh-Ubaid:
Also, I checked locally. It seems with the changes in this PR, the test at https://github.com/lcompilers/lpython/issues/2385#issuecomment-1774024304 does not fail.

I think the ASRVerify changes would be valuable if they make the test at https://github.com/lcompilers/lpython/issues/2385#issuecomment-1774024304 to fail. Then we have a Minimum Reproducible Example or MRE (https://github.com/lcompilers/lpython/issues/2385#issuecomment-1774024304) that we can use to fix LPython.

Shaikh-Ubaid:
I appreciate the contributions @arteevraina! The changes look fine to me, but do you think they are redundant?

certik:
If it is redundant, then we don't need to add it.

certik:
I think it's fine to merge. We should add these checks to all nodes, not just these two nodes.

Issue #2392, opened 2023-10-24
Incorrect Error message in ASRVerfiy


Shaikh-Ubaid:
I think this should be `"Dependency " + found_dep +  " is not* inside symbol table " `

_Originally posted by @Shaikh-Ubaid in https://github.com/lcompilers/lpython/pull/2167#discussion_r1370345894_

Shaikh-Ubaid:
@arteevraina would you like to tackle this?

arteevraina:
> @arteevraina would you like to tackle this?

Sure.

PR #2394, opened 2023-10-25
refactor: check for dependency only in parent scope

arteevraina:
Resolves: https://github.com/lcompilers/lpython/issues/2392


certik:
@arteevraina This looks good. Can you send this PR against LFortran too, to ensure everything works there? After that we can merge.

arteevraina:
> @arteevraina This looks good. Can you send this PR against LFortran too, to ensure everything works there? After that we can merge.

Sure.

certik:
Let's merge the LFortran first, then we can merge this one. For LFortran, see here: https://github.com/lfortran/lfortran/pull/2798#discussion_r1382478397.

arteevraina:
@certik Can we merge this as well ?

certik:
Yes, I think this is good, thanks!

Issue #2393, opened 2023-10-25
Design of getargs()


certik:
Let's create `SymbolicGetArg(expr e, int n)`:
```diff
--- a/src/libasr/pass/intrinsic_function_registry.h
+++ b/src/libasr/pass/intrinsic_function_registry.h
@@ -84,6 +84,7 @@ enum class IntrinsicScalarFunctions : int64_t {
     SymbolicMulQ,
     SymbolicPowQ,
     SymbolicLogQ,
+    SymbolicGetArg,
     // ...
 };
```
Which will take the expression `e`, call get_args on it, and return the `n`-th argument as an expression.

The SymEngine's getargs are slow anyway, since it returns `std::vector<Basic>` and internally SymEngine does not store the arguments in a vector, and so it must construct it (slow). Consequently, the `SymbolicGetArg` will be slow as well. We can start by using `basic_get_args`, implemented in SymEngine as:
```c++
CWRAPPER_OUTPUT_TYPE basic_get_args(const basic self, CVecBasic *args)
{
    CWRAPPER_BEGIN
    args->m = self->m->get_args();
    CWRAPPER_END
}
```
and then extract the `n`-th argument using `vecbasic_get`, implemented as:
```c++
CWRAPPER_OUTPUT_TYPE vecbasic_get(CVecBasic *self, size_t n, basic result)
{                     
    CWRAPPER_BEGIN                                                         
                                    
    SYMENGINE_ASSERT(n < self->m.size());
    result->m = self->m[n];                            
                                                              
    CWRAPPER_END
}                 
```

certik:
The other design is to introduce `SymbolicGetArgs(expr e)`:
```diff
--- a/src/libasr/pass/intrinsic_function_registry.h
+++ b/src/libasr/pass/intrinsic_function_registry.h
@@ -84,6 +84,7 @@ enum class IntrinsicScalarFunctions : int64_t {
     SymbolicMulQ,
     SymbolicPowQ,
     SymbolicLogQ,
+    SymbolicGetArgs,
     // ...
 };
 
```
This function will return a `list[S]` of all the arguments at once. Then we use regular LPython mechanism for accessing a list.

This might be a better approach.

anutosh491:
Closing as completed through #2399

PR #2396, opened 2023-10-26
Implementing symbolic get args method

anutosh491:
Trying to execute something like this 
```
from lpython import S
from sympy import pi, Symbol

def main0():
    x: S = Symbol("x")
    y: S = x**pi
    z: list[S] = y.args
    print(z[0], z[1])
```

### DIFF
--- a/src/libasr/pass/intrinsic_function_registry.h
+++ b/src/libasr/pass/intrinsic_function_registry.h
@@ -84,6 +84,7 @@ enum class IntrinsicScalarFunctions : int64_t {
     SymbolicMulQ,
     SymbolicPowQ,
     SymbolicLogQ,
+    SymbolicGetArguments,
     // ...
 };
 
@@ -150,6 +151,7 @@ inline std::string get_intrinsic_name(int x) {
         INTRINSIC_NAME_CASE(SymbolicMulQ)
         INTRINSIC_NAME_CASE(SymbolicPowQ)
         INTRINSIC_NAME_CASE(SymbolicLogQ)
+        INTRINSIC_NAME_CASE(SymbolicGetArguments)
         default : {
             throw LCompilersException("pickle: intrinsic_id not implemented");
         }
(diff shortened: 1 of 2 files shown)


anutosh491:
I'm not fully sure how we design this through the pass. As in we would have the following
1) After creating of an expression let's say `z`, we would need to extract the arguments using `basic_get_args`. That would be done using
```
    args: CPtr = vecbasic_new()
    basic_get_args(z, args)
```
2) Now we would like to end up with a `list` of type `CPtr` correct ? For that we would have to transfer all elements from the `args` (`CPtr`) variable we have to the list we would like to return. I'm thinking of how we could appraoch this. What comes across my mind is 
i) Extract the number of elements `vecbasic_size`
ii) Run a for loop. Each time define a local temp variable `temp1`
iii) Store the argument in the temporary variable using `vecbasic_get` for eg `vecbasic_get(args, 0, ele1)` 
iv) Define a temporary list which we would need and keep appending in the list after we pass. Return the list after the for loop ends.

Do you know of anything else we could do @certik ?

anutosh491:
BTW this is the design for the 2nd approach we were discussing here (https://github.com/lcompilers/lpython/issues/2393#issuecomment-1779708190)

The first design won't look too different.
In that case we are not returning a `list[CPtr]` but a `CPtr` variable only and as we would have the `nth` index which we would like to extract, it would be something as follows 
1) After creating of an expression let's say `z`, we would need to extract the arguments using `basic_get_args`. That would be done using
```
    args: CPtr = vecbasic_new()
    basic_get_args(z, args)
```
2) 
i) Now we might need a check if the argument index isn't out of bounds so we could use `vecbasic_size` to check
ii) If it is in bounds, we could create a temporary variable, `tmp` and run `vecbasic_get(args, n, tmp)` 
iii) Return this `tmp` variable !

We could maybe choose what suits us better here !

certik:
I think indeed you need to convert the `vec_basic` from SymEngine into `list[S]`. Yes, doing the loop is the way to do it.

Once you have `list[S]`, then you get everything else for free.

anutosh491:
Actually I see something strange when I was writing the code for this. Consider the following. We have `x**y` and we are extracting its arguments. For now the program tries to get the `0`th index from the args list
```
def main0():
    _x: i64 = i64(0)
    x: CPtr = empty_c_void_p()
    basic_new_stack(x)
    symbol_set(x, "x")
    _y: i64 = i64(0)
    y: CPtr = empty_c_void_p()
    basic_new_stack(y)
    symbol_set(y, "y")
    _z: i64 = i64(0)
    z: CPtr = empty_c_void_p()
    p_c_pointer(pointer(_z, i64), z)
    basic_new_stack(z)
    basic_pow(z, x, y)
    args: CPtr = vecbasic_new()
    basic_get_args(z, args)
    _ele: i64 = i64(0)
    ele: CPtr = empty_c_void_p()
    p_c_pointer(pointer(_ele, i64), ele)
    basic_new_stack(ele)
    tmp_list : list[CPtr] = []
    vecbasic_get(args, 0, ele)
    print(basic_str(ele))
    tmp_list.append(ele)
    vecbasic_get(args, 1, ele)
    print(basic_str(ele))
    tmp_list.append(ele)
    print(basic_str(tmp_list[0]))

main0()
```
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=c integration_tests/symbolics_08.py 
x
y
y
```
Why do I get `y` as the output for  `print(basic_str(tmp_list[0]))` ? Shouldn't it be `x`

anutosh491:
Okay If I just add one line to the above block
```
    tmp_list : list[CPtr] = []
    vecbasic_get(args, 0, ele1)
    print(basic_str(ele1))
    tmp_list.append(ele1)
+   print(basic_str(tmp_list[0]))
    vecbasic_get(args, 1, ele1)
    print(basic_str(ele1))
    tmp_list.append(ele1)
    print(basic_str(tmp_list[0]))
```    
I get 
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=c integration_tests/symbolics_08.py 
x
x
y
y
```
The second `x` here is due to the new line. So once `print(basic_str(tmp_list[0]))` gives `x` and the later it gives `y`

certik:
I think a slow but hopefully straightforward way to implement `print(e.args[0], e.args[1])` is to generate a new list for each call to `args`. Once this fully works, this can be upgraded to reuse the list whenever possible.

anutosh491:
I'm actually pretty confused at this point . Let's say our expression is (`x**y`) and we have the following 
```
    args: CPtr = vecbasic_new()
    basic_get_args(x, args)
    _ele1: i64 = i64(0)
    ele1: CPtr = empty_c_void_p()
    p_c_pointer(pointer(_ele1, i64), ele1)
    i: i32 = 0
    tmp_list: list[CPtr] = []
    for i in range(vecbasic_size(args)):
        vecbasic_get(args, i, ele1)
        tmp_list.append(ele1)
        print(basic_str(tmp_list[i]))
        basic_const_zero(ele1)
    print(basic_str(tmp_list[0]))
```
The output
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=c integration_tests/symbolics_08.py
x
y
0
```
I am quite clueless as to how we get this output. We are printing the elements of the list through `print(basic_str(tmp_list[i]))` and then we get the correct output and just to nullify everything I am using `basic_const_zero(ele1)`. But then after the loop if we use `print(basic_str(tmp_list[0]))` we get `0` which is pretty strange.

anutosh491:
> I think a slow but hopefully straightforward way to implement print(e.args[0], e.args[1]) is to generate a new list for each call to args. Once this fully works, this can be upgraded to reuse the list whenever possible.

I guess this would be more supportive of the first design which we were talking about (where our `SymbolicGetArguments` takes in the `symbolic expression` and the `integer`). As far as the current design is concerned we never really use `e.args[0]`. I think what we discussed what to first extract everything into a list (`list[S]`) and then only use extraction from the list to move forward.

anutosh491:
I've created a draft PR through design 1 too #2399 . I think it's a more simpler and accessible approach than this one. We can choose if we prefer a method over the other.

anutosh491:
Closing as we have an implementation for extracting args for now. (Can be opened if we plan on changing the design later)

PR #2399, opened 2023-10-28
Implementing get argument through design 1

anutosh491:
This pr is essentially trying to achieve the same thing as #2396 through a different design (Design 1 from #2393  which is https://github.com/lcompilers/lpython/issues/2393#issue-1961853101)

So essentially we can do something like the following 
```
from lpython import S
from sympy import Symbol

def main0():
    x: S = Symbol("x")
    y: S = Symbol("y")
    x = x**y
    arg1: S = x.args[1]
    print(arg1)

y
```

### DIFF
--- a/src/libasr/pass/replace_symbolic.cpp
+++ b/src/libasr/pass/replace_symbolic.cpp
@@ -1707,9 +1707,6 @@ public:
             pass_result.push_back(al, assert_stmt);
         } else if (ASR::is_a<ASR::SymbolicCompare_t>(*x.m_test)) {
             ASR::SymbolicCompare_t *s = ASR::down_cast<ASR::SymbolicCompare_t>(x.m_test);
-            SymbolTable* module_scope = current_scope->parent;
-            ASR::expr_t* left_tmp = nullptr;
-            ASR::expr_t* right_tmp = nullptr;
 
             ASR::symbol_t* basic_str_sym = declare_basic_str_function(al, x.base.base.loc, module_scope);
             left_tmp = process_with_basic_str(al, x.base.base.loc, s->m_left, basic_str_sym);
@@ -1724,6 +1721,27 @@ public:
             if (intrinsic_func->m_type->type == ASR::ttypeType::Logical) {
                 ASR::expr_t* test = process_attributes(al, x.base.base.loc, x.m_test, module_scope);
                 ASR::stmt_t *assert_stmt = ASRUtils::STMT(ASR::make_Assert_t(al, x.base.base.loc, test, x.m_msg));
+                pass_result.push_back(al, assert_stmt);
+            }
+        } else if (ASR::is_a<ASR::LogicalBinOp_t>(*x.m_test)) {
+            ASR::LogicalBinOp_t* binop = ASR::down_cast<ASR::LogicalBinOp_t>(x.m_test);
+            if (ASR::is_a<ASR::SymbolicCompare_t>(*binop->m_left) && ASR::is_a<ASR::SymbolicCompare_t>(*binop->m_right)) {
+                ASR::symbol_t* basic_str_sym = declare_basic_str_function(al, x.base.base.loc, module_scope);
+                ASR::SymbolicCompare_t *s1 = ASR::down_cast<ASR::SymbolicCompare_t>(binop->m_left);
+                left_tmp = process_with_basic_str(al, x.base.base.loc, s1->m_left, basic_str_sym);
+                right_tmp = process_with_basic_str(al, x.base.base.loc, s1->m_right, basic_str_sym);
+                ASR::expr_t* test1 =  ASRUtils::EXPR(ASR::make_StringCompare_t(al, x.base.base.loc, left_tmp,
+                    s1->m_op, right_tmp, s1->m_type, s1->m_value));
+
+                ASR::SymbolicCompare_t *s2 = ASR::down_cast<ASR::SymbolicCompare_t>(binop->m_right);
+                left_tmp = process_with_basic_str(al, x.base.base.loc, s2->m_left, basic_str_sym);
+                right_tmp = process_with_basic_str(al, x.base.base.loc, s2->m_right, basic_str_sym);
+                ASR::expr_t* test2 =  ASRUtils::EXPR(ASR::make_StringCompare_t(al, x.base.base.loc, left_tmp,
+                    s2->m_op, right_tmp, s2->m_type, s2->m_value));
+
+                ASR::expr_t *cond = ASRUtils::EXPR(ASR::make_LogicalBinOp_t(al, x.base.base.loc,
+                    test1, ASR::logicalbinopType::Or, test2, binop->m_type, binop->m_value));
+                ASR::stmt_t *assert_stmt = ASRUtils::STMT(ASR::make_Assert_t(al, x.base.base.loc, cond, x.m_msg));
                 pass_result.push_back(al, assert_stmt);
             }
         }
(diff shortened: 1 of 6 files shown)


src/libasr/pass/replace_symbolic.cpp
```diff
@@ -559,6 +559,84 @@ class ReplaceSymbolicVisitor : public PassUtils::PassVisitor<ReplaceSymbolicVisi
...
+                ASR::symbol_t* vecbasic_size_sym = declare_vecbasic_size_function(al, loc, module_scope);
+
+                // Define necessary variables
+                ASR::ttype_t* CPtr_type = ASRUtils::TYPE(ASR::make_CPtr_t(al, loc));
+                ASR::symbol_t* args_sym = ASR::down_cast<ASR::symbol_t>(ASR::make_Variable_t(
+                    al, loc, current_scope, s2c(al, "args"), nullptr, 0, ASR::intentType::Local,
+                    nullptr, nullptr, ASR::storage_typeType::Default, CPtr_type, nullptr,
+                    ASR::abiType::BindC, ASR::Public, ASR::presenceType::Required, false));
+                //if (!current_scope->get_symbol("args")) {
+                current_scope->add_symbol(s2c(al, "args"), args_sym);
```
anutosh491, line 577: I see 2 things to improve in this pr. This is one of them. What we want to do here is 
```
    args: CPtr = vecbasic_new()
    basic_get_args(x, args)
```
Hence I've created a local `args_sym` a symbol using the name "args". But I am not sure about this as to should we use a unique name of some sort here, args might be common which a user might end up using.

Secondly I think I am not able to use the `args` attribute for the second time due to some reason related to how I add it to the `current_scope`. Hence if we extract twice it gives an error 
```
    arg1: S = x.args[1]
    arg2: S = x.args[1]
    print(arg1)
    print(arg2)
```
```
  File "/home/anutosh491/lpython/lpython/src/bin/lpython.cpp", line 1869, in ??
...............
LCompilers::ASR::IntrinsicScalarFunction_t*, LCompilers::SymbolTable*, LCompilers::ASR::expr_t*)
    current_scope->add_symbol(s2c(al, "args"), args_sym);
AssertFailed: scope.find(name) == scope.end()
```
Using it once works. @Thirumalai-Shaktivel could you maybe help me out here!

src/libasr/pass/replace_symbolic.cpp
```diff
@@ -559,6 +559,84 @@ class ReplaceSymbolicVisitor : public PassUtils::PassVisitor<ReplaceSymbolicVisi
...
+                call_arg3.m_value = args;
+                call_args3.push_back(al, call_arg3);
+                ASR::expr_t* function_call2 = ASRUtils::EXPR(ASRUtils::make_FunctionCall_t_util(al, loc,
+                    vecbasic_size_sym, vecbasic_size_sym, call_args3.p, call_args3.n,
+                    ASRUtils::TYPE(ASR::make_Integer_t(al, loc, 4)), nullptr, nullptr));
+                ASR::expr_t* int_compare = ASRUtils::EXPR(ASR::make_IntegerCompare_t(al, loc, function_call2, ASR::cmpopType::LtE,
+                        x->m_args[1], ASRUtils::TYPE(ASR::make_Logical_t(al, loc, 4)), nullptr));
+                ASR::stmt_t* stmt3 = ASRUtils::STMT(make_If_t(al, loc, int_compare,
+                    if_body.p, if_body.n, else_body.p, else_body.n));
+                pass_result.push_back(al, stmt3);
```
anutosh491, line 620: Also this is the second thing which I wanted to point out. What I'm trying to do here is raise an error of some sort 
```
    if vecbasic_size(args) <= "the argument number passed":
        raise
```
I think I have the `test` ready but I need to raise an error of some sort to tell the user that the argument number queried for can't be returned or rather is not available. I am not fully sure how do I address the loopbody

anutosh491:
There are 4 statements in total we need to consider as pointed out in the pr 
```
    args: CPtr = vecbasic_new()
    basic_get_args(x, args)
    if vecbasic_size(args) <= i:
        pass
    vecbasic_get(args, i, ele)
```
Once we do these we should be good to go.

src/libasr/pass/replace_symbolic.cpp
anutosh491: Currently the loop body is empty so I assume it's using `pass` or something.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -5822,6 +5842,10 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                 using_func_attr = true;
                 return;
             }
+            if (attr == "args") {
+                using_args_attr = true;
```
certik, line 5846: Where is this variable getting reset to `false`? If it isn't, then I think that's a bug.

certik:
I think it's fine, let's get this PR to fully work. That gets something going. We can later iterate on it and do another design and compare if it is simpler, but the most important thing is to get something working, to unblock us. So this PR seems fine.

src/lpython/semantics/python_ast_to_asr.cpp
anutosh491: Addressed.

src/libasr/pass/replace_symbolic.cpp
anutosh491: I think we can't reuse the args local variable that we would like to extract some argument multiple times (for the same expression or different expressions) hence we would need a unique variable here.
```
from lpython import S
from sympy import Symbol

def main0():
    x: S = Symbol("x")
    y: S = Symbol("y")
    x = x**y
    y = x + y
    arg1: S = x.args[0]
    arg2: S = x.args[1]
    arg3: S = y.args[0]
    print(arg1)
    print(arg2)
    print(arg3)

if __name__ == "__main__":
   main0()
x
y
y

File: src/libasr/pass/replace_symbolic.cpp (hunk 2 above)
Comment (anutosh491) [reply src/libasr/pass/replace_symbolic.cpp:621]: I've addressed this too as of now but I am somewhat doubtful on the implementation of `visit_Raise` in `python_ast_to_asr.cpp` file and also how the `ErrorStop` node works in the asr. So I've addressed this but I don't know how to I add a exception statement for the same. We have the following currently 
```
def main0():
    x: S = Symbol("x")
    x = x**pi
    arg1: S = x.args[1]
    print(arg1)

pi
def main0():
    x: S = Symbol("x")
    x = x**pi
    arg1: S = x.args[2]
    print(arg1)

ERROR STOP
```

**Comment by certik** [comment]
Yes, this is another issue: the order of arguments for addition. This is determined by the hash table inside SymEngine. We could sort them lexicographically, but this will make it even more slower. Let's investigate Mathematica, if it keeps a stable order of arguments.

The way I would write the tests for now is to check that it returns `(x, y)` or `(y, x)`.

File: src/libasr/pass/replace_symbolic.cpp [hunk 4]
```diff
@@ -559,6 +559,84 @@ class ReplaceSymbolicVisitor : public PassUtils::PassVisitor<ReplaceSymbolicVisi
...
+                // Define necessary function symbols
+                ASR::expr_t* value1 = handle_argument(al, loc, x->m_args[0]);
+                ASR::symbol_t* basic_get_args_sym = declare_basic_get_args_function(al, loc, module_scope);
+                ASR::symbol_t* vecbasic_new_sym = declare_vecbasic_new_function(al, loc, module_scope);
+                ASR::symbol_t* vecbasic_get_sym = declare_vecbasic_get_function(al, loc, module_scope);
+                ASR::symbol_t* vecbasic_size_sym = declare_vecbasic_size_function(al, loc, module_scope);
+
+                // Define necessary variables
+                ASR::ttype_t* CPtr_type = ASRUtils::TYPE(ASR::make_CPtr_t(al, loc));
+                std::string args_str = current_scope->get_unique_name("_lcompilers_symbolic_argument_container");
```
anutosh491, line 572: I went over the code again and everything looks fine to me. If you have any other name for our argument container other than what I've used, you could let me know.

certik (approved):

PR #2404, opened 2023-11-01
Added support for `visit_Assert` through `basic_eq`

### LINKED ISSUE #2401: Discussing implementation for visit_assert in `replace_symbolic.cpp` pass
Consider the following example 
```
from lpython import S
from sympy import pi

def main0():
    x: S = pi
    assert(x == pi)
```
What we have been doing since the beginning is we 
1) Transform the assert block using `basic_str`, so we basically compare strings here (where `symengine_stack_variable0` stores `pi`)
`assert (basic_str(x) == basic_str(symengine_stack_variable0))`

But what would be better is to do the following `assert( basic_eq(x, symengine_stack_variable0) == 1)`

anutosh491:
Fixes #2401 

Instead of converting the `basic` variables into strings using `basic_str` and doing a string comparison, I think using `basic_eq` works better. This has been addressed through the PR

### DIFF
--- a/integration_tests/symbolics_01.py
+++ b/integration_tests/symbolics_01.py
@@ -13,4 +13,10 @@ def main0():
     assert(z == pi + y)
     assert(z != S(2)*pi + y)
 
+    # testing PR 2404
+    p: S = Symbol('pi')
+    print(p)
+    print(p != pi)
+    assert(p != pi)
+
 main0()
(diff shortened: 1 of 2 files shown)


certik:
Let's write a test that fails in master, but now works. Something like:
```
x: S = Symbol("pi")
assert x != pi
```
Which with this PR should pass, but in master fails?

certik:
Here is a test that fails in master, but works in this PR:
```
from sympy import Symbol, pi
from lpython import S

def main0():
    x: S = Symbol('pi')
    print(x)
    print(x != pi)
    assert x != pi

main0()
```

anutosh491:
Ready.

anutosh491:
cc @certik

certik (approved):
Perfect, thanks!

PR #2407, opened 2023-11-02
Fix: Punctuation mistake in error message

### LINKED ISSUE #1921: Punctuation mistake in error message
Only dataclass decorated classes and Enum subclasses are supported.

must be hyphenated (see https://www.grammarly.com/blog/hyphen/)

Only dataclass-decorated classes and Enum subclasses are supported.

pnc009:
- Fixes #1921 

cc: @certik


anutosh491 (approved):
Looks good !
Thanks

certik (approved):
Awesome, thanks for the fix! This is correct.

All our error messages can definitely be greatly improved.

PR #2412, opened 2023-11-09
Implemented visit_Return for freeing variables

anutosh491:
We need to take care of two things
1) If we have a function returning something, we would like to free all local basic variables before returning
2) As we are retuning the `_lpython_return_variable` I don't think so we need to free that.

Now we would have something like 
```
                                    (SubroutineCall
                                        2 basic_free_stack
                                        2 basic_free_stack
                                        [((Var 3 stack0))]
                                        ()
                                    )
                                    (SubroutineCall
                                        2 basic_free_stack
                                        2 basic_free_stack
                                        [((Var 3 stack1))]
                                        ()
                                    )
                                    (Return)]
```
Where `stack0` and `stack1` might be some local variables used inside the function, then followed by the `Return` statement.

### DIFF
--- a/src/libasr/pass/replace_symbolic.cpp
+++ b/src/libasr/pass/replace_symbolic.cpp
@@ -121,7 +121,9 @@ public:
 
             ASR::ttype_t *type1 = ASRUtils::TYPE(ASR::make_CPtr_t(al, xx.base.base.loc));
             xx.m_type = type1;
-            symbolic_vars_to_free.insert(ASR::down_cast<ASR::symbol_t>((ASR::asr_t*)&xx));
+            if (var_name != "_lpython_return_variable") {
+                symbolic_vars_to_free.insert(ASR::down_cast<ASR::symbol_t>((ASR::asr_t*)&xx));
+            }
             if(xx.m_intent == ASR::intentType::In){
                 symbolic_vars_to_omit.insert(ASR::down_cast<ASR::symbol_t>((ASR::asr_t*)&xx));
             }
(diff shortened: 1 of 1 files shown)


src/libasr/pass/replace_symbolic.cpp
```diff
@@ -1762,6 +1764,28 @@ class ReplaceSymbolicVisitor : public PassUtils::PassVisitor<ReplaceSymbolicVisi
             }
         }
     }
+
+    void visit_Return(const ASR::Return_t &x) {
+        SymbolTable* module_scope = current_scope->parent;
+        // freeing out variables
+        std::string new_name = "basic_free_stack";
+        ASR::symbol_t* basic_free_stack_sym = module_scope->get_symbol(new_name);
```
Thirumalai-Shaktivel, line 1772: Check if `module_scope->get_symbol(new_name)` is not `nullptr`.

Thirumalai-Shaktivel (commented):
I think it's fine, thank you!

src/libasr/pass/replace_symbolic.cpp
anutosh491: Do we return an error message if it's a `nullptr` ?

Thirumalai-Shaktivel: Nope, Let's create and add the symbol:`basic_free_stack ` to the current scope.

anutosh491: Actually I was thinking about it, as to in which cases would `basic_free_stack` be defined. And it goes like this
1) `basic_free_stack` is defined in `visit_Variable` and it only comes into play if the variable is of type `symbolic`
2) which means we only need basic_free_stack when we encounter a symbolic variable in our function.
3) The set keeping track of these symbolic variables is `symbolic_vars_to_free`. So we can say that at any points if `symbolic_vars_to_free` is non empty, it means we have encountered a symbolic variable while traversing through the pass, which means we have hit `visit_Variable` and hence `basic_free_stack` has been defined for such cases.

Thirumalai-Shaktivel (approved):

certik:
Can you add a test?

Why not calling the function to subroutine pass?

certik (approved):
I think it's fine.

certik:
This is fixing memory leaks.

PR #2414, opened 2023-11-10
Port --dump-all-passes, --dump-all-passes-fortran from LFortran

Shaikh-Ubaid:
fixes https://github.com/lcompilers/lpython/issues/2413

PR #2417, opened 2023-11-12
Remove redundant code

Shaikh-Ubaid:
This PR needs to be rebased after merging of https://github.com/lcompilers/lpython/pull/2414.


src/libasr/pickle.cpp
```diff
@@ -52,13 +52,13 @@ class ASRPickleVisitor :
...
         if (!show_intrinsic_modules &&
-                    startswith(x.m_name, "lfortran_intrinsic_")) {
+            (x.m_intrinsic || startswith(x.m_name, "numpy"))) { // Divergence between LFortran and LPython
             s.append("(");
             if (use_colors) {
                 s.append(color(style::bold));
                 s.append(color(fg::magenta));
             }
-            s.append("IntrinsicModule");
+            s.append(x.m_intrinsic ? "IntrinsicModule" : "Module"); // Divergence between LFortran and LPython
```
Thirumalai-Shaktivel, line 61: This seems okay

src/libasr/pickle.cpp
```diff
@@ -52,13 +52,13 @@ class ASRPickleVisitor :
     }
     void visit_Module(const ASR::Module_t &x) {
         if (!show_intrinsic_modules &&
-                    startswith(x.m_name, "lfortran_intrinsic_")) {
+            (x.m_intrinsic || startswith(x.m_name, "numpy"))) { // Divergence between LFortran and LPython
```
Thirumalai-Shaktivel, line 55: Here let's keep both?

Thirumalai-Shaktivel (commented):
Otherwise, looks good.

src/libasr/pickle.cpp
Shaikh-Ubaid: We can do that. I am unsure, but it may/might change the reference tests. The idea in this PR is to remove redundant/unused code, ideally keeping the other things intact. Hence, the above divergence comment.

I think it can be discussed/tackled later, not a priority at the moment.

Thirumalai-Shaktivel: But still, let's fix this the right way.
We will keep both checks; if it encounters any of the intrinsic module names, then it is marked as a IntrinsicModule.
So I would do this as:
```suggestion
            (x.m_intrinsic || startswith(x.m_name, "lfortran_intrinsic_") || startswith(x.m_name, "numpy"))) { // Divergence between LFortran and LPython
```

src/libasr/pickle.cpp
Thirumalai-Shaktivel: Also, I think you can remove the `Divergence between LFortran and LPython`

src/libasr/pickle.cpp
Thirumalai-Shaktivel: This would be synced with LFortran Libasr in the next sync.

Shaikh-Ubaid: Let me see if it changes the reference tests. If the reference tests change, I would prefer to do it in a separate PR.

Shaikh-Ubaid: The reference tests didn't change, so I think it is fine.

Thirumalai-Shaktivel (approved):

Shaikh-Ubaid:
> But still, let's fix this the right way. We will keep both checks; if it encounters any of the intrinsic module names, then it is marked as a IntrinsicModule. So I would do this as:

The old approach wasn't a wrong one. We mentioned in comments about divergence so that's like a TODO to be fixed/tackled later. Please understand that the main goal of this PR is to remove redundant/unused code and therefore I think the reference tests should not change after the removal. If the reference tests change, then it might not be redundant/unused code.

I hope you understand.

Issue #2411, opened 2023-11-08
Call `basic_free_stack` before return statement


Thirumalai-Shaktivel:
```py
from lpython import S
from sympy import pi

def func() -> S:
    print(pi)
    return pi

def test_func():
    print(func())

test_func()
```

```Clojure
(SubroutineCall
    2 basic_const_pi
    2 basic_const_pi
    [((Var 3 _lpython_return_variable))]
    ()
)
(Return)
(SubroutineCall                    ;; <== Here
    2 basic_free_stack
    2 basic_free_stack
    [((Var 3 _lpython_return_variable))]
    ()
)
(SubroutineCall
    2 basic_free_stack
    2 basic_free_stack
    [((Var 3 stack0))]
    ()
)
```

certik:
Consider this example:
```py
from lpython import S
from sympy import pi, Symbol

def func() -> S:
    return pi

def test_func():
    z: S = func()

test_func()
```
I recommend to fix it by calling the `subroutine_from_function` pass, which transforms the above into the following ASR code:
```python
from lpython import S
from sympy import pi, Symbol

def func(r: Out[S]) -> None:
    r = pi

def test_func():
    z: S
    func(z)

test_func()
```
And this code in turn should already work today, so nothing needs to be implemented in the Symbolic pass.

certik:
The first issue to fix is:
```console
$ lpython a.py --verbose
ASR Pass starts: 'nested_vars'
ASR Pass ends: 'nested_vars'
ASR Pass starts: 'global_stmts'
ASR Pass ends: 'global_stmts'
ASR Pass starts: 'init_expr'
ASR Pass ends: 'init_expr'
ASR Pass starts: 'implied_do_loops'
ASR Pass ends: 'implied_do_loops'
ASR Pass starts: 'class_constructor'
ASR Pass ends: 'class_constructor'
ASR Pass starts: 'pass_list_expr'
ASR Pass ends: 'pass_list_expr'
ASR Pass starts: 'where'
ASR Pass ends: 'where'
ASR Pass starts: 'subroutine_from_function'
ASR Pass ends: 'subroutine_from_function'
ASR Pass starts: 'array_op'
ASR Pass ends: 'array_op'
ASR Pass starts: 'symbolic'
Traceback (most recent call last):
  File "/Users/ondrej/repos/lpython/src/bin/lpython.cpp", line 1869
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir,
[...]

PR #2419, opened 2023-11-12
Added support for Out[S]

anutosh491:
Thanks to the well framed `subroutine_from_function` pass. Now we can return symbolic objects through a cummulative result of the `subroutine_from_function` and `symbolic` pass.


anutosh491:
I think the `subroutine_from_function` pass is aligned before the `symbolic` pass so we don't need to alter that!

anutosh491:
So currently we can do something like 
```
from lpython import S
from sympy import pi, Symbol

def func() -> S:
    print(Symbol("x"))
    return pi

def test_func():
    z: S = func()
    print(z)

test_func()
```
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=c examples/expr2.py 
x
pi
```
But through the llvm backend we would still be stuck as we had discussed previously (https://github.com/lcompilers/lpython/issues/2411#issuecomment-1806738228)
```
void** %_lpython_return_variable
 void*  call void @basic_const_pi(void** %_lpython_return_variable)
  %4 = load void*, void** %__libasr__created__var__0__func_call_res, align 8
 void**  call void @__module___main___func(void* %4)
 ```

**Comment by certik** [comment]
I think this is good. Can you please add a test?

Also, can you submit just the libasr change to lfortran, to ensure everything works there?

**Comment by anutosh491** [comment]
> I think this is good. Can you please add a test?
Also, can you submit just the libasr change to lfortran, to ensure everything works there?

Done.

**Comment by anutosh491** [comment]
We should also discuss as to how we would address comment (https://github.com/lcompilers/lpython/issues/2411#issuecomment-1806738228) because if you think about it, the function definition for `mrv` would look like the following 

```
def mrv(e: S, x: S) -> tuple[dict[S, S], S]:
```
Trying to replicate something like the following 
```
>>> from sympy.series.gruntz import mrv
>>> mrv(exp(x), x)
({exp(x): _Dummy_40}, {}, _Dummy_40)
```
Which means we need to return a `tuple` with a `dict` inside it and dictionaries aren't currently supported through the `C` backend, so untill we get the `llvm` backend working, we might be blocked.

certik (approved):
I think that's fine. Can you also add a test for `Out[S]` explicitly? Right now it is only tested implicitly via the subroutine_from_function.

anutosh491:
> Can you also add a test for Out[S] explicitly?

Sure.

PR #2430, opened 2023-11-25
Enable llvm tests for symbolics

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
Ready!

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1267,6 +1267,7 @@ namespace LCompilers {
                 break;
             }
             case (ASR::ttypeType::CPtr) : {
+                a_kind = 8;
```
Thirumalai-Shaktivel, line 1270: Previously, we used `a_kind = 4`, but now I set `8` to be default, this is required for Symbolic as we use `i64` pointer for `CPtr`.

certik: I thought CPtr doesn't have a kind?

Thirumalai-Shaktivel: How do we decide the malloc size for `list[CPtr]`?

anutosh491: I went through the changes and the changes do work in the correct direction. If we could know whether `a_kind` would be required or not for `list[CPtr]`, the PR can be finished accordingly .

anutosh491: cc @certik 
Could you clarify Thirumalai's comment here? I think this is close to completion once we have an idea for the same.

certik (approved):
I think this is good for now, it enables several tests and the changes are small and seem clean. I do not understand the design of a kind for `CPtr`, but that's a separate question. So I think we can merge this.

Issue #2428, opened 2023-11-18
Allow to initialize nested structs using "Transactions(transactions=[], accounts=[])"


certik:
With the following definitions:
```python
from lpython import dataclass, i64

@dataclass
class TransactionItem:
    account_idx: i64
    commodity: str
    amount: i64

@dataclass
class Transaction:
    date: str
    name: str
    cleared: bool
    comments: str
    items: list[TransactionItem]

@dataclass
class Account:
    name: str

@dataclass
class Transactions:
    transactions: list[Transaction]
    accounts: list[Account]
```
This does not work:
```python
asr: Transactions = Transactions(transactions=[], accounts=[])
```
This workaround works:
```python
transactions: list[Transaction] = []
accounts: list[Account] = []
asr: Transactions = Transactions(transactions=transactions, accounts=accounts)
```

certik:
The workaround above gives ASR, but fails to compile:
```console
$ lpython -I. process.py
ASR verify pass error: ASR verify: The symbol table was not found in the scope of `symtab`.

ASR verify pass error: ASR verify: Struct::m_derived_type 'Transaction' cannot point outside of its symbol table, owner: model
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/ondrej/repos/lpython/src/bin/lpython.cpp", line 1869
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir,
  File "/Users/ondrej/repos/lpython/src/bin/lpython.cpp", line 828
    res = fe.get_llvm3(*asr, pass_manager, diagnostics, infile);
  File "/Users/ondrej/repos/lpython/src/lpython/python_evaluator.cpp", line 71
    run_fn, infile);
  File "/Users/ondrej/repos/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 8933
    pass_manager.apply_passes(al, &asr, pass_options, diagnostics);
  File "/Users/ondrej/repos/lpython/src/libasr/pass/pass_manager.h", line 305
    apply_passes(al, asr, _passes, pass_options, diagnostics);
  File "/Users/ondrej/repos/lpython/src/libasr/pass/pass_manager.h", line 169
    + passes[i]);
LCompilersException: Verify failed in the pass: class_constructor
```

arteevraina:
Hi @certik ,
I tried to reproduce the above issue for the work around code.

```
from lpython import dataclass, i64

@dataclass
class TransactionItem:
[...]

PR #2434, opened 2023-11-28
fix: check if external symbol before checking from same scope.

arteevraina:
- [x] Resolves compilation failure for workaround mentioned here https://github.com/lcompilers/lpython/issues/2428#issuecomment-1817343501.
- [x] When the class is imported from another module as an ``ExternalSymbol``, it should omit the same scope check in ASR Verify. 

@certik Can you verify if this is the correct fix ?


certik:
I thought the ExternalSymbol should still be part of the current symbol table. I will need to investigate this.

Also let's add a test that now works.

src/libasr/asr_verify.cpp
```diff
@@ -1090,6 +1090,15 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
...
             symbol_owner = ASRUtils::symbol_name(ASRUtils::get_asr_owner(x.m_derived_type));
         }
+        
+        // Resolve the symbol in the current symbol table.
+        ASR::symbol_t* symbol = current_symtab->resolve_symbol(ASRUtils::symbol_name(x.m_derived_type));
+
+        // If the symbol is an ExternalSymbol, then omit the check.
+        if (ASR::is_a<ASR::ExternalSymbol_t>(*symbol)) {
+            return;
+        }
```
certik, line 1100: I think this change is not correct --- I think the ASR is incorrect if the ExternalSymbol is not present in the current scope. So there is a bug in the way we construct the ASR that we need to fix.

arteevraina: The symbol is there in the ``current_symtab`` imported as ``ExternalSymbol``. But, ``x.m_derived_type`` points to actual definition of the class which is inside another module, that is why ``symtab_in_scope`` fails. I think ASR is correct as the External Symbol is there in the current scope.

certik: The proper fix is described at https://github.com/lfortran/lfortran/issues/2957.

PR #2443, opened 2023-12-13
Update required packages

inkeliz:
Currently, `build0.sh` and `build1.sh` requires some packages such as re2c, bison and cmake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,10 +52,10 @@ This step involves setting up the required configuration to run the programs in 
 
 ### Linux
 
-Run the below command to install `binutils-dev` package on Linux.
+Run the below command to install `binutils-dev`, `re2c`, `bison` and `cmake` package on Linux.
 
 ```bash
-sudo apt install binutils-dev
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,10 +52,10 @@ This step involves setting up the required configuration to run the programs in
 
 ### Linux
 
-Run the below command to install `binutils-dev` package on Linux.
+Run the below command to install `binutils-dev`, `re2c`, `bison` and `cmake` package on Linux.
 
 ```bash
-sudo apt install binutils-dev
+sudo apt install binutils-dev re2c bison cmake
```
certik, line 58: One should install these using Conda. Not all Bison versions work unfortunately.

We have a Conda environment file somewhere, so we should use that.

certik:
Thanks for the PR! Let's document how to use Conda to install the dependencies, see above.

Thirumalai-Shaktivel:
Yes, I think the same.
Please address the review and mark it as "Ready for Review"

Issue #2356, opened 2023-10-02
Add new APIs in string object 


Agent-Hellboy:
there are 47 public methods in the CPython str object 
```
['capitalize', 'casefold', 'center', 'count', 'encode', 'endswith', 'expandtabs', 'find', 'format', 'format_map', 'index', 'isalnum', 'isalpha', 'isascii', 'isdecimal', 'isdigit', 'isidentifier', 'islower', 'isnumeric', 'isprintable', 'isspace', 'istitle', 'isupper', 'join', 'ljust', 'lower', 'lstrip', 'maketrans', 'partition', 'removeprefix', 'removesuffix', 'replace', 'rfind', 'rindex', 'rjust', 'rpartition', 'rsplit', 'rstrip', 'split', 'splitlines', 'startswith', 'strip', 'swapcase', 'title', 'translate', 'upper', 'zfill']
```
we have implemented 17 of them if you folks agree, I will try to implement the remaining functions

Agent-Hellboy:
Added for PR for 3 of them , will add 1 PR for 3-4 APIs

certik:
@Agent-Hellboy definitely, thank you. I recommend you first finish the PRs you sent, let's get them in. Then send more.

Agent-Hellboy:
> @Agent-Hellboy definitely, thank you. I recommend you first finish the PRs you sent, let's get them in. Then send more.

surething

PR #2447, opened 2023-12-14
Implement ``str.replace()`` and ``str.index()``

AkshayWarrier:
Relates to #2356

I also noticed that these methods don't work with constant strings
```
semantic error: 'str' object has no attribute 'split'
 --> test.py:2:7
  |
2 | print("1 2 3".split())
  |       ^^^^^^^^^^^^^^^ 
```
and I'm guessing it's because they are not being handled in ``handle_constant_string_attributes``. 
1. Why can't both constant strings and variable strings be handled by the ``_lpython_str_...`` functions? Because it looks like the methods for constant strings are directly being implemented in C++.
2. Shall I also add separate implementations for these in ``handle_constant_string_attributes``?

### DIFF
--- a/integration_tests/test_str_01.py
+++ b/integration_tests/test_str_01.py
@@ -145,6 +145,19 @@ def test_str_split():
     assert res5 == ["123"]
     # assert res6 == [""]
 
+def test_str_replace():
+    a: str = "abracadabra"
+    res: str = a.replace("a","b")
+    res1: str = a.replace("a","")
+    res2: str = a.replace("e","a") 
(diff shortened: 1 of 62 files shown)


Thirumalai-Shaktivel:
Yes, I think we handle compile-time string in python_ast_to_asr.cpp and run time string using `_lpython_...`.
What needs to be done to complete this and merge?

Kishan-Ved:
I have implemented str.replace() completely (with support for all three parameters) here: https://github.com/lcompilers/lpython/pull/2587

PR #2453, opened 2024-01-07
Don't load CPtr if the variable under consideration is an argument

anutosh491:
Most discussion for the PR was done here (#2440)


certik (approved):
I think that's fine for now.

PR #2459, opened 2024-01-10
UserOverride.cmake: fix native optflag for PowerPC

barracuda156:
@certik This is needed for the build to be feasible on PowerPC, since `-march` is unsupported there.
https://gcc.gnu.org/onlinedocs/gcc/RS_002f6000-and-PowerPC-Options.html

`-mtune` should be used instead; `-mcpu` also supported but may be unsafe and produce a broken code for some targets.

### DIFF
--- a/cmake/UserOverride.cmake
+++ b/cmake/UserOverride.cmake
@@ -9,7 +9,12 @@
 if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     # g++
     set(common "-Wall -Wextra")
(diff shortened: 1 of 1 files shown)


certik:
I am not sure if this will work all the time, but we can merge it for now, and iterate on it later if needed.

Thanks!

barracuda156:
@certik This leaves out cases of Rosetta and cross-compilation for PowerPC, but they were broken anyway.
I did not do it in one go to make it simpler. Fixing native build can be the first step.

Issue #2450, opened 2024-01-02
Seg Fault for case involving symbolic lists through llvm backend


anutosh491:
Consider the following example 
```
from lpython import S
from sympy import Symbol, pi

def mmrv() -> list[S]:
    l1: list[S] = [pi]
    return l1

def test_mrv1():
    ans: list[S] = mmrv()
    el: S = ans[0]
    print(el)

test_mrv1()
```
We get
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=llvm examples/expr2.py 
Segmentation fault
```

The same function would work for a list of any other type but fails to work for a symbolic list.

anutosh491:
After debugging this I see that the error occurs due to the extraction of element from the list `ans[0]`.
The final ASR after all passes look good to me.
While going through the llvm IR I see 
```
  %26 = load void*, void** %el, align 8
  ......
  %31 = getelementptr %list, %list* %ans, i32 0, i32 2
  %32 = load void**, void*** %31, align 8
  %33 = getelementptr inbounds void*, void** %32, i32 0
  %34 = load void*, void** %33, align 8
  call void @basic_assign(void* %26, void* %34)
  %35 = load void*, void** %el, align 8
```

Here I think the `basic_assign` assigns `%34` to `%26` but then it is not being stored back into `%el` and hence when line 
`%35 = load void*, void** %el, align 8` is executed, `%el` isn't having the required value. I think we might want to have a 
`store void** %26, void*** %el, align 8`

certik:
This works:
```
def mmrv() -> list[i32]:
    l1: list[i32] = [5]
    return l1

def test_mrv1():
    ans: list[i32] = mmrv()
    el: i32 = ans[0]
    print(el)

test_mrv1()
```
So we can compare it against the CPtr version after the symbolic pass and see what the problem is. You can then try to reproduce it by explicitly writing Python code that uses CPtr (does not use the symbolic pass) and we need to figure out why it segfaults.

anutosh491:
I was almost able to figure this out. Now that #2453 & #2454 have gone in, we have the llvm and c backend ready to deal with any symbolic container like a symbolic list.

So consider the following. We expect the program given above to transform to something like this (in python) after the **symbolic** and the **subroutine_from_function** pass
```
from lpython import S
from sympy import Symbol, pi

def mmrv(r: Out[list[S]]) -> None:
    x: S = pi
    l1: list[S] = [x]
[...]

Issue #2455, opened 2024-01-08
CI for testing through macos seems flaky


anutosh491:
The CI yesterday was passing at some attempts and failing otherwise. At a couple of instances I saw it is also running infinitely (https://github.com/lcompilers/lpython/actions/runs/7438892349/job/20243305980)

certik:
Also discussed at https://lfortran.zulipchat.com/#narrow/stream/197339-general/topic/CI.20for.20testing.20through.20macos.20seems.20flaky.20for.20Lpython.

certik:
This is now "fixed" by removing the tests. See https://github.com/lcompilers/lpython/issues/2520 for the proper fix.

PR #2465, opened 2024-01-30
Using basic_new_heap for assigning elements of symbolic list

anutosh491:
This PR adds an example to address the first TODO as pointed out here (https://github.com/lcompilers/lpython/issues/2450#issuecomment-1896363685)

We are trying to replicate this code snippet essentially 
```
from lpython import S
from sympy import Symbol, pi

def mmrv(r: Out[list[S]]) -> None:
    x: S = pi
    l1: list[S]
    l1 = [x]
    print(l1[0])
    r = l1

def test_mrv2():
    ans: list[S]
    temp: list[S]
    mmrv(ans)
    temp = ans
    print(temp[0])

test_mrv2()
```

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -720,6 +720,7 @@ RUN(NAME symbolics_12        LABELS cpython_sym c_sym llvm_sym NOFAST)
 RUN(NAME symbolics_13        LABELS cpython_sym c_sym llvm_sym NOFAST)
 RUN(NAME symbolics_14        LABELS cpython_sym llvm_sym NOFAST)
 RUN(NAME test_gruntz         LABELS cpython_sym c_sym llvm_sym NOFAST)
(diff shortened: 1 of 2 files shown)


anutosh491:
Also talking about the note which added to the mentioned comment 
`
Note: in LPython, list assignment r = l1 is always a deep copy (unlike in CPython where it is just a refcount increase), so we should initialize r first using basic_new_heap and then use basic_assign to assign the elements. That can be done later.
`

I see that this is correct and works without flaws, hence for now the line `r = l1` can be replicated using `r = l1` itself but in future we might prefer the approach pointed above.

certik (approved):
I think that this looks good.

certik:
Is this ready to be merged?

anutosh491:
> Is this ready to be merged?

I think it is ready !

certik:
The failure is due to https://github.com/lcompilers/lpython/issues/2455.

PR #2477, opened 2024-02-04
Fixing Symbolic List assignment

anutosh491:
Explaining the operations/changes that the ASR symbolic pass introduces 

1) The pass is intended to deal with list assignment. Now there can a couple cases possible when we think about it which are 
```
i) l1 = [pi, sin(x)] # listconstant assigned to list var
ii) l1 = l2 # list var assigned to another list var
```
2) Now as discussed here https://github.com/lcompilers/lpython/issues/2450#issuecomment-1896363685 , case `ii` is always a deep copy and as observed works without any issues. Though this can surely be improved
Rather than the simple assignment we could have something like 
i) Empty l1
ii) Appening elements of l1 with `basic_heap`
iii) `basic_assign` between elements of l1 & l2
But that can be taken care of later

3) The pass essentially deals with case `i`
What happens is the following (Snippet 1 is transformed into Snippet 2)
```
// Snippet 1
    l1: list[S] = [x]
```
```
// Snippet2

    # Introduce a placeholder for list[CPtr] just like we have one for CPtr
    _l1: list[CPtr] = [x]
    l1: list[CPtr] = []

    # l1 = [x]
    i: i32 = 0
    for i in range(len(_l1)):
        tmp: CPtr = basic_new_heap()
        l1.append(tmp)
        basic_assign(l1[i], _l1[i])
```

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -718,9 +718,10 @@ RUN(NAME symbolics_10        LABELS cpython_sym c_sym llvm_sym NOFAST)
 RUN(NAME symbolics_11        LABELS cpython_sym c_sym llvm_sym NOFAST)
 RUN(NAME symbolics_12        LABELS cpython_sym c_sym llvm_sym NOFAST)
 RUN(NAME symbolics_13        LABELS cpython_sym c_sym llvm_sym NOFAST)
-RUN(NAME symbolics_14        LABELS cpython_sym llvm_sym NOFAST)
+RUN(NAME symbolics_14        LABELS cpython_sym c_sym llvm_sym NOFAST)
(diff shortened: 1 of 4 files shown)


anutosh491:
I've added few comments for our understanding and I've also added the failing case raised here (https://github.com/lcompilers/lpython/issues/2450#issue-2062004663) as a test.

certik (approved):
I thin this is good. Thanks for fixing this! I think you can merge it.

anutosh491:
Thanks for the review. I think we can go forward with this. If we are looking to customize the symbolic pass even more (if we come across any improvement maybe) let's address it through subsequent PRs.

anutosh491:
cc @certik I am not sure if I have permissions to merge something untill are requirements are met (I can enable auto-merge squash or rabase when all requirements are met) .
Now that the macos build doesn't pass, I am not sure I can get this in.

PR #2484, opened 2024-02-05
Fixes empty string bug in str isalpha method

advikkabra:
In the current implementation of the `isalpha()` method, an empty string returns a `True` value, but in CPython it returns `False`.


anutosh491:
Good catch @advikkabra 
Do you see any more builtin methods where we might be going wrong in corner cases like this ? 
Would be great if you could give it a look.

Also the CI failing for the macos build isn't due to your commit, so no need to worry about it.

anutosh491 (approved):

certik (approved):

PR #2489, opened 2024-02-06
Fix isspace edge case

advikkabra:
An empty string should return `False` when the `isspace` method is called.

### DIFF
--- a/src/runtime/lpython_builtin.py
+++ b/src/runtime/lpython_builtin.py
@@ -964,6 +964,8 @@ def _lpython_str_isascii(s: str) -> bool:
     return True
 
 def _lpython_str_isspace(s:str) -> bool:
+    if (len(s) == 0):
+        return False
     ch: str 
     for ch in s:
         if ch != ' ' and ch != '\t' and ch != '\n' and ch != '\r' and ch != '\f' and ch != '\v':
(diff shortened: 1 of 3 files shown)


src/runtime/lpython_builtin.py
```diff
@@ -963,6 +963,8 @@ def _lpython_str_isascii(s: str) -> bool:
     return True
 
 def _lpython_str_isspace(s:str) -> bool:
+    if (len(s) == 0):
+        return False
```
anutosh491, line 967: Looks good. We could skip the brackets though, `if len(s) == 0` should do !
Also add a test for the same.

advikkabra:
Added the test, and also noticed a bug in the C++  implementation, so fixed the same.

anutosh491:
> also noticed a bug in the C++ implementation

Ohh what's this ? Have you tested it out ? 
Remember to raise a separate PR for every other issue you are fixing. Helps us review and keep track of the change. Nothing wrong with atomic/small commits :)

advikkabra:
It is the same edge case, just for constant empty strings as well. I will write a test for it as well.

certik (approved):
I think this is fine.

@anutosh491 can you please review & merge?

anutosh491 (approved):
Looks good, Merging !

Issue #2499, opened 2024-02-07
Const[i16] in step position of 'range' causes error message


rebcabin:
```python
from lpython import i32, i16, Const
VR_SIZE: i32 = 32_768
l: Const[i32] = VR_SIZE
n: Const[i32] = 15
m: Const[i32] = 3
k: i32
M2: Const[i32] = 5  # ~~~~~~~~~~~~~~~~~~~~~~~~~~ ATTENTION ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
A_ik: i16
jj: i32
ii: i32
i: i32
for jj in range(0, l, VR_SIZE):  # each VR-col chunk in B and C
    for ii in range(0, n, M2):  # each M2 block in A cols and B rows  # !!!!!!!!!!!!!!
        for i in range(0, M2):  # zero-out rows of C
            pass
        for k in range(0, m):  # rows of B
            for i in range(0, M2):
                pass
```

```
(lp) ┌─(~/Documents/GitHub/lpython/integration_tests)──────────────────────────────────────────────────────────────────────────────(brian@MacBook-Pro:s001)─┐
└─(07:13:11 on vector-backend ✹ ✭)──> lpython ../ISSUES/Issue2499.py                                                                     ──(Wed,Feb07)─┘
semantic error: For loop increment type should be Integer.
  --> ../ISSUES/Issue2499.py:13:5 - 18:20
   |
13 |        for ii in range(0, n, M2):  # each M2 block in A cols and B rows
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
   |
18 |                    pass
   | ...^^^^^^^^^^^^^^^^^^^^ 

```

Shaikh-Ubaid:
This https://github.com/lcompilers/lpython/pull/2501 would hopefully fix the constant step in loop issue.

There is also another issue in the above code. 
```python
VR_SIZE: i32 = 32_768
l: Const[i32] = VR_SIZE
```
`VR_SIZE` is a variable (not a constant) for which the value will be assigned at run time. So, we cannot use `VR_SIZE` as the constant/initialization value for  `l: Const[i32] = VR_SIZE`. 

Currently, we can do either of the following ways to fix the above.

`VR_SIZE` should also be a constant so that its value can be used to initialize another constant. For example:
```python
VR_SIZE: Const[i32] = 32_768
[...]

PR #2501, opened 2024-02-07
Fix const step in loop

Shaikh-Ubaid:
towards https://github.com/lcompilers/lpython/issues/2499


certik:
The CI failure is due to https://github.com/lcompilers/lpython/issues/2455.

PR #2452, opened 2024-01-05
Fixes SEGFAULT in integration_tests/elemental_01.py

Kishan-Ved:
Please check: https://github.com/lcompilers/lpython/pull/2451

PR #2502, opened 2024-02-07
Get macOS test working again

certik:
Towards #2455.

The first macOS failure in master happened in https://github.com/lcompilers/lpython/pull/2452. This PR reverts LPython to the commit before that. We'll see if it reliably passes.

If it passes, then #2452 must have broken it.

If it doesn't pass, then something in the GitHub Actions or Conda environment changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# LPython
+#  LPython
 
 LPython is a Python compiler. It is in heavy development, currently in alpha
 stage. LPython works on Windows, macOS and Linux. Some of the goals of LPython


certik:
It hangs.... So the issue is definitely not caused by https://github.com/lcompilers/lpython/pull/2452. Rather, it's either caused by something introduced before that PR, or some changes in the GitHub Actions or Conda environments.

certik:
Testing an even older commit at https://github.com/lcompilers/lpython/pull/2504

certik:
We figured it out in https://github.com/lcompilers/lpython/issues/2520.

PR #2504, opened 2024-02-07
CI: testing an even older commit

certik:


Shaikh-Ubaid:
I am slightly confused. How are we testing an older commit? I could not see any file changes in this PR.

certik:
So it passed here. I am rerunning the macOS job again. I'll run it at least 5 times and see if I can trigger the hang. If not, then this commit might work. Then we just need to bisect it.

anutosh491:
Okay I see the CI passes here. What does this tell us though ? And how do we reset to this commit !

certik:
Restarted again, running for the 3rd time. We want to be 100% sure it works.

This PR is testing the commit a5e3d042f1dea0d49c65ab933eaff086b9a66683.

Assuming this commit works, then we just have to bisect until we find the problem. I suspect it could be some segfault in the symbolic support?

certik:
It looks like it is hanging now... So event his old commit fails.

certik:
We need to go deeper: https://github.com/lcompilers/lpython/pull/2516.

PR #2506, opened 2024-02-07
Reserve to ListReserve

rajeshkotaru:
in intrinsic_function_registry.h,
changed Reserve to ListReserve,
reserve to list.reserve,
reserve to list_reserve.
at appropriate places by observing ListReverse,list.reverse,list_reverse


Shaikh-Ubaid:
`lfortran` has the latest `libasr` so this should be sent to `lfortran`. Later it easily gets ported to LPython when there is a sync PR.

rajeshkotaru:
ok

PR #2521, opened 2024-02-09
CI: Enable LLVM and C backends on macOS

certik:
Towards #2520.

PR #2522, opened 2024-02-09
CI: XX should fail

certik:
This is the same as https://github.com/lcompilers/lpython/pull/2521, but using `-j16` instead of `-j1`. It should fail almost every time.

If so, we found the bug.


certik:
Failed on 1st try. That's good. Trying 2nd time.

certik:
I tried it twice, it fails both time. So I think we are in a good shape.

PR #2525, opened 2024-02-12
Added case to support log instances in mrv function

anutosh491:
What we are trying to do is address this [case ](https://github.com/sympy/sympy/blob/680fb21363e1ea685d6fc868d15069d80ac0c97f/sympy/series/gruntz.py#L285) and this [example](https://github.com/sympy/sympy/blob/680fb21363e1ea685d6fc868d15069d80ac0c97f/sympy/series/tests/test_gruntz.py#L253)

We already have the essential tools to support it so no changes are required in the ASR pass.

### DIFF
--- a/integration_tests/test_gruntz.py
+++ b/integration_tests/test_gruntz.py
@@ -1,5 +1,5 @@
 from lpython import S
-from sympy import Symbol
+from sympy import Symbol, log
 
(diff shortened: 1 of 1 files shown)


certik (approved):

anutosh491:
Merging as CI is green .

PR #2530, opened 2024-02-12
Improved str.count() algorithm using the Knuth-Morris-Pratt algorithm

advikkabra:
Implemented the [Knuth-Morris-Pratt](https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm) algorithm for `str.count()`, similar to the `str.find()` implementation. This should improve speed significantly in many cases.

### DIFF
--- a/integration_tests/test_str_attributes.py
+++ b/integration_tests/test_str_attributes.py
@@ -71,6 +71,26 @@ def find():
     assert s2.find("we") == -1
     assert "".find("") == 0
 
+def count():
+    s: str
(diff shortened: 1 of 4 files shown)


advikkabra:
@certik Could you review this PR?

certik (approved):
@advikkabra thank you for this PR. I think it looks great, thanks for implementing the more advanced algorithm, both at compile time and runtime.

Issue #2532, opened 2024-02-14
Verify pass fails in the symbolic pass


certik:
```python
from lpython import S
from sympy import Symbol, pi

def test_main():
    x: S = Symbol('x')
    if x == pi:
        print('hi')

test_main()
```
This gives:
```console
$ lpython a.py --enable-symengine
ASR verify pass error: ASR verify: Function test_main doesn't depend on basic_eq but is found in its dependency list.
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/ondrej/repos/lpython/src/bin/lpython.cpp", line 1872
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir,
  File "/Users/ondrej/repos/lpython/src/bin/lpython.cpp", line 824
    res = fe.get_llvm3(*asr, pass_manager, diagnostics, infile);
  File "/Users/ondrej/repos/lpython/src/lpython/python_evaluator.cpp", line 71
    run_fn, infile);
  File "/Users/ondrej/repos/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 9276
    pass_manager.apply_passes(al, &asr, co.po, diagnostics);
  File "/Users/ondrej/repos/lpython/src/libasr/pass/pass_manager.h", line 299
    apply_passes(al, asr, _passes, pass_options, diagnostics);
  File "/Users/ondrej/repos/lpython/src/libasr/pass/pass_manager.h", line 164
    throw LCompilersException("Verify failed in the pass: "
LCompilersException: Verify failed in the pass: symbolic
```

certik:
The problem is that the if condition becomes missing after the symbolic pass, so the `basic_eq` is still in the dependency list, but is not being used, and thus the verify fails.

anutosh491:
Fixed by #2535

PR #2535, opened 2024-02-15
Fixed visit_if for SymbolicCompare cases

anutosh491:
The way I addresssed #2532 is by creating a new `If` node instead of editing the `m_test` parameter of the old node itself and this works. I think the original code should itself be sufficient but we haven't been able to find the error hence using this.

I've added a test for the same.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -728,6 +728,7 @@ RUN(NAME symbolics_14        LABELS cpython_sym c_sym llvm_sym NOFAST)
 RUN(NAME test_gruntz         LABELS cpython_sym c_sym llvm_sym NOFAST)
 RUN(NAME symbolics_15        LABELS c_sym llvm_sym NOFAST)
 RUN(NAME symbolics_16        LABELS cpython_sym c_sym llvm_sym NOFAST)
+RUN(NAME symbolics_17        LABELS cpython_sym c_sym llvm_sym NOFAST)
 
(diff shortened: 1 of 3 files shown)


certik (approved):
I see what the issue is. Every visit statement method must return a new result in `pass_result`. If you don't, the statement gets skipped.

Consequently, I think the first two branches will also fail. Can you write a test and fix them?

anutosh491:
Hmm, I think the first two branches do work. We can obviously get them changed but think about this.

Whenever we encounter an if block, something like `if x == pi:`

1) We first transform/modify the if body through `transform_stmts`
2) So once that is done, we don't need to add any new if stmt to pass_result
3) Rather just make a change to the `m_test` field of the if statments .
4) So `if x == pi`, changes to `if basic_eq(x, pi):` and the if body has been modified as per the pass too.

Well we have test_gruntz.py on master  
```
def mmrv(e: S, x: S) -> list[S]:
    if not e.has(x):
        list0: list[S] = []
        return list0
    elif e == x:
        list1: list[S] = [x]
        return list1
    elif e.func == log:
        arg0: S = e.args[0]
        list2: list[S] = mmrv(arg0, x)
        return list2
    else:
        raise
```
which are based on the first two conditions we have. So I think we are doing fine, but just to get rid of any unwanted errors we can address this.

certik:
Ok. If it works then let it be. I am just surprised that the last case didn't work and I don't see how it is different to the previous two cases.

Issue #2481, opened 2024-02-05
Unhandled Exception on non-existent file


rebcabin:
```
(lp) ┌─(~/Documents/GitHub/lpython/integration_tests)───────────────────────────────────────(brian@MacBook-Pro:s001)─┐
└─(11:07:00 on vector-backend)──> lpython bindc03.py                                              ──(Mon,Feb05)─┘
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/brian/Dropbox/Mac/Documents/GitHub/lpython/src/bin/lpython.cpp", line 1738
    throw LCompilers::LCompilersException("No such file or directory: " + arg_file);
LCompilersException: No such file or directory: bindc03.py
(lp) ┌─(~/Documents/GitHub/lpython/integration_tests)───────────────────────────────────────(brian@MacBook-Pro:s00
```

rebcabin:
Seems like a `good first issue` for someone to learn the error-reporting system.

syheliel:
Hi. I create a simple fix for this. Would you like take a look?

binshabenny:
It seems that the bindc03.py file is not present in the current directory.

binshabenny:
It seems that the bindc03.py file is not present in the current directory.#2481

kmr-srbh:
@certik This issue can be closed in favor of #2536 .

PR #2536, opened 2024-02-16
refine input file missing msg

syheliel:
This PR fixs issue #2481 . It prints a more concrete err message and exits immediately instead of throwing an error.

### DIFF
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -1735,7 +1735,8 @@ int main(int argc, char *argv[])
         // the first:
         std::string arg_file = arg_files[0];
         if (CLI::NonexistentPath(arg_file).empty()){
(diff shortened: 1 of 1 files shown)


src/bin/lpython.cpp
```diff
@@ -1735,7 +1735,8 @@ int main(int argc, char *argv[])
         // the first:
         std::string arg_file = arg_files[0];
         if (CLI::NonexistentPath(arg_file).empty()){
-            throw LCompilers::LCompilersException("No such file or directory: " + arg_file);
+            std::cerr << "Your input file doesn't exist. file name is: " << arg_file << std::endl;
```
certik, line 1738: I think this is probably a better message:
```suggestion
            std::cerr << "The input file does not exist: " << arg_file << std::endl;
```

certik:
Thanks!

Issue #2475, opened 2024-02-01
Python file read twice ?


rebcabin:
repro:

lldb lpython -- --show-asr tests/expr8.py

first time it's read is line 128 of lpython.cpp

```
    frame #0: 0x000000010000e794 lpython`(anonymous namespace)::emit_asr(infile="tests/expr8.py", pass_manager=0x000000016fdfe7d8, runtime_library_dir="/Users/brian/Dropbox/Mac/Documents/GitHub/lpython/src/bin/../runtime", with_intrinsic_modules=false, compiler_options=0x000000016fdfe870) at lpython.cpp:200:5
   197 	        std::string input = LCompilers::read_file(infile);
   198 	        lm.init_simple(input);
   199 	        lm.file_ends.push_back(input.size());
-> 200 	    }
   201 	    LCompilers::Result<LCompilers::LPython::AST::ast_t*> r1 = parse_python_file(
   202 	        al, runtime_library_dir, infile, diagnostics, 0, compiler_options.new_parser);
   203 	    std::cerr << diagnostics.render(lm, compiler_options);
```

second time it's read is line 125 of parser.cpp

```
    frame #0: 0x00000001000c9f78 lpython`LCompilers::LPython::parse_python_file(al=0x000000016fdfd2b0, (null)="/Users/brian/Dropbox/Mac/Documents/GitHub/lpython/src/bin/../runtime", infile="tests/expr8.py", diagnostics=0x000000016fdfd298, prev_loc=0, new_parser=true) at parser.cpp:126:49
   123 	    new_parser = true;
   124 	    LCOMPILERS_ASSERT(new_parser)
   125 	    std::string input = read_file(infile);
-> 126 	    Result<LPython::AST::Module_t*> res = parse(al, input, prev_loc, diagnostics);
   127 	    if (res.ok) {
   128 	        ast = (LPython::AST::ast_t*)res.result;
   129 	    } else {
```

intentional?

rebcabin:
I also see a read_file on line 197 of lpython.cpp. I'm still learning what's going on.

certik:
The file should only be read once. If it is read twice, that's a bug that we need to fix. There are many file reads, but only one should be executed.

rebcabin:
Looks like, during parsing, read_file is called at least twice, once to parse and once to `[gs]et_module_path`, and maybe more than once for that purpose. Some of the following breaks are redundant, but not all of them are.

I set breakpoints on every one of these lines:

```
[...]

PR #2541, opened 2024-02-17
Fix duplicate read when paring python asr

syheliel:
fix issue #2475. I create `parse_python_source` to accept source code literal input to avoid additional file read.

Note that there are still mutiple codes sharing the similar logic, like https://github.com/lcompilers/lpython/blob/main/src/bin/lpython.cpp#L254 . If this modification is accepted, I will fix them in this PR later.

### DIFF
--- a/src/lpython/parser/parser.cpp
+++ b/src/lpython/parser/parser.cpp
@@ -112,9 +112,9 @@ std::string unique_filename(const std::string &prefix) {
     return filename;
 }
 
-Result<LPython::AST::ast_t*> parse_python_file(Allocator &al,
+Result<LPython::AST::ast_t*> parse_python_source(Allocator &al,
         const std::string &/*runtime_library_dir*/,
-        const std::string &infile,
+        const std::string &source_code,
         diag::Diagnostics &diagnostics,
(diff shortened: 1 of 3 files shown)


src/lpython/parser/parser.cpp
```diff
@@ -133,5 +132,15 @@ Result<LPython::AST::ast_t*> parse_python_file(Allocator &al,
...
 
+Result<LPython::AST::ast_t*> parse_python_file(Allocator &al,
+        const std::string &runtime_library_dir,
+        const std::string &infile,
+        diag::Diagnostics &diagnostics,
+        uint32_t prev_loc,
+        [[maybe_unused]] bool new_parser) {
+    std::string input_source_code = read_file(infile);
+    return parse_python_source(al, runtime_library_dir, input_source_code, diagnostics, prev_loc, new_parser);
+}
```
Thirumalai-Shaktivel, line 143: Instead, refactor the code to not read in `parse_python_file` function in `parser.cpp`.

Thirumalai-Shaktivel (approved):
Please address the review and mark this PR as ready for review!

Thank you

Issue #2542, opened 2024-02-17
test random failed


certik:
```
167/325 Test #174: test_random_02 ...................***Failed    0.12 sec
0.7547386333122814 0.7539588431983508
Traceback (most recent call last):
  File "/home/runner/work/lpython/lpython/integration_tests/test_random_02.py", line 11, in <module>
    test_seed()
  File "/home/runner/work/lpython/lpython/integration_tests/test_random_02.py", line 9, in test_seed
    assert abs(t1 - t2) > 1e-3
           ^^^^^^^^^^^^^^^^^^^
AssertionError
```

https://github.com/lcompilers/lpython/actions/runs/7939986580/job/21680720142?pr=2540

certik:
The test is flaky: https://github.com/lcompilers/lpython/blob/496f29efc65cddd0f47721ebfed356f86b5f2389/integration_tests/test_random_02.py#L9, one can't test random numbers like that.

I thought we already tackled this issue before with @Shaikh-Ubaid in LFortran. One must test random numbers using tests designed for random numbers: they pass with very high probability (=every time in practice). The above test is low probability, that one can even trigger a failure at a CI. So that's a bad test. We need to improve it.

Shaikh-Ubaid:
A good test exists in LFortran. The test in LPython was more for completeness or as measure of extra safety.

Since, it already has tests in LFortran, do we still need to add a similar test for LPython?

certik:
Just test it in some way that is robust in LPython as well, to ensure that everything works.

syheliel:
I have added a PR for it. The new testcase's aim is to test whther the distribution is uniform. I think it's a more valuable test compared to original version

certik:
I think this is fixed now.

PR #2543, opened 2024-02-18
refine testcase for random

syheliel:
Fix issue #2542 . test the distribution of random. The error rate is lower than 0.1% (tested locally 1000 times without fail)

### DIFF
--- a/integration_tests/test_random_02.py
+++ b/integration_tests/test_random_02.py
@@ -1,11 +1,21 @@
-from lpython import f64
+from lpython import i32, f64
 import random
 
 def test_seed():
-    random.seed()
-    t1: f64 = random.random()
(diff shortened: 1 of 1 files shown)


certik:
For me the random number seed is always the same, so I get always the same answer.

The test works fine, but we need to ensure the random seed is initialized using the system clock at the beginning of the program. That's a separate issue.

The test itself is good. If it ever fails, we can relax the 5% deviation to a higher number. Thanks!

certik:
While playing with this code, I discovered two bugs:

* https://github.com/lcompilers/lpython/issues/2545
* https://github.com/lcompilers/lpython/issues/2544

The second one is just a bug in LPython. The first one however is a bug in the random seed initialization. We need to fix it.

PR #2549, opened 2024-02-18
Seeded random.random() with system clock if it is left unseeded

### LINKED ISSUE #2545: The random seed is the same across invocations
```python
import random

def test_seed():
    print(random.random())
    print(random.random())

test_seed()
```
This gives in LPython:
```console
$ lpython c.py 
7.82636925942561089e-06
1.31537788143166245e-01
$ ./c.out 
7.82636925942561089e-06
1.31537788143166245e-01
$ ./c.out
7.82636925942561089e-06
1.31537788143166245e-01
$ ./c.out
7.82636925942561089e-06
1.31537788143166245e-01
$ ./c.out
7.82636925942561089e-06
1.31537788143166245e-01
```
Compared to CPython:
```console
$ python c.py 
0.1609877035160111
0.6090058937273372
$ python c.py
0.8816157116238759
0.7730379472615071
$ python c.py
0.7851910501722984
0.926580268293728
$ python c.py
0.23538691043891513
0.3621541015534898
$ python c.py
0.09200106922546647
0.8737898472693616
```

The random number generator itself works well, but the seed is always the same. We need to initialize the seed using system clock at the beginning of the program.

advikkabra:
Fixes #2545 .
Has a `_seed_set` variable which checks if the seed has been explicitly set by the user, else will set the default seed to the system clock.


src/runtime/random.py
```diff
@@ -2,44 +2,58 @@
...
         return -x
     return x
 
+
 def random() -> f64:
     """
     Returns a random floating point number in the range [0.0, 1.0)
     """
+    if not _seed_set:
+        _lfortran_init_random_clock()
```
certik, line 38: That's one way to do it, but a better way is to call the `init_random` at the beginning of the program, not at every call. We have infrastructure for this, I think there is a function that is being called at the beginning, so we just need to call this from it. 

Initially we can call it every time, later on we can update it to only call it if `random` is used.

advikkabra: I cannot seem to find this function. Where is it, and where is it being called from?

Thirumalai-Shaktivel (commented):
Please mark this as ready for review once it is ready!

Thanks for working on it.

advikkabra:
@Thirumalai-Shaktivel  I've reverted the changes in `random.py` and created another function in `lfortran_intrinsics` which calls all initial setup. I think the PR is good to go.

Thirumalai-Shaktivel (approved):

Thirumalai-Shaktivel:
@Shaikh-Ubaid, can you review and merge this PR?

Shaikh-Ubaid:
>Initially we can call it every time, later on we can update it to only call it if random is used.

I am concerned about calling it every time even when `random()` is not used. But since we can update it later, I think it is fine.

Shaikh-Ubaid (approved):
It looks fine to me. Thanks for this!

Thirumalai-Shaktivel:
Thanks for the review, Ubaid!

PR #2551, opened 2024-02-21
Fixes #2490: Adds unescaping `\f` and the correct implementation of `isspace()` string method.

### LINKED ISSUE #2490: Unexpected output for string method on whitespace characters
When I run the following code:
```py
def f():
    b: str = " \t\n\v\f\r"
    print(b.isspace())
    print(b.islower())
    print(b.isupper())


f()
```

In Python:
```
True
False
False
```

In LPython:
```
False
True
False
```

Hence there is a bug in both the `isspace()` and the `islower()` methods.

kmr-srbh:
# Fix:
#2490 gave wrong output due to the incorrect handling of the escape sequence '\f'. Escaping the sequence was handled, but unescaping was missed. This led to the string `" \t\n\v\f\r"` being transformed as `" \t\n\v\\f\r"` in the AST. Note the extra "\\" after "v".
Screenshot from 2024-02-21 17-30-11

Hence, the answer in #2490 was thus 'not actually' wrong as non-whitespace and lowercase characters did exist. Fixes #2490 
 
# Improvement
Though the error in the above case was due to incorrect string handling, the `isspace()` method was actually not implemented correctly. The definition of a 'whitespace' character is actually broad. The Python interpreter checks for all of them in it's implementation of the method. I incorporated checking for those characters.

A huge shout-out to @advikkabra for the clever testing. :tada: 

# Note
The Black formatter has automatically formatted the Python file I worked on. Please note that only change I made was the implementation of the `isspace()` method.

Last but not the least, I thoroughly enjoyed working on the issue. :smiley:

### DIFF
--- a/src/libasr/string_utils.cpp
+++ b/src/libasr/string_utils.cpp
@@ -192,16 +192,24 @@ char* str_unescape_c(Allocator &al, LCompilers::Str &s) {
         } else if (s[idx] == '\\' && s[idx+1] == 'v') {
             x += "\v";
             idx++;
-        } else if (s[idx] == '\\' && s[idx+1] == '\\') {
+        }
+        else if (s[idx] == '\\' && s[idx + 1] == 'f') {
+            x += "\f";
+            idx++;
+        }
+        else if (s[idx] == '\\' && s[idx + 1] == '\\') {
             x += "\\";
             idx++;
-        } else if (s[idx] == '\\' && s[idx+1] == '"') {
+        }
+        else if (s[idx] == '\\' && s[idx + 1] == '"') {
             x += '"';
             idx++;
-        } else if (s[idx] == '\\' && s[idx+1] == '\'') {
+        }
+        else if (s[idx] == '\\' && s[idx + 1] == '\'') {
             x += '\'';
             idx++;
-        } else {
+        }
+        else {
             x += s[idx];
         }
     }
--- a/src/runtime/lpython_builtin.py
+++ b/src/runtime/lpython_builtin.py
@@ -1,6 +1,7 @@
-from lpython import (i8, i16, i32, i64, f32, f64, c32, c64, overload, u8,
-                     u16, u32, u64)
-#from sys import exit
+from lpython import i8, i16, i32, i64, f32, f64, c32, c64, overload, u8, u16, u32, u64
+
+# from sys import exit
+
 
 #: abs() as a generic procedure.
 #: supported types for argument:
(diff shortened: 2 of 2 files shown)


kmr-srbh:
The failing test is the one mentioned in #2542.

Thirumalai-Shaktivel:
What is the status of this PR?

kmr-srbh:
> What is the status of this PR?

It is ready for review @Thirumalai-Shaktivel.

src/runtime/lpython_builtin.py
```diff
@@ -974,16 +1103,18 @@ def _lpython_str_isupper(s: str) -> bool:
                 return False
     return is_cased_present
 
+
```
Thirumalai-Shaktivel, line 1106: Remove all the formatting changes and you can submit a new PR for it.

Thirumalai-Shaktivel, line 1106: Remove all the formatting changes and you can submit a new PR for it.

src/runtime/lpython_builtin.py
```diff
@@ -994,12 +1125,50 @@ def _lpython_str_isascii(s: str) -> bool:
...
     for ch in s:
-        if ch != ' ' and ch != '\t' and ch != '\n' and ch != '\r' and ch != '\f' and ch != '\v':
+        ch_ord = ord(ch)
+        # A Unicode character is considered a 'whitespace' if it has has a bidirectional
+        # type 'WS', 'B' or 'S'; or the category 'Zs'.
+        # Check every character in the string to see if it is a whitespace.
+        # If none of the characters match, we have atleast one character which is not
+        # a whitespace, so we return `False`.
+        if (
+            ch_ord != 9
```
Thirumalai-Shaktivel, line 1142: Thanks for handling all the cases, but it doesn't seems to be more readable.
By readable what I meant is, I have to manually check what `9` is, and confirm that `\t` is handled.
 
Was there any issue with previous changes?
Also, instead of this long condition check, why not keep the strings in list and have a single condition check?

src/libasr/string_utils.cpp
```diff
@@ -192,16 +192,24 @@ char* str_unescape_c(Allocator &al, LCompilers::Str &s) {
         } else if (s[idx] == '\\' && s[idx+1] == 'v') {
             x += "\v";
             idx++;
-        } else if (s[idx] == '\\' && s[idx+1] == '\\') {
+        }
+        else if (s[idx] == '\\' && s[idx + 1] == 'f') {
+            x += "\f";
+            idx++;
+        }
+        else if (s[idx] == '\\' && s[idx + 1] == '\\') {
```
Thirumalai-Shaktivel, line 200: Please remove the formatting changes.

Thirumalai-Shaktivel (commented):
Also, add a test

Thirumalai-Shaktivel:

PR #2556, opened 2024-02-27
Implemented Lshift

Kishan-Ved:
I thought this would be a good one to get started with LPython implementations.

Reference: https://docs.python.org/3/library/operator.html#operator.lshift


Kishan-Ved:
@certik Sir the error seems weird. Can we rerun this? 

Also, I noticed that many intrinsic functions like Trailz don't have integration tests.

integration_tests/CMakeLists.txt
```diff
@@ -816,6 +816,7 @@ RUN(NAME callback_04       IMPORT_PATH .. LABELS cpython)
 # Intrinsic Functions
 RUN(NAME intrinsics_01     LABELS cpython llvm NOFAST) # any
 RUN(NAME intrinsics_02     LABELS cpython llvm c) # floordiv
+RUN(NAME intrinsics_03     LABELS llvm) # lshift
```
Thirumalai-Shaktivel, line 819:
```suggestion
RUN(NAME intrinsics_03     LABELS cpython llvm) # lshift
```

This verifies that the test works with python as well.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -2382,6 +2384,81 @@ namespace Mod {
 
 } // namespace Mod
 
+#define i_BitRshift(n, bits, t) EXPR(ASR::make_IntegerBinOp_t(al, loc,      \
+            n, ASR::binopType::BitRShift, bits, t, nullptr))
+#define i_BitLshift(n, bits, t) EXPR(ASR::make_IntegerBinOp_t(al, loc,      \
+            n, ASR::binopType::BitLShift, bits, t, nullptr))
```
Thirumalai-Shaktivel, line 2390: Move to asr_builder.h

Thirumalai-Shaktivel: Sorry, my bad, we haven't synced LFortran changes. 
Just move this above where other macros are residing.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -2382,6 +2384,81 @@ namespace Mod {
...
+            "ASR Verify: Arguments to LShift must be of integer type",
+            x.base.base.loc, diagnostics);
+    }
+
+    static ASR::expr_t *eval_LShift(Allocator &al, const Location &loc,
+            ASR::ttype_t* t1, Vec<ASR::expr_t*> &args) {
+            int64_t a = ASR::down_cast<ASR::IntegerConstant_t>(args[0])->m_n;
+            int64_t b = ASR::down_cast<ASR::IntegerConstant_t>(args[1])->m_n;
+            int64_t val = a << b;
+            return make_ConstantWithType(make_IntegerConstant_t, val, t1, loc);
```
Thirumalai-Shaktivel, line 2410:
```suggestion
        int64_t a = ASR::down_cast<ASR::IntegerConstant_t>(args[0])->m_n;
        int64_t b = ASR::down_cast<ASR::IntegerConstant_t>(args[1])->m_n;
        int64_t val = a << b;
        return make_ConstantWithType(make_IntegerConstant_t, val, t1, loc);
```

Thirumalai-Shaktivel (commented):
Looks good, I have left some comments.

Thirumalai-Shaktivel:

Kishan-Ved:
The test fails for cpython, may I please know the reason. The error is:
```
  File "/home/kishan/Desktop/lpython/integration_tests/intrinsics_03.py", line 8, in <module>
    print(lshift(a,b))
NameError: name 'lshift' is not defined
```
The file runs when I do `./src/bin/lpython integration_tests/intrinsics_03.py` and gives the correct  output.

src/libasr/pass/intrinsic_function_registry.h
Kishan-Ved: Done.

src/libasr/pass/intrinsic_function_registry.h
Kishan-Ved: Done.

integration_tests/CMakeLists.txt
Kishan-Ved: Please check the comment below: https://github.com/lcompilers/lpython/pull/2556#issuecomment-1983179803

Thirumalai-Shaktivel:
It fails because python doesn't have builtin lshift support. I think it needs to imported from the operator module, see: https://docs.python.org/3/library/operator.html#operator.lshift. Something like:
```py
from operator import lshift
print(lshift(x, y))
```


Does `x << y` work in lpython?

Kishan-Ved:
> It fails because, python doesn't have builtin lshift support. I think it needs to be imported from the operator module, see: https://docs.python.org/3/library/operator.html#operator.lshift. Something like:
> 
> ```python
> from operator import lshift
> print(lshift(x, y))
> ```
> 
> Does `x << y` work in lpython?

I want lshift to be imported only for the python test. For LPython's implementation, I want the control to be redirected to my intrinsic function. How can I achieve this?

Kishan-Ved:
cc @Thirumalai-Shaktivel @Shaikh-Ubaid

Issue #2554, opened 2024-02-24
String to int conversion in ASR


certik:
Figure out a clean way to represent and implement string to int conversions in ASR.

Then figure out how to hook it up in LPython.

Discussed at https://lfortran.zulipchat.com/#narrow/stream/311866-LPython/topic/String.20to.20Int.20conversion.

Shaikh-Ubaid:
It looks like we already have some support for string to int conversion.
```console
% cat examples/expr2.py 
def main():
    x: str = "123"
    print(int(x))

main()
% python examples/expr2.py         
123
% lpython_in_main examples/expr2.py
123
```

Shaikh-Ubaid:
I think we just need to polish the existing support.

kmr-srbh:
@Shaikh-Ubaid I think we should support the _conversion_ mechanism which `intrinsic_int` uses, for all the integer types we have. My main concern is that doing `i32(int("12345))` is counterintuitive.

kmr-srbh:
The conversion mechanism for `int` needs to be improved too. It currently relies on `std::stoi()` and fails for anything larger than that. We only have to improve the `BigInt` library we already have.

PR #2558, opened 2024-02-27
Implement `math.factorial(n)` to calculate very large factorials with speed

kmr-srbh:
# Overview
- The current implementation of `math.factorial()` is naive. 
- Returns 0 for large factorials due to integer overflow.

Screenshot from 2024-02-27 23-12-57

# Solution
The industry standard for calculating factorial of a large number is complicated. It deals with Prime Number generation and Swing Numbers. These methods require `BigInt` for storing the large calculated value.

One fast method for going about calculating the calculation is to use an array to store the digits of the number as strings, in reverse, and do digit by digit multiplication. This method is used for implementing the algorithm.

I support the implementation of the Swing Number algorithms if one can understand and implement it correctly. The funny thing is that a Python implementation is available online.

# Improvement
- Works for large factorials
Screenshot from 2024-02-27 23-17-56
- Works for very large factorials
Screenshot from 2024-02-27 23-19-13
- Accurate
Screenshot from 2024-02-27 23-22-05

Integration tests have not been added due to the prerequisite below.

# Prerequisite
1. While the function is complete, it depends on #2554 for getting merged. Currently it just prints the output to stdout and returns 0.
2. As it currently stands, the implementation for handling values larger than `2^63 - 1` is handled by the BigInt module we have. It is broken and returns pointers to the number instead. A `vector<char>` is required for storing and working with very large values.

Note: The long assignment of digits is because LPython currently does not support list assignment like: `my_list[0:4] = [1, 2, 3]`. The assignments can be improved through the introduction of the above list assignment.


kmr-srbh:
I think the errors are okay for now because the `factorial` function returns only 0. This will be addressed.

src/runtime/math.py
```diff
@@ -15,36 +15,169 @@ def modf(x: f64) -> tuple[f64, f64]:
...
+                carry = carry // 10
+                f_size += 1
+            i += 1
+
+    result: str = ""
+    idx: i32
+    for idx in range(f_size - 1, -1, -1):
+        result += str(f[idx])
+    print(result)
+    return i64(0)
```
faze-geek, line 97: I do understand the approach, here the final result is printed out using concatenation.
I have not looked into the pre-requisites but even after a conversion from string to int, how will you return an integer here? The highest supported annotation we have is `i64`, which has a maximum value of `2^63 - 1`. 
This is not large enough to store anything over `20!`, which means this will still encounter an overflow and return a garbage value.

faze-geek: I have tried something similar in another project, let me know if there are any other approaches. Thanks :)

kmr-srbh: @faze-geek I understand your point. Internally in LPython, the method used for handling values larger than 2 ^ 63 - 1 is handled by the BigInt module we have. The part of the module which does the handling for large numbers is broken and returns a pointer to the value instead. I propose implementing the vector and string storage type for handling large values. Only then will we be able to handle values like `20!`.

I had forgotten to mention this prerequisite. Thanks for reminding me. I am adding it above.

kmr-srbh: > I have tried something similar in another project, let me know if there are any other approaches. Thanks :)

Oh yes there are! As I had mentioned above, the industry standard for scientific computing applications is to use the [PrimeSwing algorithm](https://www.luschny.de/math/factorial/SwingFactorialPy.html). The Python implementation is available online, but I do not want to type something I do not understand. :smile: 

The Python `math.factorial(n)` uses (this)[http://www.luschny.de/math/factorial/binarysplitfact.html] algorithm.

faze-geek: Thanks. Will surely explore this once the `BigInt` module gets working.

kmr-srbh:
@certik On further thought, I realized that to handle values large as shown above, we need to introduce a new data type BigInt where we store the number as digits in base 2 ^ 32 in a vector. The `vector<string>` method is not that fast. What is your take on this? If not going through BigInt, what else can we do? Please guide me.

If yes, please guide me on how can one go about creating a new data type?

kmr-srbh:
Closing this as not planned for now.

PR #2559, opened 2024-02-27
Add support for accessing values from `Const`

kmr-srbh:
# Overview
Currently, trying to access a key from a `Const dict` throws a `semantic error` stating type mismatch. An `ASR pass error` is thrown when trying to access values from a `Const list`. This was chiefly because the case of a `Const` value was not addressed in handling subscript indices. The error exists for `Const` types in general and will be addressed in subsequent commits.

Screenshot from 2024-02-26 12-21-43

# Fix
This PR checks for `Const` when visiting subscript indices and handles access through the contained type. 

#Prerequisite
This PR depends on accessing the `expr_t` of the contained type inside `Const`. I have tried high and low but could not work out a fix. The variable in question is `ASR::expr_t* value` which is passed to `visit_SubscriptIndices`.

The issue is:
- Consider a dictionary declaration: `Const[dict[str, i32]] = {"a": 1, "b": 2}`.
- `value` contains `Const[dict[str, i32]]`. 

We need `value` to get to store `dict[str, i32]`.

### DIFF
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -3680,7 +3680,8 @@ public:
         ai.m_left = nullptr;
         ai.m_right = nullptr;
         ai.m_step = nullptr;
-        if (AST::is_a<AST::Slice_t>(*m_slice)) {
+        if (AST::is_a<AST::Slice_t>(*m_slice))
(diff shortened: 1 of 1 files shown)


kmr-srbh:
The errors are okay because of the problem with `value` variable. See above.

PR #2564, opened 2024-02-29
`--jit` option to execute without creation of executable file

### LINKED ISSUE #2563: [feature request] Ability to execute the code without compiling using llvm JIT
It would be very useful if we could directly execute the `lpython` code without compiling it using the llvm JIT. It will prevent the creation of `*.tmp.o` and the executable files.

The main idea is to run `lpython --jit program.py`

---

I have a working implementation ready. If this is useful I could open a PR.

Vipul-Cariappa:
Using LLVM JIT.

Fixes #2563.

Example:
With `--jit` flag
```bash
❯ time ./src/bin/lpython --jit ./examples/expr2.py
25
./src/bin/lpython --jit ./examples/expr2.py  0.01s user 0.00s system 97% cpu 0.013 total
```
Without
```bash
❯ time ./src/bin/lpython ./examples/expr2.py      
25
./src/bin/lpython ./examples/expr2.py  0.02s user 0.01s system 99% cpu 0.037 total
```


src/bin/lpython.cpp
```diff
@@ -840,6 +840,106 @@ int compile_python_to_object_file(
...
+    }
+
+    auto llvm_start = std::chrono::high_resolution_clock::now();
+    e.add_module(std::move(res.result));
+    try {
+        e.voidfn("__module___main_____main__global_stmts");
+    } catch (LCompilers::LCompilersException&) {}
+    auto llvm_end = std::chrono::high_resolution_clock::now();
+    times.push_back(std::make_pair("LLVM to binary", std::chrono::duration<double, std::milli>(llvm_end - llvm_start).count()));
+    print_time_report(times, time_report);
```
Thirumalai-Shaktivel, line 939: Let's do this only for LLVM backend.
This seems duplicated, try utilising or simplifying the function: `compile_python_to_object_file`

Thirumalai-Shaktivel (commented):
Also, I think we should test this, try resolving my review above, later we will figure out how to test this.

Thirumalai-Shaktivel:

src/bin/lpython.cpp
```diff
@@ -1739,6 +1841,14 @@ int main(int argc, char *argv[])
             return 1;
         }
 
+        if (to_jit) {
+            compiler_options.emit_debug_info = false;
+            compiler_options.emit_debug_line_column = false;
+            compiler_options.generate_object_code = false;
+            return execute_python_using_jit(arg_file, runtime_library_dir,
+                    lpython_pass_manager, compiler_options, time_report);
+        }
```
Shaikh-Ubaid, line 1850: > One of them says error: use of undeclared identifier 'execute_python_using_jit'. Even though it is declared and defined in the preceding lines.

This should be wrapped inside `#ifdef HAVE_LFORTRAN_LLVM` `#endif`. Hopefully, it should fix the CI failure.

Vipul-Cariappa:
@Thirumalai-Shaktivel, Regarding testing out this feature. I was thinking to add the following

```python
run_test(
                f"{filename} JIT",
                "llvm",
                "lpython --jit {infile}",
                filename,
                update_reference)
```

after [llvm test at run_tests.py](https://github.com/lcompilers/lpython/blob/main/run_tests.py#L115).

But I get 

```text
raise FileNotFoundError(
FileNotFoundError: The reference json file 'tests/reference/llvm-expr14-f21489f.json' for expr14.py JIT does not exist
```
when I run it.

I do not properly understand the methodology used to test existing features. Suggestion and guidance would be helpful here.

Thirumalai-Shaktivel:
I think we need to test this using `integration_tests/run_tests.py`. It is used for testing most of our runtime tests.
Checkout `integration_tests/CMakeLists.txt`. If you are blocked, do let me know.

Thirumalai-Shaktivel (commented):
The implementation looks good so far.

Shaikh-Ubaid (commented):
This looks good! Thanks for this! Now, you just need to add an integration test for this. You can leverage the existing `EXTRA_ARGS` support. For example:

```diff
diff --git a/integration_tests/CMakeLists.txt b/integration_tests/CMakeLists.txt
index 44b703eae..1db5d10b8 100644
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -313,11 +313,10 @@ endmacro(RUN_UTIL)
 macro(RUN)
     set(options FAIL NOFAST NOMOD ENABLE_CPYTHON LINK_NUMPY NO_WARNINGS)
     set(oneValueArgs NAME IMPORT_PATH COPY_TO_BIN REQ_PY_VER)
-    set(multiValueArgs LABELS EXTRAFILES)
+    set(multiValueArgs LABELS EXTRAFILES EXTRA_ARGS)
     cmake_parse_arguments(RUN "${options}" "${oneValueArgs}"
                         "${multiValueArgs}" ${ARGN} )
 
-    set(RUN_EXTRA_ARGS "")
     set(RUN_FILE_NAME ${RUN_NAME})
 
     if (RUN_LINK_NUMPY)
@@ -826,3 +825,6 @@ COMPILE(NAME import_order_01     LABELS cpython llvm c) # any
 
 # LPython emulation mode
 RUN(NAME lpython_emulation_01    LABELS cpython NOMOD)
+
+RUN(NAME jit_01 LABELS cpython llvm
+    EXTRA_ARGS --jit)
```

Shaikh-Ubaid:
We need an integration test for this. Please mark this as "Ready for review" when ready. Thanks for working on this! I appreciate it.

integration_tests/CMakeLists.txt
```diff
@@ -101,6 +101,9 @@ macro(RUN_UTIL RUN_FAIL RUN_NAME RUN_FILE_NAME RUN_LABELS RUN_EXTRAFILES RUN_NOM
             set_target_properties(${name} PROPERTIES LINKER_LANGUAGE C)
             target_link_libraries(${name} lpython_rtlib)
             add_test(${name} ${CMAKE_CURRENT_BINARY_DIR}/${name})
+            add_test(
+                NAME "${name}_jit"
+                COMMAND ${LPYTHON} --jit ${extra_args} ${CMAKE_CURRENT_SOURCE_DIR}/${file_name}.py)
```
Shaikh-Ubaid, line 106: Are you facing any issue/challenges in implementing the approach shared in https://github.com/lcompilers/lpython/pull/2564#pullrequestreview-1926239513?

Vipul-Cariappa:
We expect 
```text
exit_02_jit
exit_04_jit
exit_02b_jit
exit_02c_jit

expr_18_jit
if_03_jit
```
to fail (i.e. exit with non-zero code).

What should I do here? When the exit code is supposed to be non-zero?

---

```text
const_03_jit
expr_13_jit
test_c_interop_02_jit
test_c_interop_03_jit
test_c_interop_05_jit
bindc_03_jit
bindc_05_jit
bindc_06_jit
structs_07_jit
structs_13_jit
structs_18_jit
structs_19_jit
structs_20_jit
sizeof_01_jit
enum_05_jit
````
fails because it has an external dependency. 

I believe we should skip these tests under JIT.

---

`test_argv_01_jit` and `print_list_tuple_03_jit` look like proper errors. Debugging required.

File: integration_tests/CMakeLists.txt (hunk 3 above)
Comment (Vipul-Cariappa) [reply integration_tests/CMakeLists.txt:106]: If I use that approach, then `RUN(NAME jit_01 LABELS cpython llvm EXTRA_ARGS --jit)` with

```cmake
add_custom_command(
                OUTPUT ${name}.o
                COMMAND ${LPYTHON} -c ${extra_args} ${CMAKE_CURRENT_SOURCE_DIR}/${file_name}.py -o ${name}.o
                DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${file_name}.py
                VERBATIM)
```

will produce

```cmake
add_custom_command(
    OUTPUT test_file.o
    COMMAND lpython -c --jit path/to/test_file.py -o test_file.o
    ...)
```

But while using `--jit` no `.o` file is generated. And `cmake` complains about it. Also, that would not technically be testing the JIT. It will fail immediately and exit the remaining tests because it would be part of the build step, not the test step.

I hope you could understand my explanation. Let me know if there are things in my explanation you could not understand.

integration_tests/CMakeLists.txt
```diff
@@ -360,6 +363,13 @@ macro(RUN)
         list(REMOVE_ITEM RUN_LABELS cpython cpython_sym) # remove cpython, cpython_sym, from --fast test
         RUN_UTIL(RUN_FAIL RUN_NAME RUN_FILE_NAME RUN_LABELS RUN_EXTRAFILES RUN_NOMOD RUN_EXTRA_ARGS RUN_COPY_TO_BIN)
     endif()
+    
+    # if (llvm IN_LIST RUN_LABELS)
+    #     unset(RUN_EXTRA_ARGS)
+    #     set(RUN_EXTRA_ARGS --jit)
+    #     set(RUN_NAME "${RUN_NAME}_JIT")
+    #     RUN_UTIL(RUN_FAIL RUN_NAME RUN_FILE_NAME RUN_LABELS RUN_EXTRAFILES RUN_NOMOD RUN_EXTRA_ARGS RUN_COPY_TO_BIN)
```
Vipul-Cariappa, line 371: I didn't intend to push this. I was testing stuff out in my local computer, I should have removed this before pushing. Sorry. Will remove it in the next commit.

Vipul-Cariappa:
@Thirumalai-Shaktivel, @Shaikh-Ubaid Look at #2595. If you could verify that they are definitely memory-related bugs it would be helpful.
I guess progress on this PR would be halted till we figure out and resolve #2595.!?

---

I did not look at the AST and ASR. I will investigate them tomorrow.

integration_tests/CMakeLists.txt
Shaikh-Ubaid: Oh, I see. Thanks for the explanation. I think then a better/simpler approach to test this could/would be by adding it as a reference test.

We might also be able to test this by using tests in `src/lpython/tests`, but that might be slightly involved. I think adding it as a reference test is a better approach.

Shaikh-Ubaid:

Vipul-Cariappa:
I believe this PR is ready for review/merge.
I will explain some aspects of the modifications I have made to run the tests:
- Add necessary options in the `cmake`(-JIT) tests to skip tests that depend on external C/C++ code. (JIT compiling code with external dependency does not make much sense. We cannot load in a static library at runtime.)
- Also testing integration with `symengine`. I am dynamically loading in `symengine.so` using `dlopen` if `--enable-symengine` is passed.
- Similarly, I am dynamically loading in `python3.so` if `--enable-cpython` is passed.
> Note: LLVM backend does not support `BindPython` ABI yet.

There are only 2 tests that fail. They fail due to segfault. And related to #2595. They are not caused by any new bug I have introduced. My code happens to trigger it.

I tried investigating the segfault, but have not found its cause yet. It is very difficult to debug the LLVM IR directly. Presently I am trying to do that. Also fixing them should not be part of this PR.

---

Also, the Windows CI is not running to completion. Please read [this comment](https://lfortran.zulipchat.com/#narrow/stream/197339-general/topic/Windows.20CI.20looks.20broken) I made in the zulip chat.

Shaikh-Ubaid:
@Vipul-Cariappa From a high level it looks good. I will review it thoroughly tomorrow.

>There are only 2 tests that fail. They fail due to segfault. And related to https://github.com/lcompilers/lpython/issues/2595. They are not caused by any new bug I have introduced. My code happens to trigger it.

We will need the CI to be green to merge this. So, we either need to get the two failing tests to work or we need to completely skip running these two tests with `--jit`(I think right now we just skip checking the return code).

certik:
Thanks!

The advantages of this approach:

* faster
* no intermediate files
* gets rid of the Clang dependency for linking

Disadvantages:

* Now we need to handle the loading of shared libraries (cpython and symengine, if used), and we do not allow static libraries

The handling of shared libraries is something we will need for interactive mode as well, so I think we need this. The current implementation works for Conda, but it will fail for, say, Ubuntu, which uses a `lib64` (I think) prefix. We will need to add options to select where these libraries are installed. This code should be refactored into some utility file, so that we don't have these platform dependent ifdefs in the main driver.


To finish this, let's get all tests passing, and then let's polish the implementation, so that we can merge it. Then we can improve upon it.

Vipul-Cariappa:
> To finish this, let's get all tests passing

> we need to completely skip running these two tests with --jit(I think right now we just skip checking the return code).

I am trying to debug and fix #2595. It would be better to fix the issues and then merge this PR. Skipping the failing tests will require quite a lot of modifications to the `integration_tests/CMakeLists.txt`. Which should later be undone, when the issue is fixed.
Please have a look at the recent comments I have made at #2595 (Help Required).

Vipul-Cariappa:
e7760b02c7fbfccba9f2384278fe8b28e0afe3b7 fixes #2595
I tested with both AddressSanitizer and Valgrind.

Vipul-Cariappa:
Ready for review. @Shaikh-Ubaid @certik

src/bin/lpython.cpp
```diff
@@ -882,12 +885,47 @@ int compile_python_to_object_file(
...
+    std::unique_ptr<LCompilers::LLVMModule> m = std::move(res.result);
+
+    if (to_jit) {
+        LCompilers::LPython::DynamicLibrary cpython_lib;
+        LCompilers::LPython::DynamicLibrary symengine_lib;
+
+        if (compiler_options.enable_cpython)
+            LCompilers::LPython::open_cpython_library(cpython_lib);
+        if (compiler_options.enable_symengine)
+            LCompilers::LPython::open_symengine_library(symengine_lib);
```
Shaikh-Ubaid, line 897: I think we should use braces so that it is consistent with the rest of the code base.
```suggestion
        if (compiler_options.enable_cpython) {
            LCompilers::LPython::open_cpython_library(cpython_lib);
        }
        if (compiler_options.enable_symengine) {
            LCompilers::LPython::open_symengine_library(symengine_lib);
        }
```

certik (approved):
I think this looks great in general. Let's test this locally by a few people before we merge.

src/bin/lpython.cpp
Shaikh-Ubaid: We should update this at every place.

src/bin/lpython.cpp
```diff
@@ -1911,6 +1959,21 @@ int main(int argc, char *argv[])
         if (endswith(arg_file, ".py"))
         {
             int err = 0;
+            if (to_jit) {
+#ifdef HAVE_LFORTRAN_LLVM
+                if (backend != Backend::llvm)
+                    std::cerr << "JIT option is only available with LLVM backend" << std::endl;
+
```
Shaikh-Ubaid, line 1966:
```suggestion
                if (backend != Backend::llvm) {
                    std::cerr << "JIT option is only available with LLVM backend" << std::endl;
                    return 1;
                }
```

src/bin/lpython.cpp
```diff
@@ -1911,6 +1959,21 @@ int main(int argc, char *argv[])
...
+                compiler_options.emit_debug_info = false;
+                compiler_options.emit_debug_line_column = false;
+                compiler_options.generate_object_code = false;
+                return compile_python_using_llvm(arg_file, "", runtime_library_dir,
+                        lpython_pass_manager, compiler_options, time_report, false, true);
+#else
+                std::cerr << "Just-In-Time Compilation of Python files requires the LLVM backend to be enabled."
+                             " Recompile with `WITH_LLVM=yes`." << std::endl;
+#endif
```
Shaikh-Ubaid, line 1975:
```suggestion
                std::cerr << "Just-In-Time Compilation of Python files requires the LLVM backend to be enabled."
                             " Recompile with `WITH_LLVM=yes`." << std::endl;
                return 1;
#endif
```

Shaikh-Ubaid (approved):
Shared two comments. Rest looks great. I checked locally and it seems to work. Amazing work @Vipul-Cariappa!

Shaikh-Ubaid:
@Vipul-Cariappa  I think we can merge this once the above two comments are handled. Thanks for fixing this!

Shaikh-Ubaid:
A slight concern I have is the printing of skipping of tests. Previously I think the output of `./integration_tests/run_tests.py` used to be clean. Now it seems to be cluttered a bit.
```console
Label Time Summary:
c           =  77.43 sec*proc (576 tests)
cpython     =  94.00 sec*proc (650 tests)
llvm        =  98.89 sec*proc (680 tests)
wasm        =   8.98 sec*proc (74 tests)
wasm_x64    =   2.18 sec*proc (44 tests)
wasm_x86    =   0.91 sec*proc (22 tests)

Total Test time (real) =  13.09 sec

The following tests did not run:
	100 - const_03_jit (Skipped)
	128 - expr_13_jit (Skipped)
	410 - test_c_interop_02_jit (Skipped)
	412 - test_c_interop_03_jit (Skipped)
	416 - test_c_interop_05_jit (Skipped)
	418 - bindc_03_jit (Skipped)
	420 - bindc_05_jit (Skipped)
	422 - bindc_06_jit (Skipped)
	480 - structs_07_jit (Skipped)
	492 - structs_13_jit (Skipped)
	502 - structs_18_jit (Skipped)
	504 - structs_19_jit (Skipped)
	506 - structs_20_jit (Skipped)
	538 - sizeof_01_jit (Skipped)
	550 - enum_05_jit (Skipped)
	650 - test_argv_01_jit (Skipped)
```
@certik Do you have any thoughts on this?

Shaikh-Ubaid:
Overall I think the PR is good to merge.

Vipul-Cariappa:
@Shaikh-Ubaid, I believe the log of skipping tests will be a reminder to implement the necessary changes required to support missing JIT features., mainly interoperability with C.

Shaikh-Ubaid:
@Vipul-Cariappa Can you rebase the branch and resolve any conflicts?

Shaikh-Ubaid (approved):

PR #2565, opened 2024-02-29
Fixing symbolic pass according to error caught in LFortran

### LINKED ISSUE #2553: Symbolic Pass failing in LFortran
This issue was brought up here (https://github.com/lfortran/lfortran/pull/3432#discussion_r1496328641)

We were lucky to not have encountered this issue in Lpython itself. Consider the following 
```
def main0():
    x: S = Symbol('x')
    y: S = Symbol('y')
    z: S = x + y
    print(z.func == Add)

main0()
```
So this is an example which is solved by the code in consideration 
https://github.com/lcompilers/lpython/blob/7796f6d91f59b38a671056cb21859c7156cd6da7/src/libasr/pass/replace_symbolic.cpp#L786-L788

anutosh491:
Fixes #2553 

Anytime before we make a call to `process_attributes`, it is essential that we check whether the logical intrinsic function is symbolic based, and that is being done in this PR through a bool check.

### DIFF
--- a/src/libasr/pass/replace_symbolic.cpp
+++ b/src/libasr/pass/replace_symbolic.cpp
@@ -266,6 +266,25 @@ public:
             {handle_argument(al, loc, value_01), handle_argument(al, loc, value_02)},
             ASRUtils::TYPE(ASR::make_Logical_t(al, loc, 4)));
     }
+
+    static inline bool is_logical_intrinsic_symbolic(ASR::expr_t* expr) {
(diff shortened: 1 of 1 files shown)


anutosh491:
cc @HarshitaKalani I think once this PR is merged and the the sync from lpython to lfortran is completed, you can go ahead and enable the symbolic pass back in lfortran. Thanks for bringing this bug to my notice.

anutosh491:
@certik I think this would do. If you see any vulnerabilites here, I shall address it in subsequent PRs. Merging !

PR #2568, opened 2024-03-03
Handling leading zeros in integer literals

### LINKED ISSUE #2002: CPython does not admit leading 0 in ints without an `o` or `[xX]`
repro:

```python
    assert 07 == 7
    assert 08 == 8
```

CPython throws, but LPython admits them

```bash
(lp) ┌─(~/CLionProjects/lpython/lasr/LP-pycharm)─────────────────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s012)─┐
└─(10:12:25 on brian-lasr ✹ ✭)──> PYTHONPATH='../../src/runtime/lpython' python lasr_lexer.py                                                             ──(Thu,Jun22)─┘
  File "/Users/brian/CLionProjects/lpython/lasr/LP-pycharm/lasr_lexer.py", line 234
    assert 07 == 7
           ^
SyntaxError: leading zeros in decimal integer literals are not permitted; use an 0o prefix for octal integers
```

hankluo6:
Fixed #2002 

I catch this error in the tokenizer because it removes trailing zeros in `lex_int`. Also, to handle the case of all zeros, we must allow the tokenizer to process the number first.

### DIFF
--- a/src/lpython/parser/tokenizer.re
+++ b/src/lpython/parser/tokenizer.re
@@ -89,7 +89,7 @@ void lex_int(Allocator &al, const unsigned char *s,
         s = s + 2;
         uint64_t n = get_value((char*)s, 2, loc);
         u.from_smallint(n);
-    } else if ((std::tolower(s[1]) == 'o')) {
+    } else if (std::tolower(s[1]) == 'o') {
         // Oct
         s = s + 2;
         uint64_t n = get_value((char*)s, 8, loc);
--- a/src/lpython/parser/tokenizer.re
+++ b/src/lpython/parser/tokenizer.re
@@ -96,6 +96,10 @@ void lex_int(Allocator &al, const unsigned char *s,
         u.from_smallint(n);
     } else {
         lex_dec_int_large(al, s, e, u);
+        if (s[0] == '0' && u.n != 0) {            
+            throw parser_local::TokenizerError(
+                "Leading zeros in decimal integer are not allowed", {loc});
+        }
     }
     return;
 }
--- a/tests/errors/test_literal.py
+++ b/tests/errors/test_literal.py
@@ -1,0 +1,2 @@
+def test_literal1():
+    x: i32 = 0123
--- a/tests/reference/tokens-test_literal-e20c024.json
+++ b/tests/reference/tokens-test_literal-e20c024.json
@@ -1,0 +1,13 @@
+{
+    "basename": "tokens-test_literal-e20c024",
+    "cmd": "lpython --no-color --show-tokens {infile} -o {outfile}",
+    "infile": "tests/errors/test_literal.py",
+    "infile_hash": "ac9e219faa40c6554983087e8ac198c323f2e5af284b8689f5608f76",
+    "outfile": null,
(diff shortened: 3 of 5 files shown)


src/lpython/parser/tokenizer.re
```diff
@@ -89,13 +89,17 @@ void lex_int(Allocator &al, const unsigned char *s,
...
         u.from_smallint(n);
-    } else if ((std::tolower(s[1]) == 'o')) {
+    } else if (std::tolower(s[1]) == 'o') {
         // Oct
         s = s + 2;
         uint64_t n = get_value((char*)s, 8, loc);
         u.from_smallint(n);
     } else {
         lex_dec_int_large(al, s, e, u);
+        if (s[0] == '0' && u.n != 0) {
```
kmr-srbh, line 99: Hello @hankluo6 ! Thanks for this!

I think this is not the right file to make the changes. Your idea is correct. Please make the same changes in `tokenizer.cpp`. :+1: 

Welcome to LPython!

kmr-srbh (changes requested):

kmr-srbh:
@certik Why is the `tokenizer.cpp` file added to gitignore? How are tokenizer bugs to be fixed then? Please guide.

src/lpython/parser/tokenizer.re
hankluo6: Hi @kmr-srbh 

Thanks for your feedback! I think `tokenizer.cpp` is generated from `tokenizer.re` using `re2c` so I made change on this file. Please let me know if I'm mistaken.

mrdaybird:
`tokenizer.cpp` file is an auto-generated file, created by re2c when you run the build0.sh script using the `tokenizer.re` file. Any changes needed, will have to be made to the `tokenizer.re` file.

kmr-srbh:
Oh! I am sorry. @mrdaybird is right. I had tested your changes through the `tokenizer.cpp` file and built using `build1.sh`. I was mistaken in my understanding of the auto-generation part. I am sorry.

Thanks for the correction @mrdaybird !

Shaikh-Ubaid (commented):
@hankluo6 This looks good to me! Thanks for this! Please add an error reference test (for example as in https://github.com/lcompilers/lpython/commit/cbb79b66655c0c6e78eedf90626a90f0e96b79a8) and we can merge this.

After adding a test, you need to update the reference tests using `./run_tests.py -u` and commit the changes/updations.

Thirumalai-Shaktivel:

hankluo6:
> @hankluo6 This looks good to me! Thanks for this! Please add an error reference test (for example as in [cbb79b6](https://github.com/lcompilers/lpython/commit/cbb79b66655c0c6e78eedf90626a90f0e96b79a8)) and we can merge this.
> 
> After adding a test, you need to update the reference tests using `./run_tests.py -u` and commit the changes/updations.

@Shaikh-Ubaid @Thirumalai-Shaktivel I added a new test, could you review it? Thanks!

Shaikh-Ubaid (approved):
Perfect! Thanks! Great work @hankluo6!

PR #2574, opened 2024-03-04
Implemented `hypot` in numpy

Kishan-Ved:
- In this PR, I have added functionality for `hypot()` in numpy. 
- This can now be imported using: `from numpy import hypot`
- The function has been tested for both 1d and 2d arrays.


certik (approved):
I think this looks good to me. Thanks!

PR #2577, opened 2024-03-06
 Error Message (Dict,Set,List ) Type Mismatch

### LINKED ISSUE #2042: Request better error message for `dict` and `set`
I often make the following stupid mistake:

```python
from lpython import (i32, dataclass)

@dataclass
class FullFormValue:
    list_i32 : list[i32]
    string   : str

foo : dict[str, FullFormValue]

DEAD_LIST : list[i32] = [-1]  # Issue

ttype    : FullFormValue = FullFormValue(DEAD_LIST, 'dimensions')
contents : FullFormValue = FullFormValue([1, 2], '')

foo = {'ttype'   , ttype,
       'contents', contents}
```

The error message is accurate, but could be more helpful:

```bash
/Users/brian/CLionProjects/lpython/src/bin/python /Users/brian/CLionProjects/lpython/lasr/LP-pycharm/issue2034.py 
semantic error: All Set values must be of the same type for now
  --> /Users/brian/CLionProjects/lpython/lasr/LP-pycharm/issue2034.py:15:7 - 16:28
   |
15 |    foo = {'ttype'   , ttype,
   |          ^^^^^^^^^^^^^^^^^^^...
...
   |
16 |           'contents', contents}
   | ...^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```

LPython probably knows that the right-hand side should be a `dict`, so it might be better if LPython told me of a type mismatch between the left-hand and right-hand sides, that the right-hand side should be a `dict` and not a `set`. Instead, it tells me that my `set` is wrong. That's correct, of course, but it leads me in the wrong direction. If LPython allowed heterogeneous `set`s, it wouldn't even report that error!

assem2002:
# Fixes #2042 
# The Fix
_Screen Shot Of The Code In The Issue Description After Fix._ 
Screenshot from 2024-03-06 20-07-53

The issue we're facing could be described as **"CONSTRUCTION WHILE ASSIGNING"** problem.
It happens when  _visit_assign_ function is invoked.
It could happen with _annassign_ function also as shown below

# Mention To Some Other Cases
The  issue appears in the following code:
```python
someVar : list[i32]
someVar = {1 ,"whatever"}
```

Screenshot from 2024-03-06 20-12-29

And also here:
```python
@dataclass
class FullFormValue:
    list_i32 : list[i32]
    string   : str
    
ttype    : str = FullFormValue(DEAD_LIST, 1)
contents : FullFormValue = FullFormValue([1, 2], '')
```
- The first code and its other forms (set,dict,list) are handled by the fix.
-  The Second code is handling with data class construction, which the fix doesn't provide the same debug message for. but data classes are constructed and get bound to a variable in the same line of code, so it would be easy to see the whole mismatch mess you've made.

**_Let me know if the object construction handling is a concern and needs to be debugged in some other way._
# Simplifying the code 
- We could write the type check code as a function under  ASRUtiles namespace, as it's redundant piece of code here. 

- I don't have deep grasp on the visit_assign function so I didn't mess with its flow because - as I think - there's type coercion happening.
-  I just added a piece of code to solve the problem, but I think the redundancy of the code we face could be handled easily by doing early type check and remove the latter one.


kmr-srbh:
Please replace "Fix Issue #2042" with "Fixes #2042" so that the PR gets linked to the issue.

assem2002:
I Edited it,  #Thanks @kmr-srbh

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -5069,6 +5069,39 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                                                    {assign_asr_target->base.loc, value_ASR_after_down_cast->base.loc})
+                                })
+                        );
+                        throw SemanticAbort();
+                    }
+                }
+            }
+
+        }
+
```
Shaikh-Ubaid, line 5102: I am unsure if this is a good approach. We already catch the `SemanticError()` in `body_visitor()`.

Shaikh-Ubaid: Thanks for this PR @assem2002. I appreciate your work. The changes in this PR look good but they seem to be more like hard coded for the specific test and might not be needed at the moment.

I think a better (and more accurate) way to resolve the issue https://github.com/lcompilers/lpython/issues/2042 is to implement error recovery mechanisms (which may/might be a larger issue to be dealt with at the moment).

Thirumalai-Shaktivel:

assem2002:
@Thirumalai-Shaktivel 

The problem we face in issue #2042 appears to be more complex.

@Shaikh-Ubaid suggested an error recovery mechanism to be implemented in the system to handle such multiple errors while doing ASR construction.

Probably we should make a bigger issue with specified tasks as long as there's no such mechanism or similar ones already implemented.

I'll draft the PR,but I think the issue should be closed unless a hard-coded fix is fine for now.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -5205,6 +5238,10 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                 this->visit_expr(*x.m_elts[i]);
                 expr = ASRUtils::EXPR(tmp);
                 if (!ASRUtils::check_equal_type(ASRUtils::expr_type(expr), type)) {
+                	//set the tmp to use it in the error message.(copied from the end of this function)              	
+        			ASR::ttype_t* list_type = ASRUtils::TYPE(ASR::make_List_t(al, x.base.base.loc, type));
+        			tmp = ASR::make_ListConstant_t(al, x.base.base.loc, list.p,
+            			list.size(), list_type);
                     throw SemanticError("All List elements must be of the same type for now",
```
Thirumalai-Shaktivel, line 5245: I have the following idea in mind, check if it's possible to implement or handle. If any blockers, do let me know.

Instead of throw, do `diag.add(diag::Diagnostic(...))` and return. Check if diag has some error and throw error based on that.

assem2002: Okay this is doable,but I need to know what benefit we'll get after replacing a throw statement with `diag.add()`,so I can reason my fix as I don't quite get what you mean .

I think I could handle the error directly without creating ASR element. If that what you imply, that would make the code change in the `visit_assign()` only,  and (`visit_set()` , `visit_dict()` , `visit_list()` ) would remain the same.

Shaikh-Ubaid:
The fix in this PR doesn't seem to be a high priority. I think we should fix it issue by adding support for error-recovery (as mentioned in https://github.com/lcompilers/lpython/issues/2042#issuecomment-1981703862), which in itself is a larger issue. Should we tackle this later @certik?

certik:
Yes, correct. We have to extend the ASR->ASR transformation to just report all errors and recover from them. We should not be catching exceptions like that. In fact ideally we should not use exceptions at all. I think we should be able to recover on the level of statements, and then have special handling in blocks of statements (functions). 

In terms of priorities, first we should deliver beta quality ASR and LLVM (via LFortran) for compiling valid code. Only then focus on beautiful error messages for invalid code.

PR #2580, opened 2024-03-07
[tests] register tests/expr17.py

khushi-411:
As per title.

Thanks for noticing it, @Thirumalai-Shaktivel!


tests/tests.toml
```diff
@@ -140,6 +140,13 @@ cpp = true
 filename = "expr16.py"
 asr = true
 
+[[test]]
+filename = "expr17.py"
+asr = true
+ast = true
+llvm = true
```
Thirumalai-Shaktivel, line 147:
```suggestion
```
Let's test only `python`.

khushi-411: okay

tests/reference/python-expr17-3b84714.stdout
```diff
@@ -0,0 +1,14 @@
...
+        print("negative value")
+    else:
+        if (a) > (0):
+            print("positive value")
+        else:
+            print("zero")
+
+
+
+
```
Shaikh-Ubaid, line 14: Excess lines here should not be present. I think this can be handled in a separate PR.

Shaikh-Ubaid (approved):
It looks good to me.  Thanks for this!

tests/reference/python-expr17-3b84714.stdout
khushi-411: Yes, will create another PR for it.

Thanks!

PR #2586, opened 2024-03-07
Add `str.isspace()` method

### LINKED ISSUE #2490: Unexpected output for string method on whitespace characters

kmr-srbh:
Fixes #2490.

#2490 gave wrong output due to the incorrect handling of the escape sequence '\f'. Escaping the sequence was handled, but unescaping was missed. This led to the string `" \t\n\v\f\r"` being transformed to `" \t\n\v\\f\r"` in the AST. Note the extra "\\" after "v". Hence, the answer in #2490 was thus 'not actually' wrong as non-whitespace and lowercase characters did exist.

# Fix
```python
def f():
    b: str = " \t\n\v\f\r"
    print(b.isspace())
    print(b.islower())
    print(b.isupper())


f()
```
```console
True
False
False
```

# Improvement
Though the error in the above case was due to incorrect string handling, the `isspace()` method was actually not implemented correctly. The definition of a 'whitespace' character is very broad. The Python interpreter checks for all of them in it's implementation of the method. I incorporated checking for those characters.


kmr-srbh:
I am adding the tests in a while. @Thirumalai-Shaktivel please see this.

src/runtime/lpython_builtin.py
```diff
@@ -1018,16 +1016,50 @@ def _lpython_str_isascii(s: str) -> bool:
...
+    
+    char: str
+    w_char: str
+    for char in s:
+        is_whitespace: bool = False
+        for w_char in WHITESPACE:
+            if char == w_char:
+                is_whitespace = True
+                break
+        if not is_whitespace:
```
Thirumalai-Shaktivel, line 1059: I thought `if char in WHITESPACE: ...` works, but it is not the case.
So, Let's keep it as the following for now.
```suggestion
        if not (char == " "  or   # SPACE
            char == "\n"     or   # LINE FEED (LF)
            char == "\r"     or   # CARRIAGE RETURN (CR)
            char == "\t"     or   # CHARACTER TABULATION (HT)
            char == "\v"     or   # VERTICAL TAB (VT)
            char == "\f"     or   # FORM FEED (FF)
            char == "\u00A0" or   # NO-BREAK SPACE
            char == "\u1680" or   # OGHAM SPACE MARK
            char == "\u2000" or   # EN QUAD
            char == "\u2001" or   # EM QUAD
            char == "\u2002" or   # EN SPACE
            char == "\u2003" or   # EM SPACE
            char == "\u2004" or   # THREE-PER-EM SPACE
            char == "\u2005" or   # FOUR-PER-EM SPACE
            char == "\u2006" or   # SIX-PER-EM SPACE
            char == "\u2007" or   # FIGURE SPACE
            char == "\u2008" or   # PUNCTUATION SPACE
            char == "\u2009" or   # THIN SPACE
            char == "\u200A" or   # HAIR SPACE
            char == "\u2028" or   # LINE SEPARATOR
            char == "\u2029" or   # PARAGRAPH SEPARATOR
            char == "\u202F" or   # NARROW NO-BREAK SPACE
            char == "\u205F" or   # MEDIUM MATHEMATICAL SPACE
            char == "\u3000"      # IDEOGRAPHIC SPACE
        ):
```

Thirumalai-Shaktivel:
Also, Let's add some tests.

src/runtime/lpython_builtin.py
kmr-srbh: Strange, but this works fine on my branch. 
```python
from lpython import i32, i64, f32, f64, dict, list, tuple, str, Const, c64

s: str = " \n\f\v\r\t"
print(s.isspace())

l: list[i32] = [1, 2, 3, 4, 5]
i: i32
for i in l:
    print(i)
```
```console
True
1
2
3
4
5
```

Thirumalai-Shaktivel: I meant we shouldn't be using a for loop here:
```py
		for w_char in WHITESPACE:
            if char == w_char:
                is_whitespace = True
                break
```

kmr-srbh: Could not get you @Thirumalai-Shaktivel.

integration_tests/test_str_attributes.py
```diff
@@ -354,6 +354,14 @@ def is_title():
     assert " ".istitle() == False
 
 def is_space():
+    s: str = " \t\n\v\f\r"
+    assert s.isspace() == True
+    assert " \t\n\v\f\r".isspace() == True
+
+    s: str = " \t\n\v\f\rabcd"
+    assert s.isspace() == False
```
Shaikh-Ubaid, line 362: `s` is declared twice here. Use different/unique names. For example:
```suggestion
    s1: str = " \t\n\v\f\r"
    assert s1.isspace() == True
    assert " \t\n\v\f\r".isspace() == True

    s2: str = " \t\n\v\f\rabcd"
    assert s2.isspace() == False
```

Shaikh-Ubaid:
@kmr-srbh  Please mark it as ready for review when ready.

kmr-srbh:
The failing test, though unrelated, was causing disruption here. It was the one which checked for invalid literals like `01`. This was fixed in a recent merged PR, so I have removed that test.

@Shaikh-Ubaid This PR is now ready. As far as the Unicode characters are concerned, they cannot be typed, but can appear when parsing strings or related work. They were added conforming to what Python calls a whitespace character.

tests/errors/test_literal.py
```diff
@@ -1,2 +0,0 @@
-def test_literal1():
-    x: i32 = 0123
```
Thirumalai-Shaktivel, line 2: Undo the deletion and update tests using `./run_tests.py -u`

kmr-srbh: I am sorry. I did not notice that this test was added with the last PR. I am undoing it. :+1:

tests/reference/tokens-test_literal-e20c024.stdout
```diff
@@ -0,0 +1,13 @@
...
+(TOKEN ")") 18:18
+(TOKEN ":") 19:19
+(NEWLINE) 20:20
+(TOKEN "indent") 21:24
+(TOKEN "identifier" x) 25:25
+(TOKEN ":") 26:26
+(TOKEN "identifier" i32) 28:30
+(TOKEN "=") 32:32
+(TOKEN "integer" 123) 34:37
+(EOF) 38:38
```
Thirumalai-Shaktivel, line 13: Also, I think this file needs to be remove manually.

kmr-srbh:
Problem persists.

tests/errors/test_literal.py
```diff
@@ -1,2 +1,6 @@
+from lpython import i32
+
 def test_literal1():
-    x: i32 = 0123
\ No newline at end of file
+    x: i32 = 0123
+
+test_literal1()
```
Thirumalai-Shaktivel, line 6: Please don't modify this file, it only tests for syntax error, so I think there is no need for the function call.

Also, do a clean build and update tests.
```console
$ git clean -dfx
$ ./build.sh
$ ./run_tests.py -u
```
Before performing this, commit all the previous unstaged changes.

kmr-srbh: Hmm..this was a bit confusing too, as some lines had function calls while some did not. Should we document this also?

Thirumalai-Shaktivel: If it's the AST or parser tests, most probably it wouldn't have function call.
Let's keep this in mind.

Thirumalai-Shaktivel: If it's the AST or parser tests, most probably it wouldn't have function call.
Let's keep this in mind.

kmr-srbh: Thank you very much for this information! :+1:

kmr-srbh:
@Thirumalai-Shaktivel could you please look as to why the tests fail? The previous messages pointed to the null char I had declared for looping through the whitespace characters. I cannot understand why the tests fail now. Please guide me.

src/runtime/lpython_builtin.py
```diff
@@ -1051,16 +1049,42 @@ def _lpython_str_isascii(s: str) -> bool:
...
+def _lpython_str_isspace(s: str) -> bool:
+    # A Unicode character is considered a 'whitespace' if it has has a bidirectional
+    # type 'WS', 'B' or 'S'; or the category 'Zs'.
     if len(s) == 0:
         return False
-    ch: str 
-    for ch in s:
-        if ch != ' ' and ch != '\t' and ch != '\n' and ch != '\r' and ch != '\f' and ch != '\v':
+    
+    char: str
```
Thirumalai-Shaktivel, line 1058: Sometimes `char` keyword is not mangled in C backend, so it fails, rename it to `ch` or `c` or something

tests/errors/test_literal.py
Thirumalai-Shaktivel: Remove all the changes in this file.

tests/reference/asr-expr8-6beda60.stderr
```diff
@@ -0,0 +1 @@
+/bin/sh: 1: lpython: not found
```
Thirumalai-Shaktivel, line 1: Remove all these new files

tests/reference/tokens-test_literal-e20c024.stdout
```diff
@@ -0,0 +1,25 @@
+(KEYWORD "from") 0:3
```
Thirumalai-Shaktivel, line 1: remove this file.

src/runtime/lpython_builtin.py
kmr-srbh: Oh! :smile:

tests/reference/asr-expr8-6beda60.stderr
kmr-srbh: Done!

tests/reference/tokens-test_literal-e20c024.stdout
kmr-srbh: Done!

Thirumalai-Shaktivel (approved):
LGTM!

PR #2588, opened 2024-03-08
[asr->python] add string visitors

khushi-411:
- [ ] add tests

### DIFF
--- a/src/libasr/codegen/asr_to_python.cpp
+++ b/src/libasr/codegen/asr_to_python.cpp
@@ -573,6 +573,58 @@ public:
         s += "len(" + s + ")";
     }
 
+    void visit_StringItem(const ASR::StringItem_t &x) {
+        std::string r = "";
+        visit_expr(*x.m_arg);
+        r += s;
+        r += "[";
+        visit_expr(*x.m_idx);
+        r += s;
+        r += "]";
+        s = r;
+    }
+
+    void visit_StringSection(const ASR::StringSection_t &x) {
+        std::string r = "";
+        visit_expr(*x.m_arg);
+        r += s;
+        if (x.m_start) {
+            r += "[";
+            visit_expr(*x.m_start);
+            r += s;
+        }
+        if (x.m_end) {
+            r += ":";
+            visit_expr(*x.m_end);
+            r += s;
+        }
+        if (x.m_step) {
+            r += ":";
+            visit_expr(*x.m_step);
+            r += s;
+            r += "]";
+        } else {
+            r += "]";
+        }
+        s = r;
+    }
+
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/asr_to_python.cpp
```diff
@@ -573,6 +573,58 @@ class ASRToLpythonVisitor : public ASR::BaseVisitor<ASRToLpythonVisitor>
...
+            r += s;
+        }
+        if (x.m_step) {
+            r += ":";
+            visit_expr(*x.m_step);
+            r += s;
+            r += "]";
+        } else {
+            r += "]";
+        }
```
Thirumalai-Shaktivel, line 608:
```suggestion
        r += "[";
        if (x.m_start) {
            visit_expr(*x.m_start);
            r += s;
        }
        r += ":";
        if (x.m_end) {
            visit_expr(*x.m_end);
            r += s;
        }
        if (x.m_step) {
            r += ":";
            visit_expr(*x.m_step);
            r += s;
        }
        r += "]";
```

I suggested this changes because of the following cases:
```
s: str = "12345"
print(s[:])
print(s[3:])
print(s[3:4:])
```

src/libasr/codegen/asr_to_python.cpp
```diff
@@ -573,6 +573,58 @@ class ASRToLpythonVisitor : public ASR::BaseVisitor<ASRToLpythonVisitor>
...
+        s = r;
+    }
+
+    void visit_StringFormat(const ASR::StringFormat_t &x) {
```
Thirumalai-Shaktivel, line 612: Does LPython represents `StringFormat` in the ASR? (If we don't support, let's implement this later? )
I think we don't properly support f-strings yet.

Thirumalai-Shaktivel (commented):
Yes, please add tests

src/libasr/codegen/asr_to_python.cpp
khushi-411: Yes, I added these because there is an entry for `StringFormat`. Here: https://github.com/lcompilers/lpython/blob/7aa70843fe3d1f6dfd711d45220ca9c152eed506/src/libasr/ASR.asdl#L305

Do you think we should keep it? Thanks

Thirumalai-Shaktivel: Yes, we have an entry in ASR.asdl, but LPython doesn't utilise it yet.
So, I think let's remove it for now. And handle it correctly once LPython supports f-strings.

src/libasr/codegen/asr_to_python.cpp
M-zhar: Leave comment for proper understanding  of code.

PR #2591, opened 2024-03-09
Improve comparison of dissimilar types

kmr-srbh:
Fixes #2593 

Comparison on dissimilar types can be handled the way CPython does - return `True` when comparing for inequality and `False` when comparing for equality. Throw a semantic error when using `>`, `>=`, `<` and `<=`.

### Checking equality
```python
integer: i32 = 1
integer_list: list[i32] = [1, 2, 3, 4, 5]
print(integer == integer_list)
```
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ ./src/bin/lpython ./examples/example.py
semantic error: Type mismatch in comparison operator, the types must be compatible
 --> ./examples/example.py:5:7
  |
5 | print(integer == integer_list)
  |       ^^^^^^^    ^^^^^^^^^^^^ type mismatch ('i32' and 'list[i32]')


Note: Please report unclear or confusing messages as bugs at
https://github.com/lcompilers/lpython/issues.
```
Similar error is thrown for other operators.

### Checking equality
```python
integer: i32 = 1
integer_list: list[i32] = [1, 2, 3, 4, 5]
print(integer == integer_list)
```
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ ./src/bin/lpython ./examples/example.py
False
```
### Checking inequality
```python
integer: i32 = 1
integer_list: list[i32] = [1, 2, 3, 4, 5]
print(integer != integer_list)
```
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ ./src/bin/lpython ./examples/example.py
True
```
### Checking greater than or equal to (>=) [similar for other operators]
```python
integer: i32 = 1
integer_list: list[i32] = [1, 2, 3, 4, 5]
print(integer >= integer_list)
```
```console
semantic error: Comparison operator not supported between instances of 'i32' and 'list[i32]'
 --> ./examples/example.py:5:7
  |
5 | print(integer >= integer_list)
  |       ^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear or confusing messages as bugs at
https://github.com/lcompilers/lpython/issues.
```

PR #2592, opened 2024-03-09
Add an intrinsic implementation of `type(object)` function

kmr-srbh:
Currently the logic is simple - get the type of the passed argument using `ASRUtils::type_to_str_python(*ttype_t)` and return it as a StringConstant. I request guidance to improve this PR.

### Get type of the object
- Variables
```python
integer_list: list[i32] = [1, 2, 3, 4]
print(type(integer_list))
```
```console
<type 'list[i32]'>
```
- Constants
```python
print(type("Hello, LPython"))
```
```console
<type 'str'>
```
### Compare object types (requires #2591 to be merged)
```python
integer: i32 = 1
integer_list: list[i32] = [1, 2, 3, 4, 5]
print(type(integer) == type(integer_list))
```
```console
False
```


kmr-srbh:
The reason which causes the above 2 tests to fail is also the one I need help with. The argument `Vec<ASR::expr_t*> args` remains unused as I get the type directly through `ASR::ttype_t* t1`. @Thirumalai-Shaktivel please have a look into the PR.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -1160,6 +1162,53 @@ static inline void verify_args(const ASR::IntrinsicScalarFunction_t& x,
...
+namespace ObjectType {
+
+     static inline void verify_args(const ASR::IntrinsicScalarFunction_t& x, diag::Diagnostics& diagnostics) {
+        ASRUtils::require_impl(x.n_args != 1,
+            "ASR Verify: type() takes only 1 argument `object`",
+            x.base.base.loc, diagnostics);
+    }
+
+    static ASR::expr_t *eval_ObjectType(Allocator &al, const Location &loc,
+            ASR::ttype_t* t1, Vec<ASR::expr_t*>& args) {
```
Thirumalai-Shaktivel, line 1175: Here is the issue:
```suggestion
            ASR::ttype_t* t1, Vec<ASR::expr_t*>& /*args*/) {
```

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -1160,6 +1162,53 @@ static inline void verify_args(const ASR::IntrinsicScalarFunction_t& x,
...
+    static inline ASR::asr_t* create_ObjectType(Allocator& al, const Location& loc,
+            Vec<ASR::expr_t*>& args,
+            const std::function<void (const std::string &, const Location &)> err) {
+        if (args.size() > 1) {
+            err("type() takes only 1 argument `object`", loc);
+        }
+
+        Vec<ASR::expr_t *> arg_values;
+        arg_values.reserve(al, 1);
+        arg_values.push_back(al, expr_value(args[0]));
```
Thirumalai-Shaktivel, line 1189:
```suggestion
        Vec<ASR::expr_t *> arg_values;
```
We don't use the expr_value(args[0]) here, as the output is dependent on the argument type.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -1160,6 +1162,53 @@ static inline void verify_args(const ASR::IntrinsicScalarFunction_t& x,
...
+        std::string object_type = "<type '" + ASRUtils::type_to_str_python(t1) + "'>";
+        return StringConstant(object_type, character(object_type.length()));
+    }
+
+    static inline ASR::asr_t* create_ObjectType(Allocator& al, const Location& loc,
+            Vec<ASR::expr_t*>& args,
+            const std::function<void (const std::string &, const Location &)> err) {
+        if (args.size() > 1) {
+            err("type() takes only 1 argument `object`", loc);
+        }
```
Thirumalai-Shaktivel, line 1185:
```suggestion
        if (args.size() == 1) {
            err("type() takes exactly 1 argument `object` for now", loc);
        }
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -7487,6 +7487,9 @@ we will have to use something else.
             throw SemanticError("Only Name or Attribute type supported in Call",
                 x.base.base.loc);
         }
+        if (call_name == "type" && x.n_args == 0) {
+            throw SemanticError("type() takes 1 required argument `object`", x.base.base.loc);
+        }
```
Thirumalai-Shaktivel, line 7492:
```suggestion
```

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -1160,6 +1162,53 @@ static inline void verify_args(const ASR::IntrinsicScalarFunction_t& x,
...
+            Vec<ASR::call_arg_t>& new_args, int64_t /*overload_id*/) {
+        declare_basic_variables("_lcompilers_objectType_" + type_to_str_python(arg_types[0]));
+        fill_func_arg("a", arg_types[0]);
+        auto result = declare(fn_name, return_type, ReturnVar);
+
+        ASR::symbol_t *f_sym = make_ASR_Function_t(fn_name, fn_symtab, dep, args,
+            body, result, ASR::abiType::Source, ASR::deftypeType::Implementation, nullptr);
+        scope->add_symbol(fn_name, f_sym);
+        return b.Call(f_sym, new_args, return_type, nullptr);
+    }
```
Thirumalai-Shaktivel, line 1207: I think the compile time value will always be used, so there is no need for instantiate function.
```suggestion
```

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -3625,6 +3674,8 @@ namespace IntrinsicScalarFunctionRegistry {
     static const std::map<int64_t,
         std::tuple<impl_function,
                    verify_function>>& intrinsic_function_by_id_db = {
+        {static_cast<int64_t>(IntrinsicScalarFunctions::ObjectType),
+            {&ObjectType::instantiate_ObjectType, &UnaryIntrinsicFunction::verify_args}},
```
Thirumalai-Shaktivel, line 3678:
```suggestion
            {nullptr, &UnaryIntrinsicFunction::verify_args}},
```

src/libasr/pass/intrinsic_function_registry.h
kmr-srbh: Thanks a lot for this! :)

src/libasr/pass/intrinsic_function_registry.h
kmr-srbh: CPython has the `type()` overloaded. One of them takes just 1 argument `object`. While discussing with you earlier in the day, I was mistakenly referring to the other function which creates a new type - `type(name, dict, **kwargs)`.

This should remain unchanged I think.

Thirumalai-Shaktivel: According to the implement in this PR, we use or handle or perform operation on the 1st argument, so I suggested the above change. Later, we can handle more arguments.

Thirumalai-Shaktivel:
Also, Let's add some tests.

src/lpython/semantics/python_ast_to_asr.cpp
kmr-srbh: This check is required as an unhandled exception is thrown for an empty function call.

src/libasr/pass/intrinsic_function_registry.h
kmr-srbh: @Thirumalai-Shaktivel did you mean to say `args.size() != 1`?

src/lpython/semantics/python_ast_to_asr.cpp
Thirumalai-Shaktivel: I think every semantics checks related to `type(..)` needs to be handled inside the `create_ObjectType` function.

src/libasr/pass/intrinsic_function_registry.h
Thirumalai-Shaktivel: Yup!

src/lpython/semantics/python_ast_to_asr.cpp
kmr-srbh: It's not @Thirumalai-Shaktivel. That's why I had to add this here in the call visitor :(

Thirumalai-Shaktivel: Share an minimum reproducible example, I will check it out.

Thirumalai-Shaktivel: Do you mean for the following example?
```py
print(type())
```

CPython and LPython both throws following errors:
```console
$ python examples/expr2.py 
  File "/Users/thirumalai/Open_Source/lpython/examples/expr2.py", line 1, in <module>
    print(type())
TypeError: type() takes 1 or 3 arguments
$ lpython examples/expr2.py 
semantic error: type() takes 1 required argument `object`
 --> examples/expr2.py:1:7
  |
1 | print(type())
  |       ^^^^^^ 

```

kmr-srbh:
```python
print(type())
```

Thirumalai-Shaktivel: My bad, yea, it throws an exception:
```console
....
  File "/Users/thirumalai/Open_Source/lpython/src/libasr/asr.h", line 49205
    LCOMPILERS_ASSERT(f != nullptr);
AssertFailed: f != nullptr
```

Thirumalai-Shaktivel: Apply the following diff to fix the bug:
```diff
diff --git a/src/lpython/semantics/python_ast_to_asr.cpp b/src/lpython/
semantics/python_ast_to_asr.cpp
index 714349d7c..9ae12641d 100644
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -7487,9 +7487,6 @@ we will have to use something else.
             throw SemanticError("Only Name or Attribute type supported
 in Call",
                 x.base.base.loc);
         }
-        if (call_name == "type" && x.n_args == 0) {
-            throw SemanticError("type() takes 1 required argument `obj
ect`", x.base.base.loc);
-        }
 
         ASR::symbol_t *s = current_scope->resolve_symbol(call_name);
         if( s && ASR::is_a<ASR::Module_t>(*s) ) {
@@ -7522,7 +7519,7 @@ we will have to use something else.
                 }
                 Vec<ASR::expr_t*> args_; args_.reserve(al, x.n_args);
                 visit_expr_list(x.m_args, x.n_args, args_);
-                if (ASRUtils::is_array(ASRUtils::expr_type(args_[0])) 
&&
+                if (x.n_args && ASRUtils::is_array(ASRUtils::expr_type
(args_[0])) &&
                     imported_functions[call_name] == "math" ) {
                     throw SemanticError("Function '" + call_name + "' 
does not accept vector values",
                         x.base.base.loc);
```

kmr-srbh:
> Also, Let's add some tests.

There is a small issue. The return type of `type()` is currently the same as the passed argument. How do we go about fixing it?

Thirumalai-Shaktivel:
Minimum reproducible example (MRE), please?

kmr-srbh:
> Minimum reproducible example (MRE), please?

```python
l: list[i32] = [1, 2, 3, 4, 5]
t: str = type(l)
```

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -1160,6 +1162,39 @@ static inline void verify_args(const ASR::IntrinsicScalarFunction_t& x,
...
+            const std::function<void (const std::string &, const Location &)> err) {
+        if (args.size() != 1) {
+            err("type() takes exactly 1 argument `object` for now", loc);
+        }
+
+        Vec<ASR::expr_t *> arg_values;
+        ASR::expr_t *m_value = eval_ObjectType(al, loc, expr_type(args[0]), arg_values);
+        return ASR::make_IntrinsicScalarFunction_t(al, loc,
+            static_cast<int64_t>(IntrinsicScalarFunctions::ObjectType),
+            args.p, args.n, 0, ASRUtils::expr_type(args[0]), m_value);
```
Thirumalai-Shaktivel, line 1191:
```suggestion
            args.p, args.n, 0, ASRUtils::expr_type(m_value), m_value);
```

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -1160,6 +1162,39 @@ static inline void verify_args(const ASR::IntrinsicScalarFunction_t& x,
 
 } // namespace BinaryIntrinsicFunction
 
+
+namespace ObjectType {
+
+     static inline void verify_args(const ASR::IntrinsicScalarFunction_t& x, diag::Diagnostics& diagnostics) {
+        ASRUtils::require_impl(x.n_args != 1,
```
Thirumalai-Shaktivel, line 1169:
```suggestion
        ASRUtils::require_impl(x.n_args == 1,
```

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -3625,6 +3660,8 @@ namespace IntrinsicScalarFunctionRegistry {
     static const std::map<int64_t,
         std::tuple<impl_function,
                    verify_function>>& intrinsic_function_by_id_db = {
+        {static_cast<int64_t>(IntrinsicScalarFunctions::ObjectType),
+            {nullptr, &UnaryIntrinsicFunction::verify_args}},
```
Thirumalai-Shaktivel, line 3664:
```suggestion
            {nullptr, &ObjectType::verify_args}},
```

Thirumalai-Shaktivel (commented):
I think the following code output would differ with CPython, we have to fix it before merging this PR.
Rest looks good.
```py
print(type(type(1)))
```

src/libasr/pass/intrinsic_function_registry.h
kmr-srbh: This fails withe the ASR Verify error stated above.

Thirumalai-Shaktivel: Apply my other suggestions, the ASR Verify error would be fixed.

src/libasr/pass/intrinsic_function_registry.h
kmr-srbh: Hmm.. this check was really confusing. It was a bit difficult to understand what we were checking. Thanks a lot for this! :+1:

src/libasr/pass/intrinsic_function_registry.h
kmr-srbh: I am in the process 😃.

kmr-srbh: @Thirumalai-Shaktivel I am interested to know the reason this worked. I guess it was because of this line - `{nullptr, &ObjectType::verify_args}}`?

kmr-srbh:
> I think the following code output would differ with CPython, we have to fix it before merging this PR. Rest looks good.
> 
> ```python
> print(type(type(1)))
> ```

Going the `print()` way?

src/libasr/pass/intrinsic_function_registry.h
Thirumalai-Shaktivel: Yes! It's because of the function `UnaryIntrinsicFunction::verify_args` which is used for the intrinsics that have same type for the input and output.

Thirumalai-Shaktivel:
Nope, I meant the case: type(type(1))
Here CPython prints as: `<class 'type'>`, But LPython using this PR changes, prints as `<type 'str'>`.
We need to fix this.

kmr-srbh:
> Nope, I meant the case: type(type(1)) Here CPython prints as: `<class 'type'>`, But LPython using this PR changes, prints as `<type 'str'>`. We need to fix this.

But this will always be the case until we create a specific type for `type()`. Imagine `type(type(type(1)))`.

Thirumalai-Shaktivel:
Check if the args[0] is IntrinsicScalarFunciton and down_cast it and check if the m_intrinsic_id is static_cast<IntrinsicScalarFunctions::ObjectType>, if so then store the output as <class 'type'>. Otherwise do eval_ObjectType, it would work for any number of nested type(..).

Later in the future, if we find any other different output or cases, we will modify it accordingly.

kmr-srbh:
> Check if the `args[0]` is `IntrinsicScalarFunciton` and `down_cast` it and check if the `m_intrinsic_id` is `static_cast<IntrinsicScalarFunctions::ObjectType>`, if so then store the output as `<class 'type'>`. Otherwise do `eval_ObjectType`, it would work for any number of nested `type(..)`.
> 
> Later in the future, if we find any other different output or cases, we will modify it accordingly.

Hmm, this seems a good idea. Let's test it out. :+1:

kmr-srbh:
I am getting this error now. I think it is because of the clean build we did for the `isspace()` branch. 
```console
CMake Error at CMakeLists.txt:10 (file):
  file STRINGS file "/home/saurabh-kumar/Projects/System/lpython/version"
  cannot be read.
.
.
.
-- Configuring incomplete, errors occurred!
```

Thirumalai-Shaktivel:
What command did you use to build `lpython`?

kmr-srbh:
> Which command did you use to build `lpython`?

When this error occurred now, I followed the complete process which we do on a fresh build. As it turns out, only `./build1.sh` threw the error. Other commands did not do anything.

Thirumalai-Shaktivel:
Before `./build1.sh`, we have to run `./build0.sh` to generate all the required files.

Or you can just run `./build.sh`

kmr-srbh:
Thanks @Thirumalai-Shaktivel it works now. It was an issue with the tag which had previously caused `./build0.sh` to fail. What is the correct way of using tags here at LPython? Always using the default one?

Thirumalai-Shaktivel:
Yes, I think git automatically picks the correct tag, see ci/version.sh

kmr-srbh:
I am pushing the tests.

kmr-srbh:
Thanks a lot for your support and guidance @Thirumalai-Shaktivel! Without you I could not have made it this far. Thank you very much!

kmr-srbh:
@Thirumalai-Shaktivel do you think this PR is ready?

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -1160,6 +1162,49 @@ static inline void verify_args(const ASR::IntrinsicScalarFunction_t& x,
...
+        if (args.size() != 1) {
+            err("type() takes exactly 1 argument `object` for now", loc);
+        }
+        ASR::expr_t *m_value = nullptr;
+        Vec<ASR::expr_t *> arg_values;
+       
+        if (ASR::is_a<ASR::IntrinsicScalarFunction_t>(*args[0])) {
+            ASR::IntrinsicScalarFunction_t *object = ASR::down_cast<ASR::IntrinsicScalarFunction_t>(args[0]);
+            if (static_cast<IntrinsicScalarFunctions>(object->m_intrinsic_id) == IntrinsicScalarFunctions::ObjectType) {
+                m_value = StringConstant("<type 'type'>", character(13)); // 13 is the length of the string "<type 'type'>"
```
Thirumalai-Shaktivel, line 1192:
```suggestion
                m_value = StringConstant("<class 'type'>", character(13)); // 13 is the length of the string "<type 'type'>"
```
```py
>>> type(type(1))
<class 'type'>
```

tests/tests.toml
```diff
@@ -410,6 +410,10 @@ asr = true
 filename = "../integration_tests/test_builtin_oct.py"
 asr = true
 
+[[test]]
+filename = "../integration_tests/test_builtin_type.py"
+asr = true
+
```
Thirumalai-Shaktivel, line 416:
```suggestion
```

Tests this using `integration_tests/CMakeLists`

test: llvm, cpython

kmr-srbh: Okay.

src/libasr/pass/intrinsic_function_registry.h
kmr-srbh: I had intentionally not used the word 'class' as we do not currently support it fully. Should we write it here?

tests/tests.toml
kmr-srbh: The cpython test fails because the `type()` has a different return type there.

src/libasr/pass/intrinsic_function_registry.h
Thirumalai-Shaktivel: Nope, we have to make sure both CPython and LPython generates the same output.
So, I suggested the above changes and test this both in CPython and LPython(llvm).

kmr-srbh: I am really sorry for the confusion. The return type of `type()` in CPython is of type - `type`, whereas here we are using a string for it. They would not be similar.

I request guidance on how do we go about this @Thirumalai-Shaktivel.

Thirumalai-Shaktivel:
Apply the following diff:
Let's test only the data types for now; later, we can handle it based on the requirements.
```diff
diff --git a/integration_tests/test_builtin_type.py b/integration_tests/test_builtin_type.py
index dea42bc74..961ed245c 100644
--- a/integration_tests/test_builtin_type.py
+++ b/integration_tests/test_builtin_type.py
@@ -1,19 +1,27 @@
-from lpython import i32, f64, list, str, dict, Const
+from lpython import i32, f64, Const

 def test_builtin_type():
     i: i32 = 42
+    f: f64 = 64.
     s: str = "Hello, LPython!"
     l: list[i32] = [1, 2, 3, 4, 5]
     d: dict[str, i32] = {"a": 1, "b": 2, "c": 3}
-    CONST_LIST: Const[list[f64]] = [12.22, 14.63, 33.82, 19.18]
+    res: str = ""

-    assert type(i) == "<type 'i32'>"
-    assert type(s) == "<type 'str'>"
-    assert type(l) == "<type 'list[i32]'>"
-    assert type(d) == "<type 'dict[str, i32]'>"
-    assert type(CONST_LIST) == "<type 'Const[list[f64]]'>"
-
-    assert type(type(i)) == "<type 'type'>"
-    assert type(type(CONST_LIST)) == "<type 'type'>"
+    res = str(type(i))
+    print(res)
+    assert res == "<class 'int'>"
+    res = str(type(f))
+    print(res)
+    assert res == "<class 'float'>"
+    res = str(type(s))
+    print(res)
+    assert res == "<class 'str'>"
+    res = str(type(l))
+    print(res)
+    assert res == "<class 'list'>"
+    res = str(type(d))
+    print(res)
+    assert res == "<class 'dict'>"

 test_builtin_type()
diff --git a/src/libasr/pass/intrinsic_function_registry.h b/src/libasr/pass/intrinsic_function_registry.h
index e70524b62..19d219f53 100644
--- a/src/libasr/pass/intrinsic_function_registry.h
+++ b/src/libasr/pass/intrinsic_function_registry.h
@@ -1173,7 +1173,24 @@ namespace ObjectType {

     static ASR::expr_t *eval_ObjectType(Allocator &al, const Location &loc,
             ASR::ttype_t* t1, Vec<ASR::expr_t*>& /*args*/) {
-        std::string object_type = "<type '" + ASRUtils::type_to_str_python(t1) + "'>";
+        std::string object_type = "<class '";
+        switch (t1->type) {
+            case ASR::ttypeType::Integer : {
+                object_type += "int"; break;
+            } case ASR::ttypeType::Real : {
+                object_type += "float"; break;
+            } case ASR::ttypeType::Character : {
+                object_type += "str"; break;
+            } case ASR::ttypeType::List : {
+                object_type += "list"; break;
+            } case ASR::ttypeType::Dict : {
+                object_type += "dict"; break;
+            } default: {
+                LCOMPILERS_ASSERT_MSG(false, "Unsupported type");
+                break;
+            }
+        }
+        object_type += "'>";
         return StringConstant(object_type, character(object_type.length()));
     }

@@ -1185,17 +1202,8 @@ namespace ObjectType {
         }
         ASR::expr_t *m_value = nullptr;
         Vec<ASR::expr_t *> arg_values;
-
-        if (ASR::is_a<ASR::IntrinsicScalarFunction_t>(*args[0])) {
-            ASR::IntrinsicScalarFunction_t *object = ASR::down_cast<ASR::IntrinsicScalarFunction_t>(args[0]);
-            if (static_cast<IntrinsicScalarFunctions>(object->m_intrinsic_id) == IntrinsicScalarFunctions::ObjectType) {
-                m_value = StringConstant("<type 'type'>", character(13)); // 13 is the length of the string "<type 'type'>"
-            }
-        }
-        else {
-            m_value = eval_ObjectType(al, loc, expr_type(args[0]), arg_values);
-        }
-
+        m_value = eval_ObjectType(al, loc, expr_type(args[0]), arg_values);
+
         return ASR::make_IntrinsicScalarFunction_t(al, loc,
             static_cast<int64_t>(IntrinsicScalarFunctions::ObjectType),
             args.p, args.n, 0, ASRUtils::expr_type(m_value), m_value);
```

tests/reference/asr-test_builtin_type-f795c34.json
```diff
@@ -0,0 +1,13 @@
...
+    "infile": "tests/../integration_tests/test_builtin_type.py",
+    "infile_hash": "81fc4815e7adc95e05982852652f2cb26658ae4e46d7b635d0d5f580",
+    "outfile": null,
+    "outfile_hash": null,
+    "stdout": "asr-test_builtin_type-f795c34.stdout",
+    "stdout_hash": "6e813b30a83bfc41b0ca67e40451b1b81edd29b56029334eb7bba89e",
+    "stderr": null,
+    "stderr_hash": null,
+    "returncode": 0
+}
```
Thirumalai-Shaktivel, line 13: Remove these two files from tests/reference

Thirumalai-Shaktivel:
@Shaikh-Ubaid, do you have any design suggestions here?

integration_tests/test_builtin_type.py
```diff
@@ -0,0 +1,27 @@
+from lpython import i32, f64, Const
```
Shaikh-Ubaid, line 1: If `Const` is unused, remove it.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -3040,6 +3040,7 @@ PyMODINIT_FUNC PyInit_lpython_module_)" + fn_name + R"((void) {
         std::string out;
         std::string indent(4, ' ');
         switch (x.m_intrinsic_id) {
+            SET_INTRINSIC_NAME(ObjectType, "type");
```
Shaikh-Ubaid, line 3043: Does the integration test added in this PR work with the `c` backend. If it works, then enable the `c` backend as well. If not, then I think we need not update the `c` backend here.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -7586,7 +7586,8 @@ we will have to use something else.
                 }
                 Vec<ASR::expr_t*> args_; args_.reserve(al, x.n_args);
                 visit_expr_list(x.m_args, x.n_args, args_);
-                if (ASRUtils::is_array(ASRUtils::expr_type(args_[0])) &&
+                
+                if (x.n_args && ASRUtils::is_array(ASRUtils::expr_type(args_[0])) &&
```
Shaikh-Ubaid, line 7590: Is this change needed for this PR? What specific case does it fix?

If this change is not directly related to this PR, but was related to some issue found while working on this PR, then I suggest reporting the specific issue found and then sending a separate PR fixing it.

Shaikh-Ubaid (commented):
I shared some comments. Overall the implementation looks in the right direction.

src/lpython/semantics/python_ast_to_asr.cpp
kmr-srbh: It avoids an exception when the function is called with no parameters.

src/libasr/codegen/asr_to_c_cpp.h
kmr-srbh: It works. I have added the `c` back-end. :+1:

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: >It avoids an exception when the function is called with no parameters.

Do you mean when the `type()` function is called with no paramaters?

src/libasr/codegen/asr_to_c_cpp.h
Shaikh-Ubaid: Perfect.

src/lpython/semantics/python_ast_to_asr.cpp
kmr-srbh: Yes. It is due to this check that we are able to check for `type()` called with no parameters in the `create_ObjectType()`. Without that check we have an unhandled exception.

Shaikh-Ubaid, line 7590: We can make it more explicit.
```suggestion
                if (x.n_args > 0 && ASRUtils::is_array(ASRUtils::expr_type(args_[0])) &&
```

Shaikh-Ubaid (commented):
We just need to fix https://github.com/lcompilers/lpython/pull/2592#discussion_r1527573997 and https://github.com/lcompilers/lpython/pull/2592#discussion_r1527589247. I think then this would be good to merge.

integration_tests/test_builtin_type.py
kmr-srbh: It was used before, but when we decided to make it work exactly like Python, we removed checks for `Const` types.

Shaikh-Ubaid (approved):
It looks good! Thanks!

PR #2594, opened 2024-03-10
Add math.sumprod() , math.frexp() , and math.isclose()

farah-salama:
## Added functions

* `math.sumprod(x, y)`: Calculates the sum of products of values from the lists x and y.
* `math.frexp(x)`: Returns the mantissa and exponent of x as the pair (m, e), where m is a float and e is an integer such that x == m * 2**e exactly.
* `math.isclose(a, b, *, rel_tol = 1e-09, abs_tol = 0.0)`: Checks if the values a and b are close to each other.

## Testing
Tests are added in `test_math.py` to ensure the correctness of each function.


src/runtime/math.py
```diff
@@ -709,3 +709,64 @@ def remainder(x: f64, y: f64) -> f64:
...
     return x - y*f64(q)
+
+
+def sumprod(x: list[f64], y: list[f64]) -> f64:
+    '''
+    Return the sum of products of values from the lists x and y.
+    '''
+    if len(x) != len(y):
+        raise ValueError("The two lists must have the same length.")
+
```
kmr-srbh, line 720: Please add the following line:
```python
i: i32
```

kmr-srbh (changes requested):
Please make the requested changes for the tests to pass.

kmr-srbh:
Hello @farah-salama! Welcome to LPython! Thanks for this PR.

Please update the test references by running 
```bash
./run_tests.py -u
```

src/runtime/math.py
farah-salama: I noticed it right after I made the pr thanks for your help!

PR #2598, opened 2024-03-11
Fix logical comparison

### LINKED ISSUE #2597: Bug: Assert failure in logical operation
```console
% cat examples/expr2.py 
print(2.0 > 0.0 and 2.0**0.5)
% python examples/expr2.py                
1.4142135623730951
% lpython examples/expr2.py --backend llvm
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/bin/lpython.cpp", line 1929
    err = compile_python_to_object_file(arg_file, tmp_o, runtime_library_dir,
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/bin/lpython.cpp", line 837
    r1 = LCompilers::LPython::python_ast_to_asr(al, lm, nullptr, *ast, diagnostics, compiler_options,
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 8149
    auto res2 = body_visitor(al, lm, *ast_m, diagnostics, unit, main_module, module_name,
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 8085
    BodyVisitor b(al, lm, unit, diagnostics, main_module, module_name, ast_overload, allow_implicit_casting);
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 4732
    visit_stmt(*x.m_body[i]);
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/python_ast.h", line 1883
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/python_ast.h", line 1772
    case stmtType::Nonlocal: { v.visit_Nonlocal((const Nonlocal_t &)x); return; }
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 6524
    AST::Call_t *c = AST::down_cast<AST::Call_t>(x.m_value);
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 7618
    visit_expr_list(x.m_args, x.n_args, args);
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 613
    this->visit_expr(*exprs[i]);
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/python_ast.h", line 1910
    void visit_expr(const expr_t &b) { visit_expr_t(b, self()); }
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/python_ast.h", line 1784
    case exprType::BoolOp: { v.visit_BoolOp((const BoolOp_t &)x); return; }
  File "/Users/ubaid/Desktop/OpenSource/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 3329
    LCOMPILERS_ASSERT(
AssertFailed: ASRUtils::check_equal_type(ASRUtils::expr_type(lhs), ASRUtils::expr_type(rhs))
```

We should rather be printing a semantic error that the types to the logical operation are not equal and also are not of type logical.

kmr-srbh:
Fixes #2597 

### Numeric output
```python
print(2.0 + 0.0 or 2.0 - 0.5)
```
```console
2.00000000000000000e+00
```
```python
print(2.0 + 0.0 and 2.0 - 0.5)
```
```console
1.50000000000000000e+00
```
### Logical output
```python
print(2.0 > 0.0 or 2.0<0.5)
```
```console
True
```
```python
print(2.0 > 0.0 and 2.0<0.5)
```
```console
False
```
### Error for invalid operands
```python
print(2.0 > 0.0 and 2.0**0.5)
```
```console
semantic error: Type mismatch: 'bool' and 'f64'. Operand should be of type 'bool'
 --> ./examples/example.py:1:21
  |
1 | print(2.0 > 0.0 and 2.0**0.5)
  |                     ^^^^^^^^ 


```
```python
print("hello" or "world")
```
```console
semantic error: Logical operation not supported on object of type 'str'
 --> ./examples/example.py:1:7
  |
1 | print("hello" or "world")
  |       ^^^^^^^ 


```


kmr-srbh:
Pushing new error tests and fixing old integration tests.

kmr-srbh:
The above change breaks https://github.com/lcompilers/lpython/pull/1506, although I think it was not properly implemented as `print(1 or 2)` fails, but
```python
a : i32
b : i32
x : i32 
y : i32 
or_op1 : i32
or_op2 : i32
a = 1
b = 2
x = 3
y = 4
or_op1 = a or b
or_op2 = x or y
print(or_op1,or_op2)
```
gives an output.

Shaikh-Ubaid:
>Throw a semantic error if both operands are not of type logical

I think we should throw an error. But let's wait for one more view on this.

kmr-srbh:
> I think we should throw an error. But let's wait for one more view on this. I shared it here https://lfortran.zulipchat.com/#narrow/stream/311866-LPython/topic/logical.20operation/near/425880351.

I support the idea of explicit casting and an error otherwise too.

Shaikh-Ubaid:
It looks like `or` is overloaded to work with logicals (for boolean operation) and other types. When other types are used, `or` works like a selective operation. It selects the first value if it evaluates to `True`, else the second value. For example, the following prints:
```console
% python
Python 3.12.1 | packaged by conda-forge | (main, Dec 23 2023, 08:01:35) [Clang 16.0.6 ] on darwin
>>> 1 or 3
1
>>> 0 or 3
3
```

>>Throw a semantic error if both operands are not of type logical?

So, we cannot throw error if both operands are not of type logical. We need to think how to handle this properly.

kmr-srbh:
> We need to think how to handle this properly.

You are right. This was handled in #1506 and it works here too. The same is true for `and`.
## or
```python
from lpython import i32, i64, f32, f64

a: i32 = 1
b: i32 = 3
print(a or b)
```
```console
1
```
Although the above is buggy and fails for this - `print(1 or 3)`.

CPython does this by assigning a truth value to every type. Does LPython have the same structure?

kmr-srbh:
> Do you mean does LPython assigns a truth value to every type?

Yes. I see, as stated by you, that we do have truth values for all the basic types like Integer, Real and Character. The fix I see can be to just let go the checks for a `LogicalConstant` as we do need to operate on non-logical constants to obtain a value, and because we do have truth values for all these types. The case of a `List`, `Dict` or a `Tuple` can be separately handled. Although I think we should support a truth value for them - it can be used as a check for emptiness.

Please tell me your thoughts about this. I see that we cannot directly down cast the values to a logical constant. How do we go about this?

Shaikh-Ubaid:
>Please tell me your thoughts about this. I see that we cannot directly down cast the values to a logical constant. How do we go about this?

I am not sure if I idea get your idea correctly.

>Please tell me your thoughts about this. I see that we cannot directly down cast the values to a logical constant. How do we go about this?

If you have any ideas/plans, please implement and push it here. I will review and we can see how to move forward with this.

For now we just need to tackle https://github.com/lcompilers/lpython/issues/2597#issuecomment-2002133981 and https://github.com/lcompilers/lpython/issues/2597#issuecomment-2002135527.

Shaikh-Ubaid:

kmr-srbh:
@Shaikh-Ubaid the latest commits resolves all the issues which we had with returning a non-logical value, but it fails for `print(2.0 > 0.0 and 2.0**0.5)` with a segmentation fault. I think I am making some mistake in comparing types? Please have a look.

kmr-srbh:
@Shaikh-Ubaid the latest commits fixes the issue for all the cases. The idea is simple, handle the case of logical operands and strings separately and let the backend do the work for the other types which support the operation. #1506 claimed to work for strings, but I think there is some issue with it.

## Working
### Numeric output
```python
print(2.0 + 0.0 or 2.0 - 0.5)
```
```console
2.00000000000000000e+00
```
```python
print(2.0 + 0.0 and 2.0 - 0.5)
```
```console
1.50000000000000000e+00
```
### Logical output
```python
print(2.0 > 0.0 or 2.0<0.5)
```
```console
True
```
```python
print(2.0 > 0.0 and 2.0<0.5)
```
```console
False
```
### Error for invalid operands
```python
print(2.0 > 0.0 and 2.0**0.5)
```
```console
semantic error: Type mismatch: 'bool' and 'f64'. Operand should be of type 'bool'
 --> ./examples/example.py:1:21
  |
1 | print(2.0 > 0.0 and 2.0**0.5)
  |                     ^^^^^^^^ 


```
```python
print("hello" or "world")
```
```console
semantic error: Logical operation not supported on object of type 'str'
 --> ./examples/example.py:1:7
  |
1 | print("hello" or "world")
  |       ^^^^^^^ 


```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3326,29 +3326,48 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                bool right_value = ASR::down_cast<ASR::LogicalConstant_t>(
+                                        ASRUtils::expr_value(rhs))->m_value;
+                bool result;
+                switch (op) {
+                    case (ASR::logicalbinopType::And): { result = left_value && right_value; break; }
+                    case (ASR::logicalbinopType::Or): { result = left_value || right_value; break; }
+                    default : {
+                        throw SemanticError("Boolean operator type not supported",
+                            x.base.base.loc);
+                    }
```
kmr-srbh, line 3356: No change was made in this section. Only indentation was added.

Shaikh-Ubaid: It seems an assert `LCOMPILERS_ASSERT(ASR::is_a<ASR::Logical_t>(*dest_type));` is removed. I suggest not doing indentation changes in this PR. Instead you can submit a separate PR for indentation changes.

kmr-srbh: @Shaikh-Ubaid Removal of that assert statement was intentional. In this case, indentation was required to add readability to the `if` block.

Shaikh-Ubaid:
I think the approach in this PR is in the right direction, but is not the correct fix. 

1. to handle the assert failure in https://github.com/lcompilers/lpython/issues/2597#issue-2178514051, we just need to throw a SemanticError that the types are not equal (inplace of the LCOMPILERS_ASSERT)
2. next we need to fix the failure in https://github.com/lcompilers/lpython/issues/2597#issuecomment-2002135527 which currently expects/supports only logical types. I think we need to extend it to support other types.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3326,29 +3326,48 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
-        ASR::ttype_t *dest_type = ASRUtils::expr_type(lhs);
+        ASR::ttype_t *dest_type = left_operand_type;
 
+        if (ASR::is_a<ASR::Character_t>(*left_operand_type)) {
+            throw SemanticError("Logical operation not supported on object of type 'str'", lhs->base.loc);
+        }
+        if (ASR::is_a<ASR::Character_t>(*right_operand_type)) {
+            throw SemanticError("Logical operation not supported on object of type 'str'", rhs->base.loc);
+        }
         if (ASRUtils::expr_value(lhs) != nullptr && ASRUtils::expr_value(rhs) != nullptr) {
```
Shaikh-Ubaid, line 3341: Inside this `if` we compute the the compile-time value. Currently it computes only for logical types. We need to extend support for other types.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3326,29 +3326,48 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
                     x.base.base.loc);
             }
         }
-        LCOMPILERS_ASSERT(
-            ASRUtils::check_equal_type(ASRUtils::expr_type(lhs), ASRUtils::expr_type(rhs)));
```
Shaikh-Ubaid, line 3330: Just replace this with an `if` statement that gives error when types are not equal.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3326,29 +3326,48 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                        && !ASR::is_a<ASR::Logical_t>(*right_operand_type)) {
+                throw SemanticError("Type mismatch: '" + ASRUtils::type_to_str_python(left_operand_type) 
+                                + "' and '" + ASRUtils::type_to_str_python(right_operand_type) 
+                                + "'. Operand should be of type 'bool'", rhs->base.loc);
+            } else if(!ASR::is_a<ASR::Logical_t>(*left_operand_type) 
+                        && ASR::is_a<ASR::Logical_t>(*right_operand_type)) {
+                throw SemanticError("Type mismatch: '" + ASRUtils::type_to_str_python(left_operand_type) 
+                                + "' and '" + ASRUtils::type_to_str_python(right_operand_type) 
+                                + "'. Operand should be of type 'bool'", lhs->base.loc);
             }
```
Shaikh-Ubaid, line 3370: This checking will not be needed once we have https://github.com/lcompilers/lpython/pull/2598/files#r1527569879.

kmr-srbh:
> I think the approach in this PR is in the right direction, but is not the correct fix.
> 
>     1. to handle the assert failure in [Bug: Assert failure in logical operation #2597 (comment)](https://github.com/lcompilers/lpython/issues/2597#issue-2178514051), we just need to throw a SemanticError that the types are not equal (inplace of the LCOMPILERS_ASSERT)
> 
>     2. next we need to fix the failure in [Bug: Assert failure in logical operation #2597 (comment)](https://github.com/lcompilers/lpython/issues/2597#issuecomment-2002135527) which currently expects/supports only logical types. I think we need to extend it to support other types.

@Shaikh-Ubaid The above 2 tasks are already solved by the latest commit.
### For 1
```python
print(2.0 > 0.0 and 2.0**0.5)
```
```console
semantic error: Type mismatch: 'bool' and 'f64'. Operand should be of type 'bool'
 --> ./examples/example.py:1:21
  |
1 | print(2.0 > 0.0 and 2.0**0.5)
  |                     ^^^^^^^^ 


```

### For 2
```python
print(1 or 3)
```
```console
1
```

src/lpython/semantics/python_ast_to_asr.cpp
kmr-srbh: Can be done for sure! But what about the back-end? Will we be removing it?

Shaikh-Ubaid:
>@Shaikh-Ubaid The above 2 tasks are already solved by the latest commit.

It might be working, but we need it to be in the right way. Currently, you explicitly check for when one type is logical and the other is not. But this check would not be needed when we do https://github.com/lcompilers/lpython/pull/2598#discussion_r1527569879. The approach I suggested is more general.

kmr-srbh:
> > @Shaikh-Ubaid The above 2 tasks are already solved by the latest commit.
> 
> It might be working, but we need it to be in the right way. Currently, you explicitly check for when one type is logical and the other is not. But this check would not be needed when we do [#2598 (comment)](https://github.com/lcompilers/lpython/pull/2598#discussion_r1527569879). The approach I suggested is more general.

So, summing up everything, the task for me are:
1. Add type check to verify both operands are equal.
2. Evaluate the logical operation at compile time for all supported types.

Am I right @Shaikh-Ubaid ?

Shaikh-Ubaid:
>Add type check to verify both operands are equal.

Yes.

>Evaluate the logical operation at compile time for all supported types.

Need not be all. Just ensure the examples that we discussed in https://github.com/lcompilers/lpython/pull/2598#issuecomment-2002559579 work.

kmr-srbh:
@Shaikh-Ubaid I have made the changes requested by you. :+1:

Shaikh-Ubaid:
Let's add some tests for this.
- one we need is an error reference test for the generation of a semantic error
- second we need is an integration test to check if `print(1 or 3)` works perfectly or not. Also ensure that we have an assert after the print.
- Feel free to test other types that we support in the selective or/and operation.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3326,31 +3326,119 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                    throw SemanticError("Boolean operation not supported on objects of type '"
+                            + ASRUtils::type_to_str_python(dest_type) + "'",
+                                x.base.base.loc);
             }
-            value = ASR::down_cast<ASR::expr_t>(ASR::make_LogicalConstant_t(
-                al, x.base.base.loc, result, dest_type));
+
+            tmp = ASR::make_LogicalBinOp_t(al, x.base.base.loc, lhs, op, rhs, dest_type, value);
         }
-        tmp = ASR::make_LogicalBinOp_t(al, x.base.base.loc, lhs, op, rhs, dest_type, value);
```
Shaikh-Ubaid, line 3353: I think the original place for constructing `tmp` is correct. `tmp` should be constructed even when the compile-time value is not available.

kmr-srbh: Done!

kmr-srbh:
@Shaikh-Ubaid , why does the `expr_value` of a `str` variable become a `nullptr`? This is causing issues here as working with string variables leads to segfault.

Shaikh-Ubaid:
>why does the expr_value of a str variable become a nullptr? 

expr_value returns the compile-time value. It becomes a nullptr, if the compile-time value is not available.

Shaikh-Ubaid:
>This is causing issues here as working with string variables leads to segfault.

Why don't you figure out the smallest test example that fails and we can figure out how to fix it from there.

kmr-srbh:
> > This is causing issues here as working with string variables leads to segfault.
> 
> Why don't you figure out the smallest test example that fails and we can figure out how to fix it from there.
### Fails for `str` variables
```python
a: str = "hello"
b: str = "world"

print(a or b)
```
```console
```

### Works for numeric variables
```python
from lpython import i32

a: i32 = 1
b: i32 = 3

print(a or b)
```
```console
1
```
```python
from lpython import f64

a: f64 = 1.33
b: f64 = 3.67

print(a and b)
```
```console
3.66999999999999993e+00
```

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: It does not seem to be updated for me. Can you construct tmp at the original/previous place it was being constructed? I think/guess that could help fix the issue that you are facing.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3326,31 +3326,119 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                    switch (op) {
+                        case (ASR::logicalbinopType::And): { result = left_value && right_value; break; }
+                        case (ASR::logicalbinopType::Or): { result = left_value || right_value; break; }
+                        default : {
+                            throw SemanticError("Boolean operator type not supported",
+                                x.base.base.loc);
+                        }
+                    }
+                    value = ASR::down_cast<ASR::expr_t>(ASR::make_LogicalConstant_t(
+                        al, x.base.base.loc, result, dest_type));
```
Shaikh-Ubaid, line 3358:
```suggestion
                    value = ASRUtils::EXPR(ASR::make_LogicalConstant_t(
                        al, x.base.base.loc, result, dest_type));
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3326,31 +3326,119 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                            result = left_value != 0 ? left_value : right_value;
+                            break;
+                        }
+                        default : {
+                            throw SemanticError("Boolean operator type not supported",
+                                x.base.base.loc);
+                        }
+                    }
+                    value = ASR::down_cast<ASR::expr_t>(ASR::make_IntegerConstant_t(
+                        al, x.base.base.loc, result, dest_type));
```
Shaikh-Ubaid, line 3382:
```suggestion
                    value = ASRUtils::EXPR(ASR::make_IntegerConstant_t(
                        al, x.base.base.loc, result, dest_type));
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3326,31 +3326,119 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                            result = left_value != 0 ? left_value : right_value;
+                            break;
+                        }
+                        default : {
+                            throw SemanticError("Boolean operator type not supported",
+                                x.base.base.loc);
+                        }
+                    }
+                    value = ASR::down_cast<ASR::expr_t>(ASR::make_RealConstant_t(
+                        al, x.base.base.loc, result, dest_type));
```
Shaikh-Ubaid, line 3406:
```suggestion
                    value = ASRUtils::EXPR(ASR::make_RealConstant_t(
                        al, x.base.base.loc, result, dest_type));
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3326,31 +3326,119 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                            result = std::strcmp(left_value, "") != 0 ? left_value : right_value;
+                            break;
+                        }
+                        default : {
+                            throw SemanticError("Boolean operator type not supported",
+                                x.base.base.loc);
+                        }
+                    }
+                    value = ASR::down_cast<ASR::expr_t>(ASR::make_StringConstant_t(
+                        al, x.base.base.loc, result, dest_type));
```
Shaikh-Ubaid, line 3430:
```suggestion
                    value = ASRUtils::EXPR(ASR::make_StringConstant_t(
                        al, x.base.base.loc, result, dest_type));
```

kmr-srbh:
@Shaikh-Ubaid I do not know why, but the latest commit, that is the addition of the tests, causes other tests related to string methods and dictionaries, tuples to fail. I request you to have a look.

src/lpython/semantics/python_ast_to_asr.cpp
kmr-srbh: Done.

kmr-srbh:
@Shaikh-Ubaid the issue related to `str` variables persists.

Shaikh-Ubaid:
Did you try this https://github.com/lcompilers/lpython/pull/2598#discussion_r1528106831?

src/lpython/semantics/python_ast_to_asr.cpp
kmr-srbh: I am pushing the changes.

kmr-srbh:
> Did you try this [#2598 (comment)](https://github.com/lcompilers/lpython/pull/2598#discussion_r1528106831)?

Yes. I had forgotten to push that change, but the segfault persists.

integration_tests/test_logical_assignment.py
```diff
@@ -0,0 +1,18 @@
...
+    _MAX_VAL: i32 = 100
+    i_var: i32 = 0 and 100
+    assert i_var == 0
+
+    _PI: f64 = 3.14
+    f_var: f64 = 2 * _PI or _PI**2
+    assert f_var == 6.28
+
+
+test_logical_assignment()
```
Shaikh-Ubaid, line 18: - Why is this test separate from the other? It seems to test `selective or` for strings, why is it called as `assignment`?

- Can you add a print before the asserts?

kmr-srbh: I created a separate test for testing assignments through logical operations that we implemented. I am adding the print statements.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -7586,8 +7673,7 @@ we will have to use something else.
                 }
                 Vec<ASR::expr_t*> args_; args_.reserve(al, x.n_args);
                 visit_expr_list(x.m_args, x.n_args, args_);
-                
-                if (x.n_args > 0 && ASRUtils::is_array(ASRUtils::expr_type(args_[0])) &&
+                if (ASRUtils::is_array(ASRUtils::expr_type(args_[0])) &&
```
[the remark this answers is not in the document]
kmr-srbh: I think this was done automatically when I pulled the new changes after yesterdays commits.

tests/errors/test_logical_compare_01.py
```diff
@@ -0,0 +1,5 @@
+def f():
+    print([1, 2, 3] or [1, 2, 3, 4])
+
+
+f()
```
Shaikh-Ubaid, line 5: I think we do not need this test. It is redundant at the moment. Both the operands of the operation are of the same type, so we do not expect any SemanticError here. It is a different case that we do not support lists in `selective or` operation for which we can open an issue.

Shaikh-Ubaid: Please ensure that you also remove the reference outputs associated with this test.

tests/tests.toml
```diff
@@ -429,6 +429,14 @@ cpp = true
...
 asr = true
 
+[[test]]
+filename = "../integration_tests/test_logical_compare.py"
+asr = true
+
+[[test]]
+filename = "../integration_tests/test_logical_assignment.py"
+asr = true
+
```
Shaikh-Ubaid, line 439: Please remove these from here.

Shaikh-Ubaid (changes requested):
Shared couple comments.

The integration tests also need to be enabled in `integration_tests/CMakeLists.txt`.

integration_tests/test_logical_assignment.py
Shaikh-Ubaid: Ok, Let's merge this test with the other integration test added in this PR.

src/lpython/semantics/python_ast_to_asr.cpp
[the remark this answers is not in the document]
Shaikh-Ubaid: Please undo it.

tests/tests.toml
Shaikh-Ubaid: >Same must be done for the error tests

We need one error reference test to check for semantic error when different types are used with the `selective or` operation.

kmr-srbh: I am sorry for asking the same question, my question stems from the fact that we have many integration_tests listed in the above file. This is a bit confusing.

integration_tests/test_logical_assignment.py
```diff
@@ -0,0 +1,22 @@
+from lpython import i32, f64
+
+
+def test_logical_assignment():
+    # Can be uncommented after fixing the segfault
+    # _LPYTHON: str = "LPython"
+    # s_var: str = "" or _LPYTHON
+    # assert s_var == "LPython"
+    # print(s_var)
```
Shaikh-Ubaid, line 9: Please open an issue for this.

integration_tests/test_logical_compare.py
```diff
@@ -0,0 +1,130 @@
...
+    # assert (s_a + s_b or s_b + s_a) == "ab"
+
+    # print(s_a[0] or s_b[-1])
+    # assert (s_a[0] or s_b[-1]) == "a"
+
+    # print(s_a[0] and s_b[-1])
+    # assert (s_a[0] and s_b[-1]) == "b"
+
+    # print(s_a + s_b or s_b + s_a + s_a[0] and s_b[-1])
+    # assert (s_a + s_b or s_b + s_a + s_a[0] and s_b[-1]) == "ab"
```
Shaikh-Ubaid, line 126: Please open an issue for this.

Shaikh-Ubaid (approved):
It looks good. Thanks!

tests/errors/test_logical_compare_01.py
```diff
@@ -0,0 +1,5 @@
+def f():
```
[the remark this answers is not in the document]
kmr-srbh: I am adding it. :+1:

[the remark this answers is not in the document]
kmr-srbh: Done.

Shaikh-Ubaid (approved):

PR #2600, opened 2024-03-13
Remove ``is_fast`` and its associated methods from ``PassManager``

czgdp1807:
https://github.com/lcompilers/lc/pull/110
https://github.com/lfortran/lfortran/pull/3646


src/libasr/pass/pass_manager.h
```diff
@@ -287,12 +286,10 @@ namespace LCompilers {
                           PassOptions& pass_options,
                           diag::Diagnostics &diagnostics) {
             if( !_user_defined_passes.empty() ) {
-                pass_options.fast = true;
```
czgdp1807, line 290: Same here. `pass_options.fast` should only be read from the command line and should not be set to true deliberately. This line caused bugs in LC and optimisations weren't being applied to LLVM because of such modifications of `pass_options`.

PR #2601, opened 2024-03-14
Add an intrinsic implementation of `input(prompt)` function

kmr-srbh:
# Working
**Without argument**
```python
name: str = input()
print("Hello,", name)
```
```console
LPython
Hello, LPython
```

**With argument**
```python
name: str = input("Your name: ")
print("Hello,", name)
```
```console
Your name: LPython
Hello, LPython
```

# Tasks
- [ ] Add tests.


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1791,6 +1791,19 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+    void generate_Input0() {
+        std::string input;
+        std::cin >> input;
+        tmp = builder->CreateGlobalStringPtr(input);
+    }
+
+    void generate_Input1(ASR::expr_t* m_arg) {
+        std::string input;
+        std::cout << ASR::down_cast<ASR::StringConstant_t>(m_arg)->m_s;
+        std::cin >> input;
```
Shaikh-Ubaid, line 1803: @kmr-srbh this reads input at `compile-time`, but rather you need the input to be read at runtime. 

For example, I think your approach would fail when just compiling the source to a binary file (called `tmp` here) as follows: 
```console
lpython$ ./src/bin/lpython ./examples/example.py -o tmp
```
I think the compiler would wait for the input but the input would actually be provided when the `tmp` file is run. That is:
```console
./tmp
>>> Now provide your input here
```

kmr-srbh: @Shaikh-Ubaid You are right. How do we go about working a fix? Please guide me.

Shaikh-Ubaid: >How do we go about working a fix?

We handle File I/O by calling into the C runtime library (`lfortran_intrinsics.c`). So for taking input from the terminal (i.e. standard I/O), I think we might need to do the same. LFortran already supports taking input from standard I/O. You can check it out for a reference. You can also checkout https://github.com/lfortran/lfortran/pull/2025.

kmr-srbh: Great resource! I have a question. How will we link it to the `input()` intrinsic here?

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -2515,6 +2517,40 @@ create_exp_macro(Exp, exp)
...
+                loc);
+        }
+        std::string dummy_string = "";
+        ASR::ttype_t* return_type = expr_type(StringConstant(dummy_string, character(dummy_string.length())));
+        return ASR::make_IntrinsicScalarFunction_t(al, loc,
+            static_cast<int64_t>(IntrinsicScalarFunctions::Input),
+            args.p, args.n, 0, return_type, nullptr);
+    }
+
+} // namespace Input
```
Shaikh-Ubaid, line 2552: I am not sure if implementing `input()` as an intrinsic registry function is a best approach. But I think we can go ahead with it for now. If we find another approach later (for example using an ASR node), we can refactor it then.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -2515,6 +2517,40 @@ create_exp_macro(Exp, exp)
...
+        if (!ASRUtils::is_character(*ASRUtils::expr_type(args[0]))) {
+            err("Argument `prompt` to input() must be of type 'str'",
+                loc);
+        }
+        std::string dummy_string = "";
+        ASR::ttype_t* return_type = expr_type(StringConstant(dummy_string, character(dummy_string.length())));
+        return ASR::make_IntrinsicScalarFunction_t(al, loc,
+            static_cast<int64_t>(IntrinsicScalarFunctions::Input),
+            args.p, args.n, 0, return_type, nullptr);
+    }
```
Shaikh-Ubaid, line 2550: You should also define an `instantiate_Input()` function here which will carry the definition of the input function.
In the input function definition, that is in its function body, you can have a statement `visit_FileRead()` (with `m_unit` as `nullptr`) that reads into the `result` variable which is the return value for your `input()` function. This return value will contain the value read from `stdin`.

src/libasr/codegen/asr_to_llvm.cpp
Shaikh-Ubaid: >How will we link it to the input() intrinsic here?

We use `visit_FileRead()` which reads from `stdin` when the `m_unit` is not provided. I shared more about implementing it here https://github.com/lcompilers/lpython/pull/2601#discussion_r1527257564. You can also check how things work in LFortran for specific details.

src/libasr/pass/intrinsic_function_registry.h
kmr-srbh: @Shaikh-Ubaid I am facing some problems calling the `visit_FileRead()`. It appears to have been declared in the `ast_to_*` files and as it turns out, is not being used anywhere in the whole code-base. I thought maybe I could call it using the `ASRToLLVMVisitor`, but again noticed that it was not being used anywhere too. An include error pops up when including the file which contains the class, stating dependency issues with the file.

But most important of all, `ASRToLLVMVisitor::visit_FileRead()` requires a `ASR::FileRead_t` object. It has got many required arguments. What should I pass to them? 

Here is my code:
```c++
static inline ASR::expr_t* instantiate_Input(Allocator &al, const Location &loc,
        SymbolTable /*scope*/, Vec<ASR::ttype_t*>& arg_types, ASR::ttype_t *return_type,
        Vec<ASR::call_arg_t>& new_args, int64_t /*overload_id*/) {
        std::string func_name = "_lcompilers_input_";
        declare_basic_variables(func_name);
        fill_func_arg("prompt", arg_types[0]);
        auto result = declare(fn_name, return_type, ReturnVar);
        ASRToLLVMVisitor::visit_FileRead(ASR::make_FileRead_t(al, loc, 0, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, 0));
    }
```

Shaikh-Ubaid: >What should I pass to them?

Just pass the unneeded parameters as `nullptr`.

>ASRToLLVMVisitor::visit_FileRead(ASR::make_FileRead_t(al, loc, 0, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, 0));

You need not call the `ASRToLLVMVisitor::visit_FileRead()` here. You just need to construct ASR which contains the FileRead node. This ASR node will automatically be visited in `visit_FileRead()` during ASR->LLVM construction.

For example, you updated instantiate_Input() would like 
```console
auto result = declare(fn_name, return_type, ReturnVar);
Vec<ASR::expr_t*> read_values;
read_values.reserve(al, 0);
read_values.push_back(al, result);
body.push_back(ASRUtils::STMT(ASR::make_FileRead_t(al, loc, 0, nullptr, nullptr, nullptr, nullptr, nullptr, read_values.p, read_values.n)));
```

kmr-srbh: @Shaikh-Ubaid `read_values.reserve(al, 0);`?

Shaikh-Ubaid: >@Shaikh-Ubaid read_values.reserve(al, 0);?

It is not the perfect code but just a guideline for you. Please update it as needed.

Checkout implementations for other `instantiate_*()` to understand how to define an intrinsic function in ASR intrinsic registry. In LFortran we have many intrinsics implemented, so you can find many examples there.

kmr-srbh: @Shaikh-Ubaid Thank you very much for all your guidance! I have learnt a lot!

kmr-srbh:
@Shaikh-Ubaid I am trying to implement `input()` using an ASR Node as I think it can be implemented there easily. I have a few questions because I get an error.
1. Is this expr declaration correct?
```
| Input(expr? prompt, ttype type, expr? value)
```
2. I handle the call inside `python_ast_to_asr.cpp`. Is it correct?
```c++
else if (call_name == "input") {
                if (x.n_args > 1) {
                    throw SemanticError("input() takes 0 or 1 argument. " 
                        + std::to_string(x.n_args) + " given.", x.base.base.loc);
                }
                args.reserve(al, x.n_args);
                visit_expr_list(x.m_args, x.n_args, args);
                Vec<ASR::expr_t*> args_expr = ASRUtils::call_arg2expr(al, args);
                std::string dummy_string = "";
                ASR::ttype_t* return_type = ASRUtils::expr_type(ASRUtils::EXPR(ASR::make_StringConstant_t(al, x.base.base.loc, s2c(al, dummy_string), 
                ASRUtils::TYPE(ASR::make_Character_t(al, x.base.base.loc, 1, dummy_string.length(), nullptr)))));
                tmp = ASR::make_Input_t(al, x.base.base.loc, *args_expr.p, return_type, nullptr);
            }
```
I get the following error:
```console
...
 File "/home/saurabh-kumar/Projects/System/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 8185, in LCompilers::LPython::BodyVisitor::visit_Call(LCompilers::LPython::AST::Call_t const&)
    tmp = make_call_helper(al, s, current_scope, args, call_name, x.base.base.loc,
  File "/home/saurabh-kumar/Projects/System/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 1087, in LCompilers::LPython::CommonVisitor<LCompilers::LPython::BodyVisitor>::make_call_helper(Allocator&, LCompilers::ASR::symbol_t*, LCompilers::SymbolTable*, LCompilers::Vec<LCompilers::ASR::call_arg_t>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, LCompilers::Location const&, LCompilers::LPython::AST::expr_t**, unsigned long, LCompilers::LPython::AST::keyword_t*, unsigned long)
    bool is_generic_procedure = ASR::is_a<ASR::GenericProcedure_t>(*s);
  File "/home/saurabh-kumar/Projects/System/lpython/src/libasr/asr.h", line 32, in bool LCompilers::ASR::is_a<LCompilers::ASR::GenericProcedure_t, LCompilers::ASR::symbol_t>(LCompilers::ASR::symbol_t const&)
    return T::class_type == x.type;
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", in __GI___sigaction()
```
3. I am I handling input in the llvm backend correctly?
```c++
void visit_Input(const ASR::Input_t &x) {
        visit_FileRead(*ASR::down_cast<ASR::FileRead_t>(ASRUtils::EXPR(ASR::make_FileRead_t(al, x.base.base.loc, 0,
            nullptr, nullptr, x.m_prompt, nullptr, nullptr, nullptr, 0))));
    }
```
Doing `ASR::down_cast<ASR::FileRead_t>(ASR::make_FileRead_t....)` does not work.

kmr-srbh:
The above code is just a rough work on which I wish to get your guidance before moving ahead. I am sure there are some mistakes.

Shaikh-Ubaid:
>Is this expr declaration correct?

You do not need to add any new node to ASR.

>I handle the call inside python_ast_to_asr.cpp. Is it correct?

In this PR, you have already implemented handling input as an intrinsic function. So you do not need any other handling for it.

>Am I handling input in the llvm backend correctly?

Since an `input` node in the ASR is not needed at the moment, you need not handle `visit_Input` in any backend.

Shaikh-Ubaid:
Just continue the implementation that we discussed in this thread https://github.com/lcompilers/lpython/pull/2601#discussion_r1527257564.

kmr-srbh:
@Shaikh-Ubaid I was trying to undo the previous intrinsic implementation and implement it through an ASR Node. So, we should continue with the Intrinsic approach only?

Shaikh-Ubaid:
>@Shaikh-Ubaid I was trying to undo the previous intrinsic implementation and implement it through an ASR Node. So, we should continue with the Intrinsic approach only?

Is there any specific reason why you want to undo the current implementation?

Shaikh-Ubaid:
>So, we should continue with the Intrinsic approach only?

Since we have it already, I would say we continue it. We are not yet sure if we want to be extending ASR.

kmr-srbh:
I tried implementing it the way you 

> > @Shaikh-Ubaid I was trying to undo the previous intrinsic implementation and implement it through an ASR Node. So, we should continue with the Intrinsic approach only?
> 
> Is there any specific reason why you want to undo the current implementation?

Yes. I am facing difficulty in getting to construct the ASR for FileRead_t. I explored the ASR Node method and found it be easier to handle.

My current code for instantiating `input()`.
```c++
static inline ASR::expr_t* instantiate_Input(Allocator &al, const Location &loc,
        SymbolTable *scope, Vec<ASR::ttype_t*>& arg_types, ASR::ttype_t *return_type,
        Vec<ASR::call_arg_t>& new_args, int64_t /*overload_id*/) {

        std::string func_name = "_lpython_input_";
        std::string fn_name = scope->get_unique_name(func_name);
        SymbolTable *fn_symtab = al.make_new<SymbolTable>(scope);
        
        Vec<ASR::expr_t*> args;
        args.reserve(al, new_args.size());
        ASRBuilder b(al, loc);
        Vec<ASR::stmt_t*> body; body.reserve(al, args.size());
        SetChar dep; dep.reserve(al, 1);
        if (scope->get_symbol(fn_name)) {
            ASR::symbol_t *s = scope->get_symbol(fn_name);
            ASR::Function_t *f = ASR::down_cast<ASR::Function_t>(s);
            return b.Call(s, new_args, expr_type(f->m_return_var), nullptr);
        }

        fill_func_arg("prompt", arg_types[0]);
        auto result = declare(fn_name, return_type, ReturnVar);
        Vec<ASR::expr_t*> read_values;
        read_values.reserve(al, 1);
        read_values.push_back(al, result);
        body.push_back(al, ASRUtils::STMT(ASR::make_FileRead_t(
            al, loc, 0, nullptr, nullptr, nullptr, nullptr, nullptr, read_values.p, read_values.n)));

        ASR::symbol_t *f_sym = make_ASR_Function_t(fn_name, fn_symtab, dep, args,
            body, result, ASR::abiType::Source, ASR::deftypeType::Implementation, nullptr);
        scope->add_symbol(fn_name, f_sym);
        return b.Call(f_sym, new_args, return_type, nullptr);
    }
```

kmr-srbh:
`input` is a statement in Fortran and hence handled accordingly in LFortran, but here in LPython, we have a function for taking inputs. I am not sure how to handle FileRead in llvm backend.

Shaikh-Ubaid:
>My current code for instantiating input().

I think it should work. Why don't you push your changes so I can also see the other changes?

Also, I think updates from LFortran have not yet been propogated to LPython. Therefore, the `instantiate_Input()` seems larger in size than it is expected to be.

The approach of adding a new ASR node `input` that you shared here https://github.com/lcompilers/lpython/pull/2601#issuecomment-2016701964 also seems fine. If you are able to get it working completely, you can send it in a separate PR.

kmr-srbh:
The issue is the same - input read at compile time when compiling to a binary.

Shaikh-Ubaid:
>The issue is the same - input read at compile time when compiling to a binary.

Which means your `instantiate_Input()` is not getting called.

kmr-srbh:
> > The issue is the same - input read at compile time when compiling to a binary.
> 
> Which means your `instantiate_Input()` is not getting called.

:( My bad @Shaikh-Ubaid. I am making the changes.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -3680,6 +3749,8 @@ namespace IntrinsicScalarFunctionRegistry {
     static const std::map<int64_t,
         std::tuple<impl_function,
                    verify_function>>& intrinsic_function_by_id_db = {
+        {static_cast<int64_t>(IntrinsicScalarFunctions::Input),
+            {nullptr, &Input::verify_args}},
```
Shaikh-Ubaid, line 3753: You need to provide the instantiate_Input here. Check how other functions are implemented and simply follow them.

kmr-srbh: Done! :+1: Thanks a lot for this @Shaikh-Ubaid! I recall setting this to nullptr when I had started the work on the function. I had totally forgotten about it. I am sorry for all the trouble.

kmr-srbh:
@Shaikh-Ubaid , there is a paradox. We need `x.m_unit` in FileRead to be a nullptr for reading from stdin, but `visit_expr_wrapper` inside the same function, checks for `x.m_unit` to NOT be a nullptr. This results into a codegen error. This block gets accessed because `x.m_fmt` is a nullptr.

```c++
...
this->visit_expr_wrapper(x.m_unit, true);
```
```c++
inline void visit_expr_wrapper(ASR::expr_t* x, bool load_ref=false) {
        // Check if *x is nullptr.
        if( x == nullptr ) {
            throw CodeGenError("Internal error: x is nullptr");
        }

        this->visit_expr(*x);
        if( x->type == ASR::exprType::ArrayItem ||
            x->type == ASR::exprType::ArraySection ||
            x->type == ASR::exprType::StructInstanceMember ) {
            if( load_ref &&
                !ASRUtils::is_value_constant(ASRUtils::expr_value(x)) ) {
                tmp = CreateLoad(tmp);
            }
        }
    }
```

Shaikh-Ubaid:
I see. It also happens for LFortran. I fixed it in https://github.com/lfortran/lfortran/pull/3694. You can bring the changes here.

kmr-srbh:
@Shaikh-Ubaid as it currently stands, we can successfully take input at runtime. All the credit goes to you.

One small thing needs to be addressed. We are not able to _prompt_ the user. I see that `make_FileRead_t` has an argument `a_iomsg`. Is that the one which will help us? Could you please guide me on how we get access to the argument user passed?

Shaikh-Ubaid:
>We are not able to prompt the user.

Simply add a `visit_Print()` statement before FileRead inside the body of the function being defined in `instantiate_Input()`.

Shaikh-Ubaid:
> Is that the one which will help us?

No, the `iomsg` is for returning the error message generated by the compiler in case of an error.

kmr-srbh:
@Shaikh-Ubaid how do we go about writing tests for this?

kmr-srbh:
I found a method here - https://forum.learncodethehardway.com/t/testing-input-and-print/1757/3. It uses the unittest.mock library. Shall I write tests using it?

Shaikh-Ubaid:
I would not use any external library. For now I think we might need to test it manually.

kmr-srbh:
So we will have no tests for this. Am I right?

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -2569,6 +2571,65 @@ create_exp_macro(Exp, exp)
 create_exp_macro(Exp2, exp2)
 create_exp_macro(Expm1, expm1)
 
+namespace Input {
+    const std::string EMPTY_STRING = "";
```
Shaikh-Ubaid, line 2575: Make this a local variable instead.

Shaikh-Ubaid:
Let's port the fortran example at https://github.com/lfortran/lfortran/blob/main/examples/chat_01.f90 in python and add it in examples folder here.

Shaikh-Ubaid:
>So we will have no tests for this. Am I right?

Yes, it seems so at the moment.

kmr-srbh:
There are some serious issues with stdin. I guess they might be present in LFortran also.

1. Input is read until the space character is encountered.
2. EOF is not handled.

Shaikh-Ubaid (commented):
Posted a minor comment below. Can you check if `input()` without any argument works?

kmr-srbh:
I think we should make this an elemental function with the new changes. I will do that in some time. A conflict has occurred here too.

src/libasr/pass/intrinsic_function_registry.h
```diff
@@ -2569,6 +2571,67 @@ create_exp_macro(Exp, exp)
...
+namespace Input {
+
+    static inline void verify_args(const ASR::IntrinsicScalarFunction_t& x,
+            diag::Diagnostics& diagnostics) {
+        ASRUtils::require_impl(x.n_args <= 1,
+            "ASR Verify: Call `input` takes 0 or 1 argument",
+            x.base.base.loc, diagnostics);
+    }
+
+    static inline ASR::asr_t* create_Input(Allocator& al, const Location& loc,
```
certik, line 2583: Let's take this function and remove it from the IntrinsicScalarFunction registory, and put it into some utility file in libasr.

Then call it from the frontend.

certik:
See my comment above. This function is not deterministic, so it should not be an IntrinsicScalarFunction. But the approach is good otherwise, so just move it per my comment and it should work.

src/libasr/pass/intrinsic_function_registry.h
kmr-srbh: @certik What can be a good place to add it? 

Also, should we instead implement this through an ASR node? We can implement it exactly in the way `print()` is handled. Please guide me.

certik: For now put this `create_Input` into `src/libasr/asr_utils.cpp`. We can later move it elsewhere if we find a better place.

I would implement it exactly as you did just now, but make it a function in `asr_utils.cpp` and call it from the frontend. So ASR does not need to be modified. We can later improve all the IO nodes in ASR, but I want to use the same nodes for all the frontends.

kmr-srbh:
Closing in favor of #2641

PR #2605, opened 2024-03-14
Add support for global `set` declaration

### LINKED ISSUE #2604: Bug: Global declaration of `set` fails
Declaring a set in the global scope fails. This however is not the case with a local scope like a function.
```python
s: set[str] = {"a", "b", "c", "a"}
s.add("d")
```
```console
code generation error: Variable type not supported 6
 --> ./examples/example.py:1:1
  |
1 | s: set[str] = {"a", "b", "c", "a"}
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


```

As it turns out, the error message is not clear too.

kmr-srbh:
Fixes #2604 

### Working
```python
s: set[str] = {"a", "b", "c", "a"}
s.add("d")
```
```console
```


Shaikh-Ubaid (commented):
Please add integration tests for this.

Shaikh-Ubaid:

tests/set1.py
```diff
@@ -12,3 +12,7 @@ def test_Set():
 
     s: str
     s = b.pop()
+
+s1: set[str] = {"a", "b", "c", "a"}
+s2: set[i32] = {1, 2, 3, 1}
+s3: set[tuple[i32, i32]] = {(1, 2), (2, 3), (4, 5)}
```
Shaikh-Ubaid, line 18: We should be adding an integration test for this, not a reference test. A reference test is only needed when you want to check/verify the intermediate output or when adding an integration test is not possible/feasible.

Since in this PR you implemented support for global variables of type `set`, we want to know/check if an example with a global `set` variable compiles and runs and produces expected output.

Shaikh-Ubaid: Also, I think we should be adding a new test for this instead of updating an old test. Old test should be updated only when there is close similarity between the old test and the new test. We should mostly be adding new tests since we would like to keep integration tests short, testing only one specific functionality in a single test.

kmr-srbh: I am very sorry for this! I overlooked the folder name. I had wanted to add this to `integration_tests` only. I am pushing the changes in a while

integration_tests/test_global_set.py
```diff
@@ -0,0 +1,5 @@
+from lpython import i32
+
+s1: set[str] = {"a", "b", "c", "a"}
+s2: set[i32] = {1, 2, 3, 1}
+s3: set[tuple[i32, i32]] = {(1, 2), (2, 3), (4, 5)}
```
Shaikh-Ubaid, line 5: Can you add some kind of print or assert statement(s) here? For example, we can have an assert on the size of the set.

kmr-srbh: Sure!

integration_tests/test_global_set.py
```diff
@@ -0,0 +1,7 @@
...
+
+assert len(s1) == 3
```
Shaikh-Ubaid, line 7: Can we add asserts on lengths of other sets `s2` and `s3`? Also, is it possible to print these sets before the assert?

kmr-srbh: Printing on `set` is not implemented till now.

Shaikh-Ubaid: Ok, let's only have asserts for now.

integration_tests/test_global_set.py
```diff
@@ -0,0 +1,9 @@
...
+assert len(s2) == 3
```
Shaikh-Ubaid, line 8: should be `4`?
```suggestion
assert len(s2) == 4
```

Shaikh-Ubaid (approved):
It looks good to me. Thanks! Great work!

PR #2609, opened 2024-03-16
Add math functions

farah-salama:
## Added functions

- math.frexp(x): Returns the mantissa and exponent of x as the pair (m, e), where m is a float and e is an integer such that x == m * 2**e exactly.
- math.isclose(a, b, *, rel_tol = 1e-09, abs_tol = 0.0): Checks if the values a and b are close to each other.

## Testing
Tests are added in test_math.py to ensure the correctness of each function.


src/runtime/math.py
```diff
@@ -709,3 +709,43 @@ def remainder(x: f64, y: f64) -> f64:
...
+    '''
+    Return True if the values a and b are close to each other and False otherwise.
+    '''
+    difference:f64 = fabs(a-b)
+    greater:f64 = max(fabs(a),fabs(b))
+    return (difference <= rel_tol*greater) or (difference <= abs_tol)
+
+
+
+
```
Shaikh-Ubaid, line 751:
```suggestion
```

integration_tests/test_math.py
```diff
@@ -253,6 +253,31 @@ def test_issue_1242():
...
+    exponent:i16
+    mantissa, exponent = frexp(x)
+    assert abs(mantissa - 0.77875) < eps and exponent == i16(3)
+
+    x = 0.8
+    mantissa, exponent = frexp(x)
+    assert abs(mantissa - 0.8) < eps and exponent == i16(0)
+
+    x = 19.74
+    mantissa, exponent = frexp(x)
```
Shaikh-Ubaid, line 268: Please add test for negative floating numbers of types `f64` and `f32`.

integration_tests/test_math.py
```diff
@@ -253,6 +253,31 @@ def test_issue_1242():
...
+    assert abs(mantissa - 0.616875) < eps and exponent == i16(5)
+
+
+def test_isclose():
+    x:f64 = 2.2130
+    y:f64 = 2.2129
```
Shaikh-Ubaid, line 274: Please also test for negative value combinations. For example:
- x is positive, y is negative.
- x is negative, y is positive
- x is negative, y is negative.

src/runtime/math.py
```diff
@@ -709,3 +709,43 @@ def remainder(x: f64, y: f64) -> f64:
     if x - y*f64(q) > y*f64(q + i64(1)) - x:
         return x - y*f64(q + i64(1))
     return x - y*f64(q)
+
+
+@overload
+def frexp(x:f64) -> tuple[f64,i16]:
```
Shaikh-Ubaid, line 715: You need not change it here, but I am just curious to know why you used `i16` to store the exponent value in case of `f64` and `i8` to store the exponent value in case of `f32`.

Shaikh-Ubaid (commented):
Great work! Thanks for implementing this. I shared couple comments above.

Shaikh-Ubaid:

src/runtime/math.py
farah-salama: I tried to get the most accurate size needed for the exponent, as for `f32` it would need 8 bits in the worst case scenario and for `f64` it would need less than 16 bits that's why I used `i8` and `i16`.

farah-salama:
> Great work! Thanks for implementing this. I shared couple comments above.

Thanks for you time, I added the changes you suggested.

src/runtime/math.py
Shaikh-Ubaid: Ok, good, thanks!

src/runtime/math.py
```diff
@@ -709,3 +709,39 @@ def remainder(x: f64, y: f64) -> f64:
...
+
+
+@overload
+def isclose(a:f64, b:f64, rel_tol:f64 = 1e-09, abs_tol:f64 = 0.0) -> bool:
+    '''
+    Return True if the values a and b are close to each other and False otherwise.
+    '''
+    difference:f64 = fabs(a-b)
+    greater:f64 = max(fabs(a),fabs(b))
+    return (difference <= rel_tol*greater) or (difference <= abs_tol)
```
Shaikh-Ubaid, line 747: From https://docs.python.org/3/library/math.html#math.isclose, it seems the formula for `isclose()` is 
```python
abs(a-b) <= max(rel_tol * max(abs(a), abs(b)), abs_tol)
```
Therefore, should we update the above code to be the following?
```suggestion
    return difference <= max(rel_tol*greater, abs_tol)
```

Shaikh-Ubaid (commented):
Thanks for the improvements. I shared one last query and it will be good to merge.

Shaikh-Ubaid (approved):
It looks good to me. Thanks for this! Amazing work!

farah-salama:
> It looks good to me. Thanks for this! Amazing work!

Thank you!

PR #2611, opened 2024-03-17
Support constant negative indices in the array

### LINKED ISSUE #2503: Strange results with negative indices
See docstrings:

```python
from lpython import (i16, i32, Const)
from numpy import empty, int16
dim: Const[i32] = 10


def foo():
    """Negative indices produce random results each run."""
    A: i16[dim] = empty((dim,), dtype=int16)
    ww: i32
    for ww in range(dim):
        A[ww] = i16(ww + 1)
    print(A[0], A[1], A[2], "...", A[-3], A[-2], A[-1])

def bar(dim_: i32):
    """Negative indices always produce zero when 'dim' is a parameter."""
    A: i16[dim_] = empty((dim_,), dtype=int16)
    ww: i32
    for ww in range(dim_):
        A[ww] = i16(ww + 1)
    print(A[0], A[1], A[2], "...", A[-3], A[-2], A[-1])


foo()
bar(10)
```

```
(lp) ┌─(~/Documents/GitHub/lpython/integration_tests)──────────────────────────────────────────────────────────────────────────────(brian@MacBook-Pro:s001)─┐
└─(10:27:56 on vector-backend ✹ ✭)──> lpython ../ISSUES/Issue2503.py                                                                     ──(Wed,Feb07)─┘
1 2 3 ... 0 -25536 611
1 2 3 ... 0 0 0
```

hankluo6:
Fixed #2503 

## Test
```python
from lpython import (i16, i32, Const)
from numpy import empty, int16
dim: Const[i32] = 10


def foo():
    """Negative indices produce random results each run."""
    A: i16[dim] = empty((dim,), dtype=int16)
    ww: i32
    for ww in range(dim):
        A[ww] = i16(ww + 1)
    print(A[0], A[1], A[2], "...", A[-3], A[-2], A[-1])

def bar(dim_: i32):
    """Negative indices always produce zero when 'dim' is a parameter."""
    A: i16[dim_] = empty((dim_,), dtype=int16)
    ww: i32
    for ww in range(dim_):
        A[ww] = i16(ww + 1)
    print(A[0], A[1], A[2], "...", A[-3], A[-2], A[-1])


foo()
bar(10)
```
```shell
(lp) hank@MacBook-Pro lpython % ./src/bin/lpython test.py           
1 2 3 ... 8 9 10
1 2 3 ... 8 9 10
```

I've introduced a new argument to reference the current subscript index as I need it to obtain the corresponding shape of the dimension. Please let me know if there is a better solution for this.


kmr-srbh:
@hankluo6 Thanks for working on this! I really appreciate it!

I had previously done some research on this issue to understand the depth of the problem, and found that there is more to it than just the negative indices returning incorrect values. Consider this line instead of the one in `foo()`.
 ```python
print(A[0], A[1], A[20], "...", A[-300], A[-0], A[-1])
```
The values passed here are invalid. Note the interesting thing - `-0`. But it compiles.
```console
1 2 -1896 ... 0 1 10
```

Things change a bit for large positive indices:
```python
print(A[1000], A[1], A[20], "...", A[-300], A[-0], A[-1])
```
```console
0 2 -22504 ... 0 1 10
```
```python
print(A[10000], A[1], A[20], "...", A[-300], A[-0], A[-1])
```
```console
Bus error (core dumped)
```

kmr-srbh:
The problem does not end here, as it turns out, this is an issue with how we handle negative indices.
For a normal list:
```python
from lpython import i32, list

A: list[i32] = [1, 2, 4, 5]
print(A[0], A[1], A[2], "...", A[-3], A[-2], A[-1])
```
The output is the same buggy one:
```console
...
```

kmr-srbh:
The above testing is done applying the changes made in this PR.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3812,6 +3812,20 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+            } else if (ASR::is_a<ASR::Array_t>(*type)) {
+                index = ASRUtils::EXPR(tmp);
+                ASR::expr_t* val = ASRUtils::expr_value(index);
+                if (val && ASR::is_a<ASR::IntegerConstant_t>(*val) && 
+                    ASRUtils::extract_physical_type(type) != ASR::array_physical_typeType::DescriptorArray) {
+                    if (ASR::down_cast<ASR::IntegerConstant_t>(val)->m_n < 0) {
+                        ASR::ttype_t *int_type = ASRUtils::TYPE(ASR::make_Integer_t(
+                                                        al, loc, 4));
+                        ASR::expr_t *neg_idx = ASRUtils::expr_value(index);
+                        ASR::expr_t *dim_size = ASR::down_cast<ASR::Array_t>(type)->m_dims[idx].m_length;
```
Shaikh-Ubaid, line 3824: This gives you the dimension size only when it is available at compile-time. For descriptor arrays, the length is available at runtime. For example, the following fails with the changes in this PR.

```console
% cat examples/expr2.py   
from lpython import i32
from numpy import empty, int32, array

def foo(x: i32[:]):
    print(x[3], x[4], x[-1], x[-2])

def main():
    x: i32[5] = empty(5, dtype=int32)
    x = array([1, 2, 3, 4, 5])
    foo(x)

main()
% python examples/expr2.py
4 5 5 4
% lpython examples/expr2.py
4 5 1 12844716
```

To get the dimension size at runtime, you can use `ArrayBound()` and pass it the appropriate dimension index (whose dimension you need).

integration_tests/array_expr_09.py
```diff
@@ -0,0 +1,22 @@
...
+
+def g():
+    a: i32[dim, dim2] = empty((dim, dim2), dtype=int32)
+    i1: i32 = 0
+    i2: i32 = 0
+    for i1 in range(dim):
+        for i2 in range(dim2):
+            a[i1, i2] = i32(i1 * dim2 + i2)
+    # a: [[0, 1, 2], [3, 4, 5]]
+    assert a[-1, -1] == 5
```
Shaikh-Ubaid, line 15: You can also print the array values before the asserts, so that it is helpful for debugging later.

Shaikh-Ubaid:
> The problem does not end here, as it turns out, this is an issue with how we handle subscript indices. The output is correct when we have only list elements, but when something else like a string literal enters, things start to break. For a normal list:
> 
> ```python
> from lpython import i32, list
> 
> A: list[i32] = [1, 2, 4, 5]
> print(A[0], A[1], A[2], "...", A[-3], A[-2], A[-1])
> ```
> 
> The output is the same buggy one:
> 
> ```
> (lp) saurabh-kumar@Awadh:~/Projects/System/lpython$ ./src/bin/lpython ./examples/example.py
> ...
> ```

It seems this PR tackles negative indices in array. I opened a separate issue for the above https://github.com/lcompilers/lpython/issues/2613.

Shaikh-Ubaid:
>Consider this line instead of the one in foo().
>print(A[0], A[1], A[20], "...", A[-300], A[-0], A[-1])

These examples do not run with CPython and throw IndexError, so they are not expected to work with LPython. In LPython we can throw a similar error when index passed is not valid, but that is or can be a separate issue.

Shaikh-Ubaid (commented):
Perfect! I shared comments https://github.com/lcompilers/lpython/pull/2611/files#r1527548439 and  https://github.com/lcompilers/lpython/pull/2611/files#r1527549155. Rest looks good. Great Work!

Shaikh-Ubaid:

hankluo6:
@kmr-srbh  Thanks for pointing that out. I think this problem can be resolved once we implement the bound check for the array. https://github.com/lcompilers/lpython/blob/0ba09ab5a0e38432be4d839761d0e92beabca94d/src/libasr/codegen/llvm_array_utils.cpp#L569-L571

hankluo6:
I've tried using `ArrayBound`, but it throws an error when combined with `IntegerBinOp`, so I used `ArraySize`, which works as expected.

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid:
```console
% cat examples/expr2.py 
from lpython import i32
from numpy import empty, int32, array

def foo(x: i32[:]):
    print(x[3], x[4], x[-1], x[-2])

def main():
    x: i32[5] = empty(5, dtype=int32)
    x = array([1, 2, 3, 4, 5])
    foo(x)

main()
% python examples/expr2.py 
4 5 5 4
% lpython examples/expr2.py
4 5 5 4
```
The above test case now works.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -3812,6 +3812,25 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                        ASR::ttype_t *int_type = ASRUtils::TYPE(ASR::make_Integer_t(
+                                                        al, loc, 4));
+                        ASR::expr_t *neg_idx = ASRUtils::expr_value(index);
+                        ASR::expr_t *dim_size;
+                        if (ASRUtils::extract_physical_type(type) != ASR::array_physical_typeType::DescriptorArray)
+                            dim_size = ASR::down_cast<ASR::Array_t>(type)->m_dims[idx].m_length;
+                        else {
+                            ASR::expr_t *idx_expr = ASRUtils::EXPR(ASR::make_IntegerConstant_t(al, loc, idx + 1, int_type));
+                            dim_size = ASRUtils::EXPR(ASRUtils::make_ArraySize_t_util(al, loc, value, idx_expr, int_type, nullptr, false));
+                        }
```
Shaikh-Ubaid, line 3829: I think we need not check for descriptor array here. The backend implementation of ArraySize would do this check and appropriately return back the dimension size.

integration_tests/CMakeLists.txt
```diff
@@ -434,6 +434,7 @@ RUN(NAME array_expr_05            LABELS cpython llvm c)
 RUN(NAME array_expr_06            LABELS cpython llvm c)
 RUN(NAME array_expr_07            LABELS cpython llvm c)
 RUN(NAME array_expr_08            LABELS cpython llvm c)
+RUN(NAME array_expr_09            LABELS cpython llvm c)
```
Shaikh-Ubaid, line 437: We should also add the test case discussed in https://github.com/lcompilers/lpython/pull/2611#discussion_r1529321057 as an integration test.

Shaikh-Ubaid (approved):
I shared two comments below. Rest looks good to me. Thanks for this! Great work!

Shaikh-Ubaid:
We can tackle the above two things in a separate PR.

src/lpython/semantics/python_ast_to_asr.cpp
hankluo6: But then, is the negative indexing still done at compile time for the fixed-size array? My understanding was that ArraySize would create a new alloca in the backend and couldn't be reduced to a constant at compile time.

Shaikh-Ubaid: >But then, is the negative indexing still done at compile time for the fixed-size array?

It would be computed at run-time.

If you want it to work using the compile-time value, then you can provide the compile time value in the `m_value` parameter of `make_ArraySize_t_util()`. For example, you can do:
```console
compile_time_dim_size = ASR::down_cast<ASR::Array_t>(type)->m_dims[idx].m_length;
ASR::expr_t *idx_expr = ASRUtils::EXPR(ASR::make_IntegerConstant_t(al, loc, idx + 1, int_type));
dim_size = ASRUtils::EXPR(ASRUtils::make_ArraySize_t_util(al, loc, value, idx_expr, int_type, compile_time_dim_size, false));
```
If the compile-time value is availabe (for example in the case of a fixed size array), it would be used. If the compile-time value is not available, then the dimension size would be computed at run-time.

hankluo6: Got it! Thank you. I will update it in another PR.

certik:
Thanks. I think this PR is fine. No overhead for positive indices and for compile time negative indices it does what the user would have to do anyway if we didn't support this feature, so I think this is fine.

Issue #2573, opened 2024-03-03
[Bug] All global variable assignments should not go to `__main__global_init`


Vipul-Cariappa:
I came across this bug while trying to debug #2457.

Look at the code and output of the two cases

Code:
```python
l: list[i32] = [1, 2, 3, 4]
print("Before Pop:", l)
x: i32 = l.pop()
print("After Pop:", l)
print("Poped Element: ", x)
```
Output:
```text
Before Pop: [1, 2, 3]
After Pop: [1, 2, 3]
Poped Element:  4
```

---

Code:
```python
l: list[i32] = [1, 2, 3, 4]
print("Before Pop:", l)
l.pop()
print("After Pop:", l)
```
Output:
```text
Before Pop: [1, 2, 3, 4]
After Pop: [1, 2, 3]
```

This is happening because the line `x: i32 = l.pop()` is compiled into the `__main__global_init` function, but it should actually be compiled into `__main__global_stmts`. Because `__main__global_init` is called before `__main__global_stmts`.

PR #2617, opened 2024-03-19
Interactive shell implementation

### LINKED ISSUE #329: Interactivity
Make LPython work in an interactive prompt.

Copy how LFortran does it. We have to remember a symbol table, and compile each prompt as we go. LFortran has all the infrastructure, I think it's not difficult to add to LPython.

Vipul-Cariappa:
Fixes #329

# TODOs
- [ ] Remember global functions and variables declared.
- [ ] Error message with proper line number and highlighting
- [ ] Intrinsic functions support
- [ ] print the value of the evaluated expression
- [ ] Import other modules
- [ ] `symengine` and `numpy` support
- [ ] Support complex data types like `list`, `tuple` and `dict`

Got the following things to work:
image

### DIFF
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -810,19 +810,20 @@ int interactive_python_repl(
     LCompilers::LocationManager lm;
     std::vector<std::pair<std::string, double>> times;
     LCompilers::PythonCompiler::EvalResult r;
-    
+
     std::string code_string;
     std::cout << ">>> ";
     size_t cell_count = 0;
     for (std::string input; std::getline(std::cin, input);) {
-        if (input == "exit" || input == "quit") // exit condition
+        if (input == "exit" || input == "quit") {
             return 0;
+        }
 
-        if ((input.rfind("def", 0) == 0) || 
-            (input.rfind("for", 0) == 0) || 
-            (input.rfind("if", 0) == 0) || 
-            (input.rfind("else", 0) == 0) || 
-            (input.rfind("elif", 0) == 0) || 
+        if ((input.rfind("def", 0) == 0) ||
+            (input.rfind("for", 0) == 0) ||
+            (input.rfind("if", 0) == 0) ||
+            (input.rfind("else", 0) == 0) ||
+            (input.rfind("elif", 0) == 0) ||
             (input.rfind("class", 0) == 0) ||
             (input.rfind('@', 0) == 0) ||
             (input.rfind(' ', 0) == 0) ||
@@ -859,11 +860,11 @@ int interactive_python_repl(
                 std::cout << ">>> ";
                 continue;
             }
-            
+
             auto evaluation_end_time = std::chrono::high_resolution_clock::now();
             times.push_back(std::make_pair("evalution " + std::to_string(cell_count), std::chrono::duration
                 <double, std::milli>(evaluation_start_time - evaluation_end_time).count()));
-        
+
         } catch (const LCompilers::LCompilersException &e) {
             std::cerr << "Internal Compiler Error: Unhandled exception" << std::endl;
             std::vector<LCompilers::StacktraceItem> d = e.stacktrace_addresses();
@@ -1934,8 +1935,6 @@ int main(int argc, char *argv[])
         compiler_options.use_colors = !arg_no_color;
         compiler_options.indent = !arg_no_indent;
 
-        lpython_pass_manager.parse_pass_arg(arg_pass, skip_pass);
-        lpython_pass_manager.use_default_passes();
 
         // if (fmt) {
         //     return format(arg_fmt_file, arg_fmt_inplace, !arg_fmt_no_color,
@@ -1983,6 +1982,8 @@ int main(int argc, char *argv[])
         }
 
         if (arg_files.size() == 0) {
+            lpython_pass_manager.parse_pass_arg(arg_pass, skip_pass);
+            lpython_pass_manager.use_default_passes();
             compiler_options.po.disable_main = true;
             compiler_options.emit_debug_line_column = false;
             compiler_options.generate_object_code = false;
@@ -2029,6 +2030,7 @@ int main(int argc, char *argv[])
         //     return emit_c_preprocessor(arg_file, compiler_options);
         // }
 
+        lpython_pass_manager.parse_pass_arg(arg_pass, skip_pass);
         if (show_tokens) {
             return emit_tokens(arg_file, true, compiler_options);
         }
@@ -2070,6 +2072,7 @@ int main(int argc, char *argv[])
             return 1;
 #endif
         }
+        lpython_pass_manager.use_default_passes();
         if (show_llvm) {
 #ifdef HAVE_LFORTRAN_LLVM
             return emit_llvm(arg_file, runtime_library_dir, lpython_pass_manager, compiler_options);
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -33,6 +33,7 @@
 #include <lpython/python_serialization.h>
 #include <lpython/parser/tokenizer.h>
 #include <lpython/parser/parser.h>
+#include <libasr/exception.h>
 
 #include <cpp-terminal/terminal.h>
 #include <cpp-terminal/prompt0.h>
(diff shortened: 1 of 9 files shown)


Vipul-Cariappa:
I want to know how should I proceed with the first to-do point "Remember global functions and variables declared". The problem is as follows

```bash
>>> def f():
...   print("called function f")
... 
>>> f()
semantic error: Function 'f' is not declared and not intrinsic
 --> input:1:1
  |
1 | f()
  | ^ 
```

Another example:
```bash
>>> x: i32 = 10
>>> print(x)
semantic error: Variable 'x' not declared
 --> input:1:1
  |
1 | print(x)
  | ^ 
```

The declared function `f` is lost by the time we evaluate (i.e. parse and execute) the function call to `f`. We should store its declaration (and `x`'s declaration) somewhere (I believe at the AST level) and include it while parsing the second part/prompt.

I want to know if there are existing methods to do this. Or should I implement it myself. I want something like.
`get_all_function_declerations(AST_t*)` and `get_all_global_declerations(AST_t*)`, and `extend_ast_with_function_decleration(AST_t*, Function_t*)` and `extend_ast_with_global_decleration(AST_t*, ...)`

Shaikh-Ubaid:
I think the idea to solve this is to keep updating the original ASR on every new interactive query run by the user.

On every run, the functions or variables defined in the previous run need to be updated. We remove the body of functions and set their `abiType` to interactive. Similarly for variables, we mark them as interactive. Later we handle the `interactive` functions/variables appropriately in the `AST->ASR`, `ASR passes` and `ASR->backend`. 

LFortran already supports interactive compilation. You can look into it for reference. Since the ASR passes and backends are shared with LFortran, I think we will mostly need interactive support for AST->ASR in LPython.

Shaikh-Ubaid:
Let's build this iteratively. For this PR, let's get something that works with the interactive shell, add tests for the implemented support and ensure the CI passes. After that, you can send more PRs to improve the existing support and/or add more features as planned in https://github.com/lcompilers/lpython/pull/2617#issue-2195119178. Ideally, small and frequent PRs that just focus on one particular support with robust tests are preferable.

For tests, you can look into how lfortran does it https://github.com/lfortran/lfortran/blob/582db3d4379eee26cfeb0721d5fda21971aaae3d/src/lfortran/tests/test_llvm.cpp#L439-L455.

Vipul-Cariappa:
It makes more sense to have multiple smaller PR than a large one. I will finish up the work with remembering context from the previous cells, i.e. remembering global functions and variables.
We can add in tests and merge this PR. The other points can be their own separate PRs.

src/lpython/python_evaluator.cpp
```diff
@@ -68,7 +200,7 @@ Result<std::unique_ptr<LLVMModule>> PythonCompiler::get_llvm3(
     Result<std::unique_ptr<LCompilers::LLVMModule>> res
         = asr_to_llvm(asr, diagnostics,
             e->get_context(), al, lpm, compiler_options,
-            run_fn, infile);
+            "", infile); // ??? What about run function
```
Vipul-Cariappa, line 203: The argument `""` is the `run_fn`. The `run_fn` is not used here. In LFortran all global statements are compiled into 1 function, but in LPython it is compiled into two different functions one for global initialization and the other to execute global statements.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4956,6 +5033,101 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
 
         tmp = asr;
     }
+    
+    void visit_Interactive(const AST::Interactive_t &x) {
+        static size_t interactive_execution_count = 0; // ???: should this be a class member variable
+        // interactive_execution_count should be the same as eval_count in PythonCompiler
```
Vipul-Cariappa, line 5039: Any suggestions on this?
Should we create `run_fn1` and `run_fn2` and share it through the `PassOptions`?
But even that may not be straightforward.

Vipul-Cariappa:
I guess we can merge the work done till this point. The redefinition of symbols is not yet supported. It looks more complicated than initially thought. We can add support for the redefinition of symbols in a later PR.

Not sure why the tests are failing.

I have left 2 comments please have a look at it.

src/lpython/semantics/python_ast_to_asr.cpp
Vipul-Cariappa: Another option would be to combine the `init` and `stmts` functions. Look at #2573

Shaikh-Ubaid:
You need to update the reference tests. You can do `./run_tests.py -u`.

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: I saw the linked issue. If combining `init` and `stmt` function solves your issue, then you can try it. Submit it as a separate PR and we can rebase this after that is merged.

Shaikh-Ubaid:
At a high level, it looks good. I will review in detail in evening today.

src/lpython/parser/parser.cpp
```diff
@@ -42,6 +42,33 @@ Result<LPython::AST::Module_t*> parse(Allocator &al, const std::string &s,
...
+
+    Location l;
+    if (p.result.size() == 0) {
+        l.first=0;
+        l.last=0;
+    } else {
+        l.first=p.result[0]->base.loc.first;
+        l.last=p.result[p.result.size()-1]->base.loc.last;
+    }
+    return LPython::AST::make_Interactive_t(al, l, p.result.p, p.result.size());
```
Shaikh-Ubaid, line 69: Any specific reason why we use `make_Interactive_t` here? It seems if we reuse the existing `parse()` function (on line `17` of this file) which returns a `Module` we can save a lot of code being added in this PR.

For example, almost all the code added in `python_ast_to_asr.cpp` would not be needed. It seems to be duplicate of existing code for `visit_Module()`.

src/lpython/semantics/python_ast_to_asr.h
```diff
@@ -8,7 +8,7 @@ namespace LCompilers::LPython {
 
     Result<ASR::TranslationUnit_t*> python_ast_to_asr(Allocator &al, LocationManager &lm, SymbolTable* symtab,
         LPython::AST::ast_t &ast, diag::Diagnostics &diagnostics, CompilerOptions &compiler_options,
-            bool main_module, std::string module_name, std::string file_path, bool allow_implicit_casting=false);
+            bool main_module, std::string module_name, std::string file_path, bool allow_implicit_casting=false, bool is_interactive=false);
```
Shaikh-Ubaid, line 11: You can also pass the `eval_count` here and have it as a class member. Then for `body visitor` inside `visit_Module` (assuming we are reusing this instead of defining `visit_Interactive`), you can use the `eval_count` for naming `global_init` and `global_stmts` if `interactive = true`. (For example simply append the eval_count to the end if interactive is true).

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -8359,20 +8555,38 @@ std::string get_parent_dir(const std::string &path) {
...
+            unit = res.result;
+        } else {
+            return res.error;
+        }
     } else {
-        return res.error;
+        ast_i = AST::down_cast2<AST::Interactive_t>(&ast);
+        if (!symtab) {
+            // Create new empty symbol table, as a module
+        }
```
Shaikh-Ubaid, line 8580: Do we need this here? Is this a work in progress?

src/lpython/python_evaluator.cpp
```diff
@@ -28,14 +33,144 @@ PythonCompiler::PythonCompiler(CompilerOptions compiler_options)
...
+    }
+
+    if (verbose) {
+        result.llvm_ir = m->str();
+    }
+
+    bool call_init = false;
+    bool call_stmts = false;
+    std::string init_fn = "__module___main_____main____lpython_interactive_init_" + std::to_string(eval_count) + "__";
+    std::string stmts_fn = "__module___main_____main____lpython_interactive_stmts_" + std::to_string(eval_count) + "__";
```
Shaikh-Ubaid, line 104: Construct this like

```
std::string base_fn_name = "_main_____main____lpython_interactive";
std::string init_fn = base_fn_name + "_init_" + std::to_string(eval_count);
std::string module_init_fn = "__module__" + init_fn;
```

Similarly for `stmts`. Now this can be reused when erasing symbol below.

src/lpython/python_evaluator.cpp
```diff
@@ -28,14 +33,144 @@ PythonCompiler::PythonCompiler(CompilerOptions compiler_options)
...
+    }
+
+    if (call_init) {
+        ASR::down_cast<ASR::Module_t>(symbol_table->resolve_symbol("__main__"))->m_symtab
+            ->erase_symbol("__main____lpython_interactive_init_" + std::to_string(eval_count) + "__");
+    }
+    if (call_stmts) {
+        ASR::down_cast<ASR::Module_t>(symbol_table->resolve_symbol("__main__"))->m_symtab
+            ->erase_symbol("__main____lpython_interactive_stmts_" + std::to_string(eval_count) + "__");
+    }
```
Shaikh-Ubaid, line 127: Reuse the variables from the suggestion above.

Shaikh-Ubaid (commented):
I shared a couple comments. 

Mainly:
- Let's reuse `Module`, that way we get to reuse a lot of other existing code
- We can merge `global_init` and `global_stmts` in a separate PR. I think we need it to solve https://github.com/lcompilers/lpython/issues/2573
- Minor refactoring

Apart from that, it looks good. Great start! Amazing work!

src/bin/lpython.cpp
```diff
@@ -1785,6 +1941,9 @@ int main(int argc, char *argv[])
         compiler_options.use_colors = !arg_no_color;
         compiler_options.indent = !arg_no_indent;
 
+        lpython_pass_manager.parse_pass_arg(arg_pass, skip_pass);
+        lpython_pass_manager.use_default_passes();
+
```
Shaikh-Ubaid, line 1946: Do you need this change for `interactivity`? 

It should not be here. `lpython_pass_manager.use_default_passes();` applies ASR passes to the output of `--show-asr`. This changes the reference results and therefore the failure at the CI.

Let's revert this back.

Shaikh-Ubaid:

src/lpython/python_evaluator.cpp
Vipul-Cariappa: Now that we have merged `global_init` and `global_stmts`, we can assign `run_fn` to `global_stmts` and use it.

Then that change would be part of this PR. You can have a look at it in successive commits.

src/bin/lpython.cpp
Vipul-Cariappa: `lpython_pass_manager.parse_pass_arg(arg_pass, skip_pass);` and `lpython_pass_manager.use_default_passes();` is required to be initilised before calling `interactive_python_repl`. I will remove it from here and put it just before calling `interactive_python_repl`.

Overall behavior will be the same as before.

src/lpython/semantics/python_ast_to_asr.cpp
Vipul-Cariappa: This is not required. There are other changes that givens the same effect as creating a `SymbolTable` here.

src/lpython/parser/parser.cpp
Vipul-Cariappa: There are some differences between `SymbolTreeVisitor.visit_Interactive` and `SymbolTreeVisitor.visit_Module`.
I will try merging both. Most probably it should be possible. Then we can remove `make_Interactive_t` and reuse `visit_Module` at the places `visit_Interactive` is used.

src/lpython/python_evaluator.cpp
```diff
@@ -1,10 +1,15 @@
...
-#include <fstream>
 
 #include <lpython/python_evaluator.h>
+#include <lpython/semantics/python_ast_to_asr.h>
 #include <libasr/codegen/asr_to_cpp.h>
 #include <libasr/exception.h>
 #include <libasr/asr.h>
+#include <lpython/pickle.h>
+#include <lpython/parser/parser.h>
+#include <string>
```
Shaikh-Ubaid, line 12: I think this should be included with other system headers at the top. I am unsure if system headers should be listed first and then the custom headers or the other way round. But I think we should atleast not mix them.

src/lpython/python_evaluator.cpp
```diff
@@ -47,9 +174,6 @@ Result<std::unique_ptr<LLVMModule>> PythonCompiler::get_llvm3(
     )
 {
 #ifdef HAVE_LFORTRAN_LLVM
-    eval_count++;
-    run_fn = "__lfortran_evaluate_" + std::to_string(eval_count);
-
```
Shaikh-Ubaid, line 52: We later plan to move the interactivity support into the LibASR so that it is shared between LFortran, LPython and LC. So, it is better that we have/use as common/generic code as possible which easily gets shared across all LCompilers.

Assuming it does not affect the newly added interactivity, I think for now it is better to have it.

src/lpython/python_evaluator.h
```diff
@@ -59,10 +71,10 @@ class PythonCompiler
     Allocator al;
 #ifdef HAVE_LFORTRAN_LLVM
     std::unique_ptr<LLVMEvaluator> e;
-    int eval_count;
+    size_t eval_count;
```
Shaikh-Ubaid, line 74: This is passed as argument to `python_ast_to_asr()`. The call to `python_ast_to_asr()` is also used by other backends. So, this should be declared outside of `#ifdef HAVE_LFORTRAN_LLVM`.

Shaikh-Ubaid (commented):
Good! Few more minor comments. I think it will be good to merge after that. We can add tests in a separate PR.

src/lpython/python_evaluator.cpp
Vipul-Cariappa: Can we keep this as part of the next PR? It may require changes in more than one place.

> I was working towards "print the value of the evaluated expression" (https://github.com/lcompilers/lpython/pull/2617#issue-2195119178) goal, and I think there might be some obstacles to doing this. But is doable.

Vipul-Cariappa:
I have made the suggested changes. Please have a look. I have also left a comment on one of your suggestions.

src/lpython/python_evaluator.cpp
Shaikh-Ubaid: >Can we keep this as part of the next PR?

We can do that.

Shaikh-Ubaid (approved):
It looks good! Thanks for this! Great work!

Shaikh-Ubaid:
Let's add few/some tests for this in a next PR!

PR #2618, opened 2024-03-20
Function Default Arguments

### LINKED ISSUE #2109: Default arguments not working ?
The following works in CP, but not LP. 

```python
from random import random

from lpython import (i8, i32, i64, f32, f64,
                     u8, u32,
                     TypeVar, Const,
                     dataclass,
                     # random
                     )
from numpy import (empty, sqrt, float32, float64,
                   int8, int32, array, # ndarray
                   )


def coin_flip(probability : f64=f64(0.5)) -> i8:
    """roll 1 bit"""
    result : i8 = i8(0)
    r : f64 = random()
    if r < probability:
        result = i8(1)
    return result


def h() -> None:
    i : i32
    for i in range(10):
        coin : i8 = coin_flip()
        print(coin, end='')
    print('')


if __name__ == "__main__":
    h()
```

```bash
(lp) ┌─(~/CLionProjects/lpython/lasr/LP-pycharm)─────────────────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s000)─┐
└─(11:36:44 on brian-lasr ✹ ✭)──> PYTHONPATH='../../src/runtime/lpython' python issue2107.py                                                          1 ↵ ──(Wed,Jul05)─┘
1
1
1
1
1
0
1
1
0
0
(lp) ┌─(~/CLionProjects/lpython/lasr/LP-pycharm)─────────────────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s000)─┐
└─(11:36:49 on brian-lasr ✹ ✭)──> ~/CLionProjects/lpython/src/bin/lpython -I. issue2107.py                                                                ──(Wed,Jul05)─┘
semantic error: Number of arguments does not match in the function call
  --> issue2107.py:26:21
   |
26 |         coin : i8 = coin_flip()
   |                     ^^^^^^^^^^^ (found: '0', expected: '1')
```

### LINKED ISSUE #2509: Incorrect handling of default argument values
```python
def main(option: bool = False):
    print("option: ", option)

main()
```

```
(lp) ┌─(~/Documents/GitHub/masr/resources/reference)────────────────────────────────────────(brian@MacBook-Pro:s001)─┐
└─(08:43:14 on main ✹ ✭)──> PYTHONPATH=".:../src/runtime/lpython:.." LPYTHON_PY_MOD_NAME="" LPYTHON_PY_MOD_PATH="_lpython-tmp-test-cpython" python ../../../lpython/ISSUES/Issue2509.py
option:  False
(lp) ┌─(~/Documents/GitHub/masr/resources/reference)────────────────────────────────────────(brian@MacBook-Pro:s001)─┐
└─(08:43:40 on main ✹ ✭)──> lpython ../../../lpython/ISSUES/Issue2509.py                          ──(Thu,Feb08)─┘
semantic error: Number of arguments does not match in the function call
 --> ../../../lpython/ISSUES/Issue2509.py:5:1
  |
5 | main()
  | ^^^^^^ (found: '0', expected: '1')
```

assem2002:
fixes #2509 
fixes #2109  
completed the implementation of default arguments in functions.
# Code Flow
- `AST::functiondef` has `arguments` object as one of its members, this object has `m_defaults` and `n_defaults` which preserve the number of default arguments and `AST::expr_t` of the default constants.
- When we're building the variables of the function in `visit_functiondef( )` , we start pushing the ASR::expr_t of the defaults as a value for the variable when we reach the difference of `n_args` and `n_defaults` .
- When we start to call the function, we check if `n_pos_args` is less than `func.n_args` , if so, we start pushing the data from varibles in ASR::function_t  to ` Vec<ASR::call_arg_t>`,starting from the index n_pos_args.

# The code handles keyword arguments.

# serious issue to handle
The solution is build upon the idea of starting to make variable for the last number of  parameters. If the user writes a code like this 
```python
def whatever(first :str ="lpython",second :str):
    print(second)

whatever("foo") # result => lpython

``` 
as the lpython would be matched with the last parameter.

This should arise a syntax error like how Cpython does.

### DIFF
--- a/integration_tests/def_func_01.py
+++ b/integration_tests/def_func_01.py
@@ -1,0 +1,68 @@
+from lpython import i32,i64
+
+def factorial_1(x: i32, y:i32 =1) ->i32 :
+    if x <= 1:
+         return y
+    return x * factorial_1(x-1)
+
+def factorial_2(x: i32, y:i32=3 ,z:i32 =2) ->i32:
+    if x ==4:
+        return x * y * z
+    return x * factorial_2(x-1)
+
+def default_func(x : str ="Hello" ,y : str = " ", z : str = "World") ->str:
+    return x + y + z
+
+
+def even_positons(iterator : i32 , to_add : str = "?")-> str:
+    if (iterator == 10): return ""
+
+    if iterator%2 == 0 :
+        return to_add + even_positons(iterator+1,"X")
+
+    return to_add +even_positons(iterator+1)
+
+
+
+def test_all():
+
+    test_01 : i32  = factorial_1(5,0)
+    print("test_01 is =>",test_01)
+    assert test_01 == 120
+
+    test_02 : i32  = factorial_1(1,5555)
+    print("test_02 is =>",test_02)
+    assert test_02 == 5555
+
+    test_03 : i32 =factorial_2(5,99999,99999)
+    print("test_03 is =>",test_03)
+    assert test_03 == 120
+
+    test_04 : i32  = factorial_2(4,-1,100)
+    print("test_04 is =>",test_04)
+    assert test_04 == -400
+
+    test_05 :str = default_func()
+    print("test_05 is =>",test_05)
+    assert test_05 == "Hello World"
+
+    test_06 :str = default_func(y = "|||",x="Hi")
+    print("test_06 is =>",test_06)
+    assert test_06 == "Hi|||World"
+
+    test_07 : str = even_positons(0)
+    print("test_07 is =>",test_07)
+    assert test_07 == "?X?X?X?X?X"
+
+    test_08 : str = even_positons(0,"W")
+    print("test_08 is =>",test_08)
+    assert test_08 == "WX?X?X?X?X"
+
+test_all()
+
+
+
+
+
+
+    
--- a/integration_tests/def_func_01.py
+++ b/integration_tests/def_func_01.py
@@ -22,7 +22,7 @@ def even_positions(iterator : i32, to_add : str = "?")-> str:
 
 
 
-def test_all():
+def test_factorial_1():
     test_00 : i32  = factorial_1(1)
     print("test_00 is =>", test_00)
     assert test_00 == 1
@@ -35,6 +35,7 @@ def test_all():
     print("test_02 is =>", test_02)
     assert test_02 == 5555
 
+def test_factorial_2():
     test_03 : i32 =factorial_2(5,99999,99999)
     print("test_03 is =>", test_03)
     assert test_03 == 120
@@ -43,6 +44,7 @@ def test_all():
     print("test_04 is =>", test_04)
     assert test_04 == -400
 
+def test_default_func():
     test_05 :str = default_func()
     print("test_05 is =>", test_05)
     assert test_05 == "Hello World"
@@ -51,19 +53,24 @@ def test_all():
     print("test_06 is =>", test_06)
     assert test_06 == "Hi|||World"
 
-    test_07 : str = even_positions(0)
+    test_07 :str = default_func(y = "++",z = "LPython")
     print("test_07 is =>", test_07)
-    assert test_07 == "?X?X?X?X?X"
+    assert test_07 == "Hello++LPython"
 
-    test_08 : str = even_positions(0,"W")
-    print("test_08 is =>", test_08)
-    assert test_08 == "WX?X?X?X?X"
+    test_8 :str = default_func("Welcome",z = "LPython")
+    print("test_8 is =>", test_8)
+    assert test_8 == "Welcome LPython"
 
-    test_09 :str = default_func(y = "++",z = "LPython")
+def test_even_positions():
+    test_09 : str = even_positions(0)
     print("test_09 is =>", test_09)
-    assert test_09 == "Hello++LPython"
+    assert test_09 == "?X?X?X?X?X"
 
-    test_10 :str = default_func("Welcome",z = "LPython")
+    test_10 : str = even_positions(0,"W")
     print("test_10 is =>", test_10)
-    assert test_10 == "Welcome LPython"
-test_all() 
+    assert test_10 == "WX?X?X?X?X"
+
+test_factorial_1()
+test_factorial_2()
+test_default_func()
+test_even_positions() 
--- a/integration_tests/def_func_01.py
+++ b/integration_tests/def_func_01.py
@@ -10,71 +10,60 @@ def factorial_2(x: i32, y:i32=3 ,z:i32 =2) ->i32:
         return x * y * z
     return x * factorial_2(x-1)
 
-def default_func(x : str ="Hello" ,y : str = " ", z : str = "World") ->str:
+def default_func(x : str ="Hello", y : str = " ", z : str = "World") ->str:
     return x + y + z
 
 
-def even_positions(iterator : i32 , to_add : str = "?")-> str:
+def even_positions(iterator : i32, to_add : str = "?")-> str:
     if (iterator == 10): return ""
-
     if iterator%2 == 0 :
         return to_add + even_positions(iterator+1,"X")
-
     return to_add +even_positions(iterator+1)
 
 
 
 def test_all():
-
     test_00 : i32  = factorial_1(1)
-    print("test_00 is =>",test_00)
+    print("test_00 is =>", test_00)
     assert test_00 == 1
 
     test_01 : i32  = factorial_1(5,0)
-    print("test_01 is =>",test_01)
+    print("test_01 is =>", test_01)
     assert test_01 == 120
 
     test_02 : i32  = factorial_1(1,5555)
-    print("test_02 is =>",test_02)
+    print("test_02 is =>", test_02)
     assert test_02 == 5555
 
     test_03 : i32 =factorial_2(5,99999,99999)
-    print("test_03 is =>",test_03)
+    print("test_03 is =>", test_03)
     assert test_03 == 120
 
     test_04 : i32  = factorial_2(4,-1,100)
-    print("test_04 is =>",test_04)
+    print("test_04 is =>", test_04)
     assert test_04 == -400
 
     test_05 :str = default_func()
-    print("test_05 is =>",test_05)
+    print("test_05 is =>", test_05)
     assert test_05 == "Hello World"
 
     test_06 :str = default_func(y = "|||",x="Hi")
-    print("test_06 is =>",test_06)
+    print("test_06 is =>", test_06)
     assert test_06 == "Hi|||World"
 
-
     test_07 : str = even_positions(0)
-    print("test_07 is =>",test_07)
+    print("test_07 is =>", test_07)
     assert test_07 == "?X?X?X?X?X"
 
     test_08 : str = even_positions(0,"W")
-    print("test_08 is =>",test_08)
+    print("test_08 is =>", test_08)
     assert test_08 == "WX?X?X?X?X"
 
     test_09 :str = default_func(y = "++",z = "LPython")
-    print("test_09 is =>",test_09)
+    print("test_09 is =>", test_09)
     assert test_09 == "Hello++LPython"
 
     test_10 :str = default_func("Welcome",z = "LPython")
-    print("test_10 is =>",test_10)
+    print("test_10 is =>", test_10)
     assert test_10 == "Welcome LPython"
-test_all()
-
-
-
-
-
-
-    
+test_all() 
--- a/integration_tests/def_func_01.py
+++ b/integration_tests/def_func_01.py
@@ -14,17 +14,21 @@ def default_func(x : str ="Hello" ,y : str = " ", z : str = "World") ->str:
     return x + y + z
 
 
-def even_positons(iterator : i32 , to_add : str = "?")-> str:
+def even_positions(iterator : i32 , to_add : str = "?")-> str:
     if (iterator == 10): return ""
 
     if iterator%2 == 0 :
-        return to_add + even_positons(iterator+1,"X")
+        return to_add + even_positions(iterator+1,"X")
 
-    return to_add +even_positons(iterator+1)
+    return to_add +even_positions(iterator+1)
 
 
 
 def test_all():
+
+    test_00 : i32  = factorial_1(1)
+    print("test_00 is =>",test_00)
+    assert test_00 == 1
 
     test_01 : i32  = factorial_1(5,0)
     print("test_01 is =>",test_01)
@@ -51,11 +55,11 @@ def test_all():
     assert test_06 == "Hi|||World"
 
 
-    test_07 : str = even_positons(0)
+    test_07 : str = even_positions(0)
     print("test_07 is =>",test_07)
     assert test_07 == "?X?X?X?X?X"
 
-    test_08 : str = even_positons(0,"W")
+    test_08 : str = even_positions(0,"W")
     print("test_08 is =>",test_08)
     assert test_08 == "WX?X?X?X?X"
 
--- a/integration_tests/def_func_01.py
+++ b/integration_tests/def_func_01.py
@@ -50,6 +50,7 @@ def test_all():
     print("test_06 is =>",test_06)
     assert test_06 == "Hi|||World"
 
+
     test_07 : str = even_positons(0)
     print("test_07 is =>",test_07)
     assert test_07 == "?X?X?X?X?X"
@@ -58,6 +59,13 @@ def test_all():
     print("test_08 is =>",test_08)
     assert test_08 == "WX?X?X?X?X"
 
+    test_09 :str = default_func(y = "++",z = "LPython")
+    print("test_09 is =>",test_09)
+    assert test_09 == "Hello++LPython"
+
+    test_10 :str = default_func("Welcome",z = "LPython")
+    print("test_10 is =>",test_10)
+    assert test_10 == "Welcome LPython"
 test_all()
 
 
--- a/src/libasr/codegen/asr_to_c_cpp.h
+++ b/src/libasr/codegen/asr_to_c_cpp.h
@@ -565,6 +565,7 @@ R"(#include <stdio.h>
                 if( is_c ) {
                     CDeclarationOptions c_decl_options;
                     c_decl_options.pre_initialise_derived_type = false;
+                    c_decl_options.do_not_initialize = true;
                     func += self().convert_variable_decl(*arg, &c_decl_options);
                 } else {
                     CPPDeclarationOptions cpp_decl_options;
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -5167,39 +5167,6 @@ public:
         ASR::expr_t* assign_asr_target_copy = assign_asr_target;
         this->visit_expr(*x.m_targets[0]);
         assign_asr_target = ASRUtils::EXPR(tmp);
-        //Construction-While-Assigning Problem
-        AST::exprType value_type_ast = (x.m_value)->type;
-        AST::exprType objects_to_check[3] { AST::exprType::Set,  // just check these 3 for now.
-        									AST::exprType::Dict,
-        									AST::exprType::List };  
-        for(AST::exprType &exp : objects_to_check){
-            if (exp == value_type_ast){
-                ASR::ttype_t *target_type = ASRUtils::expr_type(assign_asr_target);
-                try{
-                    this->visit_expr(*x.m_value);
-                }
-                catch (SemanticError &error){
-                    ASR::expr_t *value_ASR_after_down_cast = ASRUtils::EXPR(tmp);
-                    ASR::ttype_t *value_type = ASRUtils::expr_type(value_ASR_after_down_cast);
-                    if (!ASRUtils::check_equal_type(target_type, value_type)){
-                        std::string ltype = ASRUtils::type_to_str_python(target_type);
-                        std::string rtype = ASRUtils::type_to_str_python(value_type);
-                        diag.add(diag::Diagnostic(
-                                "Type mismatch in assignment, the types must be compatible",
-                                diag::Level::Error, diag::Stage::Semantic, {
-                                        diag::Label("type mismatch ('" + ltype + "' and '" + rtype + "')",
-                                                    {assign_asr_target->base.loc, value_ASR_after_down_cast->base.loc})
-                                })
-                        );
-                        throw SemanticAbort();
-                    }
-                }
-            }
-
-        }
-
-
-
         this->visit_expr(*x.m_value);
         assign_asr_target = assign_asr_target_copy;
         if (tmp) {
@@ -6671,9 +6638,6 @@ public:
                 }
             } else {
                 if (!ASRUtils::check_equal_type(ASRUtils::expr_type(value), type)) {
-                	//set the tmp to use it in the error message.(copied from the end of this function)              	
-                    ASR::ttype_t* set_type = ASRUtils::TYPE(ASR::make_Set_t(al, x.base.base.loc, type));
-                    tmp = ASR::make_SetConstant_t(al, x.base.base.loc, elements.p, elements.size(), set_type);
                     throw SemanticError("All Set values must be of the same type for now",
                                         x.base.base.loc);
                 }
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -659,7 +659,7 @@ public:
 
         // Fill the whole call_args_vec with nullptr
         // This is for error handling later on.
-         for( size_t i = 0; i < call_args_vec.max; i++ ) {
+         for( size_t i = 0; i < orig_func->n_args; i++ ) {
             ASR::call_arg_t call_arg;
             Location loc;
             loc.first = loc.last = 1;
@@ -707,17 +707,26 @@ public:
         // Filling missing arguments with their defaults passed in function definition (if present).
         for(size_t i = 0; i < orig_func->n_args; i++ ){
             if(call_args_vec.p[i].m_value == nullptr){
-                ASR::symbol_t* sym = ASR::down_cast<ASR::Var_t>(orig_func->m_args[i])->m_v;
-                std::string variable_name = ASR::down_cast<ASR::Variable_t>(sym)->m_name;
-                ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(orig_func->m_symtab->get_symbol(variable_name));
+                ASR::Variable_t* var = ASRUtils::EXPR2VAR(orig_func->m_args[i]);
                 if (var->m_symbolic_value == nullptr){
-                    call_args_vec.reserve(al, n_pos_args + n_kwargs); //deprecate Vec container, so it raise error later.
-                    break;
+                    missed_args_names+="'" + (std::string) var->m_name + "' and ";
+                    missed_args_count++;
                 }
                 else{
                     call_args_vec.p[i].m_value = var->m_symbolic_value;
                 }
             }
+        }
+        if(missed_args_count > 0){
+            missed_args_names = missed_args_names.substr(0,missed_args_names.length() - 5);
+            diag.add(diag::Diagnostic(
+                "Number of arguments does not match in the function call",
+                diag::Level::Error, diag::Stage::Semantic, {
+                    diag::Label("missing " + std::to_string(missed_args_count) + " required arguments :" + missed_args_names,
+                        {call_loc})
+                })
+            );
+            throw SemanticAbort();
         }
         return true;
     }
@@ -1158,17 +1167,32 @@ public:
                 visit_expr_list(pos_args, n_pos_args, kwargs, n_kwargs,
                                 args, rt_subs, func, loc);
             } else if (args.size() < func->n_args) {
+                std::string missed_args_names =" ";
+                size_t missed_args_count =0;
                 for (size_t def_arg = args.size(); def_arg < func->n_args; def_arg++){
-                    ASR::symbol_t* sym = ASR::down_cast<ASR::Var_t>(func->m_args[def_arg])->m_v;
-                    std::string variable_name = ASR::down_cast<ASR::Variable_t>(sym)->m_name;
-                    ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(func->m_symtab->get_symbol(variable_name));
+                    ASR::Variable_t* var = ASRUtils::EXPR2VAR(func->m_args[def_arg]);
                     if(var->m_symbolic_value == nullptr) {
-                        break;
+                        missed_args_names+= "'" + (std::string)var->m_name + "' and ";
+                        missed_args_count++;
                     }
-                    ASR::call_arg_t call_arg;
-                    call_arg.m_value = var->m_symbolic_value;
-                    call_arg.loc = (var->m_symbolic_value->base).loc;
-                    args.push_back(al,call_arg);       
+                    else{
+
+                        ASR::call_arg_t call_arg;
+                        call_arg.m_value = var->m_symbolic_value;
+                        call_arg.loc = (var->m_symbolic_value->base).loc;
+                        args.push_back(al,call_arg);       
+                    }
+                }
+                if(missed_args_count > 0){
+                    missed_args_names = missed_args_names.substr(0,missed_args_names.length() - 5);
+                    diag.add(diag::Diagnostic(
+                    "Number of arguments does not match in the function call",
+                    diag::Level::Error, diag::Stage::Semantic, {
+                        diag::Label("missing " + std::to_string(missed_args_count) + " required arguments :" + missed_args_names,
+                                {loc})
+                        })
+                    );
+                    throw SemanticAbort();
                 }
             }
             if (ASRUtils::get_FunctionType(func)->m_is_restriction) {
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -5204,6 +5204,38 @@ public:
         ASR::expr_t* assign_asr_target_copy = assign_asr_target;
         this->visit_expr(*x.m_targets[0]);
         assign_asr_target = ASRUtils::EXPR(tmp);
+        //Construction-While-Assigning Problem
+        AST::exprType value_type_ast = (x.m_value)->type;
+        AST::exprType objects_to_check[4] {AST::exprType::Set,AST::exprType::Dict, // just check these 4 for now.
+                                           AST::exprType::Tuple,AST::exprType::List};
+        for(AST::exprType &exp : objects_to_check){
+            if (exp == value_type_ast){
+                ASR::ttype_t *target_type = ASRUtils::expr_type(assign_asr_target);
+                try{
+                    this->visit_expr(*x.m_value);
+                }
+                catch (SemanticError &error){
+                    ASR::expr_t *value_ASR_after_down_cast = ASRUtils::EXPR(tmp);
+                    ASR::ttype_t *value_type = ASRUtils::expr_type(value_ASR_after_down_cast);
+                    if (!ASRUtils::check_equal_type(target_type, value_type)){
+                        std::string ltype = ASRUtils::type_to_str_python(target_type);
+                        std::string rtype = ASRUtils::type_to_str_python(value_type);
+                        diag.add(diag::Diagnostic(
+                                "Type mismatch in assignment, the types must be compatible",
+                                diag::Level::Error, diag::Stage::Semantic, {
+                                        diag::Label("type mismatch ('" + ltype + "' and '" + rtype + "')",
+                                                    {assign_asr_target->base.loc, value_ASR_after_down_cast->base.loc})
+                                })
+                        );
+                        throw SemanticAbort();
+                    }
+                }
+            }
+
+        }
+
+
+
         this->visit_expr(*x.m_value);
         assign_asr_target = assign_asr_target_copy;
         if (tmp) {
@@ -6666,6 +6698,9 @@ public:
                 }
             } else {
                 if (!ASRUtils::check_equal_type(ASRUtils::expr_type(value), type)) {
+                    //set the tmp to use it in the error message.
+                    ASR::ttype_t* set_type = ASRUtils::TYPE(ASR::make_Set_t(al, x.base.base.loc, type));
+                    tmp = ASR::make_SetConstant_t(al, x.base.base.loc, elements.p, elements.size(), set_type);
                     throw SemanticError("All Set values must be of the same type for now",
                                         x.base.base.loc);
                 }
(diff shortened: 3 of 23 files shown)


assem2002:
checks result from setting default argument to the function parameters. I don't see how to handle such a thing.

assem2002:
I think it's due to testing a function with default parameters in the wrong spot like the issue i mentioned above.
I will try to handle the AST to throw a syntax error when such a thing happens.

assem2002:
The errors arises from converting the lpython code into "c code" as c doesn't provide default arguments.
Should i change the process of converting lpython-> c , so it doesn't make a default arguments assignment?

assem2002:
If the solution seems to be doing the job correctly. I would start handling the syntax error to make the solution complete without any flaws.
I just want to make sure it's the right approach to handle such a problem.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1141,6 +1141,19 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
                 args.reserve(al, n_pos_args + n_kwargs);
                 visit_expr_list(pos_args, n_pos_args, kwargs, n_kwargs,
                                 args, rt_subs, func, loc);
+            }
+        	if((n_pos_args+ n_kwargs) < func->n_args && (args.size() < func->n_args) ){
```
Shaikh-Ubaid, line 1145:
```suggestion
        if ((n_pos_args + n_kwargs) < func->n_args && (args.size() < func->n_args) ){
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1141,6 +1141,19 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                for(size_t def_arg = (n_pos_args+ n_kwargs) ; def_arg <(func->n_args) ;def_arg++){
```
Shaikh-Ubaid, line 1146:
```suggestion
                for (size_t def_arg = (n_pos_args+ n_kwargs); def_arg < func->n_args; def_arg++){
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1141,6 +1141,19 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                    ASR::symbol_t* sym = (ASR::down_cast<ASR::Var_t>((func->m_args)[def_arg]))->m_v;
```
Shaikh-Ubaid, line 1147:
```suggestion
                    ASR::symbol_t* sym = ASR::down_cast<ASR::Var_t>(func->m_args[def_arg])->m_v;
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1141,6 +1141,19 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                    ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(sym);
+                    if(!(var->m_value)) {
```
Shaikh-Ubaid, line 1149:
```suggestion
                    if (var->m_value == nullptr) {
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4312,6 +4332,9 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
             }
             ASR::accessType s_access = ASR::accessType::Public;
             ASR::presenceType s_presence = ASR::presenceType::Required;
+        	if (i >= default_arg_index_start){
```
Shaikh-Ubaid, line 4335:
```suggestion
        if (i >= default_arg_index_start){                
```

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: Also, I think `n_pos_args + n_kwargs` and `args.size()` will be same. So checking for `args.size() < func->n_args` seems redundant.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -562,6 +562,7 @@ R"(#include <stdio.h>
                 if( is_c ) {
                     CDeclarationOptions c_decl_options;
                     c_decl_options.pre_initialise_derived_type = false;
+                    c_decl_options.do_not_initialize = true;
```
Shaikh-Ubaid, line 565: Why do we need this change?

Shaikh-Ubaid (commented):
Shared couple comments. Rest looks good. Let's add some integration tests and check if the implementation works as expected.

src/libasr/codegen/asr_to_c_cpp.h
Shaikh-Ubaid: I just read your commit message. I think its fine.

Shaikh-Ubaid:
> If the solution seems to be doing the job correctly. I would start handling the syntax error to make the solution complete without any flaws. I just want to make sure it's the right approach to handle such a problem.

You can start another PR to support the semantic error. I suspect that the implementation might require updates in the parser.

Shaikh-Ubaid:

src/lpython/semantics/python_ast_to_asr.cpp
assem2002: I wrote this part in the first time as you said, but built-in function  `complex()` made a huge problem.
The function seems to be making a call to some function without passing any positional arguments to the call ,although the call has `Vec<ASR::call_arg_t>`  parameter filled with arguments to be placed in the called function.
# Brief 
- Sometimes the call is depending on `n_pos_args + n_kwargs` as you mentioned and that should be enough, but sometimes it depends on `args.size()` .
- `n_pos_args + n_kwargs` evaluate to => 0 .... while 'args.size()' evaluate to => (**larger than 0**).

src/lpython/semantics/python_ast_to_asr.cpp
assem2002: I'll push the new reviewed code now, and I'll work on integration test to push soon.

assem2002:
> > If the solution seems to be doing the job correctly. I would start handling the syntax error to make the solution complete without any flaws. I just want to make sure it's the right approach to handle such a problem.
> 
> You can start another PR to support the semantic error. I suspect that the implementation might require updates in the parser.

I think we can handle such a thing on the parser level as you mentioned.
we can handle it by using a `default_argumets_started_ flag = false` to throw an error if ==> `default_argumets_started_ flag = true` and the current argument being built doesn't have a default argument.

Shaikh-Ubaid:
>we can handle it by using a default_argumets_started_ flag = false to throw an error if ==> default_argumets_started_ flag = true and the current argument being built doesn't have a default argument.

Go ahead and submit a PR with your approach and I will take a look.

Shaikh-Ubaid:
I suggest implementing the semantic error in a separate PR.

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: I see. I think in that case, we can directly use `if (args.size() < func->n_args)`. Can you try it and share if it works?

You also need to update the initialization in the below loop `for(size_t def_arg = args.size(); def_arg < func->n_args; def_arg++)`

assem2002: Yeah, you're right. Using `arg.size()` only would do the job.

assem2002:
> I suggest implementing the semantic error in a separate PR.

I'll do after handling this PR

> I suggest implementing the semantic error in a separate PR.

I'll right after completing this PR

assem2002:
@Shaikh-Ubaid 

I faced a test case which I can't say whether it's a valid test case or not
the following:
```python

def default_func(x : str ="Hello" ,y : str = " ", z : str = "World") ->str:
    return x + y + z

default_func(z="lpython")
```

It results in the following error:
```shell
semantic error: default_func() got multiple values for argument 'z'
  --> ../../integration_tests/def_func_01.py:49:20
   |
49 |     test_06 :str = default_func(z="lpython")
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^ 

```

should I handle this in the parsing phase to prevent user form passing keyword arguments for the last parameters, instead the used can pass keyword arguments only if it fill some sequence in the beginning of the parameter list like that:
```python
default_func(y = "///",x="Hi")
```

Shaikh-Ubaid:
>I faced a test case which I can't say whether it's a valid test case or not

If the test case runs in cpython then its a valid test case.

assem2002:
> > I faced a test case which I can't say whether it's a valid test case or not
> 
> If the test case runs in cpython then its a valid test case.

Actually it passes this testcase. I will try to find a way to handle it.

assem2002:
@Shaikh-Ubaid 
I handled the tests case
```python
def default_func(x : str ="Hello" ,y : str = " ", z : str = "World") ->str:
    return x + y + z

    test_09 :str = default_func(y = "++",z = "LPython")
    print("test_09 is =>",test_09)
    assert test_09 == "Hello++LPython"

    test_10 :str = default_func("Welcome",z = "LPython")
    print("test_10 is =>",test_10)
    assert test_10 == "Welcome LPython"
```

# Solution Flow
- If only positional arguments are passed and aren't matching `func->n_args`,  we start from the last argument in function_t and push back in `Vec<args>`, if there's no value to push then break.
- If there's keyword arguments passed, We reserve `Vec<args>` with `func->n_args` and then `visit_expr_list()` starts to place set all arguments to `nullptr`, then pushes back positional arguments, then sets keyword arguments in its desired index in `Vec<args>` , then we loop on` Vec<args> ` to see if there's a `nullptr` arg that we can have a matching default argument for.

certik:
Thanks!

To implement default arguments, the following must be done:

* At the function definition time, find the default argument and put it into the argument's Variable `symbolic_value` into the symbol table
* At the function call site, simply lookup any default arguments in ASR, and if present, use them by explicitly inserting them

The alternative design is to not even insert them and sort it out inside the function, but I would not do this, since it is not explicit. I would always insert them at the call site.

kmr-srbh:
@assem2002 could you please test for this case?
```python
from lpython import i32

def fn(defarg1: str="default", arg: i32, defarg2: str="default") -> i32:
    return 1


fn()
```

This should throw an error.

assem2002:
> @assem2002 could you please test for this case?
> 
> ```python
> from lpython import i32
> 
> def fn(defarg1: str="default", arg: i32, defarg2: str="default") -> i32:
>     return 1
> 
> 
> fn()
> ```
> 
> This should throw an error.

Yeah it should, but unfortunately it doesn't 
I talked with **Shaikh Ubaid**  about this, We agreed to push another pull request to handle this problem on the parsing stage not in the ASR stage.

assem2002:
> * symbol table

Does that mean the AST of a call to a function with default arguments will include the default arguments passed even if they weren't provided by user? If so, isn't the whole AST is built then ASR?

Or what you mean is to edit the call to 'make_call_helper()'  to include the default arguments?

assem2002:
> Thanks!
> 
> To implement default arguments, the following must be done:
> 
>     * At the function definition time, find the default argument and put it into the argument's Variable `symbolic_value` into the symbol table
> 
>     * At the function call site, simply lookup any default arguments in ASR, and if present, use them by explicitly inserting them
> 
> 
> The alternative design is to not even insert them and sort it out inside the function, but I would not do this, since it is not explicit. I would always insert them at the call site.



Does that mean the AST of a call to a function with default arguments will include the default arguments passed even if they weren't provided by user? If so, isn't the whole AST is built then ASR?

Or what you mean is to edit the call to 'make_call_helper()' to include the default arguments?

Shaikh-Ubaid:
The approach that Ondrej shared comprises of the following two steps:

### Step 1

For default function arguments, their default value should be inserted in the functions symbol table inside the Variable declaration. Precisely, the default value should be inserted in the variable's `symbolic_value` and `value` in its declaration inside the symbol table of each function.

### Step 2

For each function call, whenever we need to use the default argument value we can simply lookup the symbol table of the called function, access the `symbolic_value` (or `value`) for the variable whose default value we need and insert this default value in the function call.

Please let me know if this makes it clear.

assem2002:
> The approach that Ondrej shared comprises of the following two steps:
> ### Step 1
> 
> For default function arguments, their default value should be inserted in the functions symbol table inside the Variable declaration. Precisely, the default value should be inserted in the variable's `symbolic_value` and `value` in its declaration inside the symbol table of each function.
> ### Step 2
> 
> For each function call, whenever we need to use the default argument value we can simply lookup the symbol table of the called function, access the `symbolic_value` (or `value`) for the variable whose default value we need and insert this default value in the function call.
> 
> Please let me know if this makes it clear.

I got the first step.
does the second step should be similar to that? :
```c++
            (for size_t i = index_start_def_arguments; i < func-> n_args; i++){
                ASR::var_t* var  = ASR::down_cast<ASR::Var_t>(func->m_args[i]);
                std::string symbol_name = ASRUtils::symbol_name(arg_Var->m_v);
                ASR::symbol_t* sym = func->m_symtab->getsymbol(symbol_name);
                //then accessing the value, and pushing in call_args vector.
            }
```

Shaikh-Ubaid:
>does the second step should be similar to that?

Seems in the right direction. Also, note that you can simply get the `symbol` from `Var_t` as follows:
```console
ASR::Var_t* var  = ASR::down_cast<ASR::Var_t>(func->m_args[i]);
ASR::symbol_t* sym = var->m_v;
```

assem2002:
> Seems in the right direction. Also, note that you can simply get the `symbol` from `Var_t` as follows:

actually I'm confused at the moment. My PR handles the issue of passing the default arguments by sorting the arguments in `Vec<call_args>`,
```C++
  for (size_t def_arg = args.size(); def_arg < func->n_args; def_arg++){
      ASR::symbol_t* sym = ASR::down_cast<ASR::Var_t>(func->m_args[def_arg])->m_v;
      ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(sym);
      if (var->m_value == nullptr) {
          break;
      }
      ASR::call_arg_t call_arg;
      args.push_back(al,call_arg);
      args.p[def_arg].loc = ((var->m_value)->base).loc;
      args.p[def_arg].m_value = (var->m_value);
```

 which Ondrej thinks isn't explicit, so he is proposing to pushing arguments by looking up them in the symbol Table and get the value in symbolic value if exist.This approach could be done like that:

```C++
(for size_t i = index_start_def_arguments; i < func-> n_args; i++){
    ASR::var_t* var  = ASR::down_cast<ASR::Var_t>(func->m_args[i]);
    std::string symbol_name = ASRUtils::symbol_name(arg_Var->m_v);
    ASR::symbol_t* sym = func->m_symtab->getsymbol(symbol_name);
    //then accessing the value, and pushing in call_args vector.
}
```
which seems better i can just look up the arguments by their name instead of manipulating the logic of the arguments positions passed the `make_call_helper( )`. Using previous approach is really more explicit to the reader of the code and it would handle the issue of [keyword arguments](https://github.com/lcompilers/lpython/pull/2618#issuecomment-2022117366) that appeared in some test cases in more explicit way.

I hope I got what you mean. If I completely lost track of what you want to implement, I hope you can guide me through an article or something.

Shaikh-Ubaid:
>which Ondrej thinks isn't explicit, so he is proposing the approach of pushing arguments by looking up their values in the symbol Table and get the value in symbolic value if exist

As of now, I think the symbolic value for a functions default argument is not inserted. That is, the symbolic value for the default argument is currently `null` at the ASR level. Even if the user has provided the default value in the AST, we just skip processing it while constructing the ASR.

As a part of this PR, the first step should be to process the default values of function arguments (if any) provided in the AST and ensure that these get appropriately reflected in the `symbolic_value` of the function argument declaration in the function's symbol table.

Please let me know if this helps.

assem2002:
> processing it while constructing the ASR.

I read about interpreter design in the classic SICP, and I see what you meant to say. Though the PR's current solution is working fine, but still not the convenient  way to do such a thing.
I think the PR is close to the desired solution,I'll push some edits to meet these requirements.

assem2002:
@Shaikh-Ubaid @certik 
Now I fetch the value of the default arguments by looking up the symbol table of the scope of the function that has been called.
Let me know if that's the right approach to do so or it needs to get changed.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -704,6 +704,21 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
             call_args_vec.p[arg_pos].m_value = expr;
         }
+        //Filling missing arguments with its default argument passed in function definition (if existed).
+        for(size_t i = 0; i < orig_func->n_args; i++ ){
+            if(call_args_vec.p[i].m_value == nullptr){
+                ASR::symbol_t* sym = ASR::down_cast<ASR::Var_t>(orig_func->m_args[i])->m_v;
+                std::string variable_name = ASR::down_cast<ASR::Variable_t>(sym)->m_name;
+                ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(orig_func->m_symtab->get_symbol(variable_name));
+                if (var->m_symbolic_value == nullptr){
+                    call_args_vec.reserve(al, n_pos_args + n_kwargs); //deprecate Vec container, so it raise error later.
```
certik, line 714: Something is not right here --- why do you need to call `reserve` here?

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1139,9 +1154,23 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
         if (ASR::is_a<ASR::Function_t>(*s)) {
             ASR::Function_t *func = ASR::down_cast<ASR::Function_t>(s);
             if( n_kwargs > 0 && !is_generic_procedure ) {
-                args.reserve(al, n_pos_args + n_kwargs);
+                args.reserve(al, func->n_args);
                 visit_expr_list(pos_args, n_pos_args, kwargs, n_kwargs,
                                 args, rt_subs, func, loc);
+            }
+        	else if(args.size() < func->n_args){
```
certik, line 1161:
```suggestion
            } else if (args.size() < func->n_args) {
```

integration_tests/def_func_01.py
```diff
@@ -0,0 +1,76 @@
...
+
+def test_all():
+
+    test_01 : i32  = factorial_1(5,0)
+    print("test_01 is =>",test_01)
+    assert test_01 == 120
+
+    test_02 : i32  = factorial_1(1,5555)
+    print("test_02 is =>",test_02)
+    assert test_02 == 5555
```
certik, line 35: Add a test for not passing in the second argument and ensure it returns the default value "1".

integration_tests/def_func_01.py
```diff
@@ -0,0 +1,76 @@
...
+def factorial_2(x: i32, y:i32=3 ,z:i32 =2) ->i32:
+    if x ==4:
+        return x * y * z
+    return x * factorial_2(x-1)
+
+def default_func(x : str ="Hello" ,y : str = " ", z : str = "World") ->str:
+    return x + y + z
+
+
+def even_positons(iterator : i32 , to_add : str = "?")-> str:
```
certik, line 17: Did you mean:
```suggestion
def even_positions(iterator : i32 , to_add : str = "?")-> str:
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -704,6 +704,21 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
             call_args_vec.p[arg_pos].loc = expr->base.loc;
             call_args_vec.p[arg_pos].m_value = expr;
         }
+        //Filling missing arguments with its default argument passed in function definition (if existed).
```
certik, line 707:
```suggestion
        // Filling missing arguments with their defaults passed in function definition (if present).
```

certik (commented):
I think this is the right approach. Where is the `m_symbolic_value` being set in the first place? I couldn't see it in the PR. I left some minor comments.

Can you please also test this by having the function with default arguments in a separate module?

@Shaikh-Ubaid can you please also review it?

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -659,7 +659,7 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
 
         // Fill the whole call_args_vec with nullptr
         // This is for error handling later on.
-        for( size_t i = 0; i < n_pos_args + n_kwargs; i++ ) {
+         for( size_t i = 0; i < call_args_vec.max; i++ ) {
```
Shaikh-Ubaid, line 662: @assem2002 please, could you explain why we need this change?

integration_tests/def_func_01.py
```diff
@@ -0,0 +1,80 @@
...
+    test_10 :str = default_func("Welcome",z = "LPython")
+    print("test_10 is =>",test_10)
+    assert test_10 == "Welcome LPython"
+test_all()
+
+
+
+
+
+
```
Shaikh-Ubaid, line 79: Please remove these stray blank lines at the end of the file.

integration_tests/def_func_01.py
```diff
@@ -0,0 +1,80 @@
...
+    assert test_08 == "WX?X?X?X?X"
+
+    test_09 :str = default_func(y = "++",z = "LPython")
+    print("test_09 is =>",test_09)
+    assert test_09 == "Hello++LPython"
+
+    test_10 :str = default_func("Welcome",z = "LPython")
+    print("test_10 is =>",test_10)
+    assert test_10 == "Welcome LPython"
+test_all()
```
Shaikh-Ubaid, line 73: In this test, @assem2002 do you mind using appropriate and consistent spacing for example `my_func(x: i32 = 10, y: str = "abc")` and `print("test_09 is =>", test_09)` .

integration_tests/def_func_01.py
```diff
@@ -0,0 +1,80 @@
...
+def even_positions(iterator : i32 , to_add : str = "?")-> str:
+    if (iterator == 10): return ""
+
+    if iterator%2 == 0 :
+        return to_add + even_positions(iterator+1,"X")
+
+    return to_add +even_positions(iterator+1)
+
+
+
```
Shaikh-Ubaid, line 26:
```suggestion
```

Shaikh-Ubaid (commented):
With respect to the previous implementation, I couldn't find any/much difference. Please, could you highlight what new changes were added?

Shaikh-Ubaid:
>Where is the m_symbolic_value being set in the first place? I couldn't see it in the PR. I left some minor comments.

I think it is not being set. @assem2002 as we shared before, the `m_symbolic_value` should be first set (it should be visible for the default parameters of the function when we do `--show-asr`). Then, on function call when sufficient arguments are not provided, we simply look up the `m_symbolic_value` for the missing arguments and insert it in the function call.

assem2002:
> > Where is the m_symbolic_value being set in the first place? I couldn't see it in the PR. I left some minor comments.
> 
> I think it is not being set. @assem2002 as we shared before, the `m_symbolic_value` should be first set (it should be visible for the default parameters of the function when we do `--show-asr`). Then, on function call when sufficient arguments are not provided, we simply look up the `m_symbolic_value` for the missing arguments and insert it in the function call.

It does appear in the ASR

the symbolic_value of the variable of the function parameters is set in the following line of code 
```cpp
  if (i >= default_arg_index_start){
      size_t default_arg_index = i - default_arg_index_start;
      this->visit_expr(*(x.m_args.m_defaults[default_arg_index]));
      init_expr = ASRUtils::EXPR(tmp);  // <<<<<<<<<<<<<<<<<<<<<<<<<HERE
```
`init_expr` is the value that is going to be passed to `symbolic_value` while building the variable in `make_variable( )`

assem2002:
> > > Where is the m_symbolic_value being set in the first place? I couldn't see it in the PR. I left some minor comments.
> > 
> > 
> > I think it is not being set. @assem2002 as we shared before, the `m_symbolic_value` should be first set (it should be visible for the default parameters of the function when we do `--show-asr`). Then, on function call when sufficient arguments are not provided, we simply look up the `m_symbolic_value` for the missing arguments and insert it in the function call.
> 
> It does appear in the ASR
> 
> ```python
> def default_func(x : str ="Hello", y : str = " ", z : str = "World") ->str:
>     return x + y + z
> ```
> 
> Screenshot from 2024-04-30 02-10-35
> 
> the symbolic_value of the variable of the function parameters is set in the following line of code
> 
> ```c++
>   if (i >= default_arg_index_start){
>       size_t default_arg_index = i - default_arg_index_start;
>       this->visit_expr(*(x.m_args.m_defaults[default_arg_index]));
>       init_expr = ASRUtils::EXPR(tmp);  // <<<<<<<<<<<<<<<<<<<<<<<<<HERE
> ```
> 
> `init_expr` is the value that is going to be passed to `symbolic_value` while building the variable in `make_variable( )`
> 
> and yeah i know this PR isn't much different, but while processing a function call, instead of fetching the values directly we search the scope for the name of the arguments. I think it's a matter of clarity in the code not more than that.

@Shaikh-Ubaid

certik:
Indeed, it seems it is already set in ASR.

assem2002:
> Indeed, it seems it is already set in ASR.

In the current version on lpython repo, The ASR of any function doesn't know any information about the default arguments as it never get passed to the `function_t` object while building the ASR

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1139,9 +1154,22 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
                 visit_expr_list(pos_args, n_pos_args, kwargs, n_kwargs,
                                 args, rt_subs, func, loc);
+            } else if (args.size() < func->n_args) {
+                for (size_t def_arg = args.size(); def_arg < func->n_args; def_arg++){
+                    ASR::symbol_t* sym = ASR::down_cast<ASR::Var_t>(func->m_args[def_arg])->m_v;
+                    std::string variable_name = ASR::down_cast<ASR::Variable_t>(sym)->m_name;
+                    ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(func->m_symtab->get_symbol(variable_name));
+                    if(var->m_symbolic_value == nullptr) {
+                        break;
+                    }
```
Shaikh-Ubaid, line 1167: Why do we `break` here? Is this a case where `m_symbolic_value` is expected/needed but is not available? If so, then I think we should throw a `SemanticError()` instead of the `break`.

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: I think I get it. I think we can better use `orig_func->n_args` here.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -704,6 +704,21 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
             call_args_vec.p[arg_pos].loc = expr->base.loc;
             call_args_vec.p[arg_pos].m_value = expr;
         }
+        //Filling missing arguments with its default argument passed in function definition (if existed).
+        for(size_t i = 0; i < orig_func->n_args; i++ ){
+            if(call_args_vec.p[i].m_value == nullptr){
+                ASR::symbol_t* sym = ASR::down_cast<ASR::Var_t>(orig_func->m_args[i])->m_v;
+                std::string variable_name = ASR::down_cast<ASR::Variable_t>(sym)->m_name;
+                ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(orig_func->m_symtab->get_symbol(variable_name));
```
Shaikh-Ubaid, line 712:
```suggestion
                ASR::Variable_t* var = ASRUtils::EXPR2VAR(orig_func->m_args[i]);
```

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: Right. I think we should better throw a `SemanticError` when a default value is expected/needed, but is `nullptr`.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1139,9 +1154,22 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
             if( n_kwargs > 0 && !is_generic_procedure ) {
-                args.reserve(al, n_pos_args + n_kwargs);
+                args.reserve(al, func->n_args);
                 visit_expr_list(pos_args, n_pos_args, kwargs, n_kwargs,
                                 args, rt_subs, func, loc);
+            } else if (args.size() < func->n_args) {
+                for (size_t def_arg = args.size(); def_arg < func->n_args; def_arg++){
+                    ASR::symbol_t* sym = ASR::down_cast<ASR::Var_t>(func->m_args[def_arg])->m_v;
+                    std::string variable_name = ASR::down_cast<ASR::Variable_t>(sym)->m_name;
+                    ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(func->m_symtab->get_symbol(variable_name));
```
Shaikh-Ubaid, line 1164:
```suggestion
                    ASR::Variable_t* var = ASRUtils::EXPR2VAR(func->m_args[def_arg]);
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1139,9 +1154,22 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
         if (ASR::is_a<ASR::Function_t>(*s)) {
             ASR::Function_t *func = ASR::down_cast<ASR::Function_t>(s);
             if( n_kwargs > 0 && !is_generic_procedure ) {
-                args.reserve(al, n_pos_args + n_kwargs);
+                args.reserve(al, func->n_args);
                 visit_expr_list(pos_args, n_pos_args, kwargs, n_kwargs,
                                 args, rt_subs, func, loc);
+            } else if (args.size() < func->n_args) {
```
Shaikh-Ubaid, line 1160: After going through the rest of the changes, it seems that this `else if` block might not be required. @assem2002 Can you try removing this `else-if` block and see if things still work?

Shaikh-Ubaid (commented):
Shared couple comments. Seems like in a good direction.

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: Second thought: It might be needed. It looks like a better place for this can be inside the function `void visit_expr_list(AST::expr_t** exprs, size_t n, Vec<ASR::call_arg_t>& call_args_vec)`, but we can do that later. For now the current place is fine for me.

Shaikh-Ubaid:

assem2002:
> Please mark as "Ready for review" when ready.

@Shaikh-Ubaid 
I need your opinion on this [link](https://github.com/lcompilers/lpython/pull/2618#discussion_r1585139929) and this [link](https://github.com/lcompilers/lpython/pull/2618#discussion_r1585109447) .
to handle your suggestions on the code properly.

Shaikh-Ubaid:
https://github.com/lcompilers/lpython/pull/2618#discussion_r1585139929 is a simple code improvement suggestion. Please try it and share if it works.

For https://github.com/lcompilers/lpython/pull/2618#discussion_r1585109447, I am unsure why we have used `break`.

assem2002:
> [#2618 (comment)](https://github.com/lcompilers/lpython/pull/2618#discussion_r1585139929) is a simple code improvement suggestion. Please try it and share if it works.
> 
> For [#2618 (comment)](https://github.com/lcompilers/lpython/pull/2618#discussion_r1585109447), I am unsure why we have used `break`.

For the break thing : I loop through the not-passed function parameters and try to look for their defaults (if exist), once i notice i have a missed parameter with no default for it, i break the loop as there's no need to proceed with the loop (**NOTICE** : default values are passed sequentially at the end of the parameters list).
............
About the [modifications](https://github.com/lcompilers/lpython/pull/2618#discussion_r1585139929) I'll test it and let you know about it.

Shaikh-Ubaid:

assem2002:
@Shaikh-Ubaid 
Okay, I'll write such an error message and let you know.
but why there wasn't such a message before?

Shaikh-Ubaid:
>but why there wasn't such a message before?

Previously we just checked for mismatch in the number of arguments. With the changes in this PR, I think we can now pinpoint to which particular argument is missing.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -704,6 +704,32 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+        std::string missed_args_names;
+        size_t missed_args_count =0;
+        for(size_t i = 0; i < orig_func->n_args; i++ ){
+            if(call_args_vec.p[i].m_value == nullptr){
+                ASR::Variable_t* var = ASRUtils::EXPR2VAR(orig_func->m_args[i]);
+                if (var->m_symbolic_value == nullptr){
+                    missed_args_names+="'" + (std::string) var->m_name + "' and ";
+                    missed_args_count++;
+                }
+                else{
```
Shaikh-Ubaid, line 717:
```suggestion
                    missed_args_names+="'" + std::string(var->m_name) + "' and ";
                    missed_args_count++;
                } else{
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1139,9 +1165,37 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                size_t missed_args_count =0;
+                for (size_t def_arg = args.size(); def_arg < func->n_args; def_arg++){
+                    ASR::Variable_t* var = ASRUtils::EXPR2VAR(func->m_args[def_arg]);
+                    if(var->m_symbolic_value == nullptr) {
+                        missed_args_names+= "'" + (std::string)var->m_name + "' and ";
+                        missed_args_count++;
+                    }
+                    else{
+
+                        ASR::call_arg_t call_arg;
```
Shaikh-Ubaid, line 1182:
```suggestion
                        missed_args_names+= "'" + std::string(var->m_name) + "' and ";
                        missed_args_count++;
                    } else{
                        ASR::call_arg_t call_arg;
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1139,9 +1165,37 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                        call_arg.m_value = var->m_symbolic_value;
+                        call_arg.loc = (var->m_symbolic_value->base).loc;
+                        args.push_back(al,call_arg);       
+                    }
+                }
+                if(missed_args_count > 0){
+                    missed_args_names = missed_args_names.substr(0,missed_args_names.length() - 5);
+                    diag.add(diag::Diagnostic(
+                    "Number of arguments does not match in the function call",
```
Shaikh-Ubaid, line 1191: @assem2002 Do you mind adding an error reference test for this?

Shaikh-Ubaid (commented):
Shared some minor changes. Rest looks good. I think the PR will be good to merge after these changes are handled.

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: @assem2002 it would be great if you could update the above to follow the existing style/formatting.

Shaikh-Ubaid (commented):
Shared a comment. The added test cases look great! Thanks for this!

kmr-srbh:
I think there might be some more changes required with the PR. The following must have failed, but executes normally:
```python
from lpython import i32

def fn(arg1: i32, arg2: i32, arg3: i32=1) -> i32:
    return arg1 + arg2 + arg3

print(fn(arg3=2, 1, 3))
```
```console
6
```

CPython throws the expected `SyntaxError` due to positional arguments following keyword arguments. I found the following error when thinking of some more testcases to add.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4290,6 +4343,11 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
             std::string arg_s = arg;
             ASR::expr_t *value = nullptr;
             ASR::expr_t *init_expr = nullptr;
+            if (i >= default_arg_index_start){
+                size_t default_arg_index = i - default_arg_index_start;
+                this->visit_expr(*(x.m_args.m_defaults[default_arg_index]));
```
kmr-srbh, line 4348:
```suggestion
            AST::expr_t *def_arg = nullptr;
            size_t default_arg_index;
            if (i >= default_arg_index_start){
                default_arg_index = i - default_arg_index_start;
                def_arg = x.m_args.m_defaults[default_arg_index];
                this->visit_expr(*def_arg);
```

We can be a little readable here.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -1139,9 +1165,35 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                        ASR::call_arg_t call_arg;
+                        call_arg.m_value = var->m_symbolic_value;
+                        call_arg.loc = (var->m_symbolic_value->base).loc;
+                        args.push_back(al,call_arg);       
+                    }
+                }
+                if(missed_args_count > 0){
+                    missed_args_names = missed_args_names.substr(0,missed_args_names.length() - 5);
+                    diag.add(diag::Diagnostic(
+                    "Number of arguments does not match in the function call",
```
kmr-srbh, line 1189:
```suggestion
                    "Number of arguments in the function call do not match the function definition",
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -704,6 +704,32 @@ class CommonVisitor : public AST::BaseVisitor<Struct> {
...
+                }
+                else{
+                    call_args_vec.p[i].m_value = var->m_symbolic_value;
+                }
+            }
+        }
+        if(missed_args_count > 0){
+            missed_args_names = missed_args_names.substr(0,missed_args_names.length() - 5);
+            diag.add(diag::Diagnostic(
+                "Number of arguments does not match in the function call",
```
kmr-srbh, line 725:
```suggestion
                "Number of arguments in the function call do not match the function definition",
```

kmr-srbh (changes requested):
I have suggested some style changes below. Please have a look.

assem2002:
> I think there might be some more changes required with the PR. The following must have failed, but executes normally:
> 
> ```python
> from lpython import i32
> 
> def fn(arg1: i32, arg2: i32, arg3: i32=1) -> i32:
>     return arg1 + arg2 + arg3
> 
> print(fn(arg3=2, 1, 3))
> ```
> 
> ```
> (lp) saurabh-kumar@Awadh:~/Projects/System/lpython$ ./src/bin/lpython ./examples/example.py
> 6
> ```
> 
> CPython throws the expected `SyntaxError` due to positional arguments following keyword arguments. I found the following error when thinking of some more testcases to add.
> 
> I think this PR is specifically handling the filling of default arguments. We can focus on this issue later if this PR is not the right place to handle it.

Yeah you're right. We suggest to handle this syntax error in another issue in addition to that problem too [link](https://github.com/lcompilers/lpython/pull/2618#issuecomment-2034517728)

Shaikh-Ubaid:
The PR seems to be getting longer. So let's merge this and work iteratively on the rest of the things (if any) in a new PR.

Shaikh-Ubaid (approved):
It looks great! Thanks for this! Amazing work! I appreciate it.

Issue #2595, opened 2024-03-10
AddressSanitizer Errors for `test_dict_nested1.py` and `print_list_tuple_03.py`: requested allocation size exceeds maximum supported size


Vipul-Cariappa:
Steps to reproduce:

```bash
./src/bin/lpython ./integration_tests/test_dict_nested1.py
clang -L./src/runtime/ -llpython_runtime -lm -g -fsanitize=address -fno-omit-frame-pointer ./test_dict_nested1.out.tmp.o
LD_PRELOAD=./src/runtime/liblpython_runtime.so ./a.out
```

AddressSanitizer Output:
```bash
==119540==ERROR: AddressSanitizer: requested allocation size 0xfffffffffafafaf8 (0xfffffffffafb0af8 after adjustments for alignment, red zones etc.) exceeds maximum supported size of 0x10000000000 (thread T0)
    #0 0x6140945de9be in malloc /home/runner/work/llvm-project/llvm-project/final/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:69:3
    #1 0x776f5992ea0c in _lfortran_malloc /home/vipul/Workspace/python/lpython/src/libasr/runtime/lfortran_intrinsics.c:1645:12
    #2 0x61409461cdfe in __module___main___test_nested_dict (/home/vipul/Workspace/python/lpython/a.out+0x106dfe)
    #3 0x61409461b435 in __module___main_____main__global_stmts (/home/vipul/Workspace/python/lpython/a.out+0x105435)

==119540==HINT: if you don't care about these errors you may set allocator_may_return_null=1
SUMMARY: AddressSanitizer: allocation-size-too-big /home/runner/work/llvm-project/llvm-project/final/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:69:3 in malloc
==119540==ABORTING
```

If I run it with `allocator_may_return_null=1`

```bash
❯ ASAN_OPTIONS="allocator_may_return_null=1" LD_PRELOAD=./src/runtime/liblpython_runtime.so ./a.out
==122213==WARNING: AddressSanitizer failed to allocate 0xfffffffffafafaf8 bytes
=================================================================
==122213==ERROR: AddressSanitizer: unknown-crash on address 0xbebebebebebebebe at pc 0x5a7abf10bad1 bp 0x7ffdca059750 sp 0x7ffdca058f08
READ of size 4210752248 at 0xbebebebebebebebe thread T0
[...]

PR #2626, opened 2024-03-25
Prevent calling `abort` while handling the `SIGABRT` signal

Vipul-Cariappa:
We register `loc_abort_callback_print_stack` as a callback on the `SIGABRT` signal and `loc_segfault_callback_print_stack` on the `SIGSEGV` signal. (reference: https://github.com/lcompilers/lpython/blob/main/src/libasr/stacktrace.cpp#L530-L531)

Tracing the flow of function calls:
`loc_abort_callback_print_stack` and `loc_segfault_callback_print_stack` calls `get_stacktrace` (reference: https://github.com/lcompilers/lpython/blob/main/src/libasr/stacktrace.cpp#L524 & https://github.com/lcompilers/lpython/blob/main/src/libasr/stacktrace.cpp#L516)
`get_stacktrace` calls `get_local_addresses` (reference: https://github.com/lcompilers/lpython/blob/main/src/libasr/stacktrace.cpp#L537)
`get_local_addresses` calls `get_local_address` (reference: https://github.com/lcompilers/lpython/blob/main/src/libasr/stacktrace.cpp#L561)
`get_local_address` calls `abort`. (reference: https://github.com/lcompilers/lpython/blob/main/src/libasr/stacktrace.cpp#L142)

So while handling `SIGABRT` we end up regenerating `SIGABRT` and the cycle repeats. To prevent this, I have changed the `abort()` to `exit(1)` to immediately exit.

I observed this while working on #2564 and #2595.

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -3299,7 +3299,7 @@ void get_local_address_mac(struct Stacktrace *d) {
     printf("The stack address was not found in any shared library or"
         " the main program, the stack is probably corrupted.\n"
         "Aborting...\n");
-    abort();
+    exit(1);
 }
 #endif // HAVE_LFORTRAN_MACHO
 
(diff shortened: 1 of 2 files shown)


Shaikh-Ubaid:
I think this is a Great catch! Thanks! I am curious how did you debug this?

Shaikh-Ubaid (approved):
It looks good to me! Great work! Thanks for this!

PR #2555, opened 2024-02-26
Implementing for-else and while-else loops

advikkabra:
Adding support for for-else and while-else loops according to #2353 .
I have implemented a `while_else` pass which adds flag variables to all while else loops and adds an if check at the end of the loops.
I have also added tests for the same.

PR #2630, opened 2024-03-27
Add nested loop tests for else clause in loops

advikkabra:
Continuation of #2555 .
Nested loops were not added completely in tests.

### DIFF
--- a/integration_tests/loop_10.py
+++ b/integration_tests/loop_10.py
@@ -22,6 +22,7 @@ def no_break_for():
     for i in range(2):
         j += 1
     else:
+        print(j)
         assert j == 2
         return
     assert False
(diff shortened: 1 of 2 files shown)


integration_tests/loop_10.py
```diff
@@ -49,9 +49,53 @@ def nested_loop_for_for():
...
+        print("outer: " + str(i))
+        i += 1
+        while j < 20:
+            print("  inner: " + str(j))
+            break
+    else:
+        print("no break in outer loop")
+        return
+    assert False
+
```
Shaikh-Ubaid, line 92: Please also add tests when there is `else` in the nested loop.

Shaikh-Ubaid:
>I think we can add a reference test for this to check the control flow output

I think integration tests are fine for this.

Shaikh-Ubaid:

integration_tests/loop_10.py
```diff
@@ -49,9 +49,68 @@ def nested_loop_for_for():
...
+    for i in range(2):
+        print("outer: " + str(i))
+        for j in range(10, 12):
+            print("  inner: " + str(j))
+        else:
+            print("no break in inner loop")
+    else:
+        print("no break in outer loop")
+        return
+    assert False
```
Shaikh-Ubaid, line 105: Its good that we are adding tests. I think I found a bug in the `for-else` implementation.
```console
% git diff
diff --git a/integration_tests/loop_10.py b/integration_tests/loop_10.py
index 224c0e21d..c40d2863b 100644
--- a/integration_tests/loop_10.py
+++ b/integration_tests/loop_10.py
@@ -105,12 +105,12 @@ def nested_loop_else():
     assert False
 
 
-with_break_for()
-with_break_while()
-no_break_for()
-break_in_if_for()
-nested_loop_for_for()
-nested_loop_for_while()
-nested_loop_while_for()
-nested_loop_while_while()
+# with_break_for()
+# with_break_while()
+# no_break_for()
+# break_in_if_for()
+# nested_loop_for_for()
+# nested_loop_for_while()
+# nested_loop_while_for()
+# nested_loop_while_while()
 nested_loop_else()
% python integration_tests/loop_10.py
outer: 0
  inner: 10
  inner: 11
no break in inner loop
outer: 1
  inner: 10
  inner: 11
no break in inner loop
no break in outer loop
% lpython integration_tests/loop_10.py
outer: 0
  inner: 10
  inner: 11
outer: 1
  inner: 10
  inner: 11
no break in outer loop
```

integration_tests/loop_10.py
```diff
@@ -49,9 +49,68 @@ def nested_loop_for_for():
...
+
+def nested_loop_else():
+    i: i32
+    j: i32
+    for i in range(2):
+        print("outer: " + str(i))
+        for j in range(10, 12):
+            print("  inner: " + str(j))
+        else:
+            print("no break in inner loop")
```
Shaikh-Ubaid, line 101: You can also add assert statement on variables. For example increment a variable in this else portion and later have an assert on the value of the variable.

Similarly we can do for the above tests.

Shaikh-Ubaid (commented):
Thanks for this! I think we need to fix the following:

- There seems to be a bug in the `for-else`.
- Test cases need to be robust

integration_tests/loop_10.py
```diff
@@ -10,42 +10,91 @@ def with_break_for():
...
 
 def no_break_for():
     i: i32
+    j: i32 = 0
     for i in range(2):
-        print(i)
+        j += 1
     else:
-        print(10)
+        assert j == 2
```
Shaikh-Ubaid, line 25: Please also print the values before the `asserts`. That helps during debugging.

integration_tests/loop_10.py
```diff
@@ -10,42 +10,91 @@ def with_break_for():
...
     j: i32
     for i in range(2):
-        print("outer: " + str(i))
         for j in range(10, 20):
-            print("  inner: " + str(j))
             break
     else:
-        print("no break in outer loop")
+        return
+    assert False
```
Shaikh-Ubaid, line 48: No need for return here. Just increment some variable (inside the `loops`, before the `breaks`, etc.). You can use different variables for different loops here. 
At the end of the function print the final values for the variables and add asserts for them. I think this would make it a much more robust test. You can follow this approach for other tests as well. Thanks!

Shaikh-Ubaid (approved):
It looks good to me. Thanks!

PR #2633, opened 2024-03-28
Implement `set.discard(elem)`

kmr-srbh:
The official Python documentation describes `set.discard(elem)` as:

> Remove element elem from the set if it is present.

and `set.remove(elem)` as:

> Remove element elem from the set. Raises [KeyError](https://docs.python.org/3/library/exceptions.html#KeyError) if elem is not contained in the set.

Hence, it is obvious that both the functions do the same work, but differ only with throwing an error for an absent key. As we already had the infrastructure for `set.remove`, it was easy to base and implement `set.discard` on top of that.

The idea is simple - utilize the same back-end function, but do not throw an error for `set.discard`.
### Working

```python
from lpython import i32

my_set: set[i32] = {1, 2, 3, 4}
my_set.discard(1)
print(len(my_set))
my_set.remove(1)
```

```
3
KeyError: The set does not contain the specified element
```

The changed length and KeyError for `set.remove` prove the normal working of `set.discard`.
#### set.discard

```python
from lpython import i32

my_set: set[i32] = {1, 2, 3, 4}
my_set.discard(100012345678901234567890)
```

```
```

#### set.remove

```python
from lpython import i32

my_set: set[i32] = {1, 2, 3, 4}
my_set.remove(100012345678901234567890)
```

```
KeyError: The set does not contain the specified element
```


kmr-srbh:
@Shaikh-Ubaid this is the new PR.

Shaikh-Ubaid (approved):
It looks good. Thanks for this! Great work!

PR #2634, opened 2024-03-28
Optimize string repetition and string copying

### LINKED ISSUE #2616: Improvement: Assignment `string2: str = string1 * N` takes a very long time for a large integer N
# N = 10<sup>6</sup>
```python
string1: str = "a"
string2: str = string1 * 10**6
```
### CPython
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ time python3 ./examples/example.py

real    0m0.038s
user    0m0.021s
sys     0m0.017s
```

### LPython
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ time ./src/bin/lpython ./examples/example.py

real    0m29.449s
user    0m29.396s
sys     0m0.049s
```

# N = 10<sup>7</sup>
```python
string1: str = "a"
string2: str = string1 * 10**7
```
### CPython
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ time python3 ./examples/example.py

real    0m0.055s
user    0m0.030s
sys     0m0.025s
```

### LPython (force break)
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ time ./src/bin/lpython ./examples/example.py
^C
real    9m24.159s
user    9m23.510s
sys     0m0.144s
```

advikkabra:
Fixes #2616 

1. The current string repetition algorithm was inefficient, where `memset` could be used for single char strings, and binary exponentiation could be used for multi char strings. This is faster than copying one byte at a time.
2. `_lfortran_strcpy` was calling `strlen` in every loop iteration, which was slowing down the program significantly.

After this change, a 1e6 repetition as shown in the issue is 0.04s, which is comparable to CPython.


kmr-srbh:
Advik, I recognize and appreciate the effort you have put in to find and resolve this issue. I fear, we have increased the time instead.
### N = 10 <sup>6</sup>
```python
string1: str = "a"
string2: str = string1 * 10**6
```
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ time ./src/bin/lpython ./examples/example.py
^C
real    1m43.633s
user    1m43.501s
sys     0m0.048s
```

The operation took around 30s for 10^6 before.

kmr-srbh:
I think you might have mistakenly tested the above using CPython.

advikkabra:
It works fine on my machine, and after profiling more than 99% of the time used was in the `strlen` calls in the `strcpy` function.
```sh
advik@advik-Inspiron-7560:~/dev/oss/lpython$ time src/bin/lpython ../test.py
1000000

real    0m0.082s
user    0m0.058s
sys     0m0.020s
advik@advik-Inspiron-7560:~/dev/oss/lpython$ cat ../test.py
string1: str = "a";
string2: str = string1 * 10**6
print(len(string2))
```

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2129,21 +2131,29 @@ LFORTRAN_API int32_t _lpython_bit_length8(int64_t num)
...
         f_len = 0;
     char* dest_char = (char*)malloc(f_len+trmn_size);
-    for (int i = 0; i < n; i++) {
-        for (int j = 0; j < s_len; j++) {
-            dest_char[cntr] = (*s)[j];
-            cntr++;
+
+    if (s_len == 1) {
+        memset(dest_char, *(*s), n);
+    } else {
```
Shaikh-Ubaid, line 2144: I wonder why we need to handle the case for `s_len == 1` separately. The else portion seems general, so I think that should be enough to handle the case of `s_len == 1`.

Shaikh-Ubaid: Does the else portion fail when `s_len == 1`?

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1976,12 +1976,14 @@ LFORTRAN_API void _lfortran_strcpy(char** x, char *y, int8_t free_target)
...
-            x[0][i] = ' ';
-        }
+    size_t y_len = strlen(y);
+    size_t x_len = strlen(*x);
+    size_t i = 0;
+    for (; i < x_len && i < y_len; i++) {
+        x[0][i] = y[i];
+    }
+    for (; i < x_len; i++) {
+        x[0][i] = ' ';
```
Shaikh-Ubaid, line 1986: This is a great improvement. Thanks!

src/libasr/runtime/lfortran_intrinsics.c
advikkabra: It does not fail when `s_len == `, but `memset` is very optimized for this usecase. Setting a large block  of data to a single value is what `memset` excels in, and it is optimized for every architecture. While testing, these are the compared times:
```
100000000
Binary exponentiation: 0.086963
memset: 0.006341
```
```
400000000
Binary exponentiation: 0.211737
memset: 0.018612
```
This is after compiling with `-O3`

Shaikh-Ubaid (approved):
I think we should add a test for this with a large `n` (so that we are sure that the time taken does not break in future). We can do it in a separate PR.

It looks good to me. Thanks for this. Great work!

Shaikh-Ubaid:
>It does not fail when s_len == 1, but memset is very optimized for this usecase.

Did you trying checking this if the `else` works for `s_len == 1`? It is fine to have a separate case for `s_len == 1`, but I just want to be sure that the else works perfectly (for example it works for odd numbers like `1`).

Shaikh-Ubaid:
Let's merge this after we are sure that the `else` case works for odd numbers like `1`.

Shaikh-Ubaid:
>This test is already there in test_str_01.py in integration_tests.

Do we have a test with a large `n` like (10^6 or so)? If so, then it is good.


              I think we should add a test for this with a large `n` (so that we are sure that the time taken does not break in future). We can do it in a separate PR.

_Originally posted by @Shaikh-Ubaid in https://github.com/lcompilers/lpython/pull/2634#pullrequestreview-1975689401_

advikkabra:
How would you suggest I add this as a test? Is there a way to check if something is taking too long?

Shaikh-Ubaid:
>How would you suggest I add this as a test? Is there a way to check if something is taking too long?

I think not yet in LPython (for LFortran we have `cpu_time`). For now, by adding a test we are just trying to ensure that the test is not taking too long. 

Previously (as of the `main` branch) the time complexity of `_lfortran_strcpy()` was `O(n^2)`. That means, for `n = 10^6` the time taken could be in orders of `n^12`, which would never practically complete at the CI.

If we add a test for `n = 6` and the test completes in a finite time (I am assuming it will take `< 1s`. If it take more than `1s` then it is not much helpful and we should reduce `n`), then it verifies/ensures that the time complexity of `_lfortran_strcpy()` has improved.

Shaikh-Ubaid:
For a robust test, we should support something like `time.time()`. But it should not be part of this PR and can be done later.

advikkabra:
I have added a test for a 10**6 repetition. This would take >1 minute in the previous code, so it would be easy to catch if it goes wrong. Right now it completes it in 20ms on my machine, so it is comfortable with the new implementation. I think the PR is ready now.

Shaikh-Ubaid (approved):
Perfect! It looks good to me. Thanks for this. Great work!

PR #2642, opened 2024-04-05
issue:2481 - Unhandled Exception on non-existent file

juliazhorniakAMI:
This modification will print a specific error message ("Error: The input file does not exist") and exit with a return code of 1 if the input file doesn't exist. It catches any other exceptions thrown during the execution and prints their error message as well. Make sure to replace "your existing code here" with the relevant parts of your program.

### DIFF
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -1789,12 +1789,18 @@ int main(int argc, char *argv[])
 
         // TODO: for now we ignore the other filenames, only handle
         // the first:
-        std::string arg_file = arg_files[0];
-        if (CLI::NonexistentPath(arg_file).empty()){
-            std::cerr << "The input file does not exist: " << arg_file << std::endl;
-            return 1;
+        try {
+            std::string arg_file = arg_files[0];
+            if (CLI::NonexistentPath(arg_file).empty()) {
+                std::cerr << "Error: The input file does not exist: " << arg_file << std::endl;
+                return 1; 
+            }
+            return 0; 
+        } 
+        catch (const std::exception& e) {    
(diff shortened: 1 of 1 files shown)


kmr-srbh:
Hello @juliazhorniakAMI! Welcome to LPython. 

We already have a mechanism for detecting non-existent input files and catching errors at runtime. Could you please be more specific regarding what the PR changes? I feel this might be redundant.

Shaikh-Ubaid:
@juliazhorniakAMI Thanks for this. Could you share which specific error/issue this change fixes? Sharing an example `lpython` code would be helpful.

src/bin/lpython.cpp
```diff
@@ -1789,12 +1789,18 @@ int main(int argc, char *argv[])
...
+            std::string arg_file = arg_files[0];
+            if (CLI::NonexistentPath(arg_file).empty()) {
+                std::cerr << "Error: The input file does not exist: " << arg_file << std::endl;
+                return 1; 
+            }
+            return 0; 
+        } 
+        catch (const std::exception& e) {    
+            std::cerr << "Error: " << e.what() << std::endl;
+            return 1;
```
Shaikh-Ubaid, line 1802: @juliazhorniakAMI The change seems fine to me. Although I would like to see a test case that was failing before.

Shaikh-Ubaid:
Thanks for working on this @juliazhorniakAMI! I shared a query above. I am marking this as draft for now. Please mark "Ready for review" when ready.

kmr-srbh:
@Shaikh-Ubaid , this PR was meant to be a fix for #2481, but that was fixed by #2536. The issue is open till now though.

Issue #2614, opened 2024-03-18
Failing symbolic case example


anutosh491:
What doesn't work 
```
from lpython import S
from sympy import Symbol, log, E, Pow, exp

def mmrv(e: S, x: S) -> list[S]:
    empty_list : list[S] = []
    if e.func == Pow and e.args[0] != E:
        e1: S = S(1) * e.args[1]
        return empty_list
    elif e.func == Pow and e.args[0] == E:
        list4: list[S] = [x]
        return list4
    else:
        raise

def test_mrv():
    x: S = Symbol("x")
    ans4: list[S] = mmrv(exp(log(x)), x)
    print(ans4[0])

test_mrv()
```
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine integration_tests/test_gruntz.py 
Segmentation fault
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=c integration_tests/test_gruntz.py 
Segmentation fault
```

anutosh491:
What works 
```
from lpython import S
from sympy import Symbol, log, E, Pow, exp

def mmrv(e: S, x: S) -> list[S]:
    empty_list : list[S] = []
    if e.func == Pow and e.args[0] != E:
        e1: S = e.args[1]
        return empty_list
    elif e.func == Pow and e.args[0] == E:
        list4: list[S] = [x]
        return list4
    else:
        raise

def test_mrv():
    x: S = Symbol("x")
    ans4: list[S] = mmrv(exp(log(x)), x)
    print(ans4[0])

test_mrv()
```
```
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine integration_tests/test_gruntz.py 
x
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine --backend=c integration_tests/test_gruntz.py 
x
```

anutosh491:
Also something similar which works 
```
def mmrv(e: S, x: S) -> list[S]:
    if e.args[0] != E:
        a: S = E
        b: S = pi
        e1: S = a * b
    else:
        list4: list[S] = [x]
        return list4

def test_mrv():
    x: S = Symbol("x")
    ans4: list[S] = mmrv(exp(log(x)), x)
    print(ans4[0])

test_mrv()
```
```
[...]

PR #2643, opened 2024-04-06
Introducing basic_free_stack_if_not_null

anutosh491:
Addressing https://github.com/lcompilers/lpython/issues/2614#issuecomment-2035025925

We have been using the `basic_free_stack` out of symengine's C wrapper untill now. But before freeing out variables we need to check if they have been stack allocated previously. Hence this PR aims to replace every `basic_free_stack` call with
```
    if (var != NULL) {
        basic_free_stack(var);
    }
``


certik:
@anutosh491 what is the status here --- can you describe what the problem is?

Are you trying to address this comment: https://github.com/lcompilers/lpython/issues/2614#issuecomment-2035025925 ?

PR #2646, opened 2024-04-12
CI: Support uploading tarball to release

anutosh491:


ci/environment_linux.yml
```diff
@@ -0,0 +1,20 @@
+name: lf
```
Shaikh-Ubaid, line 1: Let's not add another environment file. Instead let's reuse the existing `ci/environment.yml` file similar to other jobs.

PR #2651, opened 2024-04-14
Add builtin functions

farah-salama:
## Functions
`str.center()` : Return centered in a string of length width. Padding is done using the specified fillchar (default is an ASCII space). 
`str.expandtabs()` : Return a copy of the string where all tab characters are replaced by one or more spaces, depending on the current column and the given tab size.
## Testing
Tests were added in `integration_tests/test_str_attributes.py`


tests/reference/run_dbg-test_raise_01-dfd86ca.stderr
```diff
@@ -1,7 +1 @@
-  File "tests/runtime_errors/test_raise_01.py", line 1
-    def test():
-  File "tests/runtime_errors/test_raise_01.py", line 4
-    test()
-  File "tests/runtime_errors/test_raise_01.py", line 2
-    raise
-ERROR STOP
+semantic error: The `runtime stacktrace` is not enabled. To get the stacktraces, re-build LPython with `-DWITH_RUNTIME_STACKTRACE=yes`
```
Shaikh-Ubaid, line 1: You should build the `lpython` project with `-DWITH_RUNTIME_STACKTRACE=yes`. You can simply use the provided `build0.sh` and `build1.sh` files for building which has this options already setup.

Shaikh-Ubaid (commented):
I would suggest rebasing your PR over the latest `main`, rebuilding the project and updating the reference tests.

At a high level this seems good. I will review again after the rebase.

tests/reference/run_dbg-test_raise_01-dfd86ca.stderr
farah-salama: I did as you suggested and this error appeared, I guess I missed something while implementing the function but I can't figure out what exactly

image

kmr-srbh: @farah-salama the issue occurs because you have not currently implemented the method for compile-time strings. You can have a look at how `startswith` is implemented and add support for them.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -6880,6 +6880,59 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                ASR::ttype_t *arg_value_type = ASRUtils::expr_type(arg_value);
+                if (!ASRUtils::is_integer(*arg_value_type)) {
+                    throw SemanticError("str.expandtabs() argument must be integer", loc);
+                }
+
+                ASR::call_arg_t value;
+                value.loc = loc;
+                value.m_value = args[0].m_value;
+                fn_args.push_back(al, value);
+            }
```
Shaikh-Ubaid, line 6935: This handles runtime implementation. We also need to support compile-time implementation inside https://github.com/lcompilers/lpython/blob/7d0cfdbec3269b3aaaeea57a012e7b9135194ba1/src/lpython/semantics/python_ast_to_asr.cpp#L6909.

src/runtime/lpython_builtin.py
```diff
@@ -1085,6 +1085,65 @@ def _lpython_str_isspace(s: str) -> bool:
...
+    result: str = ""
+    c: str
+    for c in s:
+        if c == '\t':
+            if tabsize > 0:
+                i: i32
+                for i in range(tabsize-(col % tabsize)):
+                    result += ' '
+            col = 0
+        elif (c == '\n') | (c == '\r'):
```
kmr-srbh, line 1135:
```suggestion
        elif c == '\n' or c == '\r':
```

src/runtime/lpython_builtin.py
```diff
@@ -1085,6 +1085,65 @@ def _lpython_str_isspace(s: str) -> bool:
...
+        left_padding = i32(width/2) + 1
+    else:
+        left_padding = i32(width/2)
+    i: i32 
+    for i in range(left_padding):
+        result += fillchar
+    right_padding: i32 = width - left_padding
+    result += s
+    for i in range(right_padding):
+        result += fillchar
```
kmr-srbh, line 1112: We can utilize the new improvement in string repeat added a dew days ago and simplify this to just:
```python
right_padding: i32 = width - left_padding
result += fillchar * left_padding
result += s
result += fillchar * right_padding
```

#2652 needs to be addressed to achieve it. For now we may keep the current `for` loop.

src/runtime/lpython_builtin.py
```diff
@@ -1085,6 +1085,65 @@ def _lpython_str_isspace(s: str) -> bool:
...
+    """
+    Return centered in a string of length width. 
+    Padding is done using the specified fillchar (default is an ASCII space). 
+    The original string is returned if width is less than or equal to len(s).
+    """
+    if(len(fillchar) != 1):
+        raise TypeError("The fill character must be exactly one character long")
+    if(width <= len(s)):
+        return s
+    width -= len(s)
```
kmr-srbh, line 1099:
```suggestion
    str_len: i32 = len(s)
    if width <= str_len:
        return s
    width -= str_len
```

Although `len(x)` is a O(1) operation, let's have only a single function call.

integration_tests/test_str_attributes.py
```diff
@@ -472,6 +472,24 @@ def is_numeric():
     assert "".isnumeric() == False
     assert "ab2%3".isnumeric() == False
 
+def center():
```
kmr-srbh, line 475: The current implementation supports runtime strings. Please add tests for them. By runtime strings, I mean a string stored in a variable. You can have a look at how other `is_upper()` is tested.

kmr-srbh (changes requested):
I shared some comments above. The PR looks good. 

It would be nice to have #2652 resolved as it can speed up things.

integration_tests/test_str_attributes.py
kmr-srbh: @farah-salama please update the test references also.

kmr-srbh: Could you please rebuild and update the test references again?

kmr-srbh:
@farah-salama the print statements existed for checking the tests. Please revert that change.

farah-salama:
@Shaikh-Ubaid @kmr-srbh Thank you so much for your help.
I tried to handle the compile-time implementation for the functions in `handle_constant_string_attributes` but I couldn't find a similar function implemented that takes an integer argument and I faced many errors that I couldn't solve so I changed the test cases to test the run time implementation only, I would appreciate it though if someone could tell me how to handle the compile-time implementation for such functions.

kmr-srbh:
@farah-salama I present the method to implement `center`. You may similarly follow to implement `expandtabs`. 

We first check for the number of arguments to the function.
```c++
if (args.size() != 1 || args.size() != 2) {
    throw SemanticError("str.center() takes 1 or 2 arguments",
                        loc);
}
```
We next move on to check the argument types. As `fillchar` is an optional parameter, we need an `if` block to check for `args.size()` again. When we have 1 argument, we handle only `width`. When we have 2 arguments, we handle both.

You can check for `width` being an integer type like:
```c++
ASR::expr_t *arg1 = args[0].m_value;
ASR::ttype_t *type = ASRUtils::expr_type(arg1);
if (!ASRUtils::is_integer(*type)) {
    throw SemanticError("argument  'width' to str.center() must be of type int",
                    arg1->base.loc);
}
```
We follow the above step again to check for `args[1].m_value` when we have 2 arguments.

We next store their compile time values. We do this here because we need all the arguments to be a compile-time constant to move ahead. For the `width` parameter of `center`, we can do:
```c++
ASR::IntegerConstant_t* width_arg = ASR::down_cast<ASR::IntegerConstant_t>(arg1);
int64_t width = width_arg->m_n;
```

`fillchar` is an optional parameter so we declare it as `std::string fillchar = "";` when only 1 argument is provided. We follow the above step to store it as a `StringConstant_t*` when we have 2 arguments.

We next check whether the provided arguments are compile-time constants. Handle this according to the arguments you are handling above. We do so by checking if they are not all `nullptr`:
```c++
if (ASRUtils::expr_value(arg1) != nullptr) {...}
```

Inside this `if` statement, you need to implement the same logic you have used for implementing the method.
We need an `else` clause to handle cases when we have either one or all arguments non-constants. Inside that clause, we make a call to the `_lpython_str_center` function which you have already implemented. You may have a look at how `endswith` does this.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -6880,6 +6880,59 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            str.loc = loc;
+            str.m_value = s_var;
+
+            ASR::call_arg_t value;
+            value.loc = loc;
+            value.m_value = args[0].m_value;
+            fn_args.push_back(al, str);
+            fn_args.push_back(al, value);
+
+            if(args.size()==2){
```
Shaikh-Ubaid, line 6905:
```suggestion
            if(args.size() == 2){
```

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -6880,6 +6880,59 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            if(args.size() > 1) {
+                throw SemanticError("str.expandtabs() takes at most one argument.", loc);
+            }
+            fn_call_name = "_lpython_str_expandtabs";
+            ASR::call_arg_t str;
+            str.loc = loc;
+            str.m_value = s_var;
+            fn_args.push_back(al, str);
+
+            if(args.size()==1){
```
Shaikh-Ubaid, line 6924:
```suggestion
            if(args.size() == 1){
```

src/runtime/lpython_builtin.py
```diff
@@ -1085,6 +1085,64 @@ def _lpython_str_isspace(s: str) -> bool:
...
 
+@overload
+def _lpython_str_center(s: str, width: i32, fillchar: str) -> str:
+    """
+    Return centered in a string of length width. 
+    Padding is done using the specified fillchar (default is an ASCII space). 
+    The original string is returned if width is less than or equal to len(s).
+    """
+    if(len(fillchar) != 1):
+        raise TypeError("The fill character must be exactly one character long")
```
Shaikh-Ubaid, line 1096: Considering https://github.com/lcompilers/lpython/issues/2666, I think this will not print the error message. But I think it is fine for now. The error message should start printing as soon as we fix https://github.com/lcompilers/lpython/issues/2666.

src/runtime/lpython_builtin.py
```diff
@@ -1085,6 +1085,64 @@ def _lpython_str_isspace(s: str) -> bool:
...
+    col: i32 = 0
+    result: str = ""
+    c: str
+    for c in s:
+        if c == '\t':
+            if tabsize > 0:
+                i: i32
+                iterations: i32 = tabsize - _mod(col,tabsize)
+                for i in range(iterations):
+                    result += ' '
```
Shaikh-Ubaid, line 1132: For now this is fine, but we can later refactor this to simply use `*` operator.

Shaikh-Ubaid (commented):
This looks good to me. I shared some comments and code suggestions. Let's do the code suggestions and merge this. Great work @farah-salama ! Thanks for the help @kmr-srbh !

src/runtime/lpython_builtin.py
Shaikh-Ubaid: @farah-salama do you mind updating this?

Shaikh-Ubaid: Updated.

Shaikh-Ubaid (approved):

PR #2653, opened 2024-04-16
WASM: Use release build

anutosh491:
Inspired by the patch added here(https://github.com/emscripten-forge/recipes/pull/883/files) to build lpython against the emscripten compiler package

### DIFF
--- a/build_to_wasm.sh
+++ b/build_to_wasm.sh
@@ -9,8 +9,9 @@ cp -r src/runtime/lpython src/bin/asset_dir
 
 ./build0.sh
 emcmake cmake \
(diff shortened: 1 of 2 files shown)


Shaikh-Ubaid:
Not exactly sure if the `release` build works perfectly in a browser. I will check and revert back later.

Shaikh-Ubaid:
I checked locally. I think this works. Thanks!

anutosh491:
Which means if we tag a release, we can host the recipe for lpython emscripten-forge without the additional patch. Would just make the recipe cleaner ;)

PR #2655, opened 2024-04-16
Implement `input(prompt)`

kmr-srbh:


kmr-srbh:
@Shaikh-Ubaid , there is a small issue occurring here. Because we are directing calls to `FileRead`, `input()` itself is not having a return type and value. So, an expression like `n: i32 = i32(int(input("Enter a number: "))))` leads to an exception. A simple program:
```python
name: str = input("Your name: ")
print("Hello,", name)
```
gives the following output
```console
Your name: Segmentation fault (core dumped)
```

I request you to see if the implementation is correct. Instead of a separate function `create_Input()` with a call inside `if (call_name == "input")`, I am using the complete piece of code here.

Shaikh-Ubaid:
@kmr-srbh the following works:
```console
% cat examples/expr2.py 
name: str = "                "
name = input("Your name: ")
print("Hello,", name)
% python examples/expr2.py
Your name: John
Hello, John
% lpython examples/expr2.py
Your name: John
Hello,John
```
We just needed `name` to have initial length/space so that we could write to it the string read from `stdin`.

Shaikh-Ubaid:
We need to see what other better alternative approaches can be for the workaround (for `name` length) in https://github.com/lcompilers/lpython/pull/2655#issuecomment-2059840977.

kmr-srbh:
> We need to see what other better alternative approaches can be for the workaround (for `name` length) in [#2655 (comment)](https://github.com/lcompilers/lpython/pull/2655#issuecomment-2059840977).

You are right @Shaikh-Ubaid. Also, could you please look into the return type issue?

kmr-srbh:
> Let's modify the semantics of the FileRead ASR node: if the "value" expr argument is an allocatable string, then the backend (LLVM/WASM, etc.) implementation of the FileRead node will allocate the string to hold the whole input, and assign it to it.

@certik Do you mean handling `String_t` here? 

https://github.com/lcompilers/lpython/blob/fce0b35fd59ba3e4a4cb978aab69ec3ff0a7a82a/src/libasr/codegen/asr_to_llvm.cpp#L7493-L7510

PR #2656, opened 2024-04-17
Updated test_gruntz.py

anutosh491:
This PR introduces some changes and a case in test gruntz.py 
The changes introduced are related to freeing out variables.


anutosh491:
So something that doesn't work perfectly as of now is this
```
from lpython import S
from sympy import Symbol, E

def mmrv(e: S) -> S:
    if False:
        print(E)
    else:
        return e

def test_mrv():
    x: S = Symbol("x")
    ans: S = mmrv(x)
    print(ans)

test_mrv()
```
But something that does work perfectly ( a very fair workaround ) is this
```
from lpython import S
from sympy import Symbol, E

def mmrv(e: S) -> S:
    if False:
        var: S = E
        print(var)
    else:
        return e

def test_mrv():
    x: S = Symbol("x")
    ans: S = mmrv(x)
    print(ans)

test_mrv()
```

anutosh491:
The PR also introduces a case where the base of a Pow expression might be `E`

certik (approved):

PR #2670, opened 2024-04-26
Add space between multiple strings in print

### LINKED ISSUE #2669: Bug: Printing multiple strings together does not add a space between them
```python
print("a", "b", "c", "d")

fruits: list[str] = ["apple", "guava", "mango"]
mango: str = fruits.pop()
print("I have got a", mango, "to eat!")
```

### CPython
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ python ./examples/example.py
a b c d
I have got a mango to eat!
```

### LPython
```console
abcd
I have got amangoto eat!
```

kmr-srbh:
fixes #2669 

```python
print("a", "b", "c", "d")

fruits: list[str] = ["apple", "guava", "mango"]
mango: str = fruits.pop()
print("I have got a", mango, "to eat!")
```
```console
a b c d
I have got a mango to eat!
```


kmr-srbh:
@Shaikh-Ubaid could you please rerun the CI? The failing check is unrelated to the PR.

Shaikh-Ubaid:
The approach in this PR might break `lfortran`. `lfortran` needs to not print spaces when two consecutive print arguments are strings.

```console
% cat examples/expr2.f90     
program expr2
implicit none

print *, "hi", "bye"
print *, 12, 23

end program
% gfortran examples/expr2.f90 
% ./a.out 
 hibye
          12          23
% lfortran examples/expr2.f90
hibye
12 23
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -7981,9 +7981,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         for (size_t i=0; i<x.n_values; i++) {
             if (i != 0) {
                 fmt.push_back("%s");
-                if (global_sep_space &&
-                    !(ASRUtils::is_character(*ASRUtils::expr_type(x.m_values[i]))
-                        && ASRUtils::is_character(*ASRUtils::expr_type(x.m_values[i - 1])))) {
+                if (global_sep_space) {
```
Shaikh-Ubaid, line 7984: The way to fix this such that we print spaces when two consecutive arguments are strings in LPython and not print spaces for the same in LFortran, is to support this in AST->ASR Translation.

Shaikh-Ubaid: You will need to figure out how to update the middle-ends of both lfortran and lpython such that both of them work as expected when two consecutive arguments are strings.

kmr-srbh: @Shaikh-Ubaid thanks for this! I will try to create a solution that works for both. :+1:

Shaikh-Ubaid: See if your solution works for both lpython and lfortran (test it locally on your system or send a PR to lfortran so it gets tested at the CI). If no `run` reference tests change for lfortran, then it works.

You can also test your current approach with lfortran. If it works currently, then I think its good.

PR #2685, opened 2024-05-05
Added support for addressing more exp cases

anutosh491:


anutosh491:
1) is_integer check needs to be supported
2) Symbolic comparison between numbers (is_positive/ is_negative) etc checks should be supported.

PR #2688, opened 2024-05-09
Gruntz Demo

anutosh491:
This is a draft PR trying to replicate the gruntz algorithm and a few test cases for the same.

### DIFF
--- a/integration_tests/gruntz_demo2.py
+++ b/integration_tests/gruntz_demo2.py
@@ -143,23 +143,15 @@ def mrv(e, x):
 
     """
 
-    if not e.has(x):
-        return set()
     if e == x:
         return {x}
     if e.is_Mul or e.is_Add:
         a, b = e.as_two_terms()
+        ans1 = mrv(a, x)
+        ans2 = mrv(b, x)
         return mrv_max(mrv(a, x), mrv(b, x), x)
-    if e.func == exp:
-        if e.exp == x:
-            return {e}
-        if any(a.is_infinite for a in Mul.make_args(limitinf(e.exp, x))):
-            return mrv_max({e}, mrv(e.exp, x), x)
-        return mrv(e.exp, x)
     if e.is_Pow:
         return mrv(e.base, x)
-    if isinstance(e, log):
-        return mrv(e.args[0], x)
     if e.is_Function:
         return reduce(lambda a, b: mrv_max(a, b, x), (mrv(a, x) for a in e.args))
     raise NotImplementedError(f"Can't calculate the MRV of {e}.")
@@ -182,19 +174,6 @@ def mrv_max(f, g, x):
     if f & g:
         return f | g
 
-    a, b = map(next, map(iter, (f, g)))
-
-    # The log(exp(...)) must always be simplified here.
-    la = a.exp if a.is_Exp else log(a)
-    lb = b.exp if b.is_Exp else log(b)
-
-    c = limitinf(la/lb, x)
-    if c.is_zero:
-        return g
-    if c.is_infinite:
-        return f
-    return f | g
-
 def rewrite(e, x, w):
     r"""
     Rewrites the expression in terms of the MRV subexpression.
@@ -213,8 +192,7 @@ def rewrite(e, x, w):
     Returns
     =======
 
-    tuple
-        A pair: rewritten (in `w`) expression and `\log(w)`.
+    The rewritten expression
 
     Examples
     ========
@@ -225,14 +203,12 @@ def rewrite(e, x, w):
     """
 
     Omega = mrv(e, x)
-    if not Omega:
-        return e, None  # e really does not depend on x
 
     if x in Omega:
         # Moving up in the asymptotical scale:
         with evaluate(False):
-            e = e.xreplace({x: exp(x)})
-            Omega = {s.xreplace({x: exp(x)}) for s in Omega}
+            e = e.subs(x, exp(x))
+            Omega = {s.subs(x, exp(x)) for s in Omega}
 
     Omega = list(ordered(Omega, keys=lambda a: -len(mrv(a, x))))
 
@@ -249,9 +225,9 @@ def rewrite(e, x, w):
         c = limitinf(a.exp/g.exp, x)
         b = exp(a.exp - c*g.exp)*w**c  # exponential must never be expanded here
         with evaluate(False):
-            e = e.xreplace({a: b})
+            e = e.subs(a, b)
 
-    return e, -sig*g.exp
+    return e
 
 @cacheit
 def mrv_leadterm(e, x):
@@ -274,28 +250,9 @@ def mrv_leadterm(e, x):
 
     """
 
-    if not e.has(x):
-        return e, Integer(0)
-
-    # Rewrite to exp-log functions per Sec. 3.3 of thesis.
-    e = e.replace(lambda f: f.is_Pow and f.exp.has(x),
-                  lambda f: exp(log(f.base)*f.exp))
-    e = e.replace(lambda f: f.is_Mul and sum(a.func == exp for a in f.args) > 1,
-                  lambda f: Mul(exp(Add(*(a.exp for a in f.args if a.func == exp))),
-                                *(a for a in f.args if not a.func == exp)))
-
-    # The positive dummy, w, is used here so log(w*2) etc. will expand.
-    # TODO: For limits of complex functions, the algorithm would have to
-    # be improved, or just find limits of Re and Im components separately.
     w = Dummy('w', real=True, positive=True)
-    e, logw = rewrite(e, x, w)
-
-    c0, e0 = e.leadterm(w, logx=logw)
-    if c0.has(w):
-        raise NotImplementedError(f'Cannot compute leadterm({e}, {x}). '
-                                  'The coefficient should have been free of '
-                                  f'{w}, but got {c0}.')
-    return c0.subs(log(w), logw), e0
+    e = rewrite(e, x, w)
+    return e.leadterm(w)
 
 @cacheit
 def signinf(e, x):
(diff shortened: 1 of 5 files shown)


anutosh491:
cc @certik

anutosh491:
Internally `mrv_leadterm` depends on `mrv` , `rewrite` and `leadterm`
```
mrv_leadterm(sin(1/x), x) 
1) exps = mrv(sin(1/x), x) = sin(1/x)
2) 
    w = Dummy("w", positive=True)
    f, logw = rewrite(exps, Omega, x, w)
```    
This gives us `f` which would be `sin(_w)`
3) Then a calculate the leadterm for `sin(_w)` which is `x` -> `(1, 1)`

anutosh491:
A slightly more involved example would be this 
```
In [1]: limit(sin(x)/x, x, 0)
limitinf(_x*sin(1/_x), _x) = 1
+-mrv_leadterm(_x*sin(1/_x), _x) = (1, 0)
| +-mrv(_x*sin(1/_x), _x) = set([_x])
| | +-mrv(_x, _x) = set([_x])
| | +-mrv(sin(1/_x), _x) = set([_x])
| |   +-mrv(1/_x, _x) = set([_x])
| |     +-mrv(_x, _x) = set([_x])
| +-mrv_leadterm(exp(_x)*sin(exp(-_x)), _x, set([exp(_x)])) = (1, 0)
|   +-rewrite(exp(_x)*sin(exp(-_x)), set([exp(_x)]), _x, _w) = (1/_w*sin(_w), -_x)
|     +-sign(_x, _x) = 1
|     +-mrv_leadterm(1, _x) = (1, 0)
+-sign(0, _x) = 0
+-limitinf(1, _x) = 1
```

anutosh491:
The current implementation can be simplified quite a bit 

1) limitinf technically depends on 

-   `mrv_leadterm` : To calculate leadterm of most rapidly varying subexpression
-  `sign` : To compute limitinf based on exponent of calculated leadterm

2) `mrv_leadterm` technically depends on 

- `rewrite` : To basically rewrite the expression in terms of the most rapidly varying (`mrv`) expression
- " Any general leadterm computing function": Like we have the leadterm from sympy which is actually the function that is responsible for generating the leading term's coeff and exponent

 3) `rewrite` technically depends on 

- `mrv` : To calculate the most rapidly varying term.

So we have something like this 
image

certik:
Perfect. Yes, let's do the sin(x)/x, and you can take the running example in sympy and just simplify it to the bare minimum. Then we'll port the simplified code to LPython.

anutosh491:
The latest commit fixes some import errors and any erorrs with the code and now we can get results for our simplified gruntz algorithm through sympy. I've added a test for the same saying 
```
# tests
x = Symbol('x')
ans = gruntz(sin(x)/x, x, 0)
print(ans)
```

anutosh491:
The recent commits simplifies the previous version of gruntz by quite a bit. For our use case i.e. `sin(x)/x`

1) We don't have to worry about ideas related to **tractable/intractable** functions
That would only come into picture we would be dealing with **inverse trig** or **inverse hyperbolic** functions etc

2) We don't need to worry about the `logw` parameter so that has been removed.
That would only come into picture if we would be dealing with nested logs, exponentials for example `log(log(1/x))` or `exp(exp(log(1/x))`

anutosh491:
Also talking about the `rewrite` function. The `rewrite` functions is cruicial and can't be simplified any further but it's actual requirements comes into picture when we have `exp-log` functions. For our use case we can simplify it if we want

So the `mrv` term would in almost all cases be a part of one of these classes `const`, `log`, `linear (x)` or `exp` . Now if we consider our test case or any case where we know the mrv is linear, we essentially know what the rewrite would be.
```
>>> x = Symbol('x')
>>> rewrite(x, x, w)
(1/_w, -x)
>>> rewrite(log(x), x, w)
(log(1/_w), -x)
>>> rewrite(x**2, x, w)
((1/_w)**2, -x)
>>> rewrite(exp(log(x)), x, w)
(1/_w, -x)
>>> rewrite(sin(x), x, w)
(sin(1/_w), -x)
>>> rewrite(sin(x)/x, x, w)
(sin(1/_w)/1/_w, -x)
```
All of these have `x` as their `mrv` and so the rewrite would just be substituing `1/x` in place of `x` in the expression.

anutosh491:
For the `signinf` function, we would need the mathematical function `sign`
```
def signinf(e, x):
    r"""
    Determine sign of the expression at the infinity.

    Returns
    =======

    {1, 0, -1}
        One or minus one, if `e > 0` or `e < 0` for `x` sufficiently
        large and zero if `e` is *constantly* zero for `x\to\infty`.

    """

    if not e.has(x):
        return sign(e)
    if e == x or (e.is_Pow and signinf(e.base, x) == 1):
        return S(1)     
```

I've raised a PR to symengine to add functionality for `sign` ( using `basic_sign`) here https://github.com/symengine/symengine/pull/2016

anutosh491:
The `limitinf` function would require support for the `subs` function 
```
def gruntz(e, z, z0, dir="+"):
    if str(dir) == "-":
        e0 = e.subs(z, z0 - 1/z)
    elif str(dir) == "+":
        e0 = e.subs(z, z0 + 1/z)
    else:
        raise NotImplementedError("dir must be '+' or '-'")

    r = limitinf(e0, z)
    return r 
```
 And technically the `rewrite` function would also be using `subs` or `xreplace` function. Hence I've raised a PR adding support for the `subs` attrbute here https://github.com/lcompilers/lpython/pull/2695

anutosh491:
Listing all blockers below.

integration_tests/gruntz_demo3.py
```diff
@@ -0,0 +1,184 @@
...
+        b: S = e.args[1]
+        ans1: list[S] = mrv(a, x)
+        ans2: list[S] = mrv(b, x)
+        return mrv_max(mrv(a, x), mrv(b, x), x)
+    if e.func == Pow:
+        base: S = e.args[0]
+        list3: list[S] = mrv(base, x)
+        return list3
+    if e.is_Function:
+        return reduce(lambda a, b: mrv_max(a, b, x), (mrv(a, x) for a in e.args))
```
anutosh491, line 30: Blocker 1: Need to look into this as this is somewhat restrictive.

`e.is_Function` means all cases but (`log`, `exp`, `pow`, `add`, `mul`) which is basically `trig`, `inverse trig`, `hyperbolic` etc
So maybe let's just look into limits involving `sin` for now

Also have to address this line somehow `return reduce(lambda a, b: mrv_max(a, b, x), (mrv(a, x) for a in e.args))`

integration_tests/gruntz_demo3.py
```diff
@@ -0,0 +1,184 @@
...
+    {x**5}
+
+    """
+
+    if len(f) == 0:
+        return g
+    elif len(g) == 0:
+        return f
+    else:
+        return f | g
```
anutosh491, line 49: `mrv_max` as a function is not too complex. If any list is empty return the other list and if both are non empty, just take union and return. So supporting some kind of `union` (or maybe `append`) for symbolic list (`list[S]`) might be required.

integration_tests/gruntz_demo3.py
```diff
@@ -0,0 +1,184 @@
...
+    (-1, 0)
+
+    """
+
+    #w = Dummy('w', real=True, positive=True)
+    w: S = Symbol('w')
+    #e = rewrite(e, x, w)
+    coeff_exp_list: list[S] = [S(2), S(3)]
+    #return e.leadterm(w)
+    return coeff_exp_list
```
anutosh491, line 120: Blocker 3:

We would need a leadterm function here. The leadterm for an expression `e` would end up in the form `c*x**n`.
If using sympy, we do
```
>>> 2*x**3
2*x**3
>>> _.as_coeff_exponent(x)
(2, 3)
```
But we might not have the `as_coeff_exponent` method. What we have is `args`. So we can maybe try the following workaround 
```
>>> 2*x**3
2*x**3
>>> _.args
(2, x**3)
>>> 2*x**3
2*x**3
>>> _.args
(2, x**3)
>>> c = _[0]
>>> 2*x**3/c
x**3
>>> _.args
(x, 3)
>>> n = _[1]
>>> c, n
(2, 3)
```

integration_tests/gruntz_demo3.py
```diff
@@ -0,0 +1,213 @@
...
+    The rewritten expression
+
+    Examples
+    ========
+
+    >>> rewrite(exp(x)*log(x), x, y)
+    (log(x)/y, -x)
+
+    """
+    pass
```
anutosh491, line 78: Blocker 4:

Finish rewrite function

integration_tests/gruntz_demo3.py
```diff
@@ -0,0 +1,213 @@
...
+    {x}
+
+    """
+
+    if e == x:
+        list1: list[S] = [x]
+        return list1
+    if e.func == Mul or e.func == Add:
+        a: S = e.args[0]
+        b: S = e.args[1]
```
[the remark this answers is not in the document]
anutosh491: Okay this is a blocker for the case `mrv(2*sin(x)/x, x)` or any case involving more than 1 `*` or `+` operator. In that case the `.args` can't cover for the `as_two_terms` function.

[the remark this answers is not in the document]
anutosh491: I've opened an issue for the same here. 
https://github.com/symengine/symengine/issues/2018

integration_tests/gruntz_demo2.py
```diff
@@ -0,0 +1,347 @@
...
+    {x}
+
+    """
+
+    if e == x:
+        return {x}
+    elif e.is_Integer:
+        return {}
+    elif e.is_Mul or e.is_Add:
+        a, b = e.as_two_terms()
```
anutosh491, line 151: `as_two_terms` is required here.

integration_tests/gruntz_demo3.py
```diff
@@ -0,0 +1,288 @@
...
+    if not e.has(x):
+        return sign(e)
+    if e == x:
+        return S(1)
+    if e.func == Pow:
+        base: S = e.args[0]
+        if signinf(base, x) == S(1):
+            return S(1)
+
+def leadterm(e: S, x: S) -> list[S]:
```
anutosh491, line 147: leadterm being done manually

PR #2690, opened 2024-05-09
ASR -> CPython: Add visitors for `ListConstant`, `TupleConstant` & `SetConstant`

kmr-srbh:
```python
from lpython import i32, f64

my_first_list: list[i32] = [1, 2, 3 , 4]
print(my_first_list)

my_second_list: list[str] = ["a", "b", "c", "d"]
print(my_second_list)

my_third_list: list[list[i32]] = [[1, 2], [3, 4], [5, 6]]
print(my_third_list)

my_fourth_list: list[list[f64]] = [[1.0, 2.2], [3.6, 4.9], [5.1, 6.3]]
print(my_fourth_list)

my_first_tuple: tuple[i32, str, f64] = (1, "hello", 2.4)
print(my_first_tuple)

my_second_tuple: tuple[tuple[i32, str], str] = ((1, "hello"), "world")
print(my_second_tuple)

my_first_set: set[i32] = {1, 2, 3, 2, 4}
print(my_first_set)

my_second_set: set[f64] = {1.1, 2.5, 6.8}
print(my_second_set)

my_third_set: set[str] = {"a", "b", "a", "c"}
print(my_third_set)
```

```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ ./src/bin/lpython ./examples/example.py --show-python
def __main__global_init():
    my_first_list = [1, 2, 3, 4]
    my_second_list = ["a", "b", "c", "d"]
    my_third_list = [[1, 2], [3, 4], [5, 6]]
    my_fourth_list = [[1.000000, 2.200000], [3.600000, 4.900000], [5.100000, 6.300000]]
    my_first_tuple = (1, "hello", 2.400000)
    my_second_tuple = ((1, "hello"), "world")
    my_first_set = {1, 2, 3, 2, 4}
    my_second_set = {1.100000, 2.500000, 6.800000}
    my_third_set = {"a", "b", "a", "c"}
def __main__global_stmts():
    print(my_first_list)
    print(my_second_list)
    print(my_third_list)
    print(my_fourth_list)
    print(my_first_tuple)
    print(my_second_tuple)
    print(my_first_set)
    print(my_second_set)
    print(my_third_set)
```

### TODO
- [X] Add tests


kmr-srbh:
@Shaikh-Ubaid what are your thoughts on this, is this the right way to do it? Could you please guide me on adding tests for this?

src/libasr/codegen/asr_to_python.cpp
```diff
@@ -619,6 +622,23 @@ class ASRToLpythonVisitor : public ASR::BaseVisitor<ASRToLpythonVisitor>
...
+        r += "[";
+        size_t i = 0;
+        while (i < x.n_args - 1) {
+            visit_expr(*x.m_args[i]);
+            r += s;
+            r += ", ";
+            i++;
+        }
+        visit_expr(*x.m_args[i]);
+        r += s;
```
nikabot, line 636: This is good, but why don't we follow the same method as we do for other visitors
```suggestion
        for (size_t I =0 ; I < x.n_args; i++) {
            this->visit_use_symbol(*x.m_args[i]);
            r.append(s);
            if (i < x.n_args-1) r.append(", ");
        }
```

Can you do the same for other PR's as well.
Otherwise, looks good.

Shaikh-Ubaid: Yes, we should follow similar to other visitors.

kmr-srbh: @Shaikh-Ubaid @nikabot I do not see the `visit_use_symbol` visitor used anywhere in the code-base. I think it is not yet part of LPython now.

image

src/libasr/codegen/asr_to_python.cpp
```diff
@@ -619,6 +622,23 @@ class ASRToLpythonVisitor : public ASR::BaseVisitor<ASRToLpythonVisitor>
         s = r;
     }
 
+    void visit_ListConstant(const ASR::ListConstant_t &x) {
```
Shaikh-Ubaid, line 625: Looking into your PRs https://github.com/lcompilers/lpython/pull/2690, https://github.com/lcompilers/lpython/pull/2691, https://github.com/lcompilers/lpython/pull/2692, I think the visitors implemented can have a common implementation using a templated function.

For example your `visit_List/Tuple/SetConstant()` can look like

```console
void visit_ListConstant(const ASR::ListConstant_t &x) {
    visit_AggregateConstant(x.n_args, x.m_args, "[", "]");
}

void visit_TupleConstant(const ASR::TupleConstant_t &x) {
    visit_AggregateConstant(x.n_elements, x.m_elements, "(", ")");
}

...
```

where `visit_AggregateConstant()` is the templated function. You can figure out the details for the arguments of this function.

src/libasr/codegen/asr_to_python.cpp
Shaikh-Ubaid: I think he just means `visit_expr`.

src/libasr/codegen/asr_to_python.cpp
kmr-srbh: @Shaikh-Ubaid I totally agree with you. Except `DictConstant_t`, we can have a common visitor like you mentioned for the rest data-structures. :+1:

Shaikh-Ubaid:
>Regarding tests, I think you enable tests for python in tests.toml and search for list1.py and add python=true.

Yes, let's just add one reference test that includes all three `list`, `tuple`, `set`. In the long term, we need an `integration_tests` approach for testing this.

src/libasr/codegen/asr_to_python.cpp
nikabot: Yes, I'm sorry, my bad.

kmr-srbh:
@Shaikh-Ubaid @nikabot since the visitors will be similar for 3 of 4 data-structures, shall I implement them and test them together? I think it would be fast this way. For `DictConstant`, we can have the separate PR.

Shaikh-Ubaid:
> @Shaikh-Ubaid @nikabot since the visitors will be similar for 3 of 4 data-structures, shall I implement them and test them together? I think it would be fast this way. For `DictConstant`, we can have the separate PR.

Yes.

kmr-srbh:
@Shaikh-Ubaid @nikabot what can be a better name for the test file?

src/libasr/codegen/asr_to_python.cpp
```diff
@@ -150,6 +150,15 @@ class ASRToLpythonVisitor : public ASR::BaseVisitor<ASRToLpythonVisitor>
             } case ASR::ttypeType::Logical : {
                 r = "bool";
                 break;
+            } case ASR::ttypeType::List : {
+                r = ASRUtils::type_to_str_python(t);
+                break;
+            } case ASR::ttypeType::Tuple : {
+                r = ASRUtils::type_to_str_python(t);
+                break;
+            }case ASR::ttypeType::Set : {
```
nikabot, line 159:
```suggestion
            } case ASR::ttypeType::List : 
              case ASR::ttypeType::Tuple :
              case ASR::ttypeType::Set : {
```

kmr-srbh: Please see https://github.com/lcompilers/lpython/pull/2691#discussion_r1595655576

Shaikh-Ubaid:
> what can be a better name for the test file?

`test_aggregate_constants` or something similar.

kmr-srbh:
Generated code verified using CPython interpreter 3.10.13.

Shaikh-Ubaid:
>Generated code verified using CPython interpreter v3.10.13.

@kmr-srbh the generated code does not work with CPython. There are couple things we will need to fix for it to work.

- call to `__main__global_init()` , `__main__global_stmts()` needs to be added at the end.
- ensure the variables defined in `__main__global_init()` are accessbile to `__main__global_stmts()`.
- and any other issues that we come across while fixing the above two

The above should be tackled in different PR and should not be part of this PR.

kmr-srbh:
> > Generated code verified using CPython interpreter v3.10.13.
> 
> @kmr-srbh the generated code does not work with CPython. There are couple things we will need to fix for it to work.
> 
>     * call to `__main__global_init()` , `__main__global_stmts()` needs to be added at the end.
> 
>     * ensure the variables defined in `__main__global_init()` are accessbile to `__main__global_stmts()`.
> 
>     * and any other issues that we come across while fixing the above two
> 
> 
> The above should be tackled in different PR and should not be part of this PR.

You are right @Shaikh-Ubaid. I had planned to look into this to get the CPython back-end working. A separate PR will be good! :+1:

Shaikh-Ubaid (approved):

PR #2694, opened 2024-05-10
ASR -> CPython: Add `list` method visitors

kmr-srbh:
```python
from lpython import i32, f64

# global scope
my_first_list: list[i32] = [1, 2, 3, 4, 5]
print(my_first_list)

my_first_list.append(4)
my_first_list.remove(2)
print(my_first_list.count(4))
my_first_list.insert(0, 1)
my_first_list.clear()

# local scope
def f():
    my_first_list: list[i32] = [1, 2, 3, 4, 5]
    print(my_first_list)

    my_first_list.append(4)
    my_first_list.remove(2)
    print(my_first_list.count(4))
    my_first_list.insert(0, 1)
    my_first_list.clear()


f()
```
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ ./src/bin/lpython ./examples/example.py --show-python
def __main__global_stmts():
    my_first_list = [1, 2, 3, 4, 5]
    print(my_first_list)
    my_first_list.append(4)
    my_first_list.remove(2)
    print(my_first_list.count(4))
    my_first_list.insert(0, 1)
    my_first_list.clear()
    f()
def f():
    my_first_list: list[i32]
    my_first_list = [1, 2, 3, 4, 5]
    print(my_first_list)
    my_first_list.append(4)
    my_first_list.remove(2)
    print(my_first_list.count(4))
    my_first_list.insert(0, 1)
    my_first_list.clear()
```


kmr-srbh:
@Shaikh-Ubaid could you please look into the anomaly occurring in the generated code when a function call is present inside a function, as in, `print(my_first_list.count(4))`?

src/libasr/codegen/asr_to_python.cpp
```diff
@@ -636,6 +636,52 @@ class ASRToLpythonVisitor : public ASR::BaseVisitor<ASRToLpythonVisitor>
     void visit_SetConstant(const ASR::SetConstant_t &x) {
         visit_AggregateConstant(x.n_elements, x.m_elements, "{", "}");
     }
+
+    // An aggregate visitor for list methods with 0 or 1 argument
+    void visit_UnaryListMethods(ASR::expr_t* list, std::string method_name, ASR::expr_t* arg=nullptr) {
+        std::string r = indent;
```
Shaikh-Ubaid, line 642: A function call can be of two types:
- when it returns some value (so it is like an expression)
- when it does not return value (so it is like a statement) (in this case we call it a subroutine call)

So,
- When the function call is like an expression, it does not need indent and "\n" at the end.
- When the function call is like a statement, it needs the indent and "\n" at the end.

Please update your approach as per the above. Thanks.

kmr-srbh: Thank you very much for the advice @Shaikh-Ubaid! :) I will push the changes in a while. :+1:

Shaikh-Ubaid (approved):
LGTM, thanks!

PR #2695, opened 2024-05-10
Added support for subs attribute

anutosh491:
This allows us to address the following case 
```
from lpython import S
from sympy import Symbol, sin

def main0():
    x: S = Symbol('x')
    y: S = Symbol('y')
    z: S = sin(x)
    s: S = z.subs(x, y)
    print(s)

main0()
```
```
sin(y)
(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython --enable-symengine examples/expr2.py --backend=c
sin(y)
```


certik (approved):
I think that this is fine. Thanks!

Issue #2676, opened 2024-04-30
Bug: Comparing lists using the `==` and `!=` operators leads to an LLVM code generation error


kmr-srbh:
```python
from lpython import i32

my_first_list: list[i32] = [1, 2, 3, 4]
my_second_list: list[i32] = [5, 6, 7, 8]

print(my_first_list == my_second_list)
```

```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ ./src/bin/lpython ./examples/example.py
; ModuleID = 'LFortran'
source_filename = "LFortran"

%list = type { i32, i32, i32* }
.
.
.
declare void @_lpython_call_initial_functions(i32, i8**)

attributes #0 = { argmemonly nounwind willreturn }
asr_to_llvm: module failed verification. Error:
Basic Block in function '__module___main_____main__global_stmts' does not have terminator!
label %.entry

code generation error: asr_to_llvm: module failed verification. Error:
Basic Block in function '__module___main_____main__global_stmts' does not have terminator!
label %.entry



Note: Please report unclear or confusing messages as bugs at
https://github.com/lcompilers/lpython/issues.
```

The issue is the same for `!=`. It does not occur for `>` and `<`.

Shaikh-Ubaid:
```console
% cat examples/expr2.py 
from lpython import i32

def main():
    my_first_list: list[i32] = [1, 2, 3, 4]
    my_second_list: list[i32] = [5, 6, 7, 8]

    print(my_first_list == my_second_list)

main()
% python examples/expr2.py 
False
% lpython examples/expr2.py
False
```

It seems this happens only for global list variables.

kmr-srbh:
Thanks for confirming this @Shaikh-Ubaid!

PR #2697, opened 2024-05-11
Fix LLVM code generation error in List comparsion

hankluo6:
Fixes: #2676

There was an issue where `builder0` and `builder` shared the same insertion point. As a result, `builder0` would append `alloca` instruction after instructions created by `builder`.


Shaikh-Ubaid (approved):
LGTM! Thanks!

PR #2698, opened 2024-05-12
Interactive Shell: Printing top level expression

Vipul-Cariappa:
Example:

Present behavior:
```bash
>>> 1 + 2
>>> 
```

Expected behavior (after potentially merging this PR):
```bash
>>> 1 + 2
3
>>> 
```

---

Reference for myself
Things to do:
- [ ] https://github.com/lcompilers/lpython/pull/2698#discussion_r1597676023
- [ ] https://github.com/lcompilers/lpython/pull/2698#discussion_r1597676269
- [ ] https://github.com/lcompilers/lpython/pull/2617#discussion_r1597487482

### DIFF
--- a/src/libasr/pass/global_stmts.cpp
+++ b/src/libasr/pass/global_stmts.cpp
@@ -91,6 +91,7 @@ void pass_wrap_global_stmts(Allocator &al,
                 target = return_var_ref;
                 idx++;
             } else {
+                // TODO: We will need to add support to return other ASR::ttypeType::*
                 throw LCompilersException("Return type not supported in interactive mode");
             }
             ASR::stmt_t* asr_stmt = ASRUtils::STMT(ASR::make_Assignment_t(al, loc, target, value, nullptr));
@@ -119,7 +120,7 @@ void pass_wrap_global_stmts(Allocator &al,
         /* a_body */ body.p,
         /* n_body */ body.size(),
         /* a_return_var */ (return_var ? return_var_ref : nullptr),
-        (return_var ? ASR::abiType::BindC : ASR::abiType::Source),
+        ASR::abiType::Source,
         ASR::Public, ASR::Implementation,
         nullptr,
         false, false, false, false, false,
--- a/src/lpython/python_evaluator.cpp
+++ b/src/lpython/python_evaluator.cpp
@@ -105,13 +105,48 @@ Result<PythonCompiler::EvalResult> PythonCompiler::evaluate(
     }
 
     bool call_run_fn = false;
+    std::string return_type;
     if (m->get_return_type(run_fn) != "none") {
         call_run_fn = true;
+        return_type = m->get_return_type(run_fn);
     }
 
     e->add_module(std::move(m));
     if (call_run_fn) {
-        e->voidfn(run_fn);
+        if (return_type == "integer4") {
+            int32_t r = e->int32fn(run_fn);
+            result.type = EvalResult::integer4;
+            result.i32 = r;
+        } else if (return_type == "integer8") {
+            int64_t r = e->int64fn(run_fn);
+            result.type = EvalResult::integer8;
+            result.i64 = r;
+        } else if (return_type == "real4") {
+            float r = e->floatfn(run_fn);
+            result.type = EvalResult::real4;
+            result.f32 = r;
+        } else if (return_type == "real8") {
+            double r = e->doublefn(run_fn);
+            result.type = EvalResult::real8;
+            result.f64 = r;
+        } else if (return_type == "complex4") {
+            std::complex<float> r = e->complex4fn(run_fn);
+            result.type = EvalResult::complex4;
+            result.c32.re = r.real();
+            result.c32.im = r.imag();
+        } else if (return_type == "complex8") {
+            std::complex<double> r = e->complex8fn(run_fn);
+            result.type = EvalResult::complex8;
+            result.c64.re = r.real();
+            result.c64.im = r.imag();
+        } else if (return_type == "void") {
+            e->voidfn(run_fn);
+            result.type = EvalResult::statement;
+        } else if (return_type == "none") {
+            result.type = EvalResult::none;
+        } else {
+            throw LCompilersException("FortranEvaluator::evaluate(): Return type not supported");
+        }
     }
 
     if (call_run_fn) {
--- a/src/lpython/semantics/python_ast_to_asr.cpp
+++ b/src/lpython/semantics/python_ast_to_asr.cpp
@@ -4088,8 +4088,8 @@ public:
 
         // Every module goes into a Module_t
         SymbolTable *parent_scope = current_scope;
+        ASR::Module_t* module_sym = nullptr;
         if (parent_scope->get_scope().find(module_name) == parent_scope->get_scope().end()) {
-            ASR::Module_t* module_sym = nullptr;
             current_scope = al.make_new<SymbolTable>(parent_scope);
             ASR::asr_t *tmp1 = ASR::make_Module_t(al, x.base.base.loc,
                                         /* a_symtab */ current_scope,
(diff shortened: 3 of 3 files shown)


src/libasr/pass/global_stmts.cpp
```diff
@@ -91,6 +91,7 @@ void pass_wrap_global_stmts(Allocator &al,
                 target = return_var_ref;
                 idx++;
             } else {
+                // TODO: We will need to add support to return other ASR::ttypeType::*
```
Vipul-Cariappa, line 94: We will need to handle all other cases like `list`, `set`, `dict`, etc.

src/libasr/pass/global_stmts.cpp
```diff
@@ -119,7 +120,7 @@ void pass_wrap_global_stmts(Allocator &al,
         /* a_body */ body.p,
         /* n_body */ body.size(),
         /* a_return_var */ (return_var ? return_var_ref : nullptr),
-        (return_var ? ASR::abiType::BindC : ASR::abiType::Source),
+        ASR::abiType::Source,
```
Vipul-Cariappa, line 123: I have changed this because `asr_to_llvm` changes the final function depending on ABI type (Look at https://github.com/lcompilers/lpython/blob/476d9bb47564f80158fe7e9f9b9124417fbed606/src/libasr/codegen/asr_to_llvm.cpp#L3794-L3805). We don't want the function name to change, it becomes difficult to figure out the correct function name in `PythonCompile.evaluate`.

src/libasr/pass/global_stmts.cpp
Vipul-Cariappa: Tests are going to fail due to this reason.

src/lpython/python_evaluator.cpp
```diff
@@ -105,13 +105,48 @@ Result<PythonCompiler::EvalResult> PythonCompiler::evaluate(
...
+            result.c64.re = r.real();
+            result.c64.im = r.imag();
+        } else if (return_type == "void") {
+            e->voidfn(run_fn);
+            result.type = EvalResult::statement;
+        } else if (return_type == "none") {
+            result.type = EvalResult::none;
+        } else {
+            throw LCompilersException("FortranEvaluator::evaluate(): Return type not supported");
+        }
```
Vipul-Cariappa, line 149: I was thinking we could accept `struct` as the return type here for `dict`, `list`, etc. And handle it accordingly.
But I have no idea how would we print it later.
Example:
```bash
>>> n: list[i32] = [1, 2, 3]
>>> n.append(4)
>>> n
```
We would like to print the list `n` in the last line. But it would be returned as a `struct` here.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -6691,8 +6699,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         // If tmp is a statement and not an expression
         // never cast into expression using ASRUtils::EXPR
         // Just ignore and exit the function naturally.
-        if( tmp && !ASR::is_a<ASR::stmt_t>(*tmp) ) {
-            LCOMPILERS_ASSERT(ASR::is_a<ASR::expr_t>(*tmp));
+        if( tmp && !ASR::is_a<ASR::expr_t>(*tmp) ) {
```
Vipul-Cariappa, line 6702: This is changed to keep track of the final expression executed, and printing its result.

Vipul-Cariappa:
@Shaikh-Ubaid, Please have a look at the comments I have left. Do let me know, if I am not clear.

After having a look at the comments, read the following part:

We can otherwise implement it as an ASR pass. Identify the last expression and wrap the `print` function around it. This ASR pass will only be enabled when using the interactive mode.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4103,21 +4103,29 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+            for (size_t i = 0; i < module_sym->n_dependencies; i++) {
+                if (module_sym->m_dependencies[i]) {
+                    current_module_dependencies.push_back(al, module_sym->m_dependencies[i]);
+                }
+            }
+            module_sym->m_dependencies = current_module_dependencies.p;
+            module_sym->n_dependencies = current_module_dependencies.size();
+            if (!overload_defs.empty()) {
+                create_GenericProcedure(x.base.base.loc);
+            }
```
Shaikh-Ubaid, line 4128: This portion seems to be the same as that of the `if` part. So, let's move this out of the `if-else`?

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: I didn't get this. Could you elaborate this, please?

Shaikh-Ubaid (commented):
I think LFortran already prints values for global expressions. We just need to follow how it does it. 

In LFortran, the pass `pass_wrap_global_stmt` is able to give us a function that can return a value. The same does not hold true for LPython. I think it happens because of the `make_dummy_assignment` used here

https://github.com/lcompilers/lpython/blob/476d9bb47564f80158fe7e9f9b9124417fbed606/src/lpython/semantics/python_ast_to_asr.cpp#L4840-L4850

So, we just need to figure out how to avoid creating the dummy assignment (without breaking any existing tests) and we should have it.

Shaikh-Ubaid:
>We can otherwise implement it as an ASR pass. Identify the last expression and wrap the print function around it. This ASR pass will only be enabled when using the interactive mode.

I clarified this in my review comment above.

Also, you need to update the reference tests for the CI to pass.

src/lpython/python_evaluator.cpp
Vipul-Cariappa: Ok will do it in the next commit.

src/lpython/semantics/python_ast_to_asr.cpp
Vipul-Cariappa: Yes, we should be able to do that. Will do it in the next commit.

src/lpython/semantics/python_ast_to_asr.cpp
Vipul-Cariappa: Please look at the difference in the AST generated with this change and without this change (i.e. this PR and the main branch). It may become easier to understand. Start the interactive shell with `-v` option to print the AST and LLVM IR.

With this change, the `pass_wrap_global_stmt` will generate a function that returns the result of the expression `tmp`, avoiding the modifications you specified with `make_dummy_assignment` (https://github.com/lcompilers/lpython/pull/2698#pullrequestreview-2051417737).

Vipul-Cariappa:
Replying to https://github.com/lcompilers/lpython/pull/2698#pullrequestreview-2051417737

Let's say we made the necessary modifications such that we return a value from the function generated from `pass_wrap_global_stmt`. But what if the return type is `list` or `dict`. We would not be able to print it. Code to print `list` or `dict` should be produced at the LLVM backend level.

Reference:
The function generated by `pass_wrap_global_stmt` is being called by `PythonCompiler::evaluate` (look at https://github.com/lcompilers/lpython/blob/6aed371ba12a2b55e7284483dad651aa3982600e/src/lpython/python_evaluator.cpp#L114-L149). And then being printed by `interactive_python_repl` (Look at https://github.com/lcompilers/lpython/blob/6aed371ba12a2b55e7284483dad651aa3982600e/src/bin/lpython.cpp#L890-L944).

Shaikh-Ubaid:
@Vipul-Cariappa Is this safe to close? If it is, can you close this?

Vipul-Cariappa:
Closed by #2716

Vipul-Cariappa:
Closed by #2716

PR #2699, opened 2024-05-13
Fix logical comparsion for string

### LINKED ISSUE #2615: Bug: Logical operation on `str` variables fail
#1506 claimed to handle logical operations on strings, but as it turns out, things break. #2598 adds support for string literals, but the case of a string variable still fails with a segfault. The chief reason, as pointed out by @Shaikh-Ubaid is that the `expr_value` of the operands, in case of a string variable, result into a `nullptr`.

```python
s_var1: str = "Hello"
s_var2: str = "LPython"

print(s_var1 or s_var2)
```
```console
```

The fix for this issue can help us to enable logical operations on string variables.

hankluo6:
Fixes #2615

The problem is that `_lpython_str_compare_eq` requires pointer to pointer to char as parameters. However, the previous implementation passed a null pointer, leading to a segmentation fault.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -765,8 +765,8 @@ RUN(NAME test_platform       LABELS cpython llvm llvm_jit c)
 RUN(NAME test_vars_01        LABELS cpython llvm llvm_jit)
 RUN(NAME test_version        LABELS cpython llvm llvm_jit)
 RUN(NAME logical_binop1      LABELS cpython llvm llvm_jit)
-RUN(NAME test_logical_compare           LABELS cpython llvm llvm_jit)
-RUN(NAME test_logical_assignment        LABELS cpython llvm llvm_jit)
+RUN(NAME test_logical_compare           LABELS cpython llvm llvm_jit) # TODO: Add C backend after fixing issue #2708
+RUN(NAME test_logical_assignment        LABELS cpython llvm llvm_jit) # TODO: Add C backend after fixing issue #2708
 RUN(NAME vec_01              LABELS cpython llvm llvm_jit c NOFAST)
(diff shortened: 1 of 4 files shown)


kmr-srbh:
Thanks for this @hankluo6! The PR looks good! :+1:

anutosh491 (approved):
I think these changes look good considering the `llvm` backend. So we could have this merged.
But we could do that once we make note of the fact that `test_logical_assignment` & `test_logical_compare` don't work with the `c` backend.
```
def main0():
    s_var1: str = "Hello"
    s_var2: str = "LPython"

    print(s_var1 or s_var2)

(lf) anutosh491@spbhat68:~/lpython/lpython$ lpython examples/expr2.py --backend=c
expr2__tmp__generated__.c: In function ‘main0’:
expr2__tmp__generated__.c:21:14: warning: format ‘%s’ expects argument of type ‘char *’, but argument 2 has type ‘int’ [-Wformat=]
   21 |     printf("%s\n", s_var1 || s_var2);
      |             ~^     ~~~~~~~~~~~~~~~~
      |              |            |
      |              char *       int
      |             %d
Segmentation fault
```

Could you check if we have an issue raised for the same ? If not could you please raise an issue for the same. Once that is done, let's merge this and we could then try addressing the issues through the C backend too !

hankluo6:
Thank you @anutosh491! I raised an issue #2708 for this.

anutosh491:
Could you maybe add a TODO here saying that we can add the C backend once the above issue you raised has been fixed ?
https://github.com/lcompilers/lpython/blob/fbbdbdd83e866a4ebbb7a5541bd7989be60c5fe1/integration_tests/CMakeLists.txt#L768-L769

I guess that would be sufficient enough to merge this !

PR #2700, opened 2024-05-13
`pass_wrap_global_stmts` always creates function of `abiType::BindC`

Vipul-Cariappa:
Reference https://github.com/lcompilers/lpython/pull/2698#discussion_r1597650250


Vipul-Cariappa:
cc @Shaikh-Ubaid

Shaikh-Ubaid:
This is fine. Can you submit this change to LFortran? So that we know it works with LFortran.

Shaikh-Ubaid:
When you make a PR there, just post a link to it here.

Vipul-Cariappa:
Now, WASM related tests are failing.
I will need to look into it.

Shaikh-Ubaid:
>I initially set the ABI to Source. LFortran does not like Source and tests fail there. So, I instead changed it to BindC

I think changing to `Source` is fine. But changing to `BindC` might not be. When we say `BindC`, it means that the function is implemented in `C`. For this function, we only know the prototype of the function and the function body is empty.

I think in our case, we have the function body so marking it as `BindC` only to avoid name mangle does not look right. @Vipul-Cariappa Since the name issue for you arises during interactive compilation, can you trying not mangling name when `compiler_options.interactive = true` in the llvm backend?

Vipul-Cariappa:
That might be possible. Changing `ASRToLLVMVisitor.instantiate_function` to avoid name mangling while `compiler_options.interactive = true` is set.

Vipul-Cariappa:
I guess we can close this as we have merged #2701.

PR #2701, opened 2024-05-14
avoiding name mangling while interactive is true

Vipul-Cariappa:
reference: https://github.com/lcompilers/lpython/pull/2700#issuecomment-2109194716


Vipul-Cariappa:
cc @Shaikh-Ubaid

Shaikh-Ubaid (approved):

PR #2706, opened 2024-05-16
Fix ASR verify pass error while using Interactive

Vipul-Cariappa:
Code to replicate:
```bash
❯ lpython  
>>> print(3 % 2)
1
>>> print(0)
ASR verify pass error: Function __lpython_overloaded_0___lpython_str_center doesn't depend on __lpython_overloaded_2___mod but is found in its dependency list.
  Binary file "/home/vipul/Workspace/python/lpython/src/bin/lpython", in _start()
  Binary file "/lib64/libc.so.6", in __libc_start_main_alias_2()
  Binary file "/lib64/libc.so.6", in __libc_start_call_main()
  File "/home/vipul/Workspace/python/lpython/src/bin/lpython.cpp", line 1990, in main()
    return interactive_python_repl(lpython_pass_manager, compiler_options, arg_v);
  File "/home/vipul/Workspace/python/lpython/src/bin/lpython.cpp", line 852, in interactive_python_repl()
  File "/home/vipul/Workspace/python/lpython/src/lpython/python_evaluator.cpp", line 74, in LCompilers::PythonCompiler::evaluate(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, LCompilers::LocationManager&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&)
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm, true);
  File "/home/vipul/Workspace/python/lpython/src/lpython/python_evaluator.cpp", line 153, in LCompilers::PythonCompiler::get_asr3(LCompilers::LPython::AST::ast_t&, LCompilers::diag::Diagnostics&, LCompilers::LocationManager&, bool)
    compiler_options, true, "__main__", "", false, is_interactive ? eval_count : 0);
LCompilersException: Verify failed
```

---

I am skipping the verification of function with Interactive ABI. In the above example when line 1 is executed all the functions are verified. While executing the second line we call `mark_all_variables_external` where we change the ABI type of all the previously defined/declared functions to Interactive. Therefore we can skip their verification.


Vipul-Cariappa:
cc @Shaikh-Ubaid

src/libasr/asr_verify.cpp
```diff
@@ -420,6 +420,11 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
     }
 
     void visit_Function(const Function_t &x) {
+        ASR::FunctionType_t* v_func_type = ASR::down_cast<ASR::FunctionType_t>(x.m_function_signature);
+        if (v_func_type->m_abi == abiType::Interactive) {
+            // This function would have been verified in the previous interactive pass
+            return;
+        }
```
Shaikh-Ubaid, line 427: Instead of just returning here, let's verify things that we can verify. Like I think for interactive functions we expect the function bodies to be empty, we expect the dependencies to be empty and the other things in that direction.

Vipul-Cariappa:
I found something interesting. Trying this out in interactive:

```bash
❯ lpython
>>> print(3 % 2)
1
>>> 
ASR verify pass error: The Function::m_body should be null if abi set to Interactive
  Binary file "/home/vipul/Workspace/c/lpython/src/bin/lpython", in _start()
  Binary file "/lib64/libc.so.6", in __libc_start_main_alias_2()
  Binary file "/lib64/libc.so.6", in __libc_start_call_main()
  File "/home/vipul/Workspace/c/lpython/src/bin/lpython.cpp", line 1990, in main()
    return interactive_python_repl(lpython_pass_manager, compiler_options, arg_v);
  File "/home/vipul/Workspace/c/lpython/src/bin/lpython.cpp", line 852, in interactive_python_repl()
  File "/home/vipul/Workspace/c/lpython/src/lpython/python_evaluator.cpp", line 92, in LCompilers::PythonCompiler::evaluate(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, LCompilers::LocationManager&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&)
    pass_manager, diagnostics, lm.files.back().in_filename);
  File "/home/vipul/Workspace/c/lpython/src/lpython/python_evaluator.cpp", line 194, in LCompilers::PythonCompiler::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    run_fn, infile);
  File "/home/vipul/Workspace/c/lpython/src/libasr/codegen/asr_to_llvm.cpp", line 9783, in LCompilers::asr_to_llvm(LCompilers::ASR::TranslationUnit_t&, LCompilers::diag::Diagnostics&, llvm::LLVMContext&, Allocator&, LCompilers::PassManager&, LCompilers::CompilerOptions&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    pass_manager.apply_passes(al, &asr, co.po, diagnostics);
  File "/home/vipul/Workspace/c/lpython/src/libasr/pass/pass_manager.h", line 311, in LCompilers::PassManager::apply_passes(Allocator&, LCompilers::ASR::TranslationUnit_t*, LCompilers::PassOptions&, LCompilers::diag::Diagnostics&)
    apply_passes(al, asr, _passes, pass_options, diagnostics);
LCompilersException: Verify failed in the pass: nested_vars
```

Looks like we missed clearing the body of some of the functions.

certik:
I think the code that sets all functions to interactive is incorrect ---- it leaves dependencies in there which should also be removed. The functions just become an interface?

So that code has to be fixed, not removing ASR verify, which now hide the bugs in there. 

The ASR is incorrect, because the dependencies of these functions are incorrect, and consequently other parts of our compiler will fail that assume the dependencies are correct.

In general, we don't fix ASR verify(), unless we are absolutely sure the ASR verification is incorrect.

certik:
P.S. is "Interactive" function always without a body --- an interface? If so, then this check should be added in addition to all the other checks.

src/libasr/asr_verify.cpp
```diff
@@ -420,6 +420,14 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
     }
 
     void visit_Function(const Function_t &x) {
+        ASR::FunctionType_t* v_func_type = ASR::down_cast<ASR::FunctionType_t>(x.m_function_signature);
+        if (v_func_type->m_abi == abiType::Interactive) {
+            require(x.n_body == 0,
+            "The Function::n_body should be 0 if abi set to Interactive");
+            require(x.m_body == nullptr,
+            "The Function::m_body should be null if abi set to Interactive");
+            return;
```
certik, line 429: I think this should be without the return:
```suggestion
```
Then it would be good, it will make the ASR verify stronger.

Shaikh-Ubaid:
>P.S. is "Interactive" function always without a body --- an interface?

Yes, we do that here https://github.com/lfortran/lfortran/blob/d0882b59a1002efccc8532ee235f859271ef658e/src/libasr/asr_scopes.cpp#L39-L45

In the first run for each function, its abi is source. For the next run, previously declared functions are marked as abi interactive and their bodies are removed.

src/libasr/asr_verify.cpp
Shaikh-Ubaid: >I think this should be without the return:

Yes, it should be removed. @Vipul-Cariappa  since you have issues with the dependencies for interactive, how about we make just the dependency check to be optional? Or better we make the dependency check to be having no dependencies?

And we can also update the `mark_all_variables_external()` here https://github.com/lfortran/lfortran/blob/d0882b59a1002efccc8532ee235f859271ef658e/src/libasr/asr_scopes.cpp#L39-L45 to remove the dependencies (along with the body).

certik: You can remove the dependency check in a Draft PR to see if it fixes issues.

However, to merge into master, we need to keep all these checks, they are very important. Instead, we need to just remove all dependencies when we are removing the body. To be exact, we should keep dependencies if they are used in declarations, but removing all of them might get things working pretty well and we can refine it later. The ASR checks I think should all stay, with the addition of the above ones for interactive.

Vipul-Cariappa: > To be exact, we should keep dependencies if they are used in declarations

Do we know of any case where we will require the dependencies of a function during its declaration, but it does not require the body of the function? Also a thing to note is that we would have changed the ABI to interactive.

---

I am not very familiar with dependency, can someone please explain the requirement of the dependency, and how it is used in later parts of the compiling stages? I assume we verify the dependencies by going through the function body and checking if it calls other functions and then checking if that function is in the dependency list.

Vipul-Cariappa:
> In the first run for each function, its abi is source. For the next run, previously declared functions are marked as abi interactive and their bodies are removed.

If you look at this https://github.com/lcompilers/lpython/pull/2706#issuecomment-2120601841.
To me, it looks like there is some kind of edge case where the body is not removed. Any thoughts?

src/libasr/asr_verify.cpp
certik: Yes, for example:
```fortran
interface
  subroutine s(n, x)
  integer, intent(in) :: n
  real, intent(inout) :: x(f(n))
  end subroutine
interface
```
Here `s` depends on the pure function `f`.

Vipul-Cariappa:
Please have a look at the new changes. I have undone the modifications to `asr_verify` and just updated the method in which we clear the body and dependencies of a function.
For some reason, we treat intrinsic functions differently, and need to preserve their body and dependencies for `asr_verify`.

If these changes are acceptable, I can also update lfortran/lfortran#4006 with the same changes. Testing it locally, LLVM tests did pass.

certik:
Let's keep the ASR verify changes, those were good. Just remove "return".

Vipul-Cariappa:
The reason I asked not to merge this yet is that this fix gives an error for the example @certik gave above in the interactive mode.

Code:
```bash
>>> pure function id(x) result(r)                                                                                         ┐
...     integer, intent(in) :: x                                                                                          │
...     integer :: r                                                                                                      │
...     r = x                                                                                                             │
... end function id                                                                                                 5,16  ┘
>>> subroutine sa(n, x)                                                                                                   ┐
...     integer, intent(in) :: n                                                                                          │
...     integer, intent(inout) :: x(id(n))                                                                                │
...                                                                                                                       │
...     integer :: i                                                                                                      │
...     do i = 1, size(x)                                                                                                 │
...         x(i) = x(i) + 1                                                                                               │
...     end do                                                                                                            │
... end subroutine sa                                                                                               9,18  ┘
>>> integer :: i                                                                                                    1,13  ]
ASR verify pass error: Function sa depends on id but isn't found in its dependency list.
 --> input:1:1 - 2:167
  |
1 |    integer :: i
  |    ^^^^^^^^^^^^...
...
  |
2 |    
  | ...^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ failed here


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

We get `ASR verify pass error: Function sa depends on id but isn't found in its dependency list.`, ideally that should not be the behaviour.

---

I was trying the above code out in gfortran, and get the following errors.

```fortran
program main
    implicit none
    integer :: l
    integer :: arr(10)
    integer :: x
    l = 10
    print *, "Before: "
    print *, arr
    call sa(l, arr)
    print *, "After: "
    print *, arr
contains

pure function id(x) result(r)
    integer, intent(in) :: x
    integer :: r
    r = x
end function id

subroutine sa(n, x)
    integer, intent(in) :: n
    integer, intent(inout) :: x(id(n))

    integer :: i
    do i = 1, size(x)
        x(i) = x(i) + 1
    end do
end subroutine sa

end program main
```

gfortran errors:
```bash
❯ gfortran ./e.f90 
./e.f90:22:32:

   22 |     integer, intent(inout) :: x(id(n))
      |                                1
Error: Specification function ‘id’ at (1) cannot be an internal function
./e.f90:22:32:

   22 |     integer, intent(inout) :: x(id(n))
      |                                1
Error: Specification function ‘id’ at (1) cannot be an internal function
./e.f90:22:32:

   22 |     integer, intent(inout) :: x(id(n))
      |                                1
Error: Specification function ‘id’ at (1) cannot be an internal function
```

I am new to fortran, if you could explain the meaning of these error messages and how to fix them, it would be helpful.

Vipul-Cariappa:
The following change looks like the best option to me, if we want to verify the ASR of interactive ABI that has been verified in the previous interactive evaluation.


```diff
index 4fae6739e..759269885 100644
--- a/src/libasr/asr_scopes.cpp
+++ b/src/libasr/asr_scopes.cpp
@@ -40,8 +40,6 @@ void SymbolTable::mark_all_variables_external(Allocator &al) {
                 ASR::Function_t *v = ASR::down_cast<ASR::Function_t>(a.second);
                 ASR::FunctionType_t* v_func_type = ASR::down_cast<ASR::FunctionType_t>(v->m_function_signature);
                 v_func_type->m_abi = ASR::abiType::Interactive;
-                v->m_body = nullptr;
-                v->n_body = 0;
                 break;
             }
             case (ASR::symbolType::Module) : {
```

i.e. Do not remove the body nor the dependencies, only change the ABI to interactive.

> Note: The change I have specified in this comment passes all test cases and the edge cases I have encountered so far (like the above one I mentioned).

Any thoughts/suggestions?

Shaikh-Ubaid:
>We get ASR verify pass error: Function sa depends on id but isn't found in its dependency list., ideally that should not be the behaviour.

We cleared the body of the function, but as per the declaraction/prototype, the function`sa` depends on `id`. If we simply clear out the dependencies, then this error is expected.

To fix this, I think we would need to update the dependencies after we clear the body of the function. For example, @Vipul-Cariappa can you try using `PassUtils::UpdateDependenciesVisitor` after we clear the function body?

Shaikh-Ubaid:
From here https://github.com/lcompilers/lpython/pull/2706#issuecomment-2120601841, for the following

>LCompilersException: Verify failed in the pass: nested_vars
Looks like we missed clearing the body of some of the functions.

Some passes might be adding a new function or redefining the old function, they might need updates/fixes to support interactivity.

Vipul-Cariappa:
Please look at the new changes. If these changes look good, then I can implement the same changes in the LFortran repo and add in Fortran specific tests.

Vipul-Cariappa:
Not sure why the tests fail. Locally they pass.

```bash
❯ ctest
Test project /home/vipul/Workspace/python/lpython
    Start 1: test_stacktrace
1/2 Test #1: test_stacktrace ..................   Passed    0.12 sec
    Start 2: test_lpython
2/2 Test #2: test_lpython .....................   Passed    0.15 sec

100% tests passed, 0 tests failed out of 2

Total Test time (real) =   0.27 sec
```
```bash
❯ ./src/lpython/tests/test_lpython
[doctest] doctest version is "2.4.8"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:  38 |  38 passed | 0 failed | 0 skipped
[doctest] assertions: 192 | 192 passed | 0 failed |
[doctest] Status: SUCCESS!
```

Shaikh-Ubaid:
>Not sure why the tests fail. Locally they pass.

There seems a failure somewhere, we need to debug it and get the CI to pass.

Vipul-Cariappa:
You can see that `./src/lpython/tests/test_lpython` is failing ([reference](https://github.com/lcompilers/lpython/actions/runs/9206229747/job/25323667923?pr=2706#step:10:439)). But in this https://github.com/lcompilers/lpython/pull/2706#issuecomment-2126808126, I have already specified that it is running in my local computer.

Things I tried to replicate:
- Run under `gdb` and `lldb`. Could not replicate the error.
- Run `valgrind`, there were some memory leaks, but I was searching for the use of uninitialized values. There were 2 cases where an uninitialized value was used, but to me, it did not look like it could be the source of the problem.
- Also tested release build. But still could not replicate the error.

I am not of ideas and clues.

One more thing I could try is to open a new PR that only adds the test without any other changes.

Shaikh-Ubaid:
>One more thing I could try is to open a new PR that only adds the test without any other changes.

Go ahead and try this. Let's get the test to work first.

Vipul-Cariappa:
> Go ahead and try this. Let's get the test to work first.

Please look at #2713. Once that has been merged I will try rebasing the main branch and debug the problem.

Vipul-Cariappa:
@Shaikh-Ubaid, should I make these changes in the LFortran repo and add in test cases specific to Fortran and check if the tests pass there?

Shaikh-Ubaid:
>@Shaikh-Ubaid, should I make these changes in the LFortran repo and add in test cases specific to Fortran and check if the tests pass there?

Yes, we should test it. It seems you made a PR here https://github.com/lfortran/lfortran/pull/4086.

Shaikh-Ubaid:
We also need the tests here to pass.

Shaikh-Ubaid:

Vipul-Cariappa:
3 tests from `test_llvm.cpp` fail on Windows. All the 3 tests that fail, use the modulo operator `%`. This causes LPython to import the `lpython_builtin.py`.

Running this locally on Windows, you will see the following error message:
```bash
(lp16) C:\Users\vipul\Documents\Workspace\lpython\build16>.\src\bin\lpython.exe
>>> 3 % 2
warning: The module 'lpython_builtin' located in C:\Users\vipul\Documents\Workspace\lpython\build16\src\bin/../runtime/lpython_builtin.py cannot be loaded
 --> input:1:1
  |
1 | 3 % 2
  | ^^^^^ imported here

syntax error: Token '' (of type 'dedent') is unexpected here
   --> C:\Users\vipul\Documents\Workspace\lpython\build16\src\bin/../runtime/lpython_builtin.py:153:4
    |
153 |     return sum
    |    ^

semantic error: The file 'C:\Users\vipul\Documents\Workspace\lpython\build16\src\bin/../runtime/lpython_builtin.py' failed to parse
 --> input:1:1
  |
1 | 3 % 2
  | ^^^^^


```

This bug is related to #2725

Vipul-Cariappa:
**Explanation behind the [update cmake to copy runtime python files to build dir](https://github.com/lcompilers/lpython/pull/2706/commits/5469ddf974e9fe9348aaab0babe8ac344587423d) commit**

Previous to this commit all tests passed except for `ubuntu-latest`, `macos-latest`, and `windows-2019`. One of the differences between these 3 tests and other tests is that these 3 tests are out-of-source built/compiled. At runtime, they were not able to locate the required runtime Python/LPython files. The commit copies the required runtime files to the build folder from the src folder.

Vipul-Cariappa:
With the latest commits/changes, tests still fail on Windows. But passes in all other environments.

Error ([reference](https://github.com/lcompilers/lpython/actions/runs/9383055462/job/25835834498?pr=2706#step:11:11243)):
```txt
JIT session error: Failed to materialize symbols: { (Main, { __module_lpython_builtin___lpython_overloaded_1__abs, _lcompilers_optimization_floordiv_u64, __module_lpython_builtin___lpython_overloaded_13__complex, _lcompilers_optimization_floordiv_i8, __module_lpython_builtin___lpython_overloaded_7___mod, __module_lpython_builtin___lpython_overloaded_2__round, __module_lpython_builtin___lpython_overloaded_0___mod, __module_lpython_builtin___lpython_overloaded_0__max, __module_lpython_builtin___lpython_overloaded_9___mod, __module_lpython_builtin___lpython_overloaded_3__min, _lcompilers_optimization_floordiv_u8, __module_lpython_builtin___lpython_overloaded_0___floor, __module_lpython_builtin___lpython_overloaded_8__abs, __module_lpython_builtin___lpython_overloaded_1__round, __module_lpython_builtin___lpython_overloaded_0___lpython_str_lstrip, _lcompilers_optimization_floordiv_i328, __module_lpython_builtin___lpython_overloaded_2__sum, __module_lpython_builtin___lpython_overloaded_1___lpython_str_center, __mod
JIT session error: Failed to materialize symbols: { (Main, { __real@3fe0000000000000, __real@40000000, __xmm@80000000000000008000000000000000, __real@3eb0c6f7a0b5ed8d, __real@3f000000, __real@80000000, __real@4000000000000000 }) }
```

The bug is similar to another bug I came across while using REPL on Windows.
```python
(base) PS C:\Users\vipul\Documents\Workspace\lpython\build16> .\src\bin\lpython.exe
>>> def f(x: f64) -> f64:
...   return x
...
>>> f(2.0)
2
>>> def g(x: f64) -> f64:
...   return f(x)
...
>>> g(2.0)
2
>>> def h(x: f64) -> f64:
...   return g(-x)
...
>>> h(2.0)
JIT session error: Failed to materialize symbols: { (Main, { __module___main___h }) }
JIT session error: Failed to materialize symbols: { (Main, { __xmm@80000000000000008000000000000000 }) }
```

---

I request that we wrap the failing tests with `#ifndef _WIN32`, and skip these tests on Windows. I will also open an issue regarding it. We can fix the problem of failing tests on Windows in a separate PR.

CMakeLists.txt
```diff
@@ -120,6 +120,20 @@ if (WITH_LCOMPILERS_FAST_ALLOC)
...
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/runtime/lpython_intrinsic_numpy.py" "${CMAKE_CURRENT_BINARY_DIR}/src/runtime/lpython_intrinsic_numpy.py")
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/runtime/lpython_parser.py" "${CMAKE_CURRENT_BINARY_DIR}/src/runtime/lpython_parser.py")
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/runtime/math.py" "${CMAKE_CURRENT_BINARY_DIR}/src/runtime/math.py")
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/runtime/os.py" "${CMAKE_CURRENT_BINARY_DIR}/src/runtime/os.py")
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/runtime/platform.py" "${CMAKE_CURRENT_BINARY_DIR}/src/runtime/platform.py")
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/runtime/random.py" "${CMAKE_CURRENT_BINARY_DIR}/src/runtime/random.py")
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/runtime/statistics.py" "${CMAKE_CURRENT_BINARY_DIR}/src/runtime/statistics.py")
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/runtime/sys.py" "${CMAKE_CURRENT_BINARY_DIR}/src/runtime/sys.py")
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/runtime/time.py" "${CMAKE_CURRENT_BINARY_DIR}/src/runtime/time.py")
```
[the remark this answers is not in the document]
Vipul-Cariappa: Copying is the safest thing to do. Right now it uses relative path from the executable's location to figure out the runtime file's path.

The reason I say it is a safe option is because the user building the LPython may set the `CMAKE_CURRENT_BINARY_DIR` to a completely different folder. For example, in Windows, I may want to store the source files in `D` drive and the compiled executable in `C` drive.

Vipul-Cariappa:
I have managed to fix the Windows error. Please have a look at the latest changes. I have modifying `KaleidoscopeJIT.h` according to the LLVM's [KaleidoscopeJIT.h](https://github.com/llvm/llvm-project/blob/llvmorg-16.0.6/llvm/examples/Kaleidoscope/include/KaleidoscopeJIT.h) from the 16.0.6 version.

src/runtime/lpython_builtin.py
```diff
@@ -129,7 +129,7 @@ def pow(x: f64, y: i32) -> f64:
 def pow(x: bool, y: bool) -> i32:
     if y and not x:
         return 0
-
+
```
Shaikh-Ubaid, line 132: Let's remove these spaces and instead fix our tokenizer/parser (https://github.com/lcompilers/lpython/issues/2725).

Vipul-Cariappa:
@Shaikh-Ubaid, this is ready to be merged. Please have a look at it.

src/libasr/string_utils.cpp
```diff
@@ -116,7 +116,7 @@ std::string read_file(const std::string &filename)
     std::vector<char> bytes(filesize);
     ifs.read(&bytes[0], filesize);
 
-    return std::string(&bytes[0], filesize);
+    return replace(std::string(&bytes[0], filesize), "\r\n", "\n");
```
Shaikh-Ubaid, line 119: Let's open an issue that our tokenizer/parser should be updated/fixed to produce the same tokens/parse-tree on windows and unix.

Shaikh-Ubaid (approved):
LGTM.

PR #2711, opened 2024-05-23
Add membership checks in dictionaries and sets

advikkabra:
-  Changed `In` and `NotIn` in the AST from compare operators to membership operators
- Implemented membership checks for dictionaries and sets, need to implement for lists, tuples and strings
- Searching could be further optimized, but there is a bug.


integration_tests/test_membership_01.py
```diff
@@ -0,0 +1,36 @@
...
+    assert (1 in a)
+    assert (6 not in a)
+    i = 4
+    assert (i in a)
+
+def test_str_set():
+    a: set[str] = {'a', 'b', 'c', 'e', 'f'}
+    i: str
+    assert ('a' in a)
+    # assert ('d' not in a)
```
czgdp1807, line 29: Try `assert not ('d' in a)`. It should also fail.

czgdp1807 (approved):

integration_tests/test_membership_01.py
czgdp1807: Can be fixed in subsequent PR.

PR #2714, opened 2024-05-24
CI: Upgrade SymEngine to 0.12.0

anutosh491:
This would allow us to use `basic_sign` and `basic_add_as_two_terms` and `basic_mul_as_two_terms` from symengine's C wrapper.

PR #2715, opened 2024-05-24
Implement symbolic sign function

anutosh491:
We need to use `basic_sign` for the `signinf` from the gruntz algorithm. This is a draft PR and tests can be added once #2714 is resolved.


anutosh491:
This is ready to go, enabling auto merge !

PR #2720, opened 2024-06-01
Implementing static methods in classes

tanay-man:


tanay-man:
Test code:
```
from lpython import i32

class Test:
    inside_class : i32 = 5
    def inside_fn() -> i32:
        inside_fn : i32 = 0
        print("In test")
        return 0
```

PR #2733, opened 2024-06-14
Added debug capabilities for lists and sets

advikkabra:
Now lists can be included in code with the `-g` flag.

For example, 
```py
a: list[i32] = [1, 2]
```
The above code can be compiled with `lpython name.py -g --debug-with-line-column`.

Sets can also be included.


czgdp1807 (approved):
Add an example of how to use this flag in the description of this PR. I am approving this but not enabling auto-merge.

PR #2734, opened 2024-06-14
fix array symbol duplication in interactive mode

Vipul-Cariappa:


Vipul-Cariappa:
cc @Shaikh-Ubaid

Shaikh-Ubaid (approved):

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2738,9 +2738,11 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
                 llvm_utils->get_type_from_ttype_t_util(x.m_type, module.get()));
             llvm::Constant *ptr = module->getOrInsertGlobal(x.m_name, array_type);
-            module->getNamedGlobal(x.m_name)->setInitializer(
-                llvm::ConstantStruct::get(array_type,
-                llvm::Constant::getNullValue(array_type)));
+            if (!external) {
+                module->getNamedGlobal(x.m_name)->setInitializer(
+                    llvm::ConstantStruct::get(array_type,
+                    llvm::Constant::getNullValue(array_type)));
+            }
```
Shaikh-Ubaid, line 2745: @Vipul-Cariappa can you add a test that was failing previously in interactive mode and now works?

Vipul-Cariappa: On main you will get the following error:
```bash
>>> i: i32[10]                                                                                                                         1,11  ]
>>> print(i)                                                                                                                           1,9   ]
  Binary file "/home/vipul/Workspace/python/lpython/src/bin/lpython", in _start()
  Binary file "/lib64/libc.so.6", in __libc_start_main_alias_2()
  Binary file "/lib64/libc.so.6", in __libc_start_call_main()
  File "/home/vipul/Workspace/python/lpython/src/bin/lpython.cpp", line 2061, in main()
    return interactive_python_repl(lpython_pass_manager, compiler_options, arg_v);
  File "/home/vipul/Workspace/python/lpython/src/bin/lpython.cpp", line 883, in interactive_python_repl()
  File "/home/vipul/Workspace/python/lpython/src/lpython/python_evaluator.cpp", line 128, in LCompilers::PythonCompiler::evaluate(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, LCompilers::LocationManager&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&)
    e->add_module(std::move(m));
  File "/home/vipul/Workspace/python/lpython/src/libasr/codegen/evaluator.cpp", line 252, in LCompilers::LLVMEvaluator::add_module(std::unique_ptr<LCompilers::LLVMModule, std::default_delete<LCompilers::LLVMModule> >)
    add_module(std::move(m->m_m));
LCompilersException: addModule() returned an error: Duplicate definition of symbol 'i'
```

---

I have added in the test.

Shaikh-Ubaid: The issue shared and fix in this PR might be real, but the test example shared in the above comment does not look right.

```console
(lp) ubaid@ubaids-MacBook-Pro lpython % python examples/expr2.py 
  File "/Users/ubaid/Desktop/OpenSource/lpython/examples/expr2.py", line 4, in <module>
    print(i)
NameError: name 'i' is not defined. Did you mean: 'id'?
```

It does not work with CPython. We should only support tests or cases that work with CPython first, then work with LPython regular/normal mode and finally support them in LPython interactive mode.

src/lpython/tests/test_llvm.cpp
```diff
@@ -1450,6 +1450,23 @@ def my_concat(x: str, y: str) -> str:
...
+TEST_CASE("PythonCompiler Array 1") {
+    CompilerOptions cu;
+    cu.po.disable_main = true;
+    cu.emit_debug_line_column = false;
+    cu.generate_object_code = false;
+    cu.interactive = true;
+    cu.po.runtime_library_dir = LCompilers::LPython::get_runtime_library_dir();
+    PythonCompiler e(cu);
+    LCompilers::Result<PythonCompiler::EvalResult>
+    r = e.evaluate2("i: i32[10]");
```
Shaikh-Ubaid, line 1462: In this case, I think we should have it like

```console
>>> i: i32[10] = empty(10, dtype=int32)                                                          1,36  ]
>>> print(i)  
```

Shaikh-Ubaid (approved):

PR #2744, opened 2024-06-22
feat: add a pixi file and split into two envs (host & build)

wolfv:
I was trying to do some development on `lpython`. 

It might simplify the first steps for some people :) The lockfile should create a consistent environment. I only have tested this on macOS arm64 so far though.

We create two environments:

- `compiler`: contains `cmake`, `ninja`, and the C++ compiler
- `host`: contains `llvmdev` and `zlib` (and whatever else might be linked)

A `start` task is also added to run the development version of lpython. I was pretty confused for a while which `runtime` folder to include so I also added that to the `start` command (and changed `build1.sh`).

I can fix this up for linux and windows if you all like it.


build1.sh
```diff
@@ -3,15 +3,36 @@
...
+   if [[ ! -f src/libasr/asr.h ]]; then
+       pixi run build_0
+   fi
+fi
+
+# if the outoftree env var is set to true, create a build directory
+if [[ "$OUTOFTREE" == "true" ]]; then
+    mkdir -p build
+    cd build
+fi
```
certik, line 23: Is this needed for `pixi build`? If so, why not just move this logic to `pixi.toml` itself? That way this simple script can stay build system independent.

certik (commented):
Let's remove `pixi.lock` for now, because it's huge and will get even bigger for all platforms. This can always be added later.

The move from make to Ninja might also break a few things, but we can fix those as we discover them.

Otherwise good.

PR #2746, opened 2024-06-23
add string.py with constants

wolfv:
These are a few constants copied from `cpython` `Lib/string.py`.


certik:
I think this looks great. Let's add some tests to ensure it actually works?

integration_tests/CMakeLists.txt
```diff
@@ -532,6 +532,7 @@ RUN(NAME test_str_02         LABELS cpython llvm llvm_jit c)
 RUN(NAME test_str_03         LABELS cpython llvm llvm_jit c)
 RUN(NAME test_str_04         LABELS cpython llvm llvm_jit c wasm)
 RUN(NAME test_str_05         LABELS cpython llvm llvm_jit c)
+RUN(NAME test_string_01      LABELS cpython llvm llvm_jit c)
```
certik, line 535: Either rename this to `test_string`, or (rather) rename the file `test_string.py` to `test_string_01.py`.

certik (approved):

certik (approved):
I think this is good, thanks!

integration_tests/test_string_01.py
```diff
@@ -0,0 +1,7 @@
+from string import ascii_lowercase, ascii_letters
+
+def test_string():
```
namannimmo10, line 3: You need to call the function below ;)

wolfv: ouch!

wolfv: fixed in #2788

namannimmo10: Perhaps one can write a custom flake8 plugin using https://pypi.org/project/flake8-plugin-utils/ that checks all functions that are defined, are being called.

wolfv: I don't think it has to be custom. You could probably use `ruff`, too :)

certik: LPython itself should give a warning for something like this.

PR #2752, opened 2024-07-02
Support to print lists in REPL

Vipul-Cariappa:

PR #2755, opened 2024-07-06
Underscore: `_` variable in REPL

Vipul-Cariappa:
Demo:
```python
>>> 5
5
>>> _
5
>>> _ ** _
3125
>>> 2.5
2.5
>>> _
2.5
>>> _ ** _
9.8821176880261863
```

---

This PR is built on top of #2752. Therefore #2752 should be merged first.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -205,6 +205,8 @@ public:
     std::map<llvm::Value*, llvm::Value*> strings_to_be_allocated; // (array, size)
     Vec<llvm::Value*> strings_to_be_deallocated;
 
+    uint32_t global_underscore_hash; // used in interactive mode
+
     ASRToLLVMVisitor(Allocator &al, llvm::LLVMContext &context, std::string infile,
         CompilerOptions &compiler_options_, diag::Diagnostics &diagnostics) :
     diag{diagnostics},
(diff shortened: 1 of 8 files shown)


Vipul-Cariappa:
cc @Shaikh-Ubaid

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2709,11 +2712,20 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
     }
 
     void visit_Variable(const ASR::Variable_t &x) {
+        if ((compiler_options.interactive) &&
```
Shaikh-Ubaid, line 2715:
```suggestion
        if (compiler_options.interactive && 
```
Similarly others.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2709,11 +2712,20 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
     }
 
     void visit_Variable(const ASR::Variable_t &x) {
+        if ((compiler_options.interactive) && 
+            (std::strcmp(x.m_name, "_") == 0) &&
```
Shaikh-Ubaid, line 2716: Why is it so that at some place we compare with `(std::strcmp(x.m_name, "_") == 0)` and at some we use `(std::strcmp(x.m_name, compiler_options.po.global_underscore.c_str()) == 0)`? What is the significance of `compiler_options.po.global_underscore`?

Shaikh-Ubaid:
Looks good overall. Lets merge https://github.com/lcompilers/lpython/pull/2752 and rebase and resolve conflicts in this PR. After that I will have a look again.

src/libasr/codegen/asr_to_llvm.cpp
Vipul-Cariappa: The previous code block's last expression's value is stored in `compiler_options.po.global_underscore`, which is `___main__global_stmts_<int>__`.
While generating IR for `compiler_options.po.global_underscore`, we store its hash in `global_underscore_hash`, then whenever we need a reference to `_` variable, we use `global_underscore_hash`. So `_` gets replaced with `___main__global_stmts_<int>__`.

Shaikh-Ubaid (approved):

PR #2756, opened 2024-07-06
Implement Jupyter kernel

Vipul-Cariappa:


.github/workflows/CI.yml
```diff
@@ -46,12 +46,12 @@ jobs:
...
         if: contains(matrix.os, 'windows')
         shell: bash -e -l {0}
-        run: conda install m2-bison=3.0.4 cmake=3.21.1
+        run: conda install m2-bison=3.0.4 cmake=3.21.1 xeus xeus-zmq
 
       - name: Install Linux / macOS Conda Packages
         if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'macos')
         shell: bash -e -l {0}
-        run: conda install bison=3.4 nodejs=18
+        run: conda install bison=3.4 nodejs=18 xeus xeus-zmq
```
certik, line 54: you can remove these

certik:
In default build, do not build the xeus kernel. So you need to hide it behind ifdefs. Look at how LFortran does it.

Then at the CI, just add a new job, as you already did, and build it there. That's done. 

The other CI jobs should work exactly as before, and since xeus is not installed, they won't be testing it.

src/bin/lpython.cpp
```diff
@@ -29,6 +29,7 @@
 #include <libasr/config.h>
 #include <libasr/string_utils.h>
 #include <libasr/lsp_interface.h>
+#include <lpython/python_kernel.h>
```
certik, line 32: You can probably just put this behind `#ifdef HAVE_LFORTRAN_XEUS` and that should do it for now. We can later refactor it differently.

certik:
It almost works. You can investigate the macOS failure. We need to update CI to use micromamba in a separate PR, which I am doing in https://github.com/lcompilers/lpython/pull/2758. Then we rebase this one.

ci/environment.yml
```diff
@@ -7,7 +7,8 @@ dependencies:
   - toml
   - pytest
   - jupyter
-  - xeus=1.0.1
+  - xeus=5.1.0
+  - xeus-zmq=3.0.0
   - xtl
   - nlohmann_json
   - cppzmq
```
anutosh491, line 14: Migration to `xeus` 5 and above allows us to drop the dependency on `xtl` and `cppzmq` for all downstream projects (mostly kernels) so you have remove them and also we should pin `nlohmann_json` to `3.11.3` wherever possible

certik:

PR #2761, opened 2024-07-07
Parameterised init and methods

tanay-man:
Enabled generation of parameterised init and methods in the ASR. Adding self after symtab visit in each function symtabs so as to have the correct type of the class.

TODO:
- [x]  Enable function code gen in the ASR->LLVM
- [ ] Call __init__ instead of the StructConstructor
- [ ] Modify the method calls to replace self with the calling object 
@certik @Thirumalai-Shaktivel

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -3054,6 +3054,30 @@ public:
         }
     }
 
+    void instantiate_methods(const ASR::Struct_t &x) {
+        SymbolTable *current_scope_copy = current_scope;
+        current_scope = x.m_symtab;
+        for ( auto &item : x.m_symtab->get_scope() ) {
+            if ( is_a<ASR::Function_t>(*item.second) ) {
+                ASR::Function_t *v = down_cast<ASR::Function_t>(item.second);
+                instantiate_function(*v);
+            }
+        }
+        current_scope = current_scope_copy;
+    }
+
+    void visit_methods (const ASR::Struct_t &x) {
+        SymbolTable *current_scope_copy = current_scope;
+        current_scope = x.m_symtab;
+        for ( auto &item : x.m_symtab->get_scope() ) {
+            if ( is_a<ASR::Function_t>(*item.second) ) {
+                ASR::Function_t *v = down_cast<ASR::Function_t>(item.second);
+                visit_Function(*v);
+            }
+        }
+        current_scope = current_scope_copy;
+    }
+
     void start_module_init_function_prototype(const ASR::Module_t &x) {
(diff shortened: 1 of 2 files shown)


Thirumalai-Shaktivel:
We need to translate to the following:
```fortran
! Fortran code after applying the pass: implied_do_loops
module __main__
implicit none
type :: coord
    integer(4) :: x
    integer(4) :: y
end type coord

contains

subroutine __init__(self, x, y)
class(coord), intent(inout) :: self
integer, intent(in) :: x, y
self%x = x
self%y = y
end subroutine

subroutine __main__global_stmts()
    type(coord) :: p1
    ! p1 =     coord(6, 4)
    call __init__(p1, 6, 4)
    p1%x = 10
    print *, p1%x
    print *, p1%y
end subroutine __main__global_stmts

end module __main__

program main_program
use __main__, only: __main__global_stmts
implicit none
call __main__global_stmts()
end program main_program
```

Thirumalai-Shaktivel:
Passing the argument:
```py
class test:
	def __init__(self: "coord"):
		self.x = 3
		self.y = 4

t: test = test()
t.x = 10
print(t.x)
print(t.y)
```

Fortran
```fortran
! Fortran code after applying the pass: implied_do_loops
module __main__
implicit none
type :: coord
    integer(4) :: x
    integer(4) :: y
end type coord

contains

subroutine __init__(self, x, y)
class(coord), intent(inout) :: self
integer, intent(in) :: x, y
self%x = x
self%y = y
end subroutine

subroutine __main__global_stmts()
    type(coord) :: p1
    ! p1 = coord(3, 4)
    call p1%__init__(3, 4)
    p1%x = 10
    print *, p1%x
    print *, p1%y
end subroutine __main__global_stmts

end module __main__

program main_program
use __main__, only: __main__global_stmts
implicit none
call __main__global_stmts()
end program main_program
```

Thirumalai-Shaktivel:
Simple fortran example, check the ASR:
```fortran
module mod
type :: t
contains
    procedure :: f
end type

contains

subroutine f(self)
class(t), intent(inout):: self
end subroutine

end module

program expr2
use mod
implicit none

type(t) :: t_
call t_%f()

end program
```

Thirumalai-Shaktivel:
How to handle:
```py
from lpython import i32,i64,f32

class coord:
    # TypeParameter("coord") => ClassType(coord)
    def __init__(self: "coord"):
        self.x: i32 = 3
        self.y: i32 = 4

p1: coord = coord()

# p1: coord
# ExternalSymbol p1_init => __init__
# p1_init(p1)

p1.x = 10
print(p1.x)
print(p1.y)
```

Issue #1836, opened 2023-05-18
lpython_bit_length() fails for negative numbers in release mode


Shaikh-Ubaid:
On latest main, compiling using the following commands on Mac M1
```bash
./build0.sh
cmake -DWITH_FMT=yes -DCMAKE_CXX_FLAGS_RELEASE="-Wall -Wextra -O3 -funroll-loops -DNDEBUG" -DWITH_LLVM=yes .
cmake --build . -j16
```
I am experiencing the following:
```bash
$ python integration_tests/test_bit_length.py               
7
3
7
7
$ lpython integration_tests/test_bit_length.py               
AssertionError
```

_Originally posted by @Shaikh-Ubaid in https://github.com/lcompilers/lpython/issues/1835#issuecomment-1553425715_

certik:
Awesome, I wonder if it has something to do with compiling in optimized mode, which is what the Conda package uses.

Shaikh-Ubaid:
I would like to know if this issue is a general case or if it is something experienced specifically my environment setup. Dr. Ondrej, please, could you share if you are able to reproduce this?

Vipul-Cariappa:
`bit_lenght` does not work properly in interactive mode:

```python
>>> (12).bit_length()
4
>>> u8(12).bit_length()
>>> i8(12).bit_length()
>>> x: i32 = i8(12).bit_length()
>>> x
0
```

PR #2762, opened 2024-07-08
Ci mamba fix

Vipul-Cariappa:
Builds on top of #2758


Vipul-Cariappa:
@certik, as I can't write to that branch, I will be using this branch.

Vipul-Cariappa:
I had to skip the `test_bit_length` integration test from `llvm` and `llvm_jit`. I have commented about it on an issue related to `bit_length`. see https://github.com/lcompilers/lpython/issues/1836#issuecomment-2213626743.

Vipul-Cariappa:
@certik, this might be a simpler fix than #2758.
Only had to skip `test_bit_length` integrations test, compared to skipping both `test_llvm` and the integration test.
And we need not pin every single package. I have only pined the necessary packages, and most probably conda will install the newest available versions of the other packages.

certik:

PR #2764, opened 2024-07-08
Jupyter Kernel

Vipul-Cariappa:
#2756

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -4,20 +4,19 @@ channels:
   - defaults
 dependencies:
   - llvmdev=11.1.0
-  - toml=0.10.2
-  - pytest=7.2.0
-  - jupyter=1.0.0
-  - xeus=1.0.1
-  - xtl=0.7.4
-  - nlohmann_json=3.9.1
-  - cppzmq=4.7.1
-  - jupyter_kernel_test=0.4.4
+  - xeus=5.1.0
+  - xeus-zmq=3.0.0
   - xonsh=0.13.3
-  - re2c=2.2
-  - numpy=1.23.4
+  - rapidjson
+  - nlohmann_json
+  - toml
+  - pytest
+  - jupyter
+  - jupyter_kernel_test
+  - re2c
+  - numpy
   - zlib
   - zstd
-  - ninja=1.11.0
-  - rapidjson=1.1.0
+  - ninja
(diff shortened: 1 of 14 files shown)


Vipul-Cariappa:
@certik, I have squash merged the changes make in #2762 and `cherry-pick`ed the necessery changes from #2756.

src/lpython/python_evaluator.cpp
```diff
@@ -156,11 +157,11 @@ Result<PythonCompiler::EvalResult> PythonCompiler::evaluate(
                                     ->m_symtab->get_symbol(run_fn);
             LCOMPILERS_ASSERT(fn)
             if (ASRUtils::get_FunctionType(fn)->m_return_var_type->type == ASR::ttypeType::UnsignedInteger) {
-                uint16_t r = e->execfn<uint16_t>(run_fn);
+                uint16_t r = e->execfn<int>(run_fn);
                 result.type = EvalResult::unsignedInteger2;
                 result.u32 = r;
             } else {
-                int16_t r = e->execfn<int16_t>(run_fn);
+                int16_t r = e->execfn<int>(run_fn);
```
certik, line 164: Do you know why the previous code was incorrect?

certik:
Looks like there is some issue with the bit_length implementation (https://github.com/lcompilers/lpython/issues/1836), so it's fine to disable the test. The other macOS interactive issue was fixed. I don't understand exactly the problem, but it looks like something with signed/unsigned. The change in this PR works on all platforms, so it gets us unblocked.

Let's merge it, and we can fix the rest of the issues later.

ci/environment.yml
```diff
@@ -4,20 +4,19 @@ channels:
...
+  - pytest
+  - jupyter
+  - jupyter_kernel_test
+  - re2c
+  - numpy
   - zlib
   - zstd
-  - ninja=1.11.0
-  - rapidjson=1.1.0
+  - ninja
```
certik, line 20: We have to put exact version numbers here.

certik:
I am merged this as is, since all tests pass. Now I'll submit a PR to fix the package versions.

ci/environment.yml
certik: Fixed in https://github.com/lcompilers/lpython/pull/2765.

PR #2779, opened 2024-07-19
Add `set()` for creating an empty set

advikkabra:
Currently this only supports an empty set, but should support all iterables in the future.

### DIFF
--- a/integration_tests/test_set_constructor.py
+++ b/integration_tests/test_set_constructor.py
@@ -1,0 +1,15 @@
+def test_empty_set():
+    a: set[i32] = set()
+    assert len(a) == 0
+    a.add(2)
+    a.remove(2)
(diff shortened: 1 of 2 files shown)


integration_tests/test_set_constructor.py
```diff
@@ -0,0 +1,15 @@
...
+    a.add(3)
+    assert a.pop() == 3
+
+    b: set[str] = set()
+
+    assert len(b) == 0
+    b.add('a')
+    b.remove('a')
+    b.add('b')
+    assert b.pop() == 3
```
czgdp1807, line 15: I think you forgot to call the function. :-).

czgdp1807 (changes requested):

czgdp1807 (approved):

PR #2782, opened 2024-07-20
class_test_3

### LINKED ISSUE #2781: Method calling doesn't work inside methods
Ex:
```py
from lpython import f64
from math import pi

class Circle:
    def __init__(self:"Circle", radius:f64):
        self.radius :f64 = radius
    
    def circle_area(self:"Circle")->f64:
        return pi * self.radius ** 2.0
    
    def circle_print(self:"Circle"):
        area : f64 = self.circle_area()
        print("Circle: r = ",str(self.radius)," area = ",str(area))

def main():
    c : Circle = Circle(1.0)  
    c.circle_print() 
    c.radius = 1.5   
    c.circle_print()

main()

```
This results in a SEGFAULT as the `self` variable is not initialised before function call. I think atleast the `class_type` should be set to the parent class.
@Thirumalai-Shaktivel

tanay-man:
fixes #2781


tanay-man:
@Thirumalai-Shaktivel can you please review?

Thirumalai-Shaktivel (commented):
Otherwise, LGTM

Thirumalai-Shaktivel (approved):
LGTM

PR #2796, opened 2024-08-02
BindPython ABI

Vipul-Cariappa:


certik:
The right solution here is to have an ASR->ASR pass that takes the `bind(Python)` and replaces it with calls to Python C/API. This should not be hardwired into all the backends.

If you think you can get something working with relatively short code, you can finish the PR, but I think having a pass is a better solution long term.

Vipul-Cariappa:
> The right solution here is to have an ASR->ASR pass that takes the bind(Python) and replaces it with calls to Python C/API. This should not be hardwired into all the backends.

Yes. I am implementing this using an ASR to ASR pass. The `python_bind.py` implements the ASR to ASR pass.

src/lpython/semantics/python_ast_to_asr.cpp
```diff
@@ -4428,6 +4429,55 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+            (AST::ImportFrom_t*)AST::make_ImportFrom_t(
+                al, x.base.base.loc,
+                (char*)"cpython_bindings", module_symbols,
+                pybind_funcs.size(), 0);
+        
+        SymbolTable *function_scope = current_scope;
+        current_scope = parent_scope;
+        visit_ImportFrom(*imports);
+        current_scope = function_scope;
+    }
```
certik, line 4479: Why are these needed in AST->ASR?

Ideally the initial ASR has nothing to do with Python C/API. Then the ASR->ASR pass translates `bind(python)` into C/API calls in ASR.

certik:
Otherwise it looks like the right approach, it seems you are doing the type translation to/from Python in the ASR->ASR pass already, which is perfect.

src/libasr/pass/python_bind.cpp
```diff
@@ -0,0 +1,406 @@
...
+    Vec<ASR::call_arg_t> args_Py_DecRef3;
+    args_Py_DecRef3.reserve(al, 1);
+    args_Py_DecRef3.push_back(al, {f.base.base.loc, pReturn_ref});
+    body.push_back(al, ASRUtils::STMT(ASRUtils::make_SubroutineCall_t_util(al, f.base.base.loc, sym_Py_DecRef, nullptr,
+                        args_Py_DecRef3.p, args_Py_DecRef3.n, nullptr, nullptr, false, false)));
+
+    // reassignment
+    f.m_body = body.p;
+    f.n_body = body.n;
+    ASRUtils::get_FunctionType(f)->m_abi = ASR::abiType::BindC;
```
Shaikh-Ubaid, line 360: Any reason to make the `abi` as `BindC`? I think we should be making it `Source`?

src/libasr/pass/python_bind.cpp
```diff
@@ -0,0 +1,406 @@
...
+    ASR::ttype_t *f8_type = ASRUtils::TYPE(ASR::make_Real_t(al, f.base.base.loc, 8));
+    ASR::ttype_t *ptr_t = ASRUtils::TYPE(ASR::make_CPtr_t(al, f.base.base.loc));
+
+    ASR::expr_t *conv_result = nullptr;
+    if (type->type == ASR::ttypeType::Integer) {
+        ASR::symbol_t *sym_PyLong_AsLongLong = parent_scope.get_symbol("PyLong_AsLongLong");
+        Vec<ASR::call_arg_t> args_PyLong_AsLongLong;
+        args_PyLong_AsLongLong.reserve(al, 1);
+        args_PyLong_AsLongLong.push_back(al, {f.base.base.loc, exp});
+        conv_result = ASRUtils::EXPR(ASR::make_Cast_t(al, f.base.base.loc,
```
Shaikh-Ubaid, line 27: Why do we need to do `Cast_t` here (and the rest of the types below)? Isn't the returned value by the specific function (in this case `PyLong_AsLongLong`) already of the desired type?

src/lpython/semantics/python_ast_to_asr.cpp
Shaikh-Ubaid: Exactly, even I was confused in my last meet with Vipul. Looking into the changes, I think I understand what he is doing here.

He has a new runtime module `cpython_bindings.py` which contains the declarations to the `C` functions in `Python.h` that we need. In the above change, he is just creating `AST` code for the following:

```
from cpython_bindings import Py_Initialize, Py_IsInitialized, PyRun_SimpleString, Py_DecodeLocale, PySys_SetArgv, Py_FinalizeEx, PyUnicode_FromString, PyUnicode_AsUTF8AndSize, PyImport_Import, Py_DecRef, Py_IncRef, PyObject_GetAttrString, PyTuple_New, PyTuple_SetItem, PyObject_CallObject, PyLong_AsLongLong, PyLong_AsUnsignedLongLong, PyLong_FromLongLong, PyLong_FromUnsignedLongLong, PyFloat_FromDouble, PyFloat_AsDouble, PyBool_FromLong, PyObject_IsTrue
```

When he visits the `AST` code, `lpython` reads/loads the `runtime/cpython_bindings.py` and automatically generates the declarations for the above `BindC` functions.

@Vipul-Cariappa  Initially this approach sounded interesting to me, but now after going through the changes, it seems like we should instead write the ASR code for these directly in the `python_bind` pass. I think if we simply extend the current `pybind_funcs` array with the parameter types and return type info (for example specifying types like here https://github.com/lcompilers/lpython/blob/9374feb17362d8046db1e4060d633f7ff9887100/src/libasr/codegen/asr_to_wasm.cpp#L206), we should be able to run a loop and generate the ASR nodes for declaring these. Then, we ideally would not be making any changes in the AST->ASR level. And we can also remove the `runtime/cpython_bindings.py` file.

Shaikh-Ubaid (commented):
Great work, Vipul! Shared some comments. Rest looks good.

src/libasr/pass/python_bind.cpp
```diff
@@ -0,0 +1,406 @@
...
+                                                        f.base.base.loc, 0, i4_type)), i4_type, nullptr);
+    Vec<ASR::stmt_t*> if_body;
+    if_body.reserve(al, 2);
+    ASR::asr_t *call_Py_Initialize = ASRUtils::make_SubroutineCall_t_util(al, f.base.base.loc, sym_Py_Initialize,
+                                        nullptr, nullptr, 0, nullptr, nullptr, false, false);
+    if_body.push_back(al, ASRUtils::STMT(call_Py_Initialize));
+
+    ASR::symbol_t *sym_PyRun_SimpleString = parent_scope.get_symbol("PyRun_SimpleString");
+    Vec<ASR::call_arg_t> args_PyRun_SimpleString;
+    s.from_str(al, "import sys; sys.path.append(\".\")");
```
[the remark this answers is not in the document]
Vipul-Cariappa: Yes, the `bindpy_05.py` example does not require this. But it is required for other tests example `bindpy_01.py`. 

We append the current path to CPython's module search path. If we do not do this, CPython will fail with `ModuleNotFoundError`, because it only looks for modules installed using pip or the builtins/CPython standard library.

src/libasr/pass/python_bind.cpp
Vipul-Cariappa: The signature of `PyLong_AsLongLong` is
`long long PyLong_AsLongLong(PyObject*)`. i.e. 
`i64 PyLong_AsLongLong(CPtr)`.

Python does not have `PyLongAsInt` that returns an `int` (`i32`).

The `Cast_t` is required in the following situation.
Assume the following PythonBind function
```
@pythoncall(module="mypylib")
def f(a: i32) -> i32:
    pass
```
or
```
@pythoncall(module="mypylib")
def f(a: i8) -> i8:
    pass
```
The `Cast_t` casts the result of `PyLong_AsLongLong` from `i64` to `i32` or `i8` according to the actual type required.

src/libasr/pass/python_bind.cpp
Vipul-Cariappa: Ok. Will change it to Source in the next commit.

src/lpython/semantics/python_ast_to_asr.cpp
Vipul-Cariappa: Ok. I will update the code to generate the required function declarations in the `python_bind` pass.

src/libasr/pass/python_bind.cpp
Shaikh-Ubaid: Makes sense. Thanks.

src/libasr/pass/python_bind.cpp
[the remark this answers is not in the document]
Shaikh-Ubaid: I think https://github.com/lcompilers/lpython/blob/9374feb17362d8046db1e4060d633f7ff9887100/integration_tests/bindpy_05.py#L72-L73 should help. I believe it adds the current directory `.` to the path.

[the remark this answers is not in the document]
Vipul-Cariappa: I can try this.

What does `pointer(argv1, i64)` do? What does the `pointer` function do?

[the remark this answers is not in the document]
Vipul-Cariappa: I figured it out.

I am no longer using `PyRun_SimpleString` and appending the path. I am using `Py_DecodeLocale` and `PySys_SetArgv`.

src/libasr/pass/python_bind.cpp
```diff
@@ -0,0 +1,567 @@
...
+                                    ASRUtils::intent_return_var);
+        char *fn_name = s.c_str(al);
+        ASR::asr_t *f = ASRUtils::make_Function_t_util(al, *l, current_scope, fn_name, nullptr, 0, args.p, args.n,
+                                    nullptr, 0, retval, ASR::abiType::BindC, ASR::accessType::Public,
+                                    ASR::deftypeType::Interface, nullptr, false, false, false, false, false, nullptr, 0,
+                                    false, false, false, c_header);
+
+        parent_scope->add_symbol(i.m_name, ASR::down_cast<ASR::symbol_t>(f));
+    }
+}
```
Vipul-Cariappa, line 489: I was thinking about moving these functions to `asr_utils.cpp`.
These are generic functions that can be used to create new `BindC` `Interface` function declarations.

Any thoughts? @certik @Shaikh-Ubaid.

Shaikh-Ubaid: Sounds good to me. Also see/consider if `pass_utils` can be a better option.

Vipul-Cariappa: I feel `asr_utils.cpp` is the better option. In the future, if we need to generate function declarations for using any C function, we could just use these functions. Putting it in `asr_utils.cpp` will let us use it in the AST to ASR step also (if necessary).

Vipul-Cariappa:
@Shaikh-Ubaid, I believe if the CI passes and these changes look good to you, then we can merge.

src/libasr/pass/python_bind.cpp
```diff
@@ -0,0 +1,459 @@
...
+    fns.push_back({"PyLong_AsLongLong", {ASRUtils::PTR}, ASRUtils::I64});
+    fns.push_back({"PyLong_AsUnsignedLongLong", {ASRUtils::PTR}, ASRUtils::U64});
+    fns.push_back({"PyLong_FromLongLong", {ASRUtils::I64}, ASRUtils::PTR});
+    fns.push_back({"PyLong_FromUnsignedLongLong", {ASRUtils::U64}, ASRUtils::PTR});
+    fns.push_back({"PyFloat_FromDouble", {ASRUtils::F64}, ASRUtils::PTR});
+    fns.push_back({"PyFloat_AsDouble", {ASRUtils::PTR}, ASRUtils::F64});
+    fns.push_back({"PyBool_FromLong", {ASRUtils::I32}, ASRUtils::PTR});
+    fns.push_back({"PyObject_IsTrue", {ASRUtils::PTR}, ASRUtils::I32});
+
+    bool included_cpython_funcs = false;
```
Shaikh-Ubaid, line 415: I think `included_cpython_funcs` is not needed. We can simply call ` ASRUtils::declare_functions(al, fns, *l, unit.m_symtab, "Python.h")` here. We skip the whole `python_bind` pass if `--enable-cpython` is not provided.


Also suggested here https://github.com/lcompilers/lpython/pull/2796#discussion_r1704413641.

src/libasr/pass/python_bind.cpp
```diff
@@ -0,0 +1,459 @@
...
+        }
+
+    }
+
+    if (included_cpython_funcs) {
+        PassUtils::UpdateDependenciesVisitor u(al);
+        u.visit_TranslationUnit(unit);
+    }
+}
+}
```
Shaikh-Ubaid, line 459: Minor: 
```suggestion

}  // namespace LCompilers
```

Shaikh-Ubaid (approved):
Shared one comment here https://github.com/lcompilers/lpython/pull/2796/files#r1713010953. I think its good to merge after that. Thanks for this @Vipul-Cariappa!

src/lpython/semantics/python_ast_to_asr.cpp
certik: > Initially this approach sounded interesting to me, but now after going through the changes, it seems like we should instead write the ASR code for these directly in the `python_bind` pass.

Yes, exactly.

certik (approved):
Great job. After @Shaikh-Ubaid is happy, this can be merged. It looks clean and small, the pass has less than 460 lines and it does everything. I think this is the way to go.

Is this now used by default for all Python bindings, meaning all our existing tests pass with this new approach?

src/libasr/pass/python_bind.cpp
Vipul-Cariappa: The `--enable-cpython` flag sets `CompilerOptions.enable_cpython` to true. We do not have access to `CompilerOptions` within an ASR pass. We only have access to `PassOptions`.
Should I move `enable_cpython` to `PassOptions` then? Or have it in both places in `CompilerOptions` and `PassOptions`, so that we don't break any other code.

Vipul-Cariappa:
@certik
The tests that use arrays (i.e. call into a CPython function that accepts arrays or returns an array) are skipped.
Conversion from LPython arrays to CPython arrays (numpy array) needs to be implemented. I will do it in the next PR.

src/libasr/pass/python_bind.cpp
Shaikh-Ubaid: We can move it to pass options.

Vipul-Cariappa: I have made the changes. Hope I understood it correctly and this is what you wanted.

Shaikh-Ubaid (approved):
Perfect, thanks Vipul! It looks great! I appreciate it.

PR #2798, opened 2024-08-09
Free up memory at the end of scope

advikkabra:
Free lists, dictionaries and sets when they go out of scope. This is not complete, as there are still some places where memory is being leaked.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -4846,6 +4846,7 @@ public:
             list_api->list_deepcopy(value_list, target_list,
                                     value_asr_list, module.get(),
                                     name2memidx);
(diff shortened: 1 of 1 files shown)


certik:
This is freeing some memory used by the compiler, correct?

There is a separate issue of freeing the memory at runtime, and that should be fixed by explicit ASR support for finalization: https://github.com/lfortran/lfortran/issues/3770#issuecomment-2145669584.

advikkabra:
Yes, this is memory freeing at runtime. I did not know about the explicit ASR support, so this PR can be finished when that ASR support is ready.

certik:
It would be good though to work on that, if you want to give it a shot, that would be awesome.

advikkabra:
I'll try my hand at it.

PR #2809, opened 2024-09-11
Print diagnostics in each interactive loop

### LINKED ISSUE #2757: `i: i32=5; sin(i)` should give an error message in interactive mode
```
>>> i: i32 = 5                                                           1,11  ]
>>> i                                                                    1,2   ]
5
>>> i + 6                                                                1,6   ]
11
>>> i**2                                                                 1,5   ]
25
>>> from math import sin                                                 1,21  ]
>>> sin(i)                                                               1,7   ]
>>>                                                                      1,1   ]
>>> sin(f32(i))                                                          1,12  ]
-0.95892429
```

The `sin(i)` should give an error message, but somehow it got swallowed.

### LINKED ISSUE #2807: Interactive LPython delays the response or crashes
To reproduce,

```console
(lp) namannimmo@srf lpython % lpython
>>> a: i32 = 10                                                                                                                                                                                      1,12  ]
>>> b: f64 = 20.                                                                                                                                                                                     1,13  ]
>>> a + b                                                                                                                                                                                            1,6   ]
 --> input:1:1
  |
1 | a + b
  | ^   ^ type mismatch (i32 and f64)


>>> a + i32(b)                                                                                                                                                                                       1,11  ]
30
>>> pow(a, i32(b))                                                                                                                                                                                   1,15  ]
2147483647
>>> a ** i32(b)                                                                                                                                                                                      1,12  ]
2147483647
>>> a: i32                                                                                                                                                                                           1,7   ]
 --> input:1:1
  |
1 | a: i32
  | ^^ '**' could be used instead

semantic error: Symbol is already declared in the same scope
 --> input:1:1
  |
1 | a: i32
  | ~~~~~~~~~~~ original declaration
  |
1 | a: i32
  | ^^^^^^ redeclaration

```
The suggestion was delayed in the case above, but it might crash, too.

hankluo6:
Fixed #2807
Fixed #2757

In interactive mode, some errors or suggestions were added to the diagnostics but did not throw any errors. These should also be printed.

### DIFF
--- a/src/bin/lpython.cpp
+++ b/src/bin/lpython.cpp
@@ -886,6 +886,8 @@ int interactive_python_repl(
             res = fe.evaluate(code_string, verbose, lm, pass_manager, diagnostics);
             if (res.ok) {
                 r = res.result;
+                std::cerr << diagnostics.render(lm, compiler_options);
+                diagnostics.clear();
             } else {
                 LCOMPILERS_ASSERT(diagnostics.has_error())
                 std::cerr << diagnostics.render(lm, compiler_options);


Shaikh-Ubaid:
I think this works and does the job:

```console
(lp) ubaid@DESKTOP-TEN39M0:~/projects/lpython$ ./src/bin/lpython
Interactive LPython. Experimental prototype, not ready for end users.
LPython version: 0.22.0-53-g1ef4675ae
  * Use Ctrl-D to exit
  * Use Enter to submit
  * Use Alt-Enter or Ctrl-N to make a new line
    - Editing (Keys: Left, Right, Home, End, Backspace, Delete)
    - History (Keys: Up, Down)
>>> i: i32 = 5                                                                                                                                 1,11  ]
>>> from math import sin                                                                                                                       1,21  ]
>>> sin(i)                                                                                                                                     1,7   ]
semantic error: `x` argument of `Sin` must be real or complex
 --> input:1:5
  |
1 | sin(i)
  |     ^ 


>>> pow(2, 3)                                                                                                                                  1,10  ]
 --> input:1:1
  |
1 | pow(2, 3)
  | ^^^^^^^^^ '**' could be used instead


8
>>>     
```

src/bin/lpython.cpp
```diff
@@ -886,6 +886,8 @@ int interactive_python_repl(
             res = fe.evaluate(code_string, verbose, lm, pass_manager, diagnostics);
             if (res.ok) {
                 r = res.result;
+                std::cerr << diagnostics.render(lm, compiler_options);
+                diagnostics.clear();
```
Shaikh-Ubaid, line 890: On second thought, I think this approach is good to print style info/warnings. But for printing errors (semantic errors like in https://github.com/lcompilers/lpython/issues/2757), we should fix the `python_ast_to_asr` to return the result as not ok (that is `res.ok = False`).

Shaikh-Ubaid: I created a new issue for this here https://github.com/lcompilers/lpython/issues/2810. I think this PR is good to merge.

Shaikh-Ubaid (approved):

src/bin/lpython.cpp
hankluo6: I noticed this as well. The reason is that for the intrinsic functions, we used `append_error` instead of throwing a semantic error. This could potentially be resolved by throwing the error later from `python_ast_to_asr` if the intrinsic function returns null.

Issue #1946, opened 2023-06-18
f-strings produce `SIGSEGV`


rebcabin:
repro : https://github.com/rebcabin/lpython/tree/brian-lasr/lasr/lpython/Issue1946

The following works in ordinary CPython:

```python
hexit         : str = r'[0-9a-fA-F]'
octit         : str = r'[0-7]'
digit         : str = r'[0-9]'
alpha         : str = r'[_a-zA-Z]'
asr_alpha_mer : str = r'[_a-zA-Z0-9@~]'


# Really want Pattern[str] and re.compile


hex_pat : str = fr'-?0[xX]{hexit}+'
oct_pat : str = fr'-?0{octit}+'
int_pat : str = fr'-?{digit}+'
dec_pat : str = fr'(-?{digit})M'
# Not kosher Clojure
sym_pat : str = fr'({alpha}{asr_alpha_mer}*)'
```

LPython produces a `SIGSEV` (after commenting out the compatibility block at the top):

```bash
(lp) ┌─(~/CLionProjects/lpython/lasr/lpython)────────────────────────────────────────────────────────────────────────────────────────────────────────────(brian@Golf37:s000)─┐
└─(15:51:59 on brian-lasr ✹ ✭)──> ~/CLionProjects/lpython/src/bin/lpython  Issue1946/lasr_lexer.py && ./lasr_lexer.out                                2 ↵ ──(Sun,Jun18)─┘
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/brian/CLionProjects/lpython/src/bin/lpython.cpp", line 1828
    std::string emit_file_name = basename + "__tmp__generated__.c";
  File "/Users/brian/CLionProjects/lpython/src/bin/lpython.cpp", line 783
    r1 = LCompilers::LPython::python_ast_to_asr(al, lm, *ast, diagnostics, compiler_options,
  File "/Users/brian/CLionProjects/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 7429
    diag::Diagnostics &diagnostics,
  File "/Users/brian/CLionProjects/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 7382
    target_type = ASRUtils::TYPE(ASR::make_Integer_t(al, x.base.base.loc, 8));
  File "/Users/brian/CLionProjects/lpython/src/lpython/semantics/python_ast_to_asr.cpp", line 4492
[...]

PR #2566, opened 2024-02-29
Add support for f-strings

mrdaybird:
Support f-strings in lpython, because f-strings are awesome.
Previously, expression enclosed inside the f-strings was not parsed correctly. 
TODO:
 - [x] Add new tokens to handle f-strings.
     - Divide f-strings into three parts start, middle, end to allow parsing of enclosed expression.
- [x] Add grammar for f-strings.
- [x] Add semantics for f-strings to create correct AST.
     - All the parts of f-string are merged using recursively apply add operation.
- [x] Implement visit_FormattedValue in python_ast_to_asr.cpp
     - This uses handle_intrinsic_str to cast expression to string
 - [ ] Add tests and clean up code.

PR #2819, opened 2025-01-17
Throw error for f-strings

xaerru:
Add error message when f-strings are used. 

#1946


xaerru:
f-strings are being implemented in #2566.

Issue #2668, opened 2024-04-26
Bug: Non-constant values for `sep` do not work inside `print()`


kmr-srbh:
```python
seperator: str = "#"
print(1, 2, 3, 4, 5, sep=seperator)
```
#### CPython
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ python ./examples/example.py
1#2#3#4#5
```

#### LPython
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lpython$ ./src/bin/lpython ./examples/example.py
12345
```

PR #2820, opened 2025-01-18
Fix separator inside print

### LINKED ISSUE #2677: Bug: `sep` is ignored while printing when two adjacent values are either strings or integers
The value of `sep` is ignored while printing when two adjacent values are strings.

```python
print("a", "b", "c", "d", "e", sep="-")
```
```console
abcde
```

The same is the case for integers.
```python
print(1, 2, 3, 4, 5, sep=" is less than ")
```
```console
12345
```

Please note that this is not an issue with the keyword argument itself. When two adjacent objects are necessarily not strings or integers, the value is used.

```python
print(1, 2, 3, 4, 5, "a", "b", "c", [1, 2, 3, 4], sep="->")
```
```console
12345abc->[1, 2, 3, 4]
```

The presence of a list leads to the usage of the separator.

xaerru:
Fix #2677 and #2668.

I think `global_sep_space` isn't needed because we set the sep to `" "` as the default when it's not specified.

I couldn't figure out why that if condition of two consecutive non characters was in place and also why `sep_no_space` was used.

When the if condition of two consecutive non characters was false, the empty separator was being pushed.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -8432,14 +8432,11 @@ public:
         args.push_back(nullptr); // reserve space for fmt_str
         std::vector<std::string> fmt;
         llvm::Value *sep = nullptr;
-        llvm::Value *sep_no_space = nullptr;
         llvm::Value *end = nullptr;
-        bool global_sep_space = false;
         if (x.m_separator) {
             this->visit_expr_wrapper(x.m_separator, true);
             sep = tmp;
         } else {
-            global_sep_space = true;
             sep = builder->CreateGlobalStringPtr(" ");
         }
         if (x.m_end) {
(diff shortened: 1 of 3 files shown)


kmr-srbh (commented):
Thanks for the changes @xaerru! LPython is a sister project of a modern Fortran compiler LFortran. In Fortran, we do not add a space between 2 strings implicitly. We add a space only when either the left value or right value is something other than a string. The deleted code handled the separator based on this criteria.

Please see the discussion at https://github.com/lcompilers/lpython/pull/2670/files#r1581397608 on suggestion to fix this.

xaerru:
In LFortran ASR, `Print` node was refactored to only contain `StringFormat` or a single `StringConstant`.

Also for `StringFormat`, only `FormatFortran` is implemented as of now.

https://github.com/lfortran/lfortran/issues/2543

https://github.com/lfortran/lfortran/pull/4696

I think this should be done after libasr in LPython is synced to LFortran.

Issue #2828, opened 2025-03-13
Import causes unhandles exception


swamishiju:
```py
import string
print(string.ascii_uppercase)
```

The above code runs without any issue but

```
import string
```

simply importing causes an unhandled exception

swamishiju:
Well i found another issue with imports I believe both are linked

```py
from lpython import i32


class Foo:
    def __init__(self: "Foo", a: i32):
        self.a: i32 = a

    def bar(self: "Foo") -> i32:
        return self.a


def bux() -> i32:
    f: Foo = Foo(10)
    return f.bar()


print(bux())
```

The above code works and gives output

```sh
$ lpython main.py
10
```

When we move the class definition into a separate python file

foo.py

```py
from lpython import i32


class Foo:
    def __init__(self: "Foo", a: i32):
        self.a: i32 = a

    def bar(self: "Foo") -> i32:
        return self.a


def bux() -> i32:
    f: Foo = Foo(10)
    return f.bar()
```

main.py

```py
import foo

print(foo.bux())
```

Now this code gives `Internal Compiler Error: Unhandled exception` but both give the same output with CPython

swamishiju:
I think I've found the issue. Import dependency checking doesn't ignore the local use of module variables. Below is a minimal reproduction of the error. 

foo.py
```py
a:i32 = 10
b:i32 = a + 10
```

main.py
```
import foo
```

Here the asr verifier sees that `b` needs as a dependency`a` which in turn leads foo to have itself as a dependency. This is what I think is happening, there are still a couple details to work out like how does it work when the import is used but I'll start working on a fix and open a PR.

swamishiju:
Related issue to the first one

foo.py
```py
print("Hello,World")
```

bar.py
```py
import foo
```

main.py
```py
import foo
import bar
```


```
$ python main.py
Hello,World

$ lpython -I. main.py
Hello,World
Hello,World
```

PR #2829, opened 2025-03-19
Import issues

swamishiju:
This PR is regarding the first part of [#2828](https://github.com/lcompilers/lpython/issues/2828#issue-2916997010) this does seem to fix part of the problem but I think it fails integration tests because it hangs when trying to create the lnn pkg i'm not entirely sure why


swamishiju:
Seems to be working now

integration_tests/test_import_08.py
Shaikh-Ubaid, line 1: Can you show the output of executing this with the lpython built from latest `main` branch? I would like to see this test getting failed with the current lpython.

Shaikh-Ubaid (approved):
lgtm

I think we can merge once we see output for https://github.com/lcompilers/lpython/pull/2829/files#r2004399682.

integration_tests/test_import_08.py
swamishiju: Current lpython main

```
% src/bin/lpython test.py -I.
ASR verify pass error: Module __main__ dependencies must contain string because a function present in it is getting called in __main__.
^C
```

cpython

```
% python3 test.py
Inside import
```

lpython after changes
```
% src/bin/lpython test.py -I.
Inside import
```

swamishiju:
Oh ye there is another thing i noticed. When the compiler crashes and hangs integration tests don't stop and give and error (afaik i may have waited too little). If there is no check for that I think we need to add a TLE exit on each test

swamishiju:
This PR also fixes a related problem

foo.py
```py
print("Inside import")
```

main.py
```py
import foo
```


Lpython current main
```
$ src/bin/lpython main.py -I.
ASR verify pass error: Module __main__ dependencies must contain foo because a function present in it is getting called in __main__.
^C
```


Cpython
```
$ python3 main.py
Inside import
```

Lpython after changes
```
$ src/bin/lpython main.py -I.
Inside import
```

I'll add an stdout test for this too

swamishiju:
@Shaikh-Ubaid all done i think. I'll continue checking for bugs, I have midterms coming so might be less active

swamishiju:
I'm having some problems with tests can someone help?

Shaikh-Ubaid:
You will need to remove those extra commits.

swamishiju:
@Shaikh-Ubaid just did also if you are free I had a question on Zulip

PR #2830, opened 2025-03-22
String attribute methods

swamishiju:
I am facing some problem with tests, adding zfill to builtins seems to change some sort of id for other builtins so it fails ASR tests


swamishiju:
The symbol table counter changes for __main__ when new builtin methods are added

PR #2831, opened 2025-03-26
Added initial support for Bytes type

swamishiju:
I have implemented initial support for the bytes type 

```py
a: bytes
a = b"test"
```

As of now they support assignment, function args, function return type

The implementation itself is very similar to str and most functions carry over directly


swamishiju:
I think i'll keep this draft for now i'll rebase once we have integration tests completely working again

PR #2836, opened 2025-03-30
Ubaid/pr/libasr sync

Shaikh-Ubaid:


Shaikh-Ubaid:
Seems like we need to get two workflows to pass.

Shaikh-Ubaid:
>Locally I got:

I didn't see those locally. What build steps did you use?

PR #2843, opened 2025-04-08
Fixed DSYM for mac

swamishiju:
I was facing some problems with lldb locally for lpython which i didnt with lfortran so i wrote a small fix for it

before 

```
lpython % lldb src/bin/lpython
(lldb) target create "src/bin/lpython"
warning: (arm64) /Users/swaminathshiju/lpython/src/bin/lpython empty dSYM file detected, dSYM was created with an executable with no debug info.
Current executable set to '/Users/swaminathshiju/lpython/src/bin/lpython' (arm64).
(lldb)
```

after

```
lpython % lldb src/bin/lpython
(lldb) target create "src/bin/lpython"
Current executable set to '/Users/swaminathshiju/lpython/src/bin/lpython' (arm64).
(lldb)
```


Shaikh-Ubaid (approved):
It looks like we added similar changes for lfortran [here](https://github.com/lfortran/lfortran/pull/5495). I think this should be good.

Issue #2841, opened 2025-03-31
Fix array dimension mismatch errors


Shaikh-Ubaid:
With the recent libasr sync, we are seeing the following issue with multi-dimensional arrays in LPython.

```bash
(lp) ubaid@DESKTOP-UR96JOC:~/lpython$ cat examples/expr2.py 
from lpython import i32
from numpy import empty, int32

def main0():
    x: i32[2, 5] = empty([2, 5], dtype=int32)
    print(x)

main0()
(lp) ubaid@DESKTOP-UR96JOC:~/lpython$ PYTHONPATH=./src/runtime/lpython python examples/expr2.py 
[[  947372589       22011           0           0 -1864968336]
 [      32530 -1864968272       32530          48           0]]
(lp) ubaid@DESKTOP-UR96JOC:~/lpython$ ./src/bin/lpython examples/expr2.py 
semantic error: Type mismatch in annotation-assignment, the types must be compatible
 --> examples/expr2.py:5:5
  |
5 |     x: i32[2, 5] = empty([2, 5], dtype=int32)
  |     ^              ^^^^^^^^^^^^^^^^^^^^^^^^^^ type mismatch ('i32[2,5]' and 'i32[10]')


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

Fixing the above should hopefully get several of our array integration tests working again.

swamishiju:
I can this up tmrw if its still open by then

swamishiju:
I did find a fix but i haven't wrapped my head around why it works yet I'll put up a PR tho

swamishiju:
@Shaikh-Ubaid I have identified the code causing the issue but i don't know to fix it, I would appreciate if you could help

https://github.com/lcompilers/lpython/blob/1e0d3001955aca8b93685467ec511d0ffd66f35e/src/lpython/semantics/python_ast_to_asr.cpp#L8652-L8653

In the above code, the type of ```zero->base``` becomes ```i32[10]``` instead of ```i32[2,5]```

swamishiju:
@Shaikh-Ubaid I have identified the code causing the issue but i don't know to fix it, I would appreciate if you could help

https://github.com/lcompilers/lpython/blob/1e0d3001955aca8b93685467ec511d0ffd66f35e/src/lpython/semantics/python_ast_to_asr.cpp#L8652-L8653

[...]

PR #2844, opened 2025-04-09
Fix for #2841

swamishiju:
Removed check for struct type and used same procedure for all empty arrays


swamishiju:
#2841

swamishiju:
We still need to fix print for arrays they dont seem to be working properly

https://github.com/lcompilers/lpython/issues/2841#issuecomment-2771299998

Shaikh-Ubaid:
Does this PR also fix the print issue mentioned in https://github.com/lcompilers/lpython/issues/2841#issuecomment-2771299998?

Shaikh-Ubaid:
Ok. I think this is still good. It seems to fix some our integration tests without breaking other tests.

Shaikh-Ubaid (approved):

swamishiju:
> Does this PR also fix the print issue mentioned in [#2841 (comment)](https://github.com/lcompilers/lpython/issues/2841#issuecomment-2771299998)?

@Shaikh-Ubaid  The print issues can't be fixed until we have separated libasr because it is intended fortran behavior but not for python

Shaikh-Ubaid:
>Same problem with string tests also, why don't we have libasr as part of this repo now?

Going forward, the contribution workflow that we intend to use is that any changes to libasr needs to be sent to LFortran first, approved and merged there and then we just update the libasr commit in LPython. This ensures that any changes we do to LibASR  in LPython are always compatible with that of LFortran. This also helps when we do LibASR sync, as now we just need to update the LibASR commit and get the lpython integration tests to pass.

Shaikh-Ubaid:
>The print issues can't be fixed until we have separated libasr because it is intended fortran behavior but not for python

Could you elaborate on the intended behaviour for fortran and python, maybe sharing some small examples?

swamishiju:
> > The print issues can't be fixed until we have separated libasr because it is intended fortran behavior but not for python
> 
> Could you elaborate on the intended behaviour for fortran and python, maybe sharing some small examples?

I discussed this today with Ondrej over Zulip

He said it was supposed to be handle by string_format_kind I think its probably not working 

The examples I have in mind were

In fortran
```f90 
Program Hello
integer, dimension(3,2) :: x
Print *, x
End Program Hello
```

prints 
```
   0       0           0       0           0           0
```

Like in a straight line without any like formatting

This is what i think equivalent code for python would be

```py 
x:i32[3,2] = empty([3,2], dtype=int32)
print(x)
```
which pre-sync prints formatted 
```
0 0 0
0 0 0 
```

I will look into this

PR #2848, opened 2025-04-13
Optional args fixed 

swamishiju:
Sister PR https://github.com/lfortran/lfortran/pull/6985


swamishiju:
We still need to make sure optional args aren't defined after required args after that I think it should be fully working

swamishiju:
I removed a C-test for `def_func_01` it seems to be giving some error related to strings but integration tests run fine for me locally.

swamishiju:
I dont think its needed i was facing some weird bugs i fixed later but i dont think the merge was needed. First time using submodules like this bit messy ill open new PRs

PR #2851, opened 2025-04-15
Optional args

swamishiju:
Fixed optional args for LP

Removed C test for `def_func_01` gives some error relating to `_lfortran_strcpy` not being found


swamishiju:
@certik I found my fix while looking at C back end code if I remember correctly

```py
def foo(bar:i32 = 5):
        pass
```

would give error because of the intent check, if we commented that out the generated C code was

```
void foo(int bar, bool _is_bar_present) {
         bar = 5;
}
```

I figured that something similar happens on C and LLVM back ends since commenting out the intent check for testing always gives `foo(5)` no matter what argument was used.

This code also failed asr_verify after the array_struct_temporary pass (this is the one that creates the assignment) because of this. The easiest way I thought was to simply not create an assignment for optional function arguments and use the symbolic value for simply book keeping and making the proper function calls.

swamishiju:
> How about this fix: [lfortran/lfortran#7016](https://github.com/lfortran/lfortran/pull/7016).

Yes I think makes more sense, mine was based on thinking only optional function arguments will be both Optional and In (im pretty sure but no 100%) but yes your code is definitely better. Its also more readable I think.

swamishiju:
> Thanks for being patient with our new workflow. I checked it out locally, and got on updating the submodule:
> 
> ```
> $ git submodule update
> fatal: transport 'file' not allowed
> fatal: Fetched in submodule path 'libasr', but it did not contain 8f61f6049404b0b7c46dd15ca6dbcb24022b3bc1. Direct fetching of that commit failed.
> ```
> 
> But the following two commands fixed it:
> 
> ```
> $ cd libasr
> $ git remote add swamishiju git@github.com:swamishiju/lfortran
> $ git fetch swamishiju
> $ cd ..
> $ git submodule update
> Submodule path 'libasr': checked out '8f61f6049404b0b7c46dd15ca6dbcb24022b3bc1'
> ```

No worries I had read up on how sub-modules work so that was a plus

swamishiju:
@certik There is also an existing problem with optional args from pre sync

```
def foo(bar:i32, baz:i32 = 2, raz:i32)
```

Will not be a valid function defn in CP I dont think there is a check against that

PR #2862, opened 2025-07-12
Unsigned int support

swamishiju:
Some tests for LP were commented out due to an incomplete implementation of the print for unsigned int types.
Fixing this lets us un-comment all unsigned int tests.

libasr changes: https://github.com/lfortran/lfortran/pull/8034


swamishiju:
**Note:** Failing because of StructType refactor fixed in #2863

swamishiju:
> So what is the way forward here --- update libasr to a newer version?

Yep update libasr and it should just work

PR #2864, opened 2025-07-14
Uncommented c_interop test

swamishiju:
Only uncommented the LLVM test. The C test requires a physical cast implementation in the C backend. CPython has issues with size, likely due to it still using pointer strings, which causes the "size" variable to remain uninitialized.


swamishiju:
Ok uncommenting fully needs a bunch more work since FAST also doesn't work, I'll draft for now and undraft after getting it to work better

